PR #1, opened 2016-05-12
Cleanup ZBT-WE826 DTS, add USB WWAN kmods

ndoo:
Cleanup ZBT-WE826 support:
* Refactor DTS
* Update diag.sh to blink WLAN LED (device has no WPS or status LED)
* Refactor makefile:
   * Add usb-serial-option kmods, comgt and wwan as these are sold on AliExpress with integrated WWAN miniPCIe cards and internal antennae
   * Remove usb-dwc2 kmod as device does not have USB OTG ports

### DIFF
--- a/target/linux/ramips/mt7620/profiles/zbt.mk
+++ b/target/linux/ramips/mt7620/profiles/zbt.mk
@@ -6,11 +6,11 @@
 #
 
 define Profile/ZBT-WE826
-	NAME:=ZBT-WE826 Device
+	NAME:=ZBT-WE826
 	PACKAGES:=\
-		kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
-		block-mount kmod-usb-core kmod-usb-dwc2 kmod-usb2 kmod-usb-ohci \
-		kmod-mt76
+		kmod-mt76 kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
+		block-mount kmod-usb-core kmod-usb2 kmod-usb-ohci kmod-usb-serial \
+		kmod-usb-serial-option kmod-usb-serial-wwan comgt wwan
 endef
 
 define Profile/ZBT-WE826/Description
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -144,7 +144,8 @@ get_status_led() {
 		status_led="$board:amber:cpu"
 		;;
 	oy-0001|\
-	sl-r7205)
+	sl-r7205|\
+	zbt-we826)
 		status_led="$board:green:wifi"
 		;;
 	psr-680w)
--- a/target/linux/ramips/dts/ZBT-WE826.dts
+++ b/target/linux/ramips/dts/ZBT-WE826.dts
@@ -3,122 +3,123 @@
 #include "mt7620a.dtsi"
 
 / {
-       compatible = "zbtlink,zbt-we826", "ralink,mt7620a-soc";
-       model = "ZBT-WE826";
+	compatible = "zbtlink,zbt-we826", "ralink,mt7620a-soc";
+	model = "ZBT-WE826";
 
-       chosen {
-               bootargs = "console=ttyS0,115200";
-       };
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
 
-       palmbus@10000000 {
-               gpio0: gpio@600 {
-                       status = "okay";
-               };
+	palmbus@10000000 {
+		gpio0: gpio@600 {
+			status = "okay";
+		};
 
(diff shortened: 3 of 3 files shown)


blogic:
i have staged a change in my tree that already refactors the dts files. what exactly did you change ?

ndoo:
Spaces to tabs. It's non essential though, the patches to dts file can be
skipped.

blogic:
this should be 1 change / patch. each patch should clearly state what you have changed. 

i looked at the profile last night while making some changes to timage generation code and i think all of these are not required

kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci block-mount kmod-usb-core

blogic:
i'll manually fold the changes into my staging tree today

target/linux/ramips/mt7620/profiles/zbt.mk
```diff
@@ -6,11 +6,11 @@
...
 define Profile/ZBT-WE826
-	NAME:=ZBT-WE826 Device
+	NAME:=ZBT-WE826
 	PACKAGES:=\
-		kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
-		block-mount kmod-usb-core kmod-usb-dwc2 kmod-usb2 kmod-usb-ohci \
-		kmod-mt76
+		kmod-mt76 kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
+		block-mount kmod-usb-core kmod-usb2 kmod-usb-ohci kmod-usb-serial \
+		kmod-usb-serial-option kmod-usb-serial-wwan comgt wwan
```
pepe2k: The general convention is to deploy image with only basic and really needed packages.

This model is available (directly from manufacturer) without the WWAN card, so you shouldn't add custom packages here as the user will be able to install them later.

Remember that there is no way to remove packages which are included/builtin in squashfs image - removing/uninstalling them will not free up space on the FLASH.

ndoo: Understood. The SDHC modules should remain though, as the connector is
soldered on.

blogic: On 12/05/2016 10:04, Andrew Yong wrote:
> In target/linux/ramips/mt7620/profiles/zbt.mk
> <https://github.com/lede-project/source/pull/1#discussion_r62979358>:
> 
>>  	PACKAGES:=\
>> -		kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
>> -		block-mount kmod-usb-core kmod-usb-dwc2 kmod-usb2 kmod-usb-ohci \
>> -		kmod-mt76
>> +		kmod-mt76 kmod-scsi-core kmod-scsi-generic kmod-mmc kmod-sdhci \
>> +		block-mount kmod-usb-core kmod-usb2 kmod-usb-ohci kmod-usb-serial \
>> +		kmod-usb-serial-option kmod-usb-serial-wwan comgt wwan
> 
> Understood. The SDHC modules should remain though, as the connector is
> soldered on.
> 
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/lede-project/source/pull/1/files/aa6b387161fe65b44216824a4c2c0f9f71a86ffd#r62979358>
> 

but its the wrong module. did you test the connector ?

blogic:
there is no modem ? i asked him and he said there always is.

anyhow. i am working on a profile cleanup that removes all the broken package selections and i just pulled the status led change into that tree. the rest is a NAK i am afraid.

target/linux/ramips/mt7620/profiles/zbt.mk
ndoo: Will test and open a separate PR once I find a working combination.

ndoo:
No problem. I understand where you are coming from.

The AliExpress seller will include a Huawei 3G or LTE modem for an
additional fee and PCB antennae are stuck on the inside of the chassis.

I'm not sure if that qualifies as integrated. But it's a valid point that
since they sell a model without the card, users may be better off manually
installing the packages.

Though the disk space saved isn't too significant especially factoring in
that there's a user accessible microSD slot.

pepe2k:
> I'm not sure if that qualifies as integrated. But it's a valid point that since they sell a model without the card, users may be better off manually installing the packages.

Exactly.

From my POV we shouldn't force users to have any additional packages in basic image. If the modem was soldered on the main board, I would vote for including required packages in basic image to support it. But there is mpci-e slot with USB bus only and it can be used for different hardware, like ex. GPS.

> The AliExpress seller will include a Huawei 3G or LTE modem for an additional fee and PCB antennae are stuck on the inside of the chassis.

One seller will do that, other will not: http://postimg.org/image/h3ru1zfpd/

Cheers!

PR #2, opened 2016-05-12
base-files: improve dnsmasq.time handling for dnssec

kdarbyshirebryant:
dnsmasq uses dnsmasq.time to validate if the system clock is current or not.  This process is particularly important when dnsmasq's 'dnssec-check-unsigned' option is enabled.

### DIFF
--- a/package/base-files/files/etc/init.d/sysfixtime
+++ b/package/base-files/files/etc/init.d/sysfixtime
@@ -10,8 +10,8 @@ HWCLOCK=/sbin/hwclock
 boot() {
 	start && exit 0
 
+	local maxtime="$(maxtime)"
 	local curtime="$(date +%s)"
-	local maxtime="$(find /etc -type f -exec date -r {} +%s \; | sort -nr | head -n1)"
 	[ $curtime -lt $maxtime ] && date -s @$maxtime
 }
 
@@ -22,4 +22,13 @@ start() {
 stop() {
 	[ -e "$RTC_DEV" ] && [ -e "$HWCLOCK" ] && $HWCLOCK -w -f $RTC_DEV && \
 		logger -t sysfixtime "saved '$(date)' to $RTC_DEV"
+}
+
+maxtime() {
+	local file newest
+
+	for file in $( find /etc -type f ! -path /etc/dnsmasq.time ) ; do
+		[ -z "$newest" -o "$newest" -ot "$file"] && newest=$file
+	done
+	[ "$newest" ] && date -r "$newest" +%s
 }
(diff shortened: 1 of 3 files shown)


ConorOG:
Why should fixsystime exclude any file? Can a file be too new?

kdarbyshirebryant:
dnsmasq uses the timestamp on dnsmasq.time as a decider on whether to valid dnssec time based signatures.  Once system time has gone beyond dnsmasq.time, dnsmasq considers system time valid and starts validating dnssec timestamps.  If system time is wrong and you validate timestamps, name resolution in essence stops.  This may not be so help if it is ntpd trying to do the lookup :-)  The file based solution adopted by dnsmasq used the assumption that systime would be in the 1970s on boot to then be corrected to present by ntpd.

So, if sysfixtime sets the time to newer than dnsmasq.time, time would be considered valid which may have unfortunate consequences.

A different solution would be to have ntpd signal dnsmasq when it has set time for real.  That's not the solution I coded for :-)

jow-:
@kdarbyshirebryant - the dnsmasq specific file handling should not be added as exception to sysupgrade, you should handle it like that:

- Move the service restart and `dnsmasq.time` handling into an external shell script registering a conffiles hook, see https://github.com/openwrt/luci/blob/master/libs/luci-lib-nixio/root/lib/upgrade/luci-add-conffiles.sh for an example
- Bundle that conffiles hook script with the dnsmasq-dnssec build variant instead

The `dnsmasq.time` exclusion in `sysfixtime` can stay but we can optimize the find command instead of using a slow shell parsing loop.

I think the following should work: `find /etc -type f '!' -path /etc/dnsmasq.time -exec date -r {} +%s \; | sort -nr | head -n1`

kdarbyshirebryant:
Thank you Jow - I shall take your advice and work on a better version - I'll be back, though it might take a while :-)

kdarbyshirebryant:
@jow May I bring your attention to https://patchwork.ozlabs.org/patch/521344/  which discussed (at great length!) a variety to methods for finding the latest file (excluding dnsmasq.time)   I'm reminded that the original reason for changing from the original 'find/exec' combination (indeed my original patch used a '!' -path /etc/dnsmasq.time as per your suggestion above) was to reduce CPU usage.  Justin explores this a little more in his post in the above mentioned thread.

I'm inclined to use his improved option: 
time sh -c 'local file newest; for file in $(find /etc -type f ! -path /etc/dnsmasq.time) ; do [ -z "$newest" -o "$newest" -ot "$file" ] && newest=$file ; done ; [ "$newest" ] && date -r "$newest" +%s'
1463567373
real	0m 0.09s
user	0m 0.09s
sys	0m 0.00s


time find /etc -type f '!' -path /etc/dnsmasq.time -exec date -r {} +%s \; | sort -nr | head -n1
real	0m 1.57s
user	0m 0.06s
sys	0m 0.13s

What do you think?

jow-:
I'd say take the faster find command then.

kdarbyshirebryant:
Hopefully the latest commits are more what you're looking for :-)

kdarbyshirebryant:
Hopefully the latest commits are more what you're looking for :-)

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=382779e009af7c1c688fbd98adf71fb19ce66254

PR #4, opened 2016-05-12
base-files: Enhancements to /etc/profile

VittGam:
- Update the terminal window title with the current directory and hostname, if using an xterm-compatible terminal emulator.
- Add ll, an useful alias to ls.

Cheers,
Vittorio

### DIFF
--- a/package/base-files/files/etc/profile
+++ b/package/base-files/files/etc/profile
@@ -11,8 +11,12 @@ export HOME=$(grep -e "^${USER:-root}:" /etc/passwd | cut -d ":" -f 6)
 export HOME=${HOME:-/root}
 export PS1='\u@\h:\w\$ '
 
+[ "$TERM" = "xterm" ] && export PS1='\[\e]0;\u@\h: \w\a\]'$PS1
+
 [ -x /bin/more ] || alias more=less
 [ -x /usr/bin/vim ] && alias vi=vim || alias vim=vi
+
+alias ll='ls -alF --color=auto'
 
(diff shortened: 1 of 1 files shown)


VittGam:
For the record, this patch previously contained these changes too:

- Change the default prompt to surround the current directory with spaces.
- Add an optional colored prompt, in case the user wants to enable it.
```diff
-export PS1='\u@\h:\w\$ '
+export PS1='\u@\h \w \$ '
+
+# Uncomment the following line if you want a colored prompt:
+#export PS1='\[\033[01;33m\]\u\[\033[01;34m\]@\[\033[01;32m\]\h \[\033[01;31m\]\w \[\033[01;35m\]\$ \[\033[00m\]'
```

jow-:
+1 from me, any objections?

lynxis:
I like the idea. Unsure how much trouble we get into with color.
+1

jow-:
The colored prompt changes have been left out in this PR spin

thess:
Not particularly fond of prompt and ls opts, but I use a custom /etc/profile all the time anyway.

I'm neutral

blogic:
+1, i did not like the PS1 changes as it would just get a 2nd crowd wanting back the old behaviour. i am happy now with it removed

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=1f5ba6983fa915fb4951eb9f5ab82d2b5ba41e78

PR #6, opened 2016-05-13
ipq806x: add support for Linksys EA8500

ianchi:
Tested and working:
* Flashing from original webui using the factory images (it generates one
for original firmware 1.1.4.168206 an another for 1.1.4.169978, newer US
Firmwares are not supported due to signing) 
* Sysupgrading from within OpenWrt.
* wifi 2.4GHz and 5GHz.
* wired lan
* usb storage

Pending:
* patch wifi iface mac address


blogic:
looks ok, will pull this once we have v4.4 support in the tree

plntyk:
I am a little bit unsure about ath10k stuff (the last 2 patches)

- ath10k_core has an "skip_otp=y" option - this option was working on my ath10k hardware (Sitecom WLR8100) that seemed to have some issues w. AC wireless/firmware/art partition (skip_otp not required w. the recent mac80211 from 2016-05-12).

- why is the art/caldata sized differently compared to ar71xx w. ath10k ?

your changes:
ath10kcal_extract "art" 4096 __12064__
ath10kcal_extract "art" 20480 __12064__

ar71xx: __2116__

ianchi:
Hi @plntyk, 
I will look into that option to see if I can get rid off the patch. That would be better.
In the meantime I had to add a modiffied version of the patch to match the new package version for mac80211 commited to Source.

Regarding the difference in size, I don't know. That is the size of the data as it comes preloaded in art partition from Linksys. The size is confirmed in the first two bytes of the caldata.
Archer C2600 which also uses qca99x0 has exactly the same size. Perhaps ar71xx uses a different one, like qca988x and size differs.

bendavid:
Few comments.

1) I looked at the changes in the updated drivers in 2016-05-12 and I think this should solve the issue with the OTP, since the updated logic should cover what the patch was trying to do (same patch also posted here btw http://patchwork.ozlabs.org/patch/619947/).  

2) Yes I confirm that the calibration data is a different size for QCA99x0 vs QCA988x.  I think that this is tied to the firmware itself.  (So possibly QCA9984 will have a slightly different calibration data size again since it appears to have a different firmware.)

ianchi:
@bendavid  I've tried removing the patch but the updated drivers won't work, I'm getting the same old error:
```
[   12.132467] ath10k_pci 0000:01:00.0: unable to read from the device
[   12.132492] ath10k_pci 0000:01:00.0: could not execute otp for board id check: -110
[   12.137537] ath10k_pci 0000:01:00.0: failed to get board id from otp: -110
[   12.145309] ath10k_pci 0000:01:00.0: could not probe fw (-110)
```
In the new drivers, core.c seems similar in this critical part. The problem remains as it continues to check for return code of 'ath10k_core_get_board_id_from_otp(ar)'

The option mentioned by @plntyk is also not working, as it gets checked further down the code, and ath10k_core_get_board_id_from_otp still results in an error.

The modified patch I added to his PR is somehow working. But something with the new driver has changed, and even though wlan is UP and I can connect, I get random warnings traces to the console from compat-wireless-2016-05-12/net/mac80211/rx.c (which weren't there with the previous drivers):

```
[  476.471514] ------------[ cut here ]------------                                                                                                                                           
[  476.471601] WARNING: CPU: 0 PID: 0 at compat-wireless-2016-05-12/net/mac80211/rx.c:4068 ieee80211_rx_napi+0x8c/0x8b4 [mac80211]()                                                          
[  476.475210] Modules linked in: pppoe ppp_async iptable_nat pppox ppp_generic nf_nat_ipv4 nf_conntrack_ipv6 nf_conntrack_ipv4 ipt_REJECT ipt_MASQUERADE xt_time xt_tcpudp xt_state xt_nat xh
[  476.585611] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G        W      3.18.29 #1                                                                                                              
[  476.585976] [<c021e8ac>] (unwind_backtrace) from [<c021b608>] (show_stack+0x10/0x14)                                                                                                       
[  476.593089] [<c021b608>] (show_stack) from [<c038799c>] (dump_stack+0x8c/0xa0)                                                                                                             
[  476.600899] [<c038799c>] (dump_stack) from [<c0226500>] (warn_slowpath_common+0x6c/0x88)                                                                                                   
[  476.607927] [<c0226500>] (warn_slowpath_common) from [<c02265b8>] (warn_slowpath_null+0x1c/0x24)                                                                                           
[  476.616195] [<c02265b8>] (warn_slowpath_null) from [<bf2bea08>] (ieee80211_rx_napi+0x8c/0x8b4 [mac80211])                                                                                  
[  476.624986] [<bf2bea08>] (ieee80211_rx_napi [mac80211]) from [<bf32d7b8>] (ath10k_htt_t2h_msg_handler+0x928/0x984 [ath10k_core])                                                           
[  476.634435] [<bf32d7b8>] (ath10k_htt_t2h_msg_handler [ath10k_core]) from [<bf32e1cc>] (ath10k_htt_txrx_compl_task+0x9b8/0x1178 [ath10k_core])                                              
[  476.646062] [<bf32e1cc>] (ath10k_htt_txrx_compl_task [ath10k_core]) from [<c022923c>] (tasklet_action+0xb8/0x144)                                                                          
[  476.658624] [<c022923c>] (tasklet_action) from [<c0229420>] (__do_softirq+0xf0/0x220)                                                                                                      
[  476.668868] [<c0229420>] (__do_softirq) from [<c0229800>] (irq_exit+0x98/0xe4)                                                                                                             
[  476.676682] [<c0229800>] (irq_exit) from [<c025ab64>] (__handle_domain_irq+0xac/0xd4)                                                                                                      
[  476.683797] [<c025ab64>] (__handle_domain_irq) from [<c02085f8>] (gic_handle_irq+0x40/0x5c)                                                                                                
[  476.691696] [<c02085f8>] (gic_handle_irq) from [<c0209680>] (__irq_svc+0x40/0x74)                                                                                                          
[  476.699848] Exception stack(0xc0713f48 to 0xc0713f90)                                                                                                                                      
[  476.707486] 3f40:                   ffffffed 00000000 0007af22 c020a220 c0712018 c0712000                                                                                                  
[  476.712525] 3f60: 00000000 00000000 ffffffed 00000000 c071a4b4 c0636660 c071b684 c0713f90                                                                                                  
[  476.720682] 3f80: c02190c8 c02190cc 60000013 ffffffff                                                                                                                                      
[  476.728843] [<c0209680>] (__irq_svc) from [<c02190cc>] (arch_cpu_idle+0x34/0x50)                                                                                                           
[  476.733888] [<c02190cc>] (arch_cpu_idle) from [<c0252960>] (cpu_startup_entry+0x150/0x2c8)                                                                                                 
[  476.741353] [<c0252960>] (cpu_startup_entry) from [<c06e7df8>] (start_kernel+0x510/0x51c)                                                                                                  
[  476.749425] ---[ end trace 3e350a6fd7a3d7f4 ]---              
```

ianchi:
Hi @blogic ,
I don't know if this is the right way, but I was also trying to switch to 4.4, so I rebased my branch onto your stagging tree and added a few commits to include the device in that version.

Obviously your 4.4 branch is work in progress, but I wanted to share my results with it.
Basically I'm unable to boot because the NAND device is not recognized:
```
[    1.521561] qcom-nandc 1ac00000.nand: failure submitting descs for command 255
[    2.032489] qcom-nandc 1ac00000.nand: failure submitting descs for command 144
[    2.542044] qcom-nandc 1ac00000.nand: failure submitting descs for command 144
[    3.052525] qcom-nandc 1ac00000.nand: failure submitting descs for command 144
[    3.562469] qcom-nandc 1ac00000.nand: failure submitting descs for command 144
[    3.562504] nand: No NAND device found
```

Before trying with your stagging tree, I have also been trying to upgrade to 4.4 myself, and I was having the same problem.
For qcom_mandc (which I assume as a possible culprit) instead of your patch version I was trying to use the version that is already upstream (and to do so needed to pull other patches to reach that commit). It is in my repo, branch ea8500_mtd45 (also rebased onto yours).

Anyway, it compiles ok, but I had the exact same problem. MTD is not recognized that way either.

I'm willing to help with this target if I can. Please let me know if you have any hint to try.

Thnks

blogic:
Hi,

ok, that is what i found. i hope to have time next week to look into this. i am currently waiting on receiving some test hardware. this is why the pot is pending. thanks for your patience. i think this will take another 2-3 weeks to get resolved.

bendavid:
@ianchi I see the same ath10k stack traces with the new compat-wireless on the C2600, although I don't see any noticable impact on the functionality of the wireless.

ianchi:
Perhaps @nbd168 can help us with this warning from the new version of Compat-wireless.
It is affecting both devices using QCA9984.

I copy again the log:

```
[  476.471514] ------------[ cut here ]------------                                                                                                                                           
[  476.471601] WARNING: CPU: 0 PID: 0 at compat-wireless-2016-05-12/net/mac80211/rx.c:4068 ieee80211_rx_napi+0x8c/0x8b4 [mac80211]()                                                          
[  476.475210] Modules linked in: pppoe ppp_async iptable_nat pppox ppp_generic nf_nat_ipv4 nf_conntrack_ipv6 nf_conntrack_ipv4 ipt_REJECT ipt_MASQUERADE xt_time xt_tcpudp xt_state xt_nat xh
[  476.585611] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G        W      3.18.29 #1                                                                                                              
[  476.585976] [<c021e8ac>] (unwind_backtrace) from [<c021b608>] (show_stack+0x10/0x14)                                                                                                       
[  476.593089] [<c021b608>] (show_stack) from [<c038799c>] (dump_stack+0x8c/0xa0)                                                                                                             
[  476.600899] [<c038799c>] (dump_stack) from [<c0226500>] (warn_slowpath_common+0x6c/0x88)                                                                                                   
[  476.607927] [<c0226500>] (warn_slowpath_common) from [<c02265b8>] (warn_slowpath_null+0x1c/0x24)                                                                                           
[  476.616195] [<c02265b8>] (warn_slowpath_null) from [<bf2bea08>] (ieee80211_rx_napi+0x8c/0x8b4 [mac80211])                                                                                  
[  476.624986] [<bf2bea08>] (ieee80211_rx_napi [mac80211]) from [<bf32d7b8>] (ath10k_htt_t2h_msg_handler+0x928/0x984 [ath10k_core])                                                           
[  476.634435] [<bf32d7b8>] (ath10k_htt_t2h_msg_handler [ath10k_core]) from [<bf32e1cc>] (ath10k_htt_txrx_compl_task+0x9b8/0x1178 [ath10k_core])                                              
[  476.646062] [<bf32e1cc>] (ath10k_htt_txrx_compl_task [ath10k_core]) from [<c022923c>] (tasklet_action+0xb8/0x144)                                                                          
[  476.658624] [<c022923c>] (tasklet_action) from [<c0229420>] (__do_softirq+0xf0/0x220)                                                                                                      
[  476.668868] [<c0229420>] (__do_softirq) from [<c0229800>] (irq_exit+0x98/0xe4)                                                                                                             
[  476.676682] [<c0229800>] (irq_exit) from [<c025ab64>] (__handle_domain_irq+0xac/0xd4)                                                                                                      
[  476.683797] [<c025ab64>] (__handle_domain_irq) from [<c02085f8>] (gic_handle_irq+0x40/0x5c)                                                                                                
[  476.691696] [<c02085f8>] (gic_handle_irq) from [<c0209680>] (__irq_svc+0x40/0x74)                                                                                                          
[  476.699848] Exception stack(0xc0713f48 to 0xc0713f90)                                                                                                                                      
[  476.707486] 3f40:                   ffffffed 00000000 0007af22 c020a220 c0712018 c0712000                                                                                                  
[  476.712525] 3f60: 00000000 00000000 ffffffed 00000000 c071a4b4 c0636660 c071b684 c0713f90                                                                                                  
[  476.720682] 3f80: c02190c8 c02190cc 60000013 ffffffff                                                                                                                                      
[  476.728843] [<c0209680>] (__irq_svc) from [<c02190cc>] (arch_cpu_idle+0x34/0x50)                                                                                                           
[  476.733888] [<c02190cc>] (arch_cpu_idle) from [<c0252960>] (cpu_startup_entry+0x150/0x2c8)                                                                                                 
[  476.741353] [<c0252960>] (cpu_startup_entry) from [<c06e7df8>] (start_kernel+0x510/0x51c)                                                                                                  
[  476.749425] ---[ end trace 3e350a6fd7a3d7f4 ]---              
```

bendavid:
(I think you mean QCA9980.)

nbd168:
This happens when the driver reports packets without setting the band properly in the receive status. It might be an upstream bug, or maybe the calibration data is broken.
It could also simply be a firmware bug and maybe a newer firmware helps.

bendavid:
So, interesting point:  The ath10k stack trace goes away after updating kernel from 3.18 to 4.4.  (Maybe there has been some implicit/minor incompatibilities introduced between the newer compat-wireless and the older 3.18 kernel).  Anyway, looks like the problem will solve itself when the 4.4 update converges regardless.

bendavid:
Ok, no, spoke too soon, the stack trace did eventually appear again also in 4.4.

ianchi:
The QCOM RPM driver was included upstream in v4, but a later commit (regulator: qcom: Rework to single platform device) left the smb208 regulator out.
Added a patch to fix it and update the DT.

bendavid:
Indeed, this probably explains my speed test results here
https://forum.openwrt.org/viewtopic.php?pid=326104#p326104

ianchi:
Added commit to fix buttons & wifi led.

ianchi:
@bendavid I haven't tested wifi speed with this new driver. Will do asap.

My patch enables again smb208 in the driver, but there are still some problems with pcie in DTS.
The qcom,pcie driver expects the following properties that are missing, they should reference a regulator: vdda, vdda_phy, vdda_refclk
So for each of the 3 pcie we get the following error in klog (I assume you get them too) :

```                                                                                            
[    0.315081] 1b900000.pci supply vdda not found, using dummy regulator
[    0.315178] 1b900000.pci supply vdda_phy not found, using dummy regulator
[    0.315284] 1b900000.pci supply vdda_refclk not found, using dummy regulator         
```

I've been doing some research but couldn't find the missing mappings.

bendavid:
Ok I was actually referring to the wired network speed.

ianchi:
Rebased to current trunk.
@blogic I've fixed all conflicts with your updates to ipq806x, so it should be a clean merge.

I've re-tested in EA8500 and is working ok on Kernel 4.4.

blogic:
ovrall looks good but i see a lot of small problems. i will try to look at this in detail during the coming days.

bendavid:
@ianchi, if I cherry-pick your regulator patch on top of the current master and build with kernel 4.4 for the c2600, then the cpu clock speed is set properly, but I get corruption issues over wired lan, with ssh connections for example being terminated with errors like
"Corrupted MAC on input."

This doesn't happen without the patch (but then the cpu frequency scaling doesn't work correctly and I think the cpu is stuck at a default low clockspeed).

target/linux/ipq806x/base-files/etc/board.d/02_network
```diff
@@ -24,6 +25,17 @@ db149)
 	ucidef_add_switch "switch0" \
 		"1:lan" "2:lan" "3:lan" "4:lan" "6@eth1" "5:wan" "0@eth0"
 	;;
+ea8500)
+
+	hw_mac_addr=$(mtd_get_mac_ascii devinfo hw_mac_addr)
+	
+	ucidef_add_switch "switch0" \
+		"1:lan" "2:lan" "3:lan" "4:lan" "0t@eth0" "5:wan" "0t@eth0"
```
bendavid: Does this router really only have one ethernet controller connected to the switch?  Otherwise why not use separate ethernet controllers for wan and lan as for ap148 et al?

ianchi: no it has two. But I've been unable to make them work.
This is the only way I've got wired eth working.
Any ideas to try?

bendavid: Well what happens if you use 
` 	ucidef_add_switch "switch0" \
 		"1:lan" "2:lan" "3:lan" "4:lan" "6@eth1" "5:wan" "0@eth0"
 	;;`
as above?

ianchi:
@bendavid I've been running with the regulator patch for more than a day now, with many SSH sessions and I've never got that error.

I've compared with your C2600 dts and we do have some differences, don´t know if they might be causing the trouble (and my problem with eth1 not coming up):

In phy0: ethernet-phy@0 I have one different qca,ar8327-initvals:
C2600: 0x000e4 0xaa545     /* MAC_POWER_SEL */
EA8500: 0x000e4 0x6a545     /* MAC_POWER_SEL */

Besides in gmac1: ethernet@37200000 {
I have some additional properties:
			qcom,phy_mdio_addr = <4>;
			qcom,poll_required = <1>;
			qcom,rgmii_delay = <0>;
			qcom,emulation = <0>;

When I find some time I'll try using your settings.

ianchi:
@blogic please any changes you see are needed, pls just tell me and I'll gladly correct them.

bendavid:
Ok, yes, for the c2600 I just copied the gmac sections verbatim from the ap148 dts.  Latest version is here I guess:
https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/ipq806x/patches-4.4/708-ARM-dts-qcom-add-gmac-nodes-to-ipq806x-platforms.patch;h=691ebb66baef2387e6c794b88e3e0c705661f8ec;hb=HEAD#l37

ianchi:
By the way, continuing with the wierd eth issues:
on boot I always get a trace warning from pcie2:
```
[    0.452431] qcom-pcie 1b900000.pci: link initialization failed
[    0.454293] ------------[ cut here ]------------
[    0.454321] WARNING: CPU: 0 PID: 1 at drivers/clk/qcom/clk-branch.c:97 clk_branch_toggle+0x130/0x13c()
[    0.454334] pcie2_phy_clk status stuck at 'on'
....
```
As I believe pcie2 is not used, originally I left it with disabled status (just like in C2600).
But if I do so eth0 won't come up either, so I loose all wired networking.
So, for now I have to enable it, and live with the warning.

Any ideas to try?

ianchi:
In my case, the added parameters are in line with what I found in the GPL source code provided by Linksys:
arch/arm/mach-msm/board-ipq806x.c:
```
		/* GMAC1, GMAC2 connected to switch. Attach to PHY 0 to configure switch. */
		pdata = (struct msm_nss_gmac_platform_data *)nss_gmac_1.dev.platform_data;
		pdata->phy_mdio_addr = 4;
		pdata->poll_required = 1;
		pdata->rgmii_delay = 0;
		pdata->phy_mii_type = PHY_INTERFACE_MODE_RGMII;
		pdata->emulation = 0;
		pdata->forced_speed = SPEED_1000;
		pdata->forced_duplex = DUPLEX_FULL;
		pdata->socver = socinfo_get_version();

		pdata = (struct msm_nss_gmac_platform_data *)nss_gmac_2.dev.platform_data;
		pdata->phy_mdio_addr = 0;
		pdata->poll_required = 0;
		pdata->rgmii_delay = 0;
		pdata->phy_mii_type = PHY_INTERFACE_MODE_SGMII;
		pdata->emulation = 0;
		pdata->forced_speed = SPEED_1000;
		pdata->forced_duplex = DUPLEX_FULL;
		pdata->socver = socinfo_get_version();

		platform_device_register(&nss_gmac_1);
		platform_device_register(&nss_gmac_2);
	}
```

ianchi:
I changed the mac power config for ar8337 in dts, like the rest of the boards in this target.
Now eth1 is shown as UP in dmesg
`ipq806x-gmac-dwmac 37400000.ethernet eth1: Link is Up - 1Gbps/Full - flow control off`

but still if a change from tagged vlans in eth0 to untagged in eth0 & eth1 I get no traffic at all through eth1.
Any port can be made to work in the same vlan as port 0, but none in port 6. (run tcpdump on eth1 and no incoming traffic shows up).
I have re-flashed stock rom and there both eth0 & eth1 are working.
In that original firmware they seem to control the switch not with swconfig but with what seems a proprietary tool (ssk_sh), but swconfig is indeed bundled (and only used to send a "reset").
What is strange is that swconfig shows that vlans are disabled?!
```
# swconfig dev switch0 show
Global attributes:
        enable_vlan: 0
        max_frame_size: 1518     
```

Any ideas on what can be causing this behavior? any other ideas to try and test?

bendavid:
Maybe try removing the extra qcom attributes in the gmac1/gmac2 blocks?

ianchi:
Already tried, but it is exactly the same.
I have another weird issue with eth, that perhaps it is somehow related: 
As stated eth1 is not working, but for eth0 to work I need to enable pcie2.
I can't understand why, because I don't think it is connected by that bus. Anyway, if I leave it disabled (like you do in C2600) I loose also eth0.

bendavid:
Maybe the port ordering is not as expected?  Do all four of your LAN ports work?

bendavid:
They also seem to be using only one ethernet interface for lan and wan in ddwrt
https://github.com/mirror/dd-wrt/blob/master/src/router/services/sysinit/sysinit-ipq806x.c#L431
Also for reference the dts they are using is
https://github.com/mirror/dd-wrt/blob/master/src/linux/universal/linux-4.4/arch/arm/boot/dts/qcom-ipq8064-ea8500.dts

Maybe more information here
http://www.dd-wrt.com/phpBB2/viewtopic.php?t=287976&highlight=ea8500

ianchi:
Yes @bendavid, all ports work with port0/eth0 and none with port6/eth1.

I've seen the dd-wrt firmware, in fact I've started flashing it before starting the port of Owrt/LEDE.
They have exactly the same problem.
But stock firmware doesn't. There both eth work fine.
So it must be a way to fix it.
As port6 gets recognized and eth1 is up, perhaps is related with the MII bus?
I will continue digging into Linksys board mach to see if I find something.

bendavid:
For the pci "using dummy regulator" messages, these also appear with 3.18 kernels btw.

ryao:
The warnings at `compat-wireless-2016-05-12/net/mac80211/rx.c:4068` and subsequent issues with wireless cutting out still happen in this PR.

I took @bendavid's advice of going back to an older compat-wireless by reverting the appropriate patches and fixing the build, which produced a firmware build whose only detectable problems are:

* `ath10k_pci 0000:01:00.0: Unknown eventid: 36898` is written to dmesg
* The iPhone 6 Plus appears to be limited to 802.11g speeds at 2.4GHz
* My Lenovo T520's Intel 6300N chipset attains internet speeds 68/26 on 2.4GHz when within 10', but fails to attain quite the same speeds on 5GHz within 10'. It does something like 58/23. No amount of fiddling with /etc/config/wireless will correct this. I detected this after I had flashed away the stock firmware, so I cannot compare. I would expect 5GHz to outperform 2.4GHz at close range, so there is probably an issue lurking in the driver. This is a very minor thing though.
* My chromecast fails to connect to the AP or disconnects when connected, only to fail to reconnect persistently. Note that I have yet to rule out the possibility of misconfiguration on my part on this one. I had been using the EA8500 as a pure AP and neglected to fully disable dnsmasq. I spotted information about dhcp leases on the EA8500 and have disabled dnsmasq. I have yet to verify that this makes the problems go away though.

You can see the precise changes that I have made here:

https://github.com/ryao/lede-source/commits/ea8500

ryao:
There is a patch to suppress the `ath10k_pci 0000:01:00.0: Unknown eventid: 36898` messages:

http://comments.gmane.org/gmane.linux.kernel.wireless.general/151816

I have not tested it.

ianchi:
Remove factory image tag.

Looking at Linksys code, no validation is really done on the tag, so remove it.
For US routers it can only be flashed on firmwares up to
FW_EA8500_1.1.4.168206.
Versions later than that check for gpg signature.
On non US routers it can be flashed on any version (up to now).

ianchi:
@ryao I tried with CT firmware using the new package they have recently submited to trunk (ath10k-firmware-qca99x0-ct).
It is loaded correctly and wifi works, but I'm still getting the mac80211/rx.c stack traces. They seem to only go away switching back to previous compat-wireless version, as already reported.

ianchi:
Hi @blogic, I've cleaned up all the commits to be aligned with the rest of the devices on this target.
The only way it differs with the rest is on the device's dts handling. Instead of adding via a patch, I'm adding it as a file in dts folder (perhaps it should go in files-4.4?) and compiling with the new BuildDTB feature, as is done in many other targets. I find it much clearer.
Please, tell me which way you are moving the targets and I can adapt my commit or help changing the rest of the devices in this target.
This device has been also tested by others in this thread and in the Owrt forum, and its working OK in 4.4. There are some improvements to make, but I think its generally safe to include up to here.

blogic:
@ianchi i'll look at it during the day

dissent1:
Excuse me for posting it here. I'm here after R7800 support which is based on ipq8065/qca9984. The openwrt forum thread isn't very popular and I haven't found any better place to try my luck:)
I've made some commits to get r7800 supported here https://github.com/dissent1/mirror-lede
The profile is same as r7500/d7800 with corrections to ipq8065, which are already supported, dtsi for ipq8065 I borrowed from ddwrt and corrected ath10k firmware package to support qca9984. It builds fine. Only kernel 3.18 atm.
I hesitate testing this build yet :) could you please point out my mistakes? I'm not an IT specialist at all, even not near:)

bobafetthotmail:
@dissent1 that router seems to have a UART header. (see board image here), http://www.smallnetbuilder.com/wireless/wireless-reviews/32958-netgear-r7800-nighthawk-x4s-smart-wifi-gaming-router-reviewed

to the right, above two capacitors there is a 4-pin header.
See if that works (find out what pin is the 3.3V with a multimeter and DON'T touch it ever, the others you are free to just connect a UART-USB dongle blindly to to see what pin is rx, tx and gnd).

It allows you to see what the bootloader has to say and if it is advanced enough you can also communicate with it to reflash a firmware and unbrick it. 
It is also useful to see what goes wrong when booting as all kernel messages will be printed on your serial connection.

Here there is an unbrick guide if you cannot connect to the UART. https://www.youtube.com/watch?v=NflxFlXczVk

dissent1:
@bobafetthotmail thank you very much for the reply:) but my main concern was not about recovering, but using ipq8065.dtsi from ddwrt. As freqs and voltages differ between these chipsets (8064 and 8065) I noticed that ipq8064.dtsi doesn't include cpu freqs and voltages unlike ipq8065.dtsi from ddwrt. How safe it is to use one or another cpu wise? Maybe the one from kernel git is compatible with 8065 and should be used? The dtsi in my git is inserted through patch 900 in 3.18 kernel.

blogic:
@ianchi 

i merged this into my tree with the following changes

* fixe uboot-envtools support
* moved the dts into the kernel
* dropped the wifi_rfkill check function, it is not called anywhere
* renamed the regulator patches
* fixed various whitespace errors and alphabetical ordering problems

please build an image from my staging tree to verify that everything works fine

ianchi:
Hi @blogic, I tested and it is not booting.
The problem is that the commit "ipq806x: add mangle bootargs to 4.4" is not in the tree.
Without it the root is not set and it crashes with a kernel panic.
`No filesystem could mount root`

I cherry picked it on top of your tree and rebuilt the image and now its working OK.

blogic:
pushed an update to my staging tree, could you verify that it is now working ?

heinzek:
@blogic I tested firmware to C2600 from Your current staging tree. Kernel not booting. When i revert change from _ipq806x: add mangle bootargs to 4.4_ firmware booting correctly.

blogic:
@heinzek can i get bootlog with and without the patch ?

heinzek:
With mangle bootargs: 
[http://ctrlv.it/id/6522/1421683963](url)

and nothing happens


Without mangle bootargs:
[http://ctrlv.it/id/6521/2506589860](url)

blogic:
@ianchi can you check if you can add the missing commandline parameter into the dts file and thus omitting the boot mangle stuff ? it seems ot break all other boards

ianchi:
I can't add it to dts as a static one because this model has dual boot partition mechanism, and needs from the bootloader the partition it is currently booting from.
Before finding the mangle bootargs patch from other targets (I believe also with linksys dual boot) I wrote my version with a slightly different approach. I will get it back to work and submit a new commit, for @heinzek to test if it is harmless in C2600, but some more feedback would help to focus the approach:

What I find strange from the bootlog is that no kernel message was printed at all. The patch only renames the "root=" argument in cmdline, and from your working log it seems that you have a blank cmdline anyway. And no kernel messages seems a crash earlier in the boot sequence.
@heinzek besides removing this commit, have you also reverted the corresponding changes in the config file (the ones that where set by commit: "ipq806x: add mangle bootargs options in config-4.4")?

If that is the case, I suspect that the problem might be with
CONFIG_ARM_ATAG_DTB_COMPAT=y
your bootloader might be sending something in the ATAGs that is messing with the early boot process.

If this is really the case, it will be solved by my patch, as it ignores all the ATAGs and only reads the boot partition.

heinzek:
> @heinzek besides removing this commit, have you also reverted the corresponding changes in the config file (the ones that where set by commit: "ipq806x: add mangle bootargs options in config-4.4")?

Yes.. First I revert all changes in config-4.4,
Now I test add some options like:

> CONFIG_ARM_ATAG_DTB_COMPAT_CMDLINE_EXTEND=y
> CONFIG_ARM_ATAG_DTB_COMPAT_CMDLINE_FROM_BOOTLOADER=y
> CONFIG_MANGLE_BOOTARGS=y
it works well.

When add this option:
> CONFIG_ARM_ATAG_DTB_COMPAT=y

Firmware does not start.

ianchi:
Ok then.
I will try to finish the replacement patch for mangle bootargs on the weekend.

ianchi:
Removed commits:

- ipq806x: add mangle bootargs to 4.4
- ipq806x: add mangle bootargs options in config-4.4

replaced with new ones with different approach to getting current partition:

- generic: get ubi.mtd from bootloader's kernel arguments
- ipq806x: enable ubi.mtd from bootloader in config


@heinzek can you please test if this one also works for you.

blogic:
please show me the commandline. i stillt hink this can be encoded into the dts

heinzek:
> @heinzek can you please test if this one also works for you.

Yes. Now firmware booting correctly.

http://ctrlv.it/id/6534/4281904025

blogic:
i am still not sure if a second big manlge patch is required. cant we use the existing one and add a devicetree option in the chosen node to enable the code pach ?

chosen {
     mangle-boot-args;
};

and only do the mangling of this is present ?

ianchi:
@blogic I've rabased to your tree and updated the patch adding a DT property to activate it:
```
chosen {
   append-rootblock = "ubi.mtd="; /* append to bootargs adding the root deviceblock nmbr from bootloader */
}
```

We can't keep the previous mangle_bootargs patch because it needed the 
CONFIG_ARM_ATAG_DTB_COMPAT=y
but it processed all ATAGs, and from @heinzek tests, it seems it messed with C2600 booting.

This patch makes sure only command line ATAG is processed, and with the new DT property it will only be applied to the specific devices needing it.

The previous mangle rootblock patchs were only used in mvebu and kirkwood, as they also have devices with Linksys dual boot mechanism.
I think this new patch could be used also in those plataforms and solve the dual boot problem in the same way everywhere. 

To clarify the initial problem, the bootloader can send one of these alternative arguments, depending on which partition it is booting from at that moment (the other platforms use slight variations):
```
console=ttyHSL1,115200n8 init=/sbin/init rootfstype=squashfs root=31:14
console=ttyHSL1,115200n8 init=/sbin/init rootfstype=squashfs root=31:16
```
The patch parses from the "root=..." the current partition number (rootblock), discards everything and appends to DT bootargs "ubi.mtd=XX", so it is mounted as UBI, and from there the rootfs is selected.

With this change, the patches:
995-mangle_bootargs.patch
100-find_active_root.patch
could be deleted

ryao:
My EA8500 reboots into the previous firmware when I upgrade tot his using sysupgrade via LUCI. It is booted into boot_part=1. Forcing it with `fw_setenv boot_part 2 && reboot` reboots into boot_part=1. I do not have a serial line connected, so I cannot see what is going wrong.

ianchi:
Sorry, for the bug.
Fixed the commit, and now it is working again.
Please @heinzek and @ryao can you re test it.

heinzek:
@ianchi  firmware for C2600 booting correctly.

bendavid:
@heinzek do you have the same problem I had with memory and/or wired ethernet corruption?  Should be reproducible just ssh'ing in and running dmesg or logread repeatedly (or anything which produces a non-trivial amount of output on the console)

heinzek:
Yes. Same problem on C2600:
on linux ssh:

```
Corrupted MAC on input.
ssh_dispatch_run_fatal: Connection to 192.168.1.1: message authentication code incorrect

```

On windows putty 
Incorrect MAC received on packet

bendavid:
Yup, actually one thing I didn't try is whether the same occurs over a wireless connection (could narrow down whether the corruption is occurring in memory or specifically in the wired ethernet connection)

ianchi:
I think we should try to commit this branch so that from there we work on a unified code to fix the pending issues.
If the smb patch is causing trouble in your device, we can leave that out for the meantime while we figure it out, and then re submit it in a new PR.
Alternatively, you can disable the regulator in the C2600's DT.

The bootargs patch still needs a definition, and I believe should be made the same on all three targets using it (that was why I send it as a "general" patch).
If we take the new approach, I can submit the changes to the other platforms, but I don't have the hardware to test them.
If we keep the old patch, we must find another way not to break C2600.
@blogic what do you think?

blogic:
i just pushed everything from last week prior to the boot mangle patch. please rebase any pending work on the current state and send a new PR. i will close this one now. thanks for the effort and sorry for taking so long to process it.

PR #12, opened 2016-05-15
target/sdk: Allow make download to work in SDK

dfc-d:
It's annoying to have SDK builds interrupted by download errors;
allow to fix this by making download target work in SDK

### DIFF
--- a/include/toplevel.mk
+++ b/include/toplevel.mk
@@ -166,10 +166,10 @@ printdb: FORCE
 	@$(_SINGLE)$(NO_TRACE_MAKE) -p $@ V=99 DUMP_TARGET_DB=1 2>&1
 
 download: .config FORCE
-	@+$(SUBMAKE) tools/download
-	@+$(SUBMAKE) toolchain/download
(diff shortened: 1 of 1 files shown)


blogic:
it looks like the patch just disabled the download target and the description says that is makes it work. seems like description and patch are not aligned or am i reading this wrong.

dfc-d:
On 16-05-15 02:05 AM, John Crispin wrote:
> it looks like the patch just disabled the download target and the
> description says that is makes it work. seems like description and patch
> are not aligned or am i reading this wrong.

tools and toolchain dir do not exist in SDK, therefore attempting to run
submake on those dirs results in make download failing with error.

Regards,

Daniel

dfc-d:
I'm also pretty sure target doesn't get built in SDK so it is also skipped in SDK with this patch.

nbd168:
Pushed a slightly different change to fix this issue in 470442ea0fb78333c92bea8e62a87a0cfaa6ba8c

PR #15, opened 2016-05-18
[RFC] Base patch full vs busybox

dfc-d:
This is not fully compile and image tested, but is more as a request for comments to see if I should be only concerned with this for my personal tree, or if it of interest for you folks.  I have a corresponding patchset for the packages feed that does the same kind of thing.

Basically this patchset takes the attitude towards full versions vs. busybox version of commands that seems to be implemented for bzip2, which is that we don't care about not needing --override to opkg in order to install the full version (and in fact this is preferred, since it means the full version is the same path location as the busybox version), but takes the additional step of making sure that if the full verison is removed that the busybox symlink is restored.

It also takes care of making sure sysupgrade gets the necessary files when the full version is installed into the image without the corresponding busybox applet compiled in.

### DIFF
--- a/include/package.mk
+++ b/include/package.mk
@@ -238,6 +238,20 @@ define Build/IncludeOverlay
   endef
 endef
 
+define Package/BusyBoxReplacement/Default
+  define Package/$(1)/postrm
+  #!/bin/sh
+  /bin/busybox $(4) -h 2>&1 | grep -q BusyBox && ln -sf $(if $(3),$(3)/)busybox $(2)/$(4)
+  $(5)
+  exit 0
+  endef
+  define Package/$(1)/install-location
+	$(INSTALL_DIR) $$(1)/$(2)
+	$(INSTALL_BIN) $(6) $$(1)/$(2)/$(4)
+	$(7)
+  endef
+endef
+
 define BuildPackage
   $(Build/IncludeOverlay)
   $(eval $(Package/Default))
--- a/package/base-files/files/lib/upgrade/common.sh
+++ b/package/base-files/files/lib/upgrade/common.sh
@@ -23,10 +23,17 @@ install_bin() { # <file> [ <symlink> ... ]
 	install_file $files
 	shift
 	for link in "$@"; do {
-		dest="$RAM_ROOT/$link"
-		dir="$(dirname $dest)"
-		mkdir -p "$dir"
-		[ -f "$dest" ] || ln -s $src $dest
+		# handle case of using full version of commands
+		# rather than busybox versions
+		if [ -L $link ] && readlink "$link" | grep -q "$(basename "$src")"; then
+			dest="$RAM_ROOT/$link"
+			dir="$(dirname $dest)"
+			mkdir -p "$dir"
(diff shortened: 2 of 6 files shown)


blogic:
replacing BB packages can simply be achieved by putting the full version higher up in the PATH search order. if BB applets are already on the top level then they need to be moved to a lower level int he hierarchy. cluttering the postinst/rm code looks sub optimal and ugly

dfc-d:
On 16-05-18 01:15 AM, John Crispin wrote:
> replacing BB packages can simply be achieved by putting the full version
> higher up in the PATH search order. if BB applets are already on the top

The problem is that in order to do that you have to move *all kinds of*
busybox command from the default location to a location lower in the
search path and that will *break tons o's stuff* that hard codes paths
(and there is rather too much to make this easily done).  Felix actually
commented on the fact that moving busybox commands would break things in
a patch I submitted to OpenWrt, AND that certain things *can't* be moved
(the details are in the list archives if it's really important).

IOW that solution is really not viable.

> level then they need to be moved to a lower level int he hierarchy.
> cluttering the postinst/rm code looks sub optimal and ugly

If you have a way that genuinely works and is not just a knee jerk, "I
don't like the solution, here's something I haven't thought through that
you should do", I'd love to hear it, but the reality is that it's not as
simple as you make it sound.

This is my third attempt at a solution that pleases everyone
(particularly you and Felix).

Regards,

Daniel

nbd168:
I agree with what blogic said. 
Yes, I did mention that your previous patch that changed busybox was going to break stuff, but I also mentioned what you could do to fix it:
- Add compatibility symlinks for things that absolutely need to be in /usr, e.g. /usr/bin/env
- Use the busybox config option for moving things to / instead of patching the source.
I'd prefer if you use your previous approach as a starting point instead of messing around with postrm/preinst hackery.

dfc-d:
On 16-05-18 08:23 AM, nbd168 wrote:
> I agree with what blogic said.
> Yes, I did mention that your previous patch that changed busybox was
> going to break stuff, but I also mentioned what you could do to fix it:
> 
>   * Add compatibility symlinks for things that absolutely need to be in
>     /usr, e.g. /usr/bin/env
>   * Use the busybox config option for moving things to / instead of
>     patching the source. I'd prefer if you use your previous approach as
>     a starting point instead of messing around with postrm/preinst hackery.

Ok, I misunderstood you to mean you didn't want it done that way.  In
that case I will implement this way.

I also presume that for the few cases (like env) where there is
potential collision between the busybox version and the full version
that postrm is okay (so as to avoid broken system)?

Also, for sysupgrade we update it to use the new busybox paths (in
/bin:/sbin)?

Regards,

Daniel

nbd168:
Is there *any* value in even having coreutils-env?

dfc-d:
On 16-05-18 08:36 AM, nbd168 wrote:
> Is there /any/ value in even having coreutils-env?

Good point.  I rather doubt it provides any added functionality, so just
dropping it from the coreutils package is probably smarter.  If there
are any other can't be moved commands chances are high that the full
version is irrelevant in those cases as well.

Regards,

Daniel

PR #16, opened 2016-05-18
[RFC] Patch sdk prevent reconfigure

dfc-d:
This is another in-progress, is-it-just-for-me? request for comments patch.

This patchset builds on a patch I previously implemented that prevents packages built in core from being reconfigured in the SDK (in order to prevent packages in SDK from being built with options that conflict with the core options).

There are a couple of additional patches now, that takes care of the following:

1) Makes sure that all binary packages for source packages built when building the SDK selected by default (but this can be overridden) in order to make sure that the binary packages do no become unavailable due to not having been built in core, and not being able to be built in SDK (without using the override mechanism).  This does have the caveat that it causes generating the package KConfig to take a significant amount more time than previously (what I do is generate a hidden SOURCE_<src-pkg-name> bool that defaults to y to if any of the packages that depend on <src-pkg-name> are selected (except ones such kernel which I have an exclude mechanism for, because of recursive dependencies).  I also make all packages default to m if there is no other default and it's source packages SOURCE_<pkg-name> is y.

2) Except for package selection and image generation options I have modified the configuration preservation mechanism to preserve unset (disabled) settings as well, because this can affect whether a package gets compiled with the same options as during the core build or not (and the objective is avoid packages getting built with conflicting options).


dfc-d:
Sorry, wrong patchset

PR #20, opened 2016-05-18
[RFC] network/services/uhttpd: Add Basic Auth config

dfc-d:

instance: Which uhttpd config section the auth is associated with

If instance is not specified auth section applies to
all instances of uhttpd; if instance is specified
the auth section only applies to the uhttpd config
section with the same name as specified with the
instance option.

Has been tested and working, but is it of interest?  I have an LuCI app I did for uhttpd that was accepted, and PR for an update that utilizes this support, if that makes any difference

### DIFF
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -53,6 +53,28 @@ generate_keys() {
 	}
 }
 
+create_httpauth() {
+	local cfg="$1"
+	local basecfg="$2"
+	local prefix username password
+
+	local instance
+	config_get instance "$cfg" instance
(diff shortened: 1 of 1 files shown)


jow-:
It probably makes sense to concat `/etc/httpd.conf` to the volatile `/var/etc` file here, otherwise LGTM.

dfc-d:
I've implemented that suggestion but  haven't tested yet - I'll let you know when I have.

jow-:
Seems the config append got lost again?

dfc-d:
Hmmm...must have pushed from wrong tree.  Will fix.

blogic:
no response after 3 week. please reopen when there is an update.

dfc-d:
Hmmm....I thought I had done that....will figure out what's going on.

PR #21, opened 2016-05-18
[URFC] default sdk ib: Make building minimal sdk and ib the default

dfc-d:
Make the default building a minimal sdk (consisting of only the
toolchain, kernel, and all binary packages for source packages
depended on by the toolchain and kernel and chosen target) and
minimal imagebuilder (no packages included since that would defeat
build a minimal sdk), and building everything else from the
SDK (and have images generated by imagebuilder for which enable
menuconfig because with no packages built during core build it
doesn't make sense to configure and build images).

NOTE: that I plan on removing the DEFAULT_MODULE_PACKAGES as that becomes unnecessary with the SOURCE_<src-pkg-name> logic in a patch in my RFC patch for preventing reconfiguration of already built packages in SDK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ key-build*
 .emacs.desktop*
 TAGS*~
 git-src
+.minimal-sdk
+.nas
--- a/README
+++ b/README
@@ -21,6 +21,19 @@ To build your own firmware you need to have access to a Linux, BSD or MacOSX sys
 (case-sensitive filesystem required). Cygwin will not be supported because of
 the lack of case sensitiveness in the file system.
 
+If you wish to build a minimal sdk or use the nas device type profile, then
+*before* any other commands (or else 'rm -rf .config tmp'), you need to either
+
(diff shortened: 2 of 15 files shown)


dfc-d:
Reworked a lot.

1) Removed DEFAULT_MODULE_PACKAGES
2) Added mechanism for selecting DEVICE_TYPE (of which minimal_sdk is one)
3) Removed IN_SDK and IN_IB KConfig options
4) When minimal sdk is the device type, generates a hidden KConfig option SDK_MINIMAL that turns off ALL, defaults IB to y, defaults IB_STANDALONE to n, and defaults to not building images in the core compile.  This options is unset in the SDK itself.  
5) Updated README with how to create a minimal SDK.

dfc-d:
Currently, untested, requesting comments on the approach.

dfc-d:
In testing I have discovered some issues that I am working; still I would like to know if this is even something of interest to others.

PR #22, opened 2016-05-18
Patch acl xattr

dfc-d:
Has been tested and working in some interation, however I've had to rebase hence RFC and not full PR.  Is this of interest or only for my own tree?


blogic:
kernel: Add option to make using filesystem ACL support the default
procd patch should not be part of this, also i dont like the ifedefery. why do we need it ?

system/fstools: Add option to support POSIX ACL and XATTR …	
no ifdefery please, why is it needed ?

dfc-d:
On 16-05-18 01:22 AM, John Crispin wrote:
> kernel: Add option to make using filesystem ACL support the default
> procd patch should not be part of this, also i dont like the ifedefery.

Ok
> why do we need it ?

I was trying to not impose the code on you by making it always used.  If
you're fine with (as a separate patch) just including it I can get rid
of the ifdef.

> 
> system/fstools: Add option to support POSIX ACL and XATTR …
> 
> no ifdefery please, why is it needed ?

As above.

blogic:
On 18/05/2016 07:27, Daniel Dickinson wrote:
> On 16-05-18 01:22 AM, John Crispin wrote:
>> kernel: Add option to make using filesystem ACL support the default
>> procd patch should not be part of this, also i dont like the ifedefery.
> 
> Ok
>> why do we need it ?
> 
> I was trying to not impose the code on you by making it always used. If
> you're fine with (as a separate patch) just including it I can get rid
> of the ifdef.
> 
>>


/usr/include/linux/fs.h:#define MS_NOUSER	(1<<31)

why cant we simply include that file ?

dfc-d:
On 16-05-18 02:00 AM, John Crispin wrote:
> 
> 
> /usr/include/linux/fs.h:#define MS_NOUSER (1<<31)
> 
> why cant we simply include that file ?

You hadn't done so before for the other MS_ options, so I didn't add it
for these ones?

Regards,

Daniel

blogic:
ok, we should use the include file. if there are other MS_* options then please remove them and use the include instead while at it.

dfc-d:
1) Dropped procd patch which I thought I had eliminated from my tree in any case.
2) Dropped fstools patch (will do as patch directly against fstools.git without ifdefs)

dfc-d:
Renamed as non-RFC as it has been tested in its current iteration (although to be useful it requires either mount-utils package or the patches I've submitted for fstools; the acl tool at least is also recommeneded).

dfc-d:
Realized I need to change default to n

dfc-d:
Fixed that

blogic:
@jow-  ok to merge this ?

dfc-d:
I noticed (and corrected) the following two cosmetic issues:

The config symbol PROCD_TMPFS_ACL made no sense because I decided against adding that patch / operation (option to mount tmpfs on /tmp with POSIX ACL on boot), so I removed it.

I also noticed the 3.18 kernel config has a malformed edit that effectively remove the CONFIG_REISER_FS_XATTR symbol ( 'is not set' got deleted somehow).  I fixed that.

jow-:
@blogic, @dfc-d LGTM

nbd168:
LGTM

blogic:
@dfc-d i would merge this please tell me how you tested it and if you tested it on all 3 kernels.

dfc-d:
Sorry, I forgot there are multiple versions of the kernel in the tree -
This will be a while longer.  I only tested 4.4 kernel (current ar71xx
default).

Also, I didn't test a great number of the filesystems, because I
presumed that the defaults were only flipping on well tested kernel
options (i.e. I know that the options I set are typically enabled by
default on desktops and servers, and used regularly, in fact the ones
that are likely to be troublesome (jffs2, f2fs, old nfs, and maybe
tmpfs, I made default n, although not for that reason (it was more a
case of where I thought it was likely to be wanted for normal use)).

I only tested on ar71xx, with ext4 and vfat (with vfat the test was only
to verify that it didn't break mounting with acl and user_xattr options
set, but obviously the attempts to actually set acls or attributes gives
an error message).

I used the acl tool to set and removed acls, and confirmed that default
acls were probably propagated.  Likewise with attr tool I confirmed
attributes were set and preserved.  I also unmounted/remounted and also
checked on a desktop that the fs was okay.  (For acls I created/deleted
files and directories to verify permissions were honoured or not as they
should for a given user after modifying the fs on one, then the other
(and doing the checks on the opposite side in each case), attrs don't
take as much testing since they are only key/value pairs to test with
the attr tool).

I'm not sure quite how extensive testing you are looking for, but it sounds like it my be outside the scope of my available time (although testing all three kernels as above I should be able to do, provided there aren't compilation issues).

blogic:
On 23/05/2016 14:09, Daniel Dickinson wrote:
> Sorry, I forgot there are multiple versions of the kernel in the tree -
> This will be a while longer. I only tested 4.4 kernel (current ar71xx
> default).
> 
> Also, I didn't test a great number of the filesystems, because I
> presumed that the defaults were only flipping on well tested kernel
> options (i.e. I know that the options I set are typically enabled by
> default on desktops and servers, and used regularly, in fact the ones
> that are likely to be troublesome (jffs2, f2fs, old nfs, and maybe
> tmpfs, I made default n, although not for that reason (it was more a
> case of where I thought it was likely to be wanted for normal use)).
> 
> I only tested on ar71xx, with ext4 and vfat (with vfat the test was only
> to verify that it didn't break mounting with acl and user_xattr options
> set, but obviously the attempts to actually set acls or attributes gives
> an error message).
> 
> I used the acl tool to set and removed acls, and confirmed that default
> acls were probably propagated. Likewise with attr tool I confirmed
> attributes were set and preserved. I also unmounted/remounted and also
> checked on a desktop that the fs was okay. (For acls I created/deleted
> files and directories to verify permissions were honoured or not as they
> should for a given user after modifying the fs on one, then the other
> (and doing the checks on the opposite side in each case), attrs don't
> take as much testing since they are only key/value pairs to test with
> the attr tool).
> 
> I'm not sure quite how extensive testing you are looking for, but it
> sounds like it my be outside the scope of my available time (although
> testing all three kernels as above I should be able to do, provided
> there aren't compilation issues).
> 

too much text for what should be a yes/no answer.

you patch 3 kernel configs, so please test all 3 configs. you should
build 1 image for each of the kernels using any target you want.

dfc-d:
Posted via github since my mail to github got lost somehow.

You ask how I tested, so I answered.  It wasn't clear to me that you only actually wanted an answer to the question about testing all three kernels.

PR #23, opened 2016-05-18
[RFC] package/utils/mdadm: Fix missing conffile and add initscript

dfc-d:
Start arrays on boot, stop them on shutdown, and enable
monitoring to syslog.

Has been tested in some iteration however I had to rebase, but because it seems simple enough that it shouldn't be a problem, I submit anyway


blogic:
lgtm

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=5c847fc84d238de1f35ab20ade2b841a60d314ab

PR #24, opened 2016-05-18
target/imagebuilder: Fix sorting package list breaks opkg dependency …

dfc-d:
…handling for provides

When imagebuild sorts package lists it breaks opkg's ability to realize
that a providers for a Provides has already been installed, when the sort
results in the provider being later in the list of packages that a package
which depends on a Provides (and hence the provider is not yet installed
for opkg to realize the provider was available doesn't not handle the case
of a package that is to be installed satisfying a dependency, only one that
is already installed (or which it schedules to be installed, which in the
absence of an installed provider is whichever provider happens to be the
default)

It'll probably be quite a while before I get a chance to do the real fix, which is to patch upstream's opkg so it's suitable for LEDE, hence this workaround for now.

### DIFF
--- a/target/imagebuilder/files/Makefile
+++ b/target/imagebuilder/files/Makefile
@@ -104,7 +104,7 @@ _call_info: FORCE
 	echo 'Available Profiles:'
 	echo; $(PROFILE_LIST)
 
-BUILD_PACKAGES:=$(sort $(DEFAULT_PACKAGES) $(USER_PACKAGES) $($(USER_PROFILE)_PACKAGES) kernel)
+BUILD_PACKAGES:=$(USER_PACKAGES) $(sort $(DEFAULT_PACKAGES) $($(USER_PROFILE)_PACKAGES) kernel)
 # "-pkgname" in the package list means remove "pkgname" from the package list
 BUILD_PACKAGES:=$(filter-out $(filter -%,$(BUILD_PACKAGES)) $(patsubst -%,%,$(filter -%,$(BUILD_PACKAGES))),$(BUILD_PACKAGES))
 PACKAGES:=


blogic:
@nbd168

jow-:
I agree with the interim fix but can you please elaborate on the change from `$($(USER_PROFILE)_PACKAGES)` to `$(foreach profile,$(PROFILES_BUILD),$($(profile)_PACKAGES))` ?

This looks unrelated.

dfc-d:
On 16-05-18 07:53 AM, Jo-Philipp Wich wrote:
> I agree with the interim fix but can you please elaborate on the change
> from |$($(USER_PROFILE)_PACKAGES)| to |$(foreach
> profile,$(PROFILES_BUILD),$($(profile)_PACKAGES))| ?
> 
> This looks unrelated.

You're right.  I must have leaked that from another patch series.  I
will update.

Regards,

Daniel

dfc-d:
Fixed (but not yet tested).

lynxis:
what about reproducible builds. is USER_PACKAGE everytime in the same order or can we sort those seperat?

dfc-d:
On 16-05-18 05:25 PM, Alexander Couzens wrote:
> what about reproducible builds. is USER_PACKAGE everytime in the same
> order or can we sort those seperat?

As stated in the commit message the user supplies the packages in the
order of their chosing, so if they care about reproducible builds, they
can feed the list in the same order every time.

Regards,

Daniel

dfc-d:
Oh, I guess I edited the commit message and dropped that info. Sorry.

jow-:
LGTM with latest changes

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=1c5a2aed6eca6fdefe1be2bed509f79fd0c734a2

jow-:
Minor nitpick for the future: use a short prefix for the subject, e.g. `imagebuilder: ...` instead of `target/imagebuilder: ...` - this seems to be convention people settled on so I adjusted your commits to shorten the subject prefix.

dfc-d:
On 16-05-18 05:52 PM, Jo-Philipp Wich wrote:
> Closed #24 <https://github.com/lede-project/source/pull/24>.


Thanks, I appreciate your having the confidence to pull the changes into
you test tree so that they can get tested before I necessarily would
have a chance to.  I admit that's partly why I do RFC's for things I
have tested - it doesn't make sense to block on *my* doing a full test,
if others agree that it looks good enough to pull in for their own
tests, before I might have a chance to get to it.

Regards,

Daniel

dfc-d:
On 16-05-18 05:55 PM, Jo-Philipp Wich wrote:
> Minor nitpick for the future: use a short prefix for the subject, e.g.
> |imagebuilder: ...| instead of |target/imagebuilder: ...| - this seems
> to be convention people settled on so I adjusted your commits to shorten
> the subject prefix.

I'm happy to do that; I was following what I understood to be previous
openwrt convention (or at least seemed to be for some time).

Regards,

Daniel

dfc-d:
And in the case of this one; I had used the same thing before rebasing,
already, so it's basically tested, just not double-checked.

PR #25, opened 2016-05-18
[RFC] base: Add option to prefer full vs busybox

dfc-d:
Add option to allow to prefer the use of full version
and not even build busybox version of commands.

Note that for sysupgrade this depends on the patch in the busybox RFC series I posted.

Not sure this is of any actual intrest to LEDE team (although based on folks trying to do things like make OpenWrt/LEDE self-hosting there are at least some who will like the concept).

### DIFF
--- a/config/Config-build.in
+++ b/config/Config-build.in
@@ -19,6 +19,12 @@ menu "Global build settings"
 		bool "Select all userspace packages by default"
 		default n
 
+	config PREFER_FULL_VS_BUSYBOX
+		bool "Prefer full versions of commands vs Busybox version"
+		depends on FEED_packages
+		select PACKAGE_prefer-full
+		default y
+
 	config SIGNED_PACKAGES
 		bool "Cryptographically signed package lists"
 		default y
--- a/package/system/prefer-full/Makefile
+++ b/package/system/prefer-full/Makefile
@@ -1,0 +1,111 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=prefer-full
+PKG_VERSION:=1.0
+PKG_RELEASE=1
+
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=
+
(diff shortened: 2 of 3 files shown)


nbd168:
Some issues with that:
- Config-defaults.in is a generated file. Please change the script to regenerate it appropriately.
- default y is inappropriate for this option.
- prefer-full is a really weird and confusing package name

How about this:
- create a coreutils-full package in the coreutils makefile which selects all the right coreutils packages
- create a hidden symbol HAS_COREUTILS in the coreutils makefile (default y)
- create a config symbol in busybox config submenu 'Use coreutils instead of busybox replacements', which depends on HAS_COREUTILS and selects coreutils-full and the defaults change.

dfc-d:
On 16-05-18 08:34 AM, nbd168 wrote:
> Some issues with that:
> 
>   * Config-defaults.in is a generated file. Please change the script to
>     regenerate it appropriately.

Ok, didn't realize that.

>   * default y is inappropriate for this option.

Urgh, that was unintentional leak from testing.

>   * prefer-full is a really weird and confusing package name

Ok

> 
> How about this:
> 
>   * create a coreutils-full package in the coreutils makefile which
>     selects all the right coreutils packages
>   * create a hidden symbol HAS_COREUTILS in the coreutils makefile
>     (default y)
>   * create a config symbol in busybox config submenu 'Use coreutils
>     instead of busybox replacements', which depends on HAS_COREUTILS and
>     selects coreutils-full and the defaults change.
> 

The main problem with that is that the full version also come from
things like util-linux, netcat, and procps-ng, etc.

OTOH making the (renamed) package a packages feed package and having a
hidden symbol makes sense.

Regards,

Daniel

nbd168:
You can make util-linux, netcat, procps-ng configurable separately if you need them. That way dealing with dependencies is also easier, since not everybody is doing ./scripts/feeds install -a

PR #26, opened 2016-05-18
[RFC] package/kernel: Build it87 hardware monitor module

dfc-d:
Add packaging of it87 hardware monitor kernel module.  It is
a common thermal and voltage monitor that is in many x86
(at least) devices, and is just another i2c hwmon module.

Has been tested in some iteration, however not since I've rebased.  Simple enough to submit anyway.

### DIFF
--- a/package/kernel/linux/modules/hwmon.mk
+++ b/package/kernel/linux/modules/hwmon.mk
@@ -93,6 +93,20 @@ endef
 $(eval $(call KernelPackage,hwmon-ina2xx))
 
 
+define KernelPackage/hwmon-it87
+  TITLE:=IT87 monitoring support
+  KCONFIG:=CONFIG_SENSORS_IT87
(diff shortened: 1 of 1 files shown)


lynxis:
LGTM

package/kernel/linux/modules/hwmon.mk
```diff
@@ -93,6 +93,20 @@ endef
 $(eval $(call KernelPackage,hwmon-ina2xx))
 
 
+define KernelPackage/hwmon-it87
+  TITLE:=it87 monitoring support
```
stintel: Minor nit: other modules use uppercase characters before the number, I would prefer the same here for consistency

dfc-d: Done

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=bda35954aa8b14001cec4b2a3358a30ea17db857

PR #27, opened 2016-05-18
script/symlink-tree.sh: Fix missing config dir

dfc-d:
Apparently symlink-tree has not been used in quite some time as it
fails to symlink the always required config dir

Also, if we pulled from git but .git is missing we get
many error messages on the symlinked tree without this
patch (which symlinks .git, if present)

### DIFF
--- a/scripts/symlink-tree.sh
+++ b/scripts/symlink-tree.sh
@@ -4,6 +4,7 @@
 
 FILES="
 	BSDmakefile
+	config
 	Config.in
 	LICENSE
 	Makefile
@@ -18,6 +19,9 @@ FILES="
 	target
 	toolchain
 	tools"
+
+OPTIONAL_FILES="
+	.git"
 
 if [ -f feeds.conf ] ; then
 	FILES="$FILES feeds.conf"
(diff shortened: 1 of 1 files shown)


blogic:
@nbd168

jow-:
The change adds a redundant `exit 0` otherwise it looks logically sound, I just have no idea what this script is used for.

dfc-d:
On 16-05-18 08:42 AM, Jo-Philipp Wich wrote:
> The change adds a redundant |exit 0| otherwise it looks logically sound,
> I just have no idea what this script is used for.

It's so you can have differently configure builds of the same arch (or
different arch, but that's not so relevant now since you can build
multiple archs in the same tree, which is probably why it hasn't got
much use), or you just want to keep your builds in separate trees
without cloning multiple trees (e.g. if you want to have a single tree
where you commit, but have the changes reflected in all build trees).
An alternative is to use ./script/env and switch configs between builds;
you say pah-tae-tah, I say pah-tae-toe.

Regards,

Daniel

dfc-d:
Removed extra exit 0

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=2682cb4830b6f0f4ace63a11431144633fc18f7a

PR #28, opened 2016-05-18
[RFC] package/system/fstools: Add dosfsck use for vfat check

dfc-d:
dosfsck is command line compatible with e2fsck, but is for vfat, so we can easily support checking vfat filesystems using the same code as for ext* systems

Have not actually image tested yet, but seem simple enought that it shoud 'Just Work'


blogic:
rework this to

        if (!strncmp(pr->id->name, "vfat", 4)) {
           chkfs = dosfsck
        else if (!strncmp(pr->id->name, "ext4", 3)) {
           chkfs = e2fsck
        else
                ULOG_ERR("check_filesystem: %s is not supported\n", pr->id->name);
                return;
        }

and send a patch against the fstools.git to the mailing lit please

PR #29, opened 2016-05-18
[RFC] package/network/services/odhcpd: per-host leasetime

dfc-d:
Allow per-host leasetime which is handy when you want to have
short leasetime for hosts not specifically configured, and a longer
leasetime for hosts fully known to system.

Tested in some interation, but not since rebase, however it will probably need to be turned into patch againt odhcpd git tree anyway, hence request comments, first.

### DIFF
--- a/package/network/services/odhcpd/patches/110-add-per-host-leasetime.patch
+++ b/package/network/services/odhcpd/patches/110-add-per-host-leasetime.patch
@@ -1,0 +1,218 @@
+Index: odhcpd-2015-11-19/src/config.c
+===================================================================
+--- odhcpd-2015-11-19.orig/src/config.c	2016-04-09 07:07:14.432237095 -0400
++++ odhcpd-2015-11-19/src/config.c	2016-04-09 07:07:14.428237095 -0400
+@@ -100,6 +100,7 @@
+ 	LEASE_ATTR_MAC,
+ 	LEASE_ATTR_DUID,
+ 	LEASE_ATTR_HOSTID,
++	LEASE_ATTR_LEASETIME,
+ 	LEASE_ATTR_NAME,
+ 	LEASE_ATTR_MAX
+ };
+@@ -110,6 +111,7 @@
+ 	[LEASE_ATTR_MAC] = { .name = "mac", .type = BLOBMSG_TYPE_STRING },
+ 	[LEASE_ATTR_DUID] = { .name = "duid", .type = BLOBMSG_TYPE_STRING },
+ 	[LEASE_ATTR_HOSTID] = { .name = "hostid", .type = BLOBMSG_TYPE_STRING },
++	[LEASE_ATTR_LEASETIME] = { .name = "leasetime", .type = BLOBMSG_TYPE_STRING },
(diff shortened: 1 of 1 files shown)


blogic:
@sbyx

sbyx:
Thanks

sbyx:
Looks good to me

blogic:
@sbyx can you merge this into the odhcpd tree ?
@dfc-d  once merged i assume we'll need some script cruft to actually use this option

dfc-d:
On 16-05-18 01:43 AM, John Crispin wrote:
> @sbyx <https://github.com/sbyx> can you merge this into the odhcpd tree ?
> @dfc-d <https://github.com/dfc-d> once merged i assume we'll need some
> script cruft to actually use this option

Nope, just add

config host
	option name 'somehost'
	option ip '192.168.1.2'
	option mac 'AA:BB:CC:DD:EE:FF'
	option leasetime 24h

Just like with the equivalent dnsmasq option that was added.

Regards,

Daniel

dfc-d:
On 16-05-18 01:46 AM, Daniel Curran-Dickinson wrote:
> On 16-05-18 01:43 AM, John Crispin wrote:
>> @sbyx <https://github.com/sbyx> can you merge this into the odhcpd tree ?
>> @dfc-d <https://github.com/dfc-d> once merged i assume we'll need some
>> script cruft to actually use this option
> 
> Nope, just add
> 
> config host
> 	option name 'somehost'
> 	option ip '192.168.1.2'
> 	option mac 'AA:BB:CC:DD:EE:FF'
> 	option leasetime 24h
> 
> Just like with the equivalent dnsmasq option that was added.
> 

Or duid for ipv6 obviously.

Regards,

Daniel

blogic:
ok, in that case could you send a patch against web.git to add that info and if odhcpd is not part of the docs/ folder add that page/section ?

blogic:
https://www.lede-project.org/docs/uci_dnsmasq.html  this page would be the right one i guess. i think we should also rename it to uci_dhcp.txt

dfc-d:
On 16-05-18 01:56 AM, John Crispin wrote:
> ok, in that case could you send a patch against web.git to add that info
> and if odhcpd is not part of the docs/ folder add that page/section ?

Will do.

dfc-d:
On 16-05-18 01:57 AM, John Crispin wrote:
> https://www.lede-project.org/docs/uci_dnsmasq.html this page would be
> the right one i guess. i think we should also rename it to uci_dhcp.txt

Actually the docs already mentioned this option because it was added to
dnsmasq but not (until now) odhcpd, so there was actually a
documentation error until now, for the odhcpd case.

Regards,

Daniel

blogic:
good. we need to be vigilant and try to enforce that uci option changes always come with the according docs change to avoid this kind of situation in future.

blogic:
@sbyx can you pull this ?

sbyx:
Done. https://github.com/sbyx/odhcpd/commit/e63a2837acfaf54c97a9db466bf485e9f7024fe8

PR #30, opened 2016-05-18
[URFC] include/version.mk package/base-files: Add standard os-release file

dfc-d:
/etc/os-release is the standard distribution release information
file, therefore add it (and image configuration options for
fields not previously present in LEDE).  After a release or
two the non-standard openwrt_release and openwrt_version
files could be removed (that is with this patch we consider
them deprecated in favour of the standard file, albeit with some
fields not present in other distributions).

Works locally, however the lldpd part should probably get wider testing, and I'm not sure it's of interest anyway.


lynxis:
Hi Daniel,
I like the idea of using /etc/os-release, but I don't think we can remove /etc/openwrt_* soon. But deprecated it for sure.

dfc-d:
On 16-05-17 09:33 PM, Alexander Couzens wrote:
> Hi Daniel,
> I like the idea of using /etc/os-release, but I don't think we can
> remove /etc/openwrt_* soon. But deprecated it for sure.

If I removed then I screwed up the patch - I had intended to leave them
in place for now in case of other dependencies (but deprecate for
removal in a release or two).

Regards,

Daniel

dfc-d:
Just realized I pushed PR from wrong tree; pushed actual version I meant to submit

dfc-d:
One thing I'm thinking is moved the non-standard fields (like TAINTS and REVISION) to device_info or keep openwrt_release for fields that are  non-standard (I'm not sure if having non-standard fields will break any third party uses or not as lldpd is currently the only one I really am aware of; mostly it's likely scripts like my personal provisioning scripts that use /etc/os-release).

blogic:
please also split the lldp patc into a separate commit

dfc-d:
Split into two commits.

jow-:
Once we reworked LuCI to use `/etc/os-release` I see no reason to not drop `/etc/openwrt_*`. In retrospect, it was a unlucky decision anyway to incorperate the project/distribution name into the release file.

jow-:
Regarding the nonstandard fields; I found the following quote in the man page:

> Operating system vendors may extend the file format and introduce new fields. It is highly 
> recommended to prefix new fields with an OS specific name in order to avoid name clashes. 
> Applications reading this file must ignore unknown fields. Example: 
> "DEBIAN_BTS="debbugs://bugs.debian.org/""

So we can just use `LEDE_TAINTS` and `LEDE_REVISION` or similar.

dfc-d:
On 16-05-18 08:13 AM, Jo-Philipp Wich wrote:
> Regarding the nonstandard fields; I found the following quote in the man
> page:
> 
>     Operating system vendors may extend the file format and introduce
>     new fields. It is highly
>     recommended to prefix new fields with an OS specific name in order
>     to avoid name clashes.
>     Applications reading this file must ignore unknown fields. Example:
>     "DEBIAN_BTS="debbugs://bugs.debian.org/""
> 
> So we can just use |LEDE_TAINTS| and |LEDE_REVISION| or similar.

I also noticed on that man page the BUILD_ID field which could be used
instead of LEDE_REVISION, now that I've found it.

Regards,

Daniel

dfc-d:
On 16-05-18 08:31 AM, Daniel Curran-Dickinson wrote:
> On 16-05-18 08:13 AM, Jo-Philipp Wich wrote:
>> Regarding the nonstandard fields; I found the following quote in the man
>> page:
>>
>>     Operating system vendors may extend the file format and introduce
>>     new fields. It is highly
>>     recommended to prefix new fields with an OS specific name in order
>>     to avoid name clashes.
>>     Applications reading this file must ignore unknown fields. Example:
>>     "DEBIAN_BTS="debbugs://bugs.debian.org/""
>>
>> So we can just use |LEDE_TAINTS| and |LEDE_REVISION| or similar.
> 
> I also noticed on that man page the BUILD_ID field which could be used
> instead of LEDE_REVISION, now that I've found it.

Also, unless we think LuCI could benefit from BUG_URL and SUPPORT_URL
that don't actually need to be added as they are optional.

Regards,

Daniel

dfc-d:
I've updated with the following:

* /etc/os-release is a relative symlink to ../usr/lib/os-release per standard
* Added (with LEDE_ prefix) the fields from device_info, which could also be removed.
* BUILD_ID instead of REVISION
* LEDE_BOARD, LEDE_TAINTS for the corresponding fields from openwrt_release
* VERSION_ID is lowercase per standard
* ID_LIKE is now 'lede openwrt' to reflect the fact that the build is based  first on  (or is) lede and then on openwrt (since lede is a fork of openwrt).

jow-:
A side effect of removing the lldpd patch is that single quotes in /etc/os-release are not supported anymore so you must double quote the fields.

include/version.mk
```diff
@@ -43,6 +45,13 @@ VERSION_MANUFACTURER:=$(if $(VERSION_MANUFACTURER),$(VERSION_MANUFACTURER),LEDE)
 VERSION_MANUFACTURER_URL:=$(call qstrip_escape,$(CONFIG_VERSION_MANUFACTURER_URL))
 VERSION_MANUFACTURER_URL:=$(if $(VERSION_MANUFACTURER_URL),$(VERSION_MANUFACTURER_URL),http://www.lede-project.org/)
 
+VERSION_BUG_URL:=$(call qstrip_escape,$(CONFIG_VERSION_BUG_URL))
+VERSION_BUG_URL:=$(if $(VERSION_BUG_URL),$(VERSION_BUG_URL),https://www.lede-project.org/development.html)
+
+VERSION_SUPPORT_URL:=$(call qstrip_escape,$(CONFIG_VERSION_SUPPORT_URL))
+VERSION_SUPPORT_URL:=$(if $(VERSION_SUPPORT_URL),$(VERSION_SUPPORT_URL),http://lists.infradead.org/mailman/listinfo/lede-dev)
+
+VERSION_PRODUCT:=$(call qstrip_escape,$(CONFIG_VERSION_PRODUCT))
```
jow-: Your change duplicates this line.

dfc-d: Hi Jo,

Thanks for looking at this.  I think you've been doing more with RFC's
than I had intended - I called them RFC's because I was looking for
comments before going through the full build and test cycle with
something that wasn't what was wanted.

>> +
>> +VERSION_PRODUCT:=$(call qstrip_escape,$(CONFIG_VERSION_PRODUCT))
> 

Will push a fix.

dfc-d:
On 16-05-18 01:30 PM, Jo-Philipp Wich wrote:
> A side effect of removing the lldpd patch is that single quotes in
> /etc/os-release are not supported anymore so you must double quote the
> fields.

I had missed that one, thanks.

Regards,

Daniel

dfc-d:
Fixed, will let you know once I've tested (unless of course there are any other comments first).

hnyman:
I tried your patch but I get compile error due to a missing file.
```
/Openwrt/lede/staging_dir/host/bin/sed: can't read /Openwrt/lede/build_dir/target-mips_34kc_musl-1.1.14/linux-ar71xx_generic/base-files/ipkg-mips_34kc/base-files/var/lib/os-release: No such file or directory
```
Looking at the source, I think that you have mixed up /var and /usr:
Base-files Makefile wants to manipulate $(1)/var/lib/os-release while you later create the file as usr/lib/os-release

I don't also quite get the point of having both /usr/lib/os_release and also /etc/os_release file. Why not just /etc/os_release ?

dfc-d:
Marking as Untested Request for Comments as this is primarily for gathering comments about the concept and testing still needs to be (I will shortly be pushing a fix the above and some other issues I discovered while testing; lldpd is still untested).

dfc-d:
Oh, I actually see what happened here - this is a case where I had a working patch I thought was ready, but made changes based on comments and what I saw myself, and didn't test the changes and notify that the changes were untested.  Anyway it's now an in-progress URFC instead of ready-to-go.

dfc-d:
Pushed the fix for var vs usr error, but the lldpd part still needs verified.

hnyman:
The include/version.mk part works ok after the fix.

```
NAME="LEDE"
VERSION="r393, Reboot"
ID="lede"
ID_LIKE="lede openwrt"
PRETTY_NAME="LEDE Reboot r393"
VERSION_ID="r393"
HOME_URL="http://www.lede-project.org/"
BUG_URL="https://www.lede-project.org/development.html"
SUPPORT_URL="http://lists.infradead.org/mailman/listinfo/lede-dev"
BUILD_ID="r393"
LEDE_BOARD="ar71xx/generic"
LEDE_TAINTS="no-all busybox"
LEDE_DEVICE_MANUFACTURER="LEDE"
LEDE_DEVICE_MANUFACTURER_URL="http://www.lede-project.org/"
LEDE_DEVICE_PRODUCT="Generic"
LEDE_DEVICE_REVISION="v0"
LEDE_RELEASE="HEAD"
```
However, I am not quite happy with all the data contents. The rambling below may fall partially outside the core scope of this PR... :-(

Like I wrote in http://lists.infradead.org/pipermail/lede-dev/2016-May/000271.html I feel that the currently available items do not fit well into all build types:
* trunk snapshots/builds: Trunk codename, no release number, source revision
* branch builds before release: Branch codename, release number known but no release yet, source revision
* branch release builds: Branch codename, official release number, source revision
* branch builds after a release: Branch codename, last release number known + changes after it, source revision

Simulating a branch release leads into these items:
```
VERSION="16.05, Reboot"
PRETTY_NAME="Lede Reboot 16.05"
VERSION_ID="16.05"
BUILD_ID="r393"
LEDE_RELEASE="16.05"
```
Same trunk items again for comparison:
```
VERSION="r393, Reboot"
PRETTY_NAME="LEDE Reboot r393"
VERSION_ID="r393"
BUILD_ID="r393"
LEDE_RELEASE="HEAD"
```

Due to the handling of input for VERSION_NUMBER (either release number like '16.05' or revision like 'r333' as a fallback) and VERSION_CODE (either release number like '16.05' or 'HEAD' as a fallback)  the contents will not look the same in branch builds and trunk, and parsing of detailed version information from the fields can be difficult.

The displayed version data should look informative especially for private builds from both trunk and branches, as understanding the source revision of private builds is important for bug reports etc.

I feel that PRETTY_NAME should always (or never) contain the revision r393. Currently it is pretty hard to prettily combine data for displaying to the user in Luci, and thus I wrote https://github.com/openwrt/luci/commit/f1190e2d07f76735e0927704de7a5cf8b37998d0 to get the revision shown in branch builds.

I also wonder if we should somehow utilize the 'git describe': ``` reboot-393-gaf1e70b ``` ?

dfc-d:
I appreciate your feedback (this is exactly the kind of reason I did a URFC).  I thinks that we should version for trunk be the ${LAST_RELEASE}.99 or maybe 99.${LAST_RELEASE} which should be easy enough to parse to be useful.  For trunk, or for someone following the post-release stable branch (actually for releases too), BUILD_ID=${REVISION} which should be made available for troubleshooting.  Currently REVISION is based on the reboot tag (AIUI).  In fact, except for a patch I have in my own tree I don't think REVISION falls back to a HEAD SHA1 if reboot doesn't exist, but rather ending being "unknown"

    try_version || try_git || try_hg || REV="unknown"

    try_git() {
        git rev-parse --git-dir >/dev/null 2>&1 || return 1
        REV="$(git describe --match reboot | sed "s/reboot-\([0-9]*\)-.*/\1/g")"
        REV="${REV:+r$REV}"
        [ -n "$REV" ]
    }

I'm not sure it makes sense to display the REVISION to the user a normal 'overview' page (and I think even current full LuCI revision is rather ugly for a UI), although there should definitely be a way to get at that information for problem reports (like the proposed tools for pre-populating a bug tracker with data gathered via a LuCI app or page in admin-full, perhaps using initially from a shell script CLI users can use too)

I think maybe BUILD_ID should actually the HEAD SHA1, and there should be e.g. LEDE_REVISION which is based on the reboot tag (so one can easily see whether a build is stock or custom).

hnyman:
> I'm not sure it makes sense to display the REVISION to the user a normal 'overview' page (and I think even current full LuCI revision is rather ugly for a UI), although there should definitely be a way to get at that information for problem reports (like the proposed tools for pre-populating a bug tracker with data gathered via a LuCI app or page in admin-full, perhaps using initially from a shell script CLI

Having a rather detailed version string visible in Luci and easily available for copy-paste simplifies things. Even novice users find that easily, which I do not believe for data collection shell scripts... We should not over-complicate bug reporting by hiding version info into technical tools.

dfc-d:
First a clarification - the idea with the data collection script is for me to add a friendly LuCI frontend to it so that a user can copy and paste (e.g. for IRC) OR if submitting a problem ticket, having a bug tracker automatically populated with information from the script - the only reason for having the backend be cli is so that cli users aren't left in the cold (I'm thinking the script could have a --json flag that would make it convenient for Ajax, but cli users get more human friendly output).

I wouldn't reduce the visible information *until* that is available, but once it is I think it should become the preferred method (because presumably any such script would be designed with input from the devs who tell me or whoever writes the script what they actually think is relevant/important and therefore would want from it; this of course requires to the devs to actually give useful direction in this regard) of getting the information from the system for helping with troubleshooting, since it should be a 'one-stop-shop' for the things that are likely to relevant.

Obviously sometimes additional information will be needed, but having a baseline of of information (kind of like Ubuntu's apport attachments) that is available (although I like the CentOS thing of being able to edit and/or omit information if one has privacy/security concerns with giving the information).

PR #31, opened 2016-05-18
target/sdk: Allow to configure feeds in SDK

dfc-d:
We may want more feeds in SDK than in core, hence allow to configure feeds in SDK.


dfc-d:
Apparently I forgot to comment the commit

PR #33, opened 2016-05-18
Full versions vs. BusyBox

dfc-d:
Hi Felix, John,

Does this look more like what you are thinking?  If so I will test it before issuing a real PR.

A couple of notes:

1) Based on a grep of an ar71xx rootfs and of the package directory the only hard-coded references to busybox applets in /usr seems to be for ntpd and crond in their respecitve initscripts.
2) Based on that I'm wondering if makes sense to drop the /usr/bin/env and /usr/bin/awk symlinks too (since the only breakage might be in packages and third party scripts, not anything critical to device operation).

### DIFF
--- a/package/base-files/files/lib/upgrade/common.sh
+++ b/package/base-files/files/lib/upgrade/common.sh
@@ -49,11 +49,11 @@ supivot() { # <new_root> <old_root>
 run_ramfs() { # <command> [...]
 	install_bin /bin/busybox /bin/ash /bin/sh /bin/mount /bin/umount	\
 		/sbin/pivot_root /sbin/reboot /bin/sync /bin/dd	/bin/grep       \
-		/bin/cp /bin/mv /bin/tar /usr/bin/md5sum "/usr/bin/[" /bin/dd	\
-		/bin/vi /bin/ls /bin/cat /usr/bin/awk /usr/bin/hexdump		\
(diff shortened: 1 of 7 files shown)


blogic:
i like the idea but i think this should be merged post release. right now we want to stabilise things and this had the potential to break lots of stuff and cause unwanted side effects

dfc-d:
On 16-05-19 02:42 AM, John Crispin wrote:
> i like the idea but i think this should be merged post release. right
> now we want to stabilise things and this had the potential to break lots
> of stuff and cause unwanted side effects

Makes sense.

Regards,

Daniel

dfc-d:
FYI, while I agree to leave this until post-release because LuCI and some packages in the feeds are impacted, as far as core is concerned all hard-coded references have already been removed, except the paths to the daemons mentioned above, and in fact in core there is not a need for /usr/bin/env or /usr/bin/awk symlinks either.

I'm removing the [RFC] tag because it has been used successfully (including sysupgrade) by me (although obviously wider testing is needed, and as mentioned there are various non-core things that will break).

dfc-d:
Actually there is one place that there is hardcoding, but it is to the /bin:/sbin and that is in the openvpn compile options.  Just for reference.

jow-:
Great work, we should keep `/usr/bin/env` though as its the agreed-upon approach to portably write scripts targeting interpreters which might reside in different locations on different systems.

PR #34, opened 2016-05-18
[URFC] Patch samba vfs acl config options

dfc-d:
NOTE: This patch series has just been split up and is not yet tested; it is purely a request for comments before doing the heavy lifting of testing and debugging.

This patch series

1) Adds mDNS advertisements for Samba
2) Moves Samba under Network|Filesystems menu for menuconfig
3) Add ACL and XATTR support configuration option
4) Adds netatalk compatibility support configuration option

Any comments on improvements prior to submitting would be welcome.

### DIFF
--- a/package/network/services/samba36/Makefile
+++ b/package/network/services/samba36/Makefile
@@ -36,7 +36,7 @@ define Package/samba36-server
   SUBMENU:=Filesystem
   TITLE:=Samba 3.6 SMB/CIFS server
   URL:=http://www.samba.org/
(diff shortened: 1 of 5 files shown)


blogic:
lgtm, patc h3 would need a close look when posted after testing

nbd168:
netatalk support should not be enabled automatically if the netatalk package is selected. That will cause extra bloat on the samba package when built from the buildbot.

blogic:
nbd requested changes and there was no response after 3 week. please reopen when there is an update.

dfc-d:
I thought I had pushed those changes, but apparently not, will do that soon.

dfc-d:
Actually for this and the other one, I have the comments I was lookin for now, I'll only reopen once they're actually tested and ready now

PR #37, opened 2016-05-19
[URFC] Replace busybox versions of commands with full versions

dfc-d:
Hi,

Does this look more like what you'd like to see?  I had to update the script for Config-defaults.in and and Config-defaults.in because you can't select !SYMBOL which is what would be required to avoid needing to modify Config-defaults.

I've only tested the menuconfig part for now (I imagine there will be breakage in actual use due to the presence of hardcoded paths, hence I have marked the whole feature EXPERIMENTAL in the menuconfig).

### DIFF
--- a/package/base-files/files/lib/upgrade/common.sh
+++ b/package/base-files/files/lib/upgrade/common.sh
@@ -49,11 +49,11 @@ supivot() { # <new_root> <old_root>
 run_ramfs() { # <command> [...]
 	install_bin /bin/busybox /bin/ash /bin/sh /bin/mount /bin/umount	\
 		/sbin/pivot_root /sbin/reboot /bin/sync /bin/dd	/bin/grep       \
-		/bin/cp /bin/mv /bin/tar /usr/bin/md5sum "/usr/bin/[" /bin/dd	\
-		/bin/vi /bin/ls /bin/cat /usr/bin/awk /usr/bin/hexdump		\
-		/bin/sleep /bin/zcat /usr/bin/bzcat /usr/bin/printf /usr/bin/wc \
-		/bin/cut /usr/bin/printf /bin/sync /bin/mkdir /bin/rmdir	\
-		/bin/rm /usr/bin/basename /bin/kill /bin/chmod
+		/bin/cp /bin/mv /bin/tar /bin/md5sum "/bin/[" /bin/dd	\
+		/bin/vi /bin/ls /bin/cat /bin/awk /bin/hexdump		\
+		/bin/sleep /bin/zcat /bin/bzcat /bin/printf /bin/wc \
+		/bin/cut /bin/sync /bin/mkdir /bin/rmdir	\
+		/bin/rm /bin/basename /bin/kill /bin/chmod
 
 	install_bin /bin/uclient-fetch /bin/wget
 	install_bin /sbin/mtd
--- a/package/base-files/files/lib/upgrade/common.sh
+++ b/package/base-files/files/lib/upgrade/common.sh
@@ -23,10 +23,25 @@ install_bin() { # <file> [ <symlink> ... ]
 	install_file $files
 	shift
 	for link in "$@"; do {
-		dest="$RAM_ROOT/$link"
-		dir="$(dirname $dest)"
-		mkdir -p "$dir"
-		[ -f "$dest" ] || ln -s $src $dest
+		# Handle the case of using full versions of
+		# commands instead of busybox version; this
+		# only matters when busybox version is not
+		# compiled in (and hence /{s,}bin link doesn't
+		# exist)
+		if [ -L "$link" ]; then
+			dest="$RAM_ROOT/$link"
+			dir="$(dirname $dest)"
+			mkdir -p "$dir"
+			[ -f "$dest" ] || ln -s $src $dest
+		else
+			# We *only* have the full version, which
+			# is always in /usr/{s,}bin
+			$install_file /usr/$link
+			# For sysupgrade to work in case of full
+			# versions without busybox version we
+			# must ensure there are no hardcoded paths,
+			# so that PATH is always honoured
+		fi
 	}; done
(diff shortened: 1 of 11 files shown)


blogic:
your PR should explain what it does and what you changed to the previous version., right now i would need to spend a lot of time to read everything, then read what you posted yesterday, work out what changed etc pp...

dfc-d:
It's almost completely different.  I followed Felix's suggestion and created e.g. util-linux-busybox package that selects the util-linux commands for which the busybox versions are included by default, and in that package have a hidden config option HAS_UTIL_LINUX.  

In the busybox config I do two things:

In Config.in I depend on HAS_UTIL_LINUX for the config option BUSYBOX_USE_UTIL_LINUX, which selects util-linux-busybox when enabled

In addition I have modified the defaults creation script and run it, so that for the busybox commands for which there are full versions I make the building the busybox applet default n if BUSYBOX_USE_UTIL_LINUX (for ones from util-linux).

So if BUSYBOX_USE_FULL_VERSIONS config flag is set (default n and marked EXPERIMENTAL), all the various BUSYBOX_USE_xxx that are available in the currently installed packages (e.g. base and any installed from feeds) are selected which means that those busybox applets don't get included and the packages for the full versions are selected (and can't be deselected, except by turning of the appropriate BUSYBOX_USE_xxx option, so that busybox applet gets built).

I've done this for all the full versions I know about (although I have not yet created the corresponding patches on the e.g. coreutils in the packages feed, which means those applets will not have the option of BUSYBOX_USE_xxx yet).

dfc-d:
I've marked this as UFRC (Untested Request For Comments) because this is an untested (so far) patchset that I'd like to verify is headed in right direction for it's purposes before taking the time and effort to test and debug it.

I've created a corresponding PR (https://github.com/openwrt/packages/pull/2730) against the packages feed for the applicable packages.

package/base-files/files/lib/upgrade/common.sh
```diff
@@ -23,10 +23,25 @@ install_bin() { # <file> [ <symlink> ... ]
...
+		# exist)
+		if [ -L "$link" ]; then
+			dest="$RAM_ROOT/$link"
+			dir="$(dirname $dest)"
+			mkdir -p "$dir"
+			[ -f "$dest" ] || ln -s $src $dest
+		else
+			# We *only* have the full version, which
+			# is always in /usr/{s,}bin
+			$install_file /usr/$link
```
jow-: Looks like a typo.

package/utils/busybox/convert_defaults.pl
```diff
@@ -1,5 +1,106 @@
...
+
+	return \@newdefault;
+}
+
+sub skip_applet_if_prefer_full($$) {
+	my $applet = shift;
+	my $default = shift;
+
+	my %skip_hash = (
+	   "BUSYBOX_USE_BRIDGE", [ "BRCTL" ],
```
jow-: I wonder if we can somehow populate that from package metadata. Having this exception list here looks like a long term maintenance annoyance.

package/utils/util-linux/Makefile
```diff
@@ -35,6 +35,31 @@ define Package/util-linux/Default
   URL:=http://www.kernel.org/pub/linux/utils/util-linux/
 endef
 
+define Package/util-linux-busybox
+$(call Package/util-linux/Default)
+  TITLE:=Full util-linux commands instead of busybox replacements
```
jow-: A typo? Or is `util-linux-busybox` indeed replacing busybox applets? That'd be very confusing...

package/base-files/files/lib/upgrade/common.sh
```diff
@@ -23,10 +23,25 @@ install_bin() { # <file> [ <symlink> ... ]
...
-		dest="$RAM_ROOT/$link"
-		dir="$(dirname $dest)"
-		mkdir -p "$dir"
-		[ -f "$dest" ] || ln -s $src $dest
+		# Handle the case of using full versions of
+		# commands instead of busybox version; this
+		# only matters when busybox version is not
+		# compiled in (and hence /{s,}bin link doesn't
+		# exist)
+		if [ -L "$link" ]; then
```
jow-: I do not understand why we need this special casing?

package/utils/util-linux/Makefile
dfc-d: Combined with the busybox option above (marked experimental, default n), it allows that instead of compiling the image with busybox applets for which util-linux full versions are available, busybox will be compiled without the applets and only util-linux versions will be present.

package/base-files/files/lib/upgrade/common.sh
dfc-d: It's because it's for use with the above busybox option which make busybox not include the normal busybox applets and instead selects packages (which packages depends on which options you enable, and what packages are available as replacements in your build) with full versions.

package/base-files/files/lib/upgrade/common.sh
dfc-d: Doh! Yeah it is.

package/utils/busybox/convert_defaults.pl
dfc-d: The one issue I can with that (otherwise I think something like that could work, although I don't think the metadata would be enough; you'd need to scan install sections I think) is that it would require the tree to have the various replacement packages installed from feeds (for ones from packages).

package/utils/util-linux/Makefile
dfc-d: Without the option it installs the packages from util-linux for which applets are normally built and since /usr/{bin,sbin} is earlier in the PATH the full versions will be present, although the busybox versions will still be present in {/bin,/sbin}

dfc-d: It has a strict dependency on the busybox PR for getting rid of /usr for busybox.

PR #38, opened 2016-05-19
ramips: cleanup for Hame MPR-A2

obsy:
- disable all ethernet ports except port 0 on MPR-A2

Port 0 is the only ethernet port on this router, so disable all other PHYs in order to save power.

- don't use a VLAN for the single ethernet port of the MPR-A2

Like A5-V11, this router only has one ethernet port.


blogic:
pulled into my staging tree, thanks !

PR #43, opened 2016-05-22
Centralize setting of all version info to include/version.mk

hnyman:
It has been troublesome in the past that release version info has been set in three places: include/toplevel.mk, include/version.mk and /etc/banner. 

Centralize setting all version information to include/version.mk
* Set RELEASE env variable in include/version.mk instead of toplevel.mk.
  Stop exporting the variable.
* Remove hardcoded release name from /etc/banner. Set it dynamically along other version info

* Use "Reboot" as the release nickname, as that has been shown to users in /etc/banner since the LEDE branding commit fa69553900516944fc43cf7b377105ea1f6d3db0. So far there has been a deviation as the console banner has shown "Reboot" while Luci has shown "Designated Driver" based on /etc/openwrt_release that has contained "Designated Driver" from RELEASE. This patch applies "Reboot" to everywhere. No more Designated Driver. (Hopefully devs figure out a new naming scheme soon. Reboot sounds boring)

This PR fixes some of the issues that I listed in http://lists.infradead.org/pipermail/lede-dev/2016-May/000271.html 

**Explanations:**

* Other version options (release number, revision) at patched dynamically to /etc/banner, so do the same to the release nickname
* I made a search for the usage of the RELEASE env variable, but I did not find any other use than one line in version.mk, so I moved the variable definition there. However, I still used a separate variable for defining the name version.mk instead of just patching the line ``` VERSION_NICK:=$(if $(VERSION_NICK),$(VERSION_NICK),$(RELEASE)) ```, as a variable definition on its own line enables easier parsing of that option in a build script. (E.g. I embedd a small additional detailed version file in my build, and it is great to have the version name easily available for my build scripts already before the actual make process starts)
* I also stopped exporting the RELEASE variable, as I did not identify any usage of that. 
* Imagebuilder does its own magic for REVISION, but I did not notice any special usage of RELEASE. Both SDK and imagebuilder Makefiles include the version.mk right at the top, so the variable definition should be normally usable. (But I have not actually tested this change with the imagebuilder or SDK.)

**End result:**

After the changes the key files in a live router look like this:

/etc/banner:
```
     _________
    /        /\      _    ___ ___  ___
   /  LE    /  \    | |  | __|   \| __|
  /    DE  /    \   | |__| _|| |) | _|
 /________/  LE  \  |____|___|___/|___|                      lede-project.org
 \        \   DE /
  \    LE  \    /  -----------------------------------------------------------
   \  DE    \  /    Reboot (HEAD, r324)
    \________\/    -----------------------------------------------------------
```
/etc/openwrt_release:
```
DISTRIB_ID='LEDE'
DISTRIB_RELEASE='HEAD'
DISTRIB_REVISION='r324'
DISTRIB_CODENAME='reboot'
DISTRIB_TARGET='ar71xx/generic'
DISTRIB_DESCRIPTION='LEDE Reboot r324'
DISTRIB_TAINTS='no-all busybox'
```
Luci bottom banner:
```
Powered by LuCI Master (git-16.135.27427-6415c04) / LEDE Reboot r324 
```
Luci overview page:
```
System
Hostname	lede
Model	NETGEAR WNDR3800
Firmware Version	LEDE Reboot r324 / LuCI Master (git-16.135.27427-6415c04)
Kernel Version	4.4.11 
```


jow-:
LGTM +1

nbd168:
LGTM

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=eedfe6465ca176ff53bd4beda41f9371041c8053

PR #49, opened 2016-05-23
sdk: Allow to change ALL* package settings in SDK

dfc-d:
It is desirable to be able to use ./scripts/feeds install -a
when in the SDK without being forced to build a great number
of packages that are not actually wanted.

We therefore add the option of changing the various ALL package
build options so that we can default to packages not being built
unless we select them.

I find this quite useful and have tested it.

### DIFF
--- a/target/sdk/files/Config.in
+++ b/target/sdk/files/Config.in
@@ -1,3 +1,19 @@
+menu "Global build settings"
+
+	config ALL_NONSHARED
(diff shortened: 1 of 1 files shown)


jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=a232e84e4ad45248baef231713be0cbfc0e9f1fa

PR #50, opened 2016-05-23
ramips: Add status led for Xiaomi MiWiFi Nano

obsy:

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -186,7 +186,8 @@ miniembplug)
 	set_wifi_led "$board:red:wlan"
 	set_usb_led "$board:green:mobile"
 	;;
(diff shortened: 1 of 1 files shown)


blogic:
merged, thanks !

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -186,7 +186,8 @@ miniembplug)
 	set_wifi_led "$board:red:wlan"
 	set_usb_led "$board:green:mobile"
 	;;
-miwifi-mini)
+miwifi-mini|\
+miwifi-nano)
```
ssreekanth: I see that this was already handled (see 3 lines below). Why do you need this again here?

obsy: Yes, this is definitely mistake.

PR #56, opened 2016-05-25
dnsmasq: stop SIGHUP from enabling DNSSEC timechecks

kdarbyshirebryant:
This patch stops SIGHUP from enabling dnssec timechecks if disabled by
use of --dnssec-no-timecheck option.  --dnssec-timestamp continues to
work correctly, using the timestamp file as the arbiter of correct
time.

Enabling dnssec timechecks now requires restarting dnsmasq without
the --dnssec-no-timecheck configuration option.

This closes an unintended side-effect of accidentally enabling
time checks when using --dnssec-timestamp and the system time does not
yet match Internet time, a situation which if using
--dnssec-check-unsigned
would lead to external name resolution via dnsmasq stopping entirely.

### DIFF
--- a/package/network/services/dnsmasq/patches/220-dnssec-disable-timecheck-hup.patch
+++ b/package/network/services/dnsmasq/patches/220-dnssec-disable-timecheck-hup.patch
@@ -1,0 +1,25 @@
+This patch stops SIGHUP from enabling dnssec timechecks if disabled by
+use of --dnssec-no-timecheck option.  --dnssec-timestamp continues to
+work correctly, using the timestamp file as the arbiter of correct
+time.
+
+Enabling dnssec timechecks now requires restarting dnsmasq without
+the --dnssec-no-timecheck configuration option.
+
+This closes an unintended side-effect of accidentally enabling
+time checks when using --dnssec-timestamp and the system time does not
+yet match Internet time, a situation which if using --dnssec-check-unsigned
+would lead to external name resolution via dnsmasq stopping entirely.
+
+Signed-off-by: Kevin Darbyshire-Bryant <kevin@darbyshire-bryant.me.uk>
+--- a/src/dnsmasq.c
++++ b/src/dnsmasq.c
+@@ -1088,7 +1088,7 @@ static void sig_handler(int sig)
+       int event, errsave = errno;
+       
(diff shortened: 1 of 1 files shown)


ConorOG:
This is directly contrary to the documentation and intended behaviour.

kdarbyshirebryant:
It is not clear from the documentation that using '--dnssec-timestamp' also enables the SIGHUP method to indicated dnssec signatures should be time checked.  The point being a SIGHUP could have the unintended consequence of enabling dnssec time checking before ntpd has the chance to set the clock correctly.  Ntpd would then be in the position of not being able to resolve ntp server names and hence never be able to set the clock.

kdarbyshirebryant:
I should add that LEDE uses the '--dnssec-timestamp' method of time validation and not '--dnssec-no-timecheck'

ConorOG:
Is it perhaps worth testing the other method?

I see the standard busybox ntpd is set to call ntpd-hotplug script on time updates.

It would be easy to get that to sighup dnsmasq, which would be started with --dnssec-no-timecheck.

kdarbyshirebryant:
It's certainly an option but that quickly gets messy too: What happens if dnsmasq is restarted (crashed)?  Who signals it to say the time is actually ok after it  has restarted?  You end up implementing an external 'time ok' flag system and change how dnsmasq starts accordingly.  That's why I thought a timestamp file handled by dnsmasq itself was a better idea and suggested it to Simon.  Simon did all the heavy lifting and code.  But I may well drop this PR as I've thought of a better way to implement the change.  I need to practice my C as well, so hopefully a patch upstream soon :-)

ConorOG:
root@OpenWrt:/etc/init.d# ntpd --help
BusyBox v1.23.2 (2016-02-10 17:23:04 GMT) multi-call binary.
...
	-S PROG	Run PROG after stepping time, stratum change, and every 11 mins

kdarbyshirebryant:
Hmmm.  It looks so easy.  Why hasn't it been done before?

jow-:
Not sure if other NTP implementations provide similar hooks so if you start to rely on it then make sure that it all still works somehow without.

ConorOG:
Where is the troublesome SIGHUP coming from?

I thought originally system updating was involved in all this?

PR #57, opened 2016-05-25
[URFC] Prevent reconfiguration of already built packages in SDK

dfc-d:
I seem to have accidentally closed this when I updated the branch.  It's still an Untested Request For Comments, although I've now got some testing done (hence the update with some fixes).

Anyway, this patchset is designed to ensure that packages that are built during the core build don't get built with different options during and SDK build, creating a situation where it's non-deterministic what functionality you will get (and which may cause runtime errors due to conflicting options).


FreifunkUFO:
i so would not be able to (re-)compile most packages any more, because they were built on the core build yet?
sometimes SDK may be used to rebuild packages with more or other options, so thats forbidden now?

f.i. on freifunk: being compatible with freifunk-gluon i sometimes need to recompile fastd additionally with the server-option or alfred with the unzip-method. both packets are on our "core build"

dfc-d:
On Thu, 2016-06-09 at 06:58 -0700, FreifunkUFO wrote:
> i so would not be able to (re-)compile most packages any more, because
> they were built on the core build yet?

Actually the way it works is that when generating the SDK a file
'ignored-packages.mk' is generated and placed in $(TOPDIR) of the SDK.
If you remove a package from ignored-packages.mk then it's configuration
can be chagned.

> sometimes SDK may be used to rebuild packages with more or other
> options, so thats forbidden now?

The reason for implementing this is that changing options can change the
ABI of packages, which means that unless you have some means of ensure
that the package comes from the SDK build instead of going out and
grabbing the core build that you may or may not get packages that are
compatible with each other (think for example what happens if the SDK
repo has a temporary connectivity issue but the repo does not; you could
end up with a broken package set on the device).
> 
> f.i. on freifunk: being compatible with freifunk-gluon i sometimes
> need to recompile fastd additionally with the server-option or alfred
> with the unzip-method. both packets are on our "core build"

One thing I'm wondering (have to investigate) is whether the fact that
the SDK at least repackages core packages during it's build anyway, if
it's possible to use *only* the SDK repo and simply not use the core
repo; that would eliminate the need for this patch, and it would be easy
enough to ensure that images built with a combination of SDK + IB didn't
include the core repo by default, through a modification of patch I'm
working on for handling creating the opkg default feeds config when
building the SDK (while leaving the core feeds config untouched).
> 
Regards,

Daniel 
>

PR #61, opened 2016-05-27
My current fixes for the DGN3500, other AR9, other Lantiq, and other MIPS

danielg4:
So far, I have tested this only on a DGN3500 without connecting the ADSL2. Mohammed Berdai has agreed to run additional tests on additional devices under additional conditions tomorrow, on the 28th.


danielg4:
@blogic Patchset updated after NAK

PR #62, opened 2016-05-28
ramips:add widora support(MT7688AN,16M FLASH,128MB DRAM,1WAN,1T1R,1SD…

widora:
Widora and linkit7688 are very similar in configuration except for FLASH capacity,Yet another open source hardware by me.thanks.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -45,6 +45,7 @@ ramips_setup_interfaces()
 	ht-tm02|\
 	linkits7688 | \
 	linkits7688d | \
+	widora | \
 	m2m|\
 	microwrt|\
 	mpr-a2|\
@@ -288,7 +289,8 @@ ramips_setup_macs()
 		wan_mac=$(macaddr_add "$lan_mac" 1)
 		;;
 	linkits7688 | \
-	linkits7688d)
+	linkits7688d| \
+	widora)
 		wan_mac=$(mtd_get_mac_binary factory 4)
 		lan_mac=$(mtd_get_mac_binary factory 46)
 		;;
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -124,7 +124,8 @@ get_status_led() {
 		status_led="$board:blue:system"
 		;;
 	linkits7688| \
-	linkits7688d)
+	linkits7688d| \
+	wiora)
 		[ "$1" = "upgrade" ] && status_led="mediatek:orange:wifi"
 		;;
 	m2m)
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -238,6 +238,9 @@ ramips_board_detect() {
 			name="linkits7688"
 		fi
 		;;
+	*"Widora NEO")
+		name="widora"
+		;;
 	*"Memory 2 Move")
 		name="m2m"
 		;;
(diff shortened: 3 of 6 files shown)


target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -45,6 +45,7 @@ ramips_setup_interfaces()
 	ht-tm02|\
 	linkits7688 | \
 	linkits7688d | \
+	widora | \
```
pepe2k: Please, keep boards in alphabetical order and use "|\" instead of " | \".

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -124,7 +124,8 @@ get_status_led() {
 		status_led="$board:blue:system"
 		;;
 	linkits7688| \
-	linkits7688d)
+	linkits7688d| \
+	wiora)
```
pepe2k: Typo - should be "widora".

target/linux/ramips/base-files/lib/ramips.sh
```diff
@@ -238,6 +238,9 @@ ramips_board_detect() {
 			name="linkits7688"
 		fi
 		;;
+	*"Widora NEO")
+		name="widora"
```
pepe2k: Are you sure about the board name? Why not use ex. "widora-neo"? That would save you potential problems in future if you decide to make other board named ex. "Widora XYZ".

target/linux/ramips/base-files/lib/upgrade/platform.sh
```diff
@@ -72,6 +72,7 @@ platform_check_image() {
 	jhr-n926r|\
 	linkits7688|\
 	linkits7688d|\
+	widora|\
```
pepe2k: Please, keep boards in alphabetical order.

target/linux/ramips/dts/WIDORA.dts
```diff
@@ -0,0 +1,170 @@
...
+		device_type = "memory";
+		reg = <0x0 0x8000000>;
+	};
+
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		wifi {
+			label = "mediatek:orange:wifi";
```
pepe2k: Please, follow general convention and use labels with board name, like in many others DTS files.

target/linux/ramips/image/mt7688.mk
```diff
@@ -4,7 +4,7 @@
 
 define Device/LinkIt7688
   DTS := LINKIT7688
-  IMAGE_SIZE := $(ralink_default_fw_size_32M)
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
```
pepe2k: Why are you changing this?

target/linux/ramips/base-files/lib/ramips.sh
pepe2k: I was right, I just found Widora BIT board.

target/linux/ramips/base-files/etc/board.d/02_network
widora: oh,It's my problem

target/linux/ramips/base-files/lib/ramips.sh
widora: I will change to Widora-NEO,thx

target/linux/ramips/base-files/lib/upgrade/platform.sh
widora: Oh,It's my problem.

target/linux/ramips/dts/WIDORA.dts
widora: Oh,yes

target/linux/ramips/image/mt7688.mk
widora: typo,It's my problem,T.T

PR #64, opened 2016-05-28
TL-WR841N v11: update LEDs

Borromini:
The v11 has some extra LEDs compared to the v9/v10, this adds support for them into LEDE.

Split out into a separate machine and updated the affected scripts as requested.

Signed-off by Stijn Segers <francesco.borromini@inventati.org>


Borromini:
Already had a chat with lynxis on IRC about this, he thought it might be better to create a new machine id and a new setup() routine. I added it to the v9 code since the 841N v8 machine code does something similar for the TL-MR3420.

jow-:
LGTM

Noltari:
LGTM

NeoRaider:
The `/etc/diag.sh` entry is still missing.

Borromini:
Fixed the missing diag.sh entry, updated the pull request.

NeoRaider:
I've found one more issue: You set active_low to 0 for the power led, but that would mean the LED is off when the system thinks it is on, and vice-versa.

Instead, you should use the power led instead of qss in `/etc/diag.sh` to display the status, then you'll get the expected behaviour.

Also, I'd recommend naming the led `tp-link:green:system` instead of `tp-link:green:power`, as most machine files use that name.

Borromini:
The person who tested said 0 set the power LED to green, 1 sets it to off. Shouldn't it be active_low = 0 then?

NeoRaider:
No, "active_low = 1" means that the LED is on when the GPIO is low (i.e. set to 0). This is generally the case for all TP-Link hardware I know.

Borromini:
I asked guerby to test (I've updated the pull request as per your recommendations, except for the active_low value) and leave a message here so I can hopefully make the final adaptations needed.

Thanks for your help already NeoRaider!

guerby:
Tested the patch on 841n v11.1 : power led blinks during boot then goes off. LAN/WAN/wifi LED work. "lock" LED is off.

Borromini:
Thanks, will fix the patch this evening, at work now. Did you get the lock LED to work when testing the GPIOs? I don't see that in your e-mail (apologies if I overlooked).

guerby:
I could not get the "lock" LED to light up, GPIO 3 is taken. Is there another way to trigger this LED?

NeoRaider:
The code registers the LED as `tp-link:green:qss`, you can set it by writing 0 or 255 to `/sys/class/leds/tp-link:green:qss/brightness`.

guerby:
I confirm with latest patch that the qss LED works fine using NeoRaider commands :)

Borromini:
Ok, thanks. I've updated the patch so everything NeoRaider pointed out should be fixed. Do i need to change anything for tp-link:green:qss or is this something that userspace programs should handle?

If not, I think this can go in (with the blessing of the overlords of course ;) ).

NeoRaider:
`tp-link:green:qss` is not used for anything by default, so you don't need to do anything with it. I'll take this through my staging tree.

NeoRaider:
The commit is missing a Signed-off-by.

Borromini:
I've added it to my first comment, is that OK? Can't find anywhere else to do it.

NeoRaider:
The Signed-off-by should be at the bottom of your commit message (the PR doesn't really need one). git can automatically add this when you commit using the `-s` option.

Well, that's for future contributions, I'll just copy your Signed-off-by from the PR to the commit message this time :)

NeoRaider:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/neoraider/staging.git;a=commit;h=d2a91f9853a9bfb47cb5d6e3078372f3f85557c5

PR #80, opened 2016-06-01
ramips: Netgear EX2700 image fixes

jclehner:
This PR addresses two issues with the Netgear EX2700 images:

1. Calling `cat foo bar | dd ... conv=sync` *may* cause unexpected results, depending on the version of `dd`. Extraneous padding will render the image unbootable.

2. Rebrand the fakeroot image (`OpenWrt fakeroot` -> `LEDE fakeroot`)


jow-:
LGTM. @blogic @nbd168 ?

blogic:
merged into my staging tree

PR #81, opened 2016-06-01
ipq806x: C2600, Fix system LED, Add support C2600v1.1 to kernel 3.18

heinzek:
Change power led to system led.


jow-:
@heinzek - your PR appears to contain a bunch of random commits "Delete 806-ARM-qcom-dts-c2600-Change-power-led-to-system-led.patch" appears to undo what " ipq806x: c2600 Fix system LED" did (which appears to be present twice) and "Add support for Macronix mx25u25635f, used in C2600 v1.1" is completely unrelated to the subject.

Can you try to do a cleanly rebased PR with the relevant commits squashed into one?

blogic:
i doubt that these patches actually work. 

the new script is patched into the kernel tree and the dts change looks like it was done by hand and the line offsets were not changed.

i will manually rework this and apply it to my staging tree

@heinzek please provide a valid SoB line so we can appply the fix after rebasing it.

blogic:
@heinzek 

 i added a patch to my staging tree --> https://git.lede-project.org/lede/blogic/staging.git

please test if it is correct and let me know so i can push it to source.git

heinzek:
Firmware is booting correctly
System led is working.
Bootlog: http://pasted.co/9ec48413
Video: https://youtu.be/gVh4nW9WEz8

Thanks for adding this fixes.

blogic:
never been sent a youtube video as a report :-) thanks for etsting

PR #82, opened 2016-06-01
brcm63xx: update to linux 4.4 and switch to new profile definitions system

Noltari:
CC @KanjiMonster


blogic:
@Noltari :-)

Noltari:
Added another commit to remove obsolete scripts causing the following warnings: https://gist.github.com/Noltari/6d20d8bcbe236caf516c0a73f5477d00
Also added support for inverting ath9k LED polarity, needed on the HG556a.

target/linux/brcm63xx/Makefile
```diff
@@ -11,7 +12,7 @@ BOARD:=brcm63xx
 BOARDNAME:=Broadcom BCM63xx
 SUBTARGETS:=generic smp
 FEATURES:=squashfs usb atm pci pcmcia usbgadget
-KERNEL_PATCHVER:=4.1
+KERNEL_PATCHVER:=4.4
 MAINTAINER:=Jonas Gorski <jogo@openwrt.org>
 
 include $(INCLUDE_DIR)/target.mk
```
KanjiMonster: (because github doesn't allow commenting on the commit log):

You are doing three things here:
a) switch to 4.4
b) drop 4.1 support
c) refresh config-4.4

Please split it accordingly, and add commit logs (well, no need for the refresh one). The switch to 4.4 should say if you tested it.

target/linux/brcm63xx/profiles/huawei.mk
```diff
@@ -1,65 +0,0 @@
...
-endef
-$(eval $(call Profile,HG553))
-
-define Profile/HG556a_AB
-  NAME:=Huawei EchoLife HG556a (version A/B - Atheros)
-  PACKAGES:=kmod-ath9k wpad-mini \
-	kmod-usb2 kmod-usb-ohci kmod-ledtrig-usbdev
-endef
-define Profile/HG556a_AB/Description
-  Package set optimized for Huawei HG556a version A/B (Atheros).
```
KanjiMonster: What happens to the descriptions? Most are trivial and are "Package set for device foo", but e.g. the Atheros/Ralink information for 556 Rev a/b/c would be good to keep.

target/linux/brcm63xx/base-files/lib/preinit/05_init_interfaces_brcm63xx
```diff
@@ -4,45 +4,33 @@
...
+	td-w8900gb |\
+	usr9108 |\
+	v2500v_bb)
 		ifname=eth1
 		;;
+	*)
+		ifname=eth0
+		;;
 	esac
 }
```
KanjiMonster: How does this "fix" this? It changes the "default" from eth1 to eth0, which might be sensible, but it doesn't explain this change at all in the commit log.

target/linux/brcm63xx/patches-4.4/421-BCM63XX-add-led-pin-for-ath9k.patch
```diff
@@ -44,6 +46,6 @@
  
  void pci_enable_ath9k_fixup(unsigned slot, u32 offset,
 -	unsigned endian_check) __init;
-+	unsigned endian_check, int led_pin) __init;
++	unsigned endian_check, int led_pin, bool led_active_high) __init;
  
  #endif /* _PCI_ATH9K_FIXUP */
```
KanjiMonster: Please split this into two commits, one adding active led support, one making hg556a use it. Bonus points for a third commit adding it for dsl-274x1-f1, which needs it as well.

target/linux/brcm63xx/Makefile
Noltari: ACK

target/linux/brcm63xx/profiles/huawei.mk
Noltari: I don't know if there's a way to add a description with the new profile definition system, but I will check if that's possible...

target/linux/brcm63xx/base-files/lib/preinit/05_init_interfaces_brcm63xx
Noltari: This sets the default ifname to eth0 and only sets it to eth1 for the specific devices that need it.
(will add this to the commit description)

target/linux/brcm63xx/patches-4.4/421-BCM63XX-add-led-pin-for-ath9k.patch
Noltari: ACK

Noltari:
Patches updated with @KanjiMonster's feedback

PR #85, opened 2016-06-04
Remove some genericfiles

BWhitten:
In the current kernels brcmfmac-sdio.h and ath9k_platform.h exist but are overwritten with target/linux/generic/files.

Would it be better to patch the existing files and not overwrite?
I merged the diffs into one patch per kernel ver but it could equally be a series.


jow-:
LGTM - @nbd168 ?

nbd168:
The brcmfmac-sdio.h removal makes sense, but moving ath9k_platform.h to kernel patches makes maintenance harder wrt. mac80211 updates for no good reason.

blogic:
@BWhitten can you update your PR to address felix's comments ?

BWhitten:
Certainly, that should have done it.

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=f945797396ef17a4755bbde15286b75a87ae540c

PR #94, opened 2016-06-06
lantiq-xrx200: Fritzbox WLAN 3370 - Unhardcode MAC-Address

LipkeGu:
This Pr removes the hardcoded MAC-Address in the Fritzbox 3370 dts file.

signed-off-by: Guido Lipke <lipkegu@gmail.com>

### DIFF
--- a/target/linux/lantiq/dts/FRITZ3370.dts
+++ b/target/linux/lantiq/dts/FRITZ3370.dts
@@ -220,7 +220,6 @@
 		#address-cells = <1>;
 		#size-cells = <0>;
 		reg = <0>;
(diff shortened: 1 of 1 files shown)


blogic:
this will lead to having random macs. i think this should not be merged until we can read the macs from tffs2

LipkeGu:
```
<mkresin> lipkegu: you can remove the mac-address = [ 00 11 22 33 44 55 ] line from the dts as well
```

blogic:
that does not change the fact that this should be done in combination with a tffs2 patch to set the mac.

PR #95, opened 2016-06-06
lantiq: Fritzbox WLAN 7320 - Unhardcode MAC-Address

LipkeGu:
This PR removes the hardcoded MAC-Address for the PHY.

signed-off-by: Guido Lipke <lipkegu@gmail.com>

### DIFF
--- a/target/linux/lantiq/dts/FRITZ7320.dts
+++ b/target/linux/lantiq/dts/FRITZ7320.dts
@@ -88,7 +88,6 @@
 
 		etop@E180000 {
 			phy-mode = "mii";
-			mac-address = [ 00 11 22 33 44 55 ];
 		};
 
 		ifxhcd@E101000 {


blogic:
this will lead to having random macs. i think this should not be merged until we can read the macs from tffs2

LipkeGu:
```
<mkresin> lipkegu: you can remove the mac-address = [ 00 11 22 33 44 55 ] line from the dts as well
```

blogic:
that does not change the fact that this should be done in combination with a tffs2 patch to set the mac.

LipkeGu:
i wrote this also that it can produce problems with dhcp and routing :| because on every reboot the mac gets a new MAC. the tffs support will be needed for each box, as far i know. in the tffs are more values stored as the MACs

LipkeGu:
@i can upload the gpl sources for the avm tffs driver so that you can take a look :)

LipkeGu:
https://github.com/LipkeGu/avm-gpl/tree/master/drivers/char/tffs-2.0

blogic:
i will close this PR and the other as it is more a feature report/wishlist request.

PR #100, opened 2016-06-08
ramips / mediatek audio related patches

igvtee:
This patch sets focus on audio function on ramips / mediatek.
It includes updates on i2c / i2s / gdma (hsdma on mt7621) drivers.
I use simple-card as ASOC platform drivers. so we only need to modify
dts file to bind different codecs. verified on below two boards.
DCH-M225  mt7620 + wm8960
DuZun-DM06  mt7628 + wm8960

but need following command to correctly init codec path and volume.
`alsactl init`
`amixer sset "Left Output Mixer PCM" on`
`amixer sset "Right Output Mixer PCM" on`
use speaker-test and mpg123 to verify headphone output function is ok
 
TODO:
- now use mediatek refclk function to provide MCLK for wm8960 codec.
  but the refclk clock speed is hardcode now. only support output 12Mhz clock.
  should use clock framework to rewrite it.

- i also try to use madplay to play mp3. but it seems have mmap problems.
  can't work now.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -172,6 +172,10 @@ ramips_setup_interfaces()
 	wrh-300cr)
 		ucidef_set_interface_lan "eth0"
 		;;
+	duzun-dm06)
+		ucidef_add_switch "switch0" \
+			"1:lan" "0:wan" "6@eth0"
+		;;
 	e1700|\
 	mt7620a_mt7530)
 		ucidef_add_switch "switch1" \
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -151,6 +151,9 @@ ramips_board_detect() {
 	*"Dovado Tiny AC")
 		name="tiny-ac"
 		;;
+	*"DuZun DM06")
+		name="duzun-dm06"
+		;;
 	*"E1700")
 		name="e1700"
 		;;
--- a/target/linux/ramips/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/base-files/lib/upgrade/platform.sh
@@ -50,6 +50,7 @@ platform_check_image() {
 	dir-620-a1|\
 	dir-620-d1|\
 	dir-810l|\
+	duzun-dm06|\
 	e1700|\
 	esr-9753|\
 	ex2700|\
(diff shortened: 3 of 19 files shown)


blogic:
oh nice. thanks a lot i will look at this in detail tomorrow !

blogic:
Hi,

some comments

* replacing i2c driver with sdk driver is not possible, please fix the new i2c driver instead
* dts prefix fro mediatek is mediatek and not mtk
* pcm clock is named pcmi2c should be pcmi2s ?
* please explain what you changed in the i2s driver, it is a huge change without explaination
* please explain what you changed in gdma driver 
thanks !

igvtee:
Hi,

I fix the dts prefix to mediatek and change name to pcmi2s.

I just improve i2c driver not copy SDK driver. what i do to i2c is.
- add rt_i2c structure to store driver data
- rewrite read/write check function and add i2c error status check. so we don't need to wait until time out.
- add 10 bits address support. according to the data sheet i think it is possible. but i haven't verify it.
- The most important is start transfer only need once. otherwise it cause I2C_STARTERR status.
- add set i2c clock speed register by dts options "clock-frequency". not just hard code it.
- about mt7621 i2c driver. i just copy i2c-ralink.c and change register names. and the hardware 
  don't support error status. so i remove it. but the logic is the same.

gdma/hsdma drivers.
- fix compiler error. device_control operation not support now.
- add old chips support 8 channels. new chips support 16 channels.
- add mt7621 hsdma driver. data sheet says it has two channels. but after test only one channel.
- add memory to memory DMA support. i use dmatest kernel module to verify this function.
  on rt305x it will copy more data. on mt7621 only two channels can works at the same time.
  these two chips maybe have hardware bugs. because on other chips don't have these bugs.
- use tasklet to handle remaining dma requests.

i2s drivers.
- remove mt7620-wm8960.c use simple card and DTS to do it
- add old chips support
- add 12Mhz refclk setup. this is hard code. need use clock framework rewrite it.
- add interrupt error status support for debug. default disable it. because cause to many interrupts.
- add setup bclk suport not hard code it.
- add 24 bits support for mt7628. not verified.
- use regmap api to control registers
- add txdma-req/rxdma-req DTS info for DMA use.

blogic:
please add the description of what the patches do into the actual patches themselves and fold the pcmi2c fix and mtk->mediatek rename into the patches that introduce the incorrect naming.

blogic:
pulled into my staging tree, will look at the "alsactl/amixer" bit today

PR #103, opened 2016-06-09
openssl: tweaks to reduce size

dirkfeytons:
This set of patches introduces a couple of config options to leave out stuff from OpenSSL. Additionally the hardcoded configure options have been extended to omit some rarely used functionality and algorithms.
I'm not sure about the defaults for the introduced config options so please take a closer look at them.
In my case with these changes libcrypto shrunk by about 100K and libssl by about 22K.
Note that I'm introducing two (small) patches to OpenSSL but these have been sent upstream and will probably be included in the next 1.0.2 release.


dirkfeytons:
Please don't merge this yet. I'm processing some comments I received on the openwrt-devel list.

dirkfeytons:
Pull request has been updated.

neheb:
why is PSK enabled by default?

dirkfeytons:
@neheb: see the discussion on the openwrt-devel list: https://lists.openwrt.org/pipermail/openwrt-devel/2016-June/041752.html
The argument was basically backwards compatibility.

jow-:
Did you reach a consensus here wrt. mosquitto? I tend to ack this one.

dirkfeytons:
@jow-: I kept the default to 'y' because of the backwards compatibility argument. Personally I don't have a specific opinion; for us this argument is not relevant.
But I also sent another pull request in https://github.com/openwrt/packages/pull/2833 to update the mosquitto Makefile to depend on these new OpenSSL config options so a build that includes mosquitto will ensure OpenSSL is including the correct functionality.

karlp:
I would generally feel that if you're using the "big" ssl library, OpenSSL, then you should get all the parts of openssl by default.  even if that's "big"  If you are truly concerned about size, there's mbedtls, polarssl, cyassl/wolfssl, gnutls, libressl.  I'm against cutting out parts of the only big full featured library just because it's big and full featured.

champtar:
@karlp if there are users we can enable stuff again
I'm not a mosquitto user so I'll disable SRP and PSK in my own build, buildbot will still do full build, everyone is happy, LGTM

karlp:
In it's current form I have no objections.  I kinda feel that you're adding options that make a minor difference on the huge ssl lib, and more options are more things for people to screw up, but as long as the defaults are still fine, I can't really complain about it .)

dirkfeytons:
It's partly about the size but also about reducing potential attack surface.

nbd168:
pushed with modifications in r1093. I decided to re-enable bf (it's used in some openvpn configurations), and I removed the temporary unconditional addition of no-srp which was removed in the last commit.

champtar:
Hi all
@nbd168 can you push this to OpenWrt, if not @dirkfeytons can you resend the modified patchset to OpenWrt, so it's easier to maintain the packages feeds

hnyman:
Looks like some packages are still broken due to this partial disabling of openssl functionality. There are at least three packages (emailrelay, nail, pen) where buildbot complains about "SSLv3_method" at the LEDE buildbot, but the packages compile ok in the Openwrt buildbot with the more "full" openssl. That makes me to suspect this PR as the reason, as https://github.com/lede-project/source/commit/41da31ac2ca2306e9cff070f650f6012af753517 changed the handling of "no-ssl3-method" option.

Errors are like these: 
https://downloads.lede-project.org/snapshots/faillogs/mips_34kc/packages/emailrelay/ssl/compile.txt
```
ssl.o: In function `ssl_create_context':
ssl.c:(.text+0x4a): undefined reference to `SSLv3_method'
```
https://downloads.lede-project.org/snapshots/faillogs/mips_34kc/packages/nail/compile.txt
```
openssl.o: In function `ssl_open':
openssl.c:(.text+0x6fa): undefined reference to `SSLv3_client_method'
Makefile:92: recipe for target 'mailx' failed
```

https://downloads.lede-project.org/snapshots/faillogs/mips_34kc/packages/pen/compile.txt
```
ssl.o: In function `ssl_create_context':
ssl.c:(.text+0x4a): undefined reference to `SSLv3_method'
Makefile:434: recipe for target 'pen' failed
make[5]: *** [pen] Error 1

```

neheb:
Changing SSLv3_method to TLSv1_method would do the trick.

PR #105, opened 2016-06-09
update tools/sparse to support llvm 3.5.0

stsydow:
sparse-0.5.0 won't link against lvm 3.5.0 as  "llvm-config --libs"
does not include system libs like pthreads
This problem got fixed but not released yet, 
so we need to fetch from their git repo.

### DIFF
--- a/tools/sparse/Makefile
+++ b/tools/sparse/Makefile
@@ -5,14 +5,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=sparse
-PKG_VERSION:=0.5.0
+PKG_VERSION:=0.5-git40791b94
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=git://git.kernel.org/pub/scm/devel/sparse/sparse.git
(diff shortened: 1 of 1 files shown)


blogic:
keeping PKG_VERSION:=0.5.0 seems wrong

stsydow:
How should I call to make the update clear?
The following options come to my mind:
sparse-0.5.0+
sparse-0.5.0-rc1 - but there is no official version
or sparse-0.5.git-40791b94

jow-:
I usually use something like `<baseversion>+git-<shorthash>` e.g. `0.5+git-40791b94`

jow-:
@stsydow - I merged this PR into my staging tree and fixed it up:
 - squashed the two commits into one
 - rewrapped commit message
 - added signed-off-by
 - removed commented lines from makefile

See https://git.lede-project.org/?p=lede/jow/staging.git;a=commitdiff;h=b6e38eeb358d7ef71b140188e617719813c15e13 and tell me whether it is okay.

stsydow:
Thanks, it´s fine!

PR #108, opened 2016-06-10
Move at91 bootloader to packages

BWhitten:
I think the at91 is the only target which has bootloaders in image/, these couple of patches removes an old version of uboot and puts an up to date version in packages/boot.

@claudyus I'm not sure what to do with dfboot, should I move/remove as well? It could be replaced with at91bootstrap which I think is the successor.


blogic:
looks good, do you have time to see if the at91 v4.4 update in my staging tree works ?

blogic:
@BWhitten merged into my staging tree. i had to rebase it to work with my image building code changes. please test if it still works.

blogic:
@BWhitten updated my staging tree. please let me know if it is working for you so i can push it to source.git

BWhitten:
@blogic Will do this evening, thanks.

BWhitten:
So I can boot, couple of comments.
It generates "lede-at91-legacy-<DTB>-squashfs-uImage", this file is actually just the kernel+DTB no fs in there probably want to drop it from the name :)
I couldn't work out how to make board specific UBIFS with set parameters, I used to set <BOARD>_UBIFS_OPTS but this build appears to only want to use UBIFS_OPTS.

Thanks!

blogic:
@BWhitten ok, will look at the image name tomorrow and add a ubi example.

BWhitten:
Github mangled the comment, should be lede-at91-legacy-**DTB**-squashfs-uImage and **BOARD**_UBIFS_OPTS

That may make more sense :) cheers.

BWhitten:
@blogic Do you need to me to test anything for this PR?

jow-:
Integrated with 6638946 and 6b31c4c.

PR #110, opened 2016-06-12
[URFC] Add os-release distro info file to deprecate openwrtisms for this

dfc-d:
Seem to have accidentally closed the pull request, so reopen.  The lldpd part is untested, the rest works.

This patchset adds /usr/lib/os-release and the symlink from /etc/os-release (for third parties that don't check /usr/lib/os-release as they should per standard, when /etc/os-release doesn't exist).  BUILD_ID contains the revision VERSION_ID is the %v version.

An open issue is the question of making the output consistent for both release and snapshot builds (i.e. so that parsers don't have to deal with a revision vs. a version number) for the %v version.


jow-:
LGTM will merge it to my staging tree. I'll cherry pick the os-release changes first and follow up with the lldpd patch removal later to give the os-release stuff some time to settle in.

jow-:
Merged, https://git.lede-project.org/?p=lede/jow/staging.git;a=commitdiff;h=14f772730c651922e6d24e7a26d80c68dffd2bb8

dfc-d:
On Tue, 2016-06-21 at 05:39 -0700, Jo-Philipp Wich wrote:
> LGTM will merge it to my staging tree. I'll cherry pick the os-release
> changes first and follow up with the lldpd patch removal later to give
> the os-release stuff some time to settle in.
> 
Do we want to do something about the mentioned issue of version numbers
being a r-hash for non-release builds, vs an actual version number for
release builds (to help people who want to parse the VERSION_ID  in
scripts)? 
>

jow-:
Yes but thats a different scope (and a lot of opportunity for bikeshedding...)

I think build ID should be a full "git describe" output and version number will likely be a date designation (2016.06 or something)

PR #111, opened 2016-06-12
[URFC] Pull request full vs busybox

dfc-d:
Some PR got closed when updating me tree, so recreating.

This PR is mostly tested, despite the title (but does have some changes since last test, hence URFC for now), and I've been able to use busybox without applets being in /usr without issues.

By the time it's post-release time this should be fully tested and able to be integrated (along with corresponding packages feeds patches).


dfc-d:
Switching to new email & account for LEDE work, so will reopen PR from that account after release (since this and it's dependency on busybox in /bin have been slated for post-release).

PR #112, opened 2016-06-12
More OpenWrt -> LEDE renames

dfc-d:
Branding: Rename openwrt to lede where appropriate
(although probably some unnecessary changes were made due
to not looking all that closely and doing a relatively mindless
(but not automatic) search and replace).

Probably this commit is incomplete and also has some things
that need to be changed back, but this is a first cut.

Note that due to a removed file (a target base-files/config/system file), this is a rebased version of the patch and not what I last tested with.


blogic:
we need to look at the changes closely. i already noticed that the lzma-loader gets rebranded which is not ok

LipkeGu:
I can remember on Luci there is also some with OpenWRT flagged, maybe we should use a Variable which is holding the Distribution name

dfc-d:
1) Revert lzma-loader patch (the issue was probably the mistaken edit of the provenance of the loader).  I originally made those changes because some time ago my ar71xx devices weren't booting with an LEDE tree with partial renaming done; it is not clear exactly was the issues as there were changes in trunk between the initial test and the test with the renamed lzma-loader.  Once I build, I will report whether reverting the lzma-loader patch in it's entirety breaks booting or not.
2) In descriptive text for finding information from images/flash that could be either OpenWrt or LEDE, use 'OpenWrt/LEDE' to describe the discovered information.
3) Still prefer LEDE when referring to things like the 'LEDE tree', or 'building an LEDE image' in text docs.

jow-:
Is it really appropriate to replace "SlugOS"? Was that name simply carried over from the initial port to OpenWrt?

dfc-d:
On Sun, 2016-06-12 at 02:05 -0700, John Crispin wrote:
> we need to look at the changes closely. i already noticed that the
> lzma-loader gets rebranded which is not ok
> 
I presume the issue is needing to double-check attributions are not
inadvertently changed, not that it doesn't work (my ar71xx devices
booted fine with the renamed loader, in combination with the rest of the
changeset, so either my devices don't actually use lzma-loader, or it
works with the rename).

I'll go over changeset and look for possible misattribution errors,
arising from the changes.

Regards,

Daniel

dfc-d:
On Mon, 2016-06-13 at 02:27 -0700, Jo-Philipp Wich wrote:
> Is it really appropriate to replace "SlugOS"? Was that name simply
> carried over from the initial port to OpenWrt?
> 
IIRC Openwrt replaced SlugOS with OpenWrt; I just changed OpenWrt to
LEDE.

Regards,

Daniel 
>

blogic:
looks like this no longer applies additionally no one picked this up. reopen please when you have time again.

PR #113, opened 2016-06-12
Allow default package selection to be chosen by a config option rather than make variable

dfc-d:
Has been menuconfig's but not yet built (hence Untested Request for Comments).

Prior to this patch doing something that changed the DEFAULT_
package selections required removing $(TOPDIR)/tmp and
regenerating the metadata.  This patch fixes and and uses it
to make DEVICE_TYPE a configuration option (which allows
for different default package selection for e.g. router
vs NAS).  This should also make it easier to allow for
different build options depending on target device type.

This was motivated primarily by an issue I noticed when trying to create a minimal SDK and then use it to build packages for images generated by ImageBuilder with CONFIG_ALL turned off.  The lack of sane default packages meant manual package selection was necessary.

As it turns out the fix also allows to make DEVICE_TYPE more useful by allowing it to be a configuration option which can be influenced by the profile (or even device with the new multi profile work).

### DIFF
--- a/Config.in
+++ b/Config.in
@@ -19,6 +19,8 @@ source "target/Config.in"
 
 source "config/Config-images.in"
 
+source "config/Config-device.in"
+
 source "config/Config-build.in"
 
 source "config/Config-devel.in"
@@ -32,3 +34,5 @@ source "target/sdk/Config.in"
 source "target/toolchain/Config.in"
 
 source "tmp/.config-package.in"
+
+source "tmp/.config-defaults.in"
--- a/config/Config-device.in
+++ b/config/Config-device.in
@@ -1,0 +1,76 @@
+# Copyright (C) 2006-2013 OpenWrt.org
+# Copyright (C) 2016 LEDE Project
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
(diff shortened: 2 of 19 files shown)


dfc-d:
I've successfully used this to build images now.

nbd168:
The comment that I made in #114 also applies here.

dfc-d:
From the updated commit message:

Previously it was impossible to change the default package selection
from within menuconfig and friends.  This patch creates a KConfig
structure that allows selecting a config option (currently DEVICE_TYPE)
to select packages to be built by default.  This is different than selecting
packages via select in that defaults can be overridden by user.  Basically
there are two tiers: the DEVICE_TYPE selects which of the set of all
possible defaults is actually enabled as a default selection, and the usual
packages metadata make a package default to build if the default selection
is enabled.

It needs to be retested as I have tweaked out some unnecessary changes.

Even if DEVICE_TYPE goes away, for the purposes of building a minimal SDK (only the packages absolutely required to build an SDK) something like this is needed in order to allow it to be a menuconfig option.

I've reviewed the files that were changed and I don't see a way to meaningfully split up this patch unless you allow for intermediates steps with various degrees of brokenness.

PR #114, opened 2016-06-12
Allow to build a minimal SDK/IB

dfc-d:
This is a rebase of an accidentally closed PR which is based on #113.  It allows one to build an SDK consists only of the packages necessary to build the SDK/IB combo suitable for building all other required packages and creating the target images.

### DIFF
--- a/Config.in
+++ b/Config.in
@@ -19,6 +19,8 @@ source "target/Config.in"
 
 source "config/Config-images.in"
 
+source "config/Config-device.in"
+
 source "config/Config-build.in"
 
 source "config/Config-devel.in"
@@ -32,3 +34,5 @@ source "target/sdk/Config.in"
 source "target/toolchain/Config.in"
 
 source "tmp/.config-package.in"
+
+source "tmp/.config-defaults.in"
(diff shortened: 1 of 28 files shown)


dfc-d:
I've been successfully using to create minimal SDKs, build packages using the SDK, and IB to create images.

blogic:
@nbd168 comments ?

nbd168:
Review of this is made hard by the fact that you mix a lot of different changes into a single commit.
For a lot of changes the reasons you had for making them are simply not obvious.

Things like regenerating the metadata if defaults change can be achieved simply by adding an extra dependency for metadata scan.

For selecting the device type, I'd prefer having a set of reasonable metapackages (which can be installed via opkg) instead of further increasing the use of the DEVICE_TYPE stuff, which never really worked well in practice anyway.

What this has to do with building a minimal SDK/IB is not obvious to me either. Please start splitting up your stuff into smaller, semantically distinct commits and pull requests.

dfc-d:
The main focus of this patch (building the minimal package set needed to build the toolchain is incompatible with the DEVICE_TYPE removal patch and is deprecated by it.  There are some useful elements that can be managed as separate patches.

PR #115, opened 2016-06-12
ramips: Add support for Archer C50

heinzek:
Tested and working:

- Flashing from original webui using the sysupgrade image

- Sysupgrading from within OpenWrt.

- wifi 2.4GHz and 5GHz.

- wired lan/wan

- usb storage

Pending:

- WPS led & WAN green not work, WAN orange led blink when cable is connected to port LAN2.


blogic:
Thanks for the patch, I merged into my staging tree after a few tiny whitespace cleanups. i also added your SoB to each of the commits.

PR #119, opened 2016-06-14
dnsmasq: Change dnssec handling to use busybox ntpd hotplug mechanism

kdarbyshirebryant:
Change dnsmasq's dnssec time check handling to use time validity
indicated by ntpd.

As described really. Rip all the dnsmasq timestamp file handling out including the conditional file save as part of sysupgrade. Instead use the ntpd hotplug mechanism to create a 'time is valid' flag for dnsmasq to pick up when it is started. ntpd hotplug also restarts dnsmasq on first 'stratum' change.

dnsmasq is started with --dnssec-no-timecheck if time is not valid - as soon as time is valid then it is restarted without that flag. I'm not a fan of dnsmasq's SIGHUP handling to indicate time is valid and ideally I'd patch that out...I've been rejected on that a couple of times now already but.... ;-)

Tested on Archer C7. A better (more secure) mechanism than a file created in /tmp/status is required for the flag.

Will rebase & add SOB after the no doubt numerous rounds of tweaking :-)


jow-:
The approach looks sane to me, so LGTM. I did not perform any runtime testing though.

kdarbyshirebryant:
Thanks @jow-   I'd like some help/feedback on a more 'secure' way of generating this 'time is ok' status flag.

Also I'd like to enhance the dnsmasq init script to only check the time status file if 'ntp client' is enabled.  How can I check if ntp client is enabled?

ConorOG:
What's the problem with SIGHUP?

kdarbyshirebryant:
Here we go again.  As I've explained before:  My concern is that SIGHUPing dnsmasq with '--dnssec --dnssec-check-unsigned --dnssec-no-timecheck' with the time incorrectly set would lead to a situation where all dns lookups would fail.  ntp would then be unable to resolve names and thus never be able to set the time correctly.

I personally think it safer/better to start dnsmasq either in a 'check timestamp' or 'don't check timestamp' mode and not have dnsmasq respond to an external signal that switches from one to the other.

ConorOG:
You're not trusting dnsmasq to behave as it is documented to do? Test it. Find a bug there. Fix it.

Someone else will come along later and say "why not SIGHUP?" It's explicitly mentioned in the dnsmasq docs as the way to handle this. But now the reason is clear, at this point in time, attached to this commit.

I am merely considering the long term maintenance of code, which for an open source project is the burden of the community, long after someone pushes a commit.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -709,7 +706,7 @@ start_service() {
 
 	procd_add_jail dnsmasq ubus log
 	procd_add_jail_mount $CONFIGFILE $TRUSTANCHORSFILE $HOSTFILE /etc/passwd /etc/group /etc/TZ /dev/null /dev/urandom /etc/dnsmasq.conf /tmp/dnsmasq.d /tmp/resolv.conf.auto /etc/hosts /etc/ethers $EXTRA_MOUNT
-	procd_add_jail_mount_rw /var/run/dnsmasq/ /tmp/dhcp.leases $TIMESTAMPFILE
+	procd_add_jail_mount_rw /var/run/dnsmasq/ /tmp/dhcp.leases $TIMEVALIDFILE
```
champtar: TIMEVALIDFILE is not needed inside the jail

kdarbyshirebryant: Yes, good point!  Will amend.

kdarbyshirebryant:
Conor, I trust dnsmasq to behave, what I don't trust is some other moron signalling it to do something when they shouldn't and without realising the DNSSEC implications.  SIGHUP has quite a lot hanging on it already.

I get your point, I'd like to know that you at least acknowledge mine.  And you'll also (hopefully) note that this PR does NOT contain anything that changes dnsmasq's SIGHUP handling.

ConorOG:
Yes, SIGHUP does have a lot of other responsibilities. I'm happy that issue is now explicit.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -15,7 +15,7 @@ ADD_LOCAL_HOSTNAME=1
 CONFIGFILE="/var/etc/dnsmasq.conf"
 HOSTFILE="/tmp/hosts/dhcp"
 TRUSTANCHORSFILE="/usr/share/dnsmasq/trust-anchors.conf"
-TIMESTAMPFILE="/etc/dnsmasq.time"
+TIMEVALIDFILE=/var/state/dnsmasqsec
```
champtar: I've just checked one of my OpenWRT CC router, /var/state is world writable
Can someone double check on a recent LEDE image

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -627,11 +629,6 @@ start_service() {
 	mkdir -p /var/lib/misc
 	touch /tmp/dhcp.leases
 
-	if [ ! -f "$TIMESTAMPFILE" ]; then
-		touch "$TIMESTAMPFILE"
-		chown dnsmasq.dnsmasq "$TIMESTAMPFILE"
-	fi
-
```
champtar: Maybe we can leave a small cleanup (if exists then rm) for one release?

kdarbyshirebryant:
@champtar I believe the latest changes address both of your concerns.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -16,6 +16,7 @@ CONFIGFILE="/var/etc/dnsmasq.conf"
 HOSTFILE="/tmp/hosts/dhcp"
 TRUSTANCHORSFILE="/usr/share/dnsmasq/trust-anchors.conf"
 TIMESTAMPFILE="/etc/dnsmasq.time"
+TIMEVALIDFILE=/tmp/ntp.d/dnsmasqsec
```
champtar: I've just sent a patch on the ML to fix the /var/state is world writable,
can you put TIMEVALIDFILE where it was (/var/state/dnsmasqsec)

kdarbyshirebryant: just saw that go through.  Excellent!  Thank you.  Makes my life easier :-)

package/network/services/dnsmasq/files/dnsmasqsec.hotplug
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+TIMEVALIDFILE=/tmp/ntp.d/dnsmasqsec
+
+[ "$ACTION" = stratum ] || exit 0
```
hnyman: You might check for either stratum or step action. When I look at my own system log where I have enabled ntpd event logging, the step event comes first (when the time is initially set) and then a bit later the first stratum event to tell that the quality of the ntpd time has changed. However, there may be more than 30 seconds between the initial time set event and the first stratum event. It might be useful to restart dnsmasq already earlier.

Log extract from the last boot: 
```
root@lede:~# logread | grep ntpd
Thu Jun 16 11:01:40 2016 user.notice ntpd: Time set, stratum=16 interval=32 offset=3593.824949
Thu Jun 16 11:02:12 2016 user.notice ntpd: Stratum change, stratum=2 interval=32 offset=0.000069
```

champtar: In that case maybe only support step, see warning https://git.busybox.net/busybox/tree/networking/ntpd.c#n939

hnyman: > In that case maybe only support step

Checking only for step does not sound right, as based on the process description there might not be a step event if the time is right enough at ntpd startup (e.g. due to a real-time clock in the device). At least that is how I read the algorithm description: https://git.busybox.net/busybox/tree/networking/ntpd.c#n85

And real-life behaviour matches that. Restarting sysntpd creates only a stratum event:
```
Thu Jun 16 17:46:11 2016 user.notice root: RESTART NTPD
Thu Jun 16 17:46:15 2016 user.notice ntpd: Stratum change, stratum=2 interval=1 offset=0.004251
```

kdarbyshirebryant: Back to only checking stratum.  And that's where it stays.  It works.

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=fe395080d3d68da9d9ce96407d9c2a2c284ac33c

PR #121, opened 2016-06-15
Replace Mangle Bootargs patch with ATAG DTB mangle 

ianchi:
For dual boot devices (i.e. some new Linksys routers) the active boot partition is needed from the bootloader's kernel argument.
The current patchs that provide the functionality (generic/995-mangle-bootargs + platform/100-find-active-root) also enable the passing of other ATAGs from bootloader to kernel and appending to DT info.
When used in ipq806x for Linksys EA8500, it worked for that device but broke C2600. (Details in PR #6 ).
To solve the issue, a modified version of the patch is used, that intercepts the bootargs arguments earlier in the boot process, effectively ignoring all ATAGs, and only appending the rootblock on the specific devices that enable the feature with a new property in DT (under "chosen/append-rootblock").

The command-line arguments provided by the boot loader will be appended to a new device tree property: bootloader-args (just for information). If there is a property "append-rootblock" in DT under /chosen and a root= option in bootloaders command line it will be parsed and added to DT bootargs with the form: <append-rootblock>XX.
Only command line ATAG will be processed, the rest of the ATAGs sent by bootloader will be ignored.

This patchset also replaces the old one in Kirkwood an Mvebu (the only other platforms that used it), so as to have a unified approach in all targets.

Tested to compile and run on ipq806x, only tested to compile in kirkwood/mvebu (for lack of hardware).


blogic:
i still dont understand why we cannot use the original mangle patch that you proposed and simply enable it via a devicetree property

ianchi:
@blogic the problem is that I couldn't find a better way to simply enable it via DT.
Sorry, for the lenghty answer that follows, but I'd like to explain it as clear as possible and to have any ideas to improve it:
The original patch works perfectly ok for EA8500, the problem is that it broke C2600. This in itself is a litlle strange, as kirkwood/mvebu also share the patch between linksys dual boot devices and other devices that don't seem to be afected by it. I don't exactly know why it faild so early at the booting that it dind't print any message to console.

That patch does three things:

1. Enable merging **all** ATAG information received by the bootloader with the info on the DT.
This is standar linux code (no patch here) enabled by config options (CONFIG_ARM_ATAG_DTB_COMPAT=y + CONFIG_ARM_ATAG_DTB_COMPAT_CMDLINE_FROM_BOOTLOADER=y)
This occurs very early in the boot process, it is called from asm code in head.S

2.  When the code reaches init/main.c the patch "995-mangle_bootargs"  "activates" and from the appended command line (DT bootargs + bootloader's args) it renames **only** parameter "root=..." with "mangled_rootblock=..."

3. The second patch "100-find_active_root" just renames that partition number to "ubi" when the DT partition info is parsed by the driver.


Now, the patch as it is can only have code affecting from step 2, but at this point all bootloader's information has already been merged with DT, and I found no way of knowing what info came from where in order to simply enable it with a new DT property.

So I added code in step 1, so that no info is effectibly merged into the DT and only copy the active partition if the special property is set in DT.
Once done that, the old patch in step 2 is completely redundant ando so I eliminated it.
By the way, instead of needing a step 3 also patching the driver, I simple used a standard ubi feature, renaming the rootblock "ubi.mtd=" (and not the bogus mangle_rootblock).

Regarding the impact in other platforms, it was only for the sake of unifing the approach. Both patches can perfectly coexist, and we can enable this one only for ipq806x.
I left the changes as separate commits, so they can easily be ignored.

blogic:
ok, thanks for the explanation. what is still unclear to me, which patch do i now merge into the ipq806x folder ? i would prefer a single commit that will only change code or add a patch inside the ipq806x/ folder.

ianchi:
Updated commits, rebasing with master to solve conflicts with latest commits and removed all commits affecting other platforms.

ianchi:
@blogic I also squashed everything in only one commit, and moved the patch to the ipq806x folder

ianchi:
Closed as it was merged into trunk

KanjiMonster:
Just saw this now, unfortunately this will have broken serial output on most (all?) ipq806x boards because QSDK uses a different tty name (ttyMSM0 instead of ttyS0 iirc or the other way round), and using the bootloader bootargs linux will then try to use the non-existent tty set by the original bootargs. There's currently no way from linux to overwrite that AFAICT.

ianchi:
@KanjiMonster, but this patch doesn't consider anything send by the bootloader.
That is exactly its purpose, to have the data available (to extract current root partition numbr) but in a way that no data at all reaches the kernel, so no harm is done.
Only root block if enabled by dts

KanjiMonster:
After reports that R7500 wasn't booting with newer LEDE/OpenWrt images and investigating, I found that this commit (0ddcbee2615197c8905408422b4a8a9fe7dbe954) broke it, reverting makes the R7500 boot again.

What seems to break the kernel is the bootloader supplied bootargs, which is
`console=ttyHSL1,115200n8 ubi.mtd=rootfs root=mtd:ubi_rootfs`

It doesn't boot at all, seems to hang somewhere very early.

ianchi:
@KanjiMonster, that is strange, this patch intercepts the bootloader bootargs, so that they don't reach the kernel. It's the main change from the original Mangle Bootargs patch.
Do you have some serial output, or it doesn't produce any?

I will go over it, to see if I find the cause.

KanjiMonster:
There is no serial output at all - with just the `console=` bootargs it boots fine, so the issue likely isn't the command line reaching the kernel, but the mangle code breaking with that command line.

PR #127, opened 2016-06-16
This is the WIP stuff for the DGN3500

danielg4:
@blogic

### DIFF
--- a/package/network/config/ltq-adsl-app/files/dsl_control
+++ b/package/network/config/ltq-adsl-app/files/dsl_control
@@ -1,7 +1,7 @@
 #!/bin/sh /etc/rc.common
 # Copyright (C) 2012 OpenWrt.org
 
-START=99
+START=49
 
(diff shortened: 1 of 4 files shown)


blogic:
@danielg4 patches 2/4 and 3/4 are missing the description explaining the bug that they are fixing.

danielg4:
2/4 does not address a bug, merely a logic error, possibly affecting only ath5k boards where under certain conditions, a nonexistent macaddr would be assumed correct.
3/4 puts br2684ctl init after ADSL init instead of before, so that the ESI is set at the right time, and for consistency with the PTM driver.

blogic:
put a proper description into the patches so that they can be merged. you know how to created patches, i have seen you do it before. the patches need to be in the expected format so that they can be merged.

blogic:
i merged 1/4 and 3/4 into my staging tree. please opena  new PR when the other patches are ready.

PR #128, opened 2016-06-16
ar71xx: WDR4300: Fixed default VLAN order

DPini:
Reordered the VLANs so the LAN ports are set to VLAN 1 and the WAN
port is set to VLAN 2, as in the other routers in the config file.
Moreover, this model had this VLAN mapping in OpenWRT Chaos Calmer. It seems
that the VLAN were switched when fixing a bug in the port mapping ( OpenWRT
 changeset 47799 )

Signed-off-by: David Pinilla Caparrós <dpinitux@gmail.com>


blogic:

DPini:
Thanks!

PR #130, opened 2016-06-17
let perf link against libunwind

stsydow:
Let perf link against libunwind to get better backtraces.

### DIFF
--- a/package/devel/perf/Makefile
+++ b/package/devel/perf/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=perf
 PKG_VERSION:=$(LINUX_VERSION)
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


blogic:
the commit is missing a valid Signed-oof-by line. can you please provide it so we can merge the patch

stsydow:
Ok there it is.
I put it in my commit template.

stsydow:
Please wait. It doesnt work yet.

jow-:
@stsydow - do you like to keep this PR open or will you propose another one if you sorted out the changes?

stsydow:
So this changeset does not break anything but perfs feature and library detection does not enable libunwind. There have been changes in April to improve backtraces for mips, so I hope it will work with a newer Kernel. 
Sorry for the missing details. I was in a hurry when I discovered the problem.
I would also submit i again or we can open an issue an do it after the next release.

blogic:
just open a new PR when it is tetsted/working properly please

PR #131, opened 2016-06-17
scksum: simple checksum for ath10k QCA99x0

ianchi:
In ath10k QCA99x0 verifies calibration data checksum and throws an error if it does not match.
So when patching mac address, it is also needed to patch cksum info.
This utility implements the simple checksum method used in calibration data files.

The PR also enables its use for ipq806x/EA8500.

### DIFF
--- a/package/utils/scksum/Makefile
+++ b/package/utils/scksum/Makefile
@@ -1,0 +1,42 @@
+#
+# Copyright (C) 2016 LEDE
+# Copyright (C) 2016 Adrian Panella
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=simple-cksum
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/simple-cksum
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Utility to calculate simple checksum
+  MAINTAINER:=Adrian Panella <ianchi74@outlook.com>
+endef
+
+define Package/simple-cksum/description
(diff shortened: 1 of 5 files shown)


nbd168:
Instead of fixing up the checksum, why not just leave the caldata alone and use sysfs to change the MAC address? That's what the mac80211 patch 120-cfg80211_allow_perm_addr_change.patch is for.

ianchi:
That sounds good and way simpler.
How does it work? I suppose I have to do:
´´´
echo $mymacaddr > /sys/class/ieee80211/phy0/macaddress  
´´´
But this needs to be done on every boot before wifi detect or wifi up are run, isn't it?
Where should I add it,in preinit or in an init.d script before S20_network?

nbd168:
You could add this as a hotplug script to run when a new ieee80211 phy is detected

blogic:
please open a new PR for the hotplug script based solution

ianchi:
Hi @nbd168 / @blogic , sorry to bother with such a simple question, but I'm having some trouble figuring out the hotplug mechanism. I enabled debuglevel 4, to get all events logged from procd, to see how to trap  ieee80211 phy addition, but I see no event for that, and what confused me even more: not even the firmware request event, which I know is indeed being called.
The only one printed for this devices are net/wlan? iface removal+addition and net/wlan?/queues/tx-?.
What am I missing?, from where else are hotplug scripts called? I see that /proc/sys/kernel/hotplug is set to null, so apparently not from std kernel mechanism.

ianchi:
I managed to track down the issue a little bit further.
It seems that there are some invocations of hotplug that are executed but not logged, in spite of debuglevel being set to 4.

It seems that there is a transition in procd life cycle, where it stops logging to kernel log, but still the system log is not started, and so we get no log at all for some time, till it starts logging again in syslog. 
Is this normal behavior? should I enable this log somewhere else?

Anyway, the phy hotplug script is working to change the mac address. After some cleaning up I'll submit it in a new PR.

PR #135, opened 2016-06-18
[RFC] Move append-dtb to common image-commands

BWhitten:
This build step is used by various targets, move it to a common section.

For the brcm63xx and lantiq targets where DTSs are in a separate target directory, an extra variable is introduced to allow their building.
I compile tested kirkwood and brcm63xx to make sure it worked as expected with custom dir and without.

I am unsure what to do with bcm53xx when the image builder is used (IB?)


blogic:
looks good. i asked a few people to test it o 53xx. i have merged this into my staging tree already.

BWhitten:
I see this has been pushed, thanks! Closing PR.

PR #136, opened 2016-06-18
ipq806x: add initial support for Netgear R7800

dissent1:

The commit adds initial support for Netgear R7800
The device boots and is accessible, networking is okay.

Known issues:
- USB does not function yet
- Not all buttons function
- WIFI support by ath10k driver (QCA9984, compat-wireless) has to be implemnted into LEDE seperately
- not all regulators clocks are supported by LEDE drivers

Seems there should be more changes into other linux drivers related to ipq806x.

For further development I need help from community, that unfortunately I can't get on forums.

Made a new profile file to choose ath10k firmware.

### DIFF
--- a/target/linux/ipq806x/Makefile
+++ b/target/linux/ipq806x/Makefile
@@ -20,7 +20,7 @@ DEFAULT_PACKAGES += \
 	kmod-ata-core kmod-ata-ahci kmod-ata-ahci-platform \
 	kmod-usb-core kmod-usb-ohci kmod-usb2 kmod-ledtrig-usbdev \
 	kmod-usb3 kmod-usb-dwc3-qcom kmod-usb-phy-qcom-dwc3 \
-	kmod-ath10k ath10k-firmware-qca99x0 wpad-mini \
+	kmod-ath10k wpad-mini \
 	uboot-envtools
 
 $(eval $(call BuildTarget))
--- a/target/linux/ipq806x/base-files/etc/board.d/01_leds
+++ b/target/linux/ipq806x/base-files/etc/board.d/01_leds
@@ -19,13 +19,14 @@ c2600)
 	ucidef_set_led_default "general" "general" "ledgnr:blue" "1"
 	;;
 d7800 |\
-r7500)
-	ucidef_set_led_usbdev "usb1" "USB 1" "r7500:white:usb1" "1-1"
-	ucidef_set_led_usbdev "usb2" "USB 2" "r7500:white:usb3" "3-1"
-	ucidef_set_led_netdev "wan" "WAN" "r7500:white:wan" "eth0"
-	ucidef_set_led_ide "esata" "eSATA" "r7500:amber:esata"
-	ucidef_set_led_default "wps" "WPS" "r7500:white:wps" "0"
-	ucidef_set_led_default "rfkill" "rfkill" "r7500:white:rfkill" "0"
+r7500 |\
+r7800)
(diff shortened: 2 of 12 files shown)


ianchi:
Hi, please don't remove ath10k-firmware-qca99x0, it is needed by C2600 and EA8500.
On the image makefile you are removing all devices' titles.

dissent1:
That is made on purpose, they are moved into profile/01-radio...mk file with right firmware for each device

ianchi:
But just recently all profiles where unified by @blogic so that images for all devices are generated at once sharing the main build.
Why go back, only to save one driver? In that case there are other things that are not needed by all devices.
But none of these devices are space constrained, so not really much gain there.
Are you having conflicts between qca99x0 and qca9984?

dissent1:
I haven't tested, but there should be no conflicts actually, just wanted to make firmware a bit cleaner, considering that firmware for qca988x already gets built in by default. 
I don't know the whole story behind titles in makefile, but if it makes some things easier I can revert it back, no problem.
Maybe there is a way to choose appropriate package in image/Makefile?

blogic:
* please do not remove DEVICE_TITLE and add a profiles folder. instead add a DEVICE_PACKAGES:= $firmware package to the image makefile code.
* the changes to target/linux/ipq806x/patches-4.4/168-ARM-qcom-add-smb208-DT.patch look wrong
* target/linux/ipq806x/base-files/lib/upgrade/platform.sh seems to have spaces instead of tabs

if you could make those changes please, the rest looks good.

dissent1:
I've adjusted the code.
Is it only me getting QCA99X0 board-2.bin md5 mismatch during the building process?

blogic:
merged into my staging tree, please verify that it work so i can push it to master

dissent1:
Yes, it builds and boots as intended. Thank you very much.

ianchi:
@blogic I think that adding qca99x0 driver via DEVICE_PACKAGES is not working.
I remember a similar problem you reported for IB (commented in PR #154), but apparently it also occurs from standard make.
The dependency isn't included.

Now EA8500 image is not longer working, and I believe the same is happening to C2600 according to forum's post. (Perhaps @bendavid can confirm)

blogic:
@nbd168: looks like the image building code is not working causing lots of headaches for lots of people

dissent1:
@blogic @ianchi  Weird that I don't have such problems on r7800, 5ghz is working and firmware for QCA988X and kmod-ath10k gets built in as before.
I have problems with 2.4ghz, but guess it's not related to the above

ianchi:
@dissent1, the QCA988x firmware is selected by default by ath10k, which in turns is set as a default package of the ipq target; so that firmware is always included, even for devices that doesn't need it (like EA8500).
The problem for r7800 should be qca9984, which is specified in DEVICE_PACKAGES. Are you sure it is included in the default images and not by a custom selection in menuconfig?

dissent1:
@ianchi exactly. My only steps are:
make menuconfig -> select target device
make defconfig
make menuconfig -> add basic luci collection
make
That's all. I get qca9984 firmware included without extra steps. The same for c2600 and qca99x0 when I buildtested.
Maybe we are talking about different things?

dissent1:
Btw, what do you think about switching cpu_type in makefile to cortex-a15, instead of cortex-a9. Krait-400 cores in ipq806x are very similar to cortex-a15.

PR #137, opened 2016-06-20
download: add download info file

feckert:
Add download info file to get short information where the packages are
downloaded from. If you add a local mirror to your config. It is now
easy possible to see if an offline build is possible.

### DIFF
--- a/include/download.mk
+++ b/include/download.mk
@@ -11,6 +11,8 @@ LEDE_GIT = https://git.lede-project.org
 
 DOWNLOAD_RDEP=$(STAMP_PREPARED) $(HOST_STAMP_PREPARED)
 
+DL_INFO_FILE=$(BIN_DIR)/download.txt
+
 # Try to guess the download method from the URL
 define dl_method
 $(strip \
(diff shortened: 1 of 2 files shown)


blogic:
few comments ..

cant these 2 lines be put into the code calling DownloadMethod/* ? 
*+		echo "Url: $(URL)^$(VERSION) File:$(FILE) Md5sum: " >> $(DL_INFO_FILE) && \
 +		md5sum $(DL_DIR)/$(FILE) | cut -d ' ' -f 1 >> $(DL_INFO_FILE); \

it looks like the file will never be cleared but will continue to grow forever ?

feckert:
1. It was a little bit unclear where to at this lines globaly? But that is right -> avoiding duplicate code
2. You are right it will never be deleted it crows forever.
Is this a problem?
If yes -> when should the file be deleted? For your Information the file will only be updated if the requested file is not in the dl directory.

hnyman:
> You are right it will never be deleted it crows forever. 

The approach to append with deleting anything means that the file will accumulate every new version of e.g. procd, ubus etc., generating a rather long list, which will sooner or later contain quite many old discarded versions.

I think that "make clean" should clear the file. Then the file would contain the new stuff that was needed for the current build, not all discarded stuff downloaded during the past months. Your patch creates the file in BIN_DIR, right? In that case "make clean" clears it, I think.

feckert:
@hnyman yes if its in the BIN_DIR and if i make a "make clean" then the BIN_DIR will be deleted and so the download.txt as well.
@blogic is this ok for you?

blogic:
@feckert no it is not ok for me hence my last comment ;)

feckert:
@blogic do you meen

> cant these 2 lines be put into the code calling DownloadMethod/* ?
echo "Url: $(URL)^$(VERSION) File:$(FILE) Md5sum: " >> $(DL_INFO_FILE) && \
md5sum $(DL_DIR)/$(FILE) | cut -d ' ' -f 1 >> $(DL_INFO_FILE); \

if you meen this: i have to look to get this into the calling code


or

> it looks like the file will never be cleared but will continue to grow forever ?

if you mean this: It will be deleted on a `make clean`

blogic:
i dont like that make clean is required but i can live with that. the prior part witht he code duplication needs to be resolved before we can merge the patch.

feckert:
@blogic I have resolved the code duplication as you suggest.

blogic:
@nbd168 any comments ?

nbd168:
I think it would be better if the infos are dumped from the package makefile context (to get it even if files are present in dl/ already) into a per-source-directory file in staging_dir, then merged into bin/ after the build is done.
That way the file doesn't grow continously, it works with or without make clean, with or without files present in dl/ and it always contains up to date information.

feckert:
The reason for this patch is to see if an offline build is possible with my selected packages. I want to see the download sources if i add an mirror to the config. The best is to see only this mirror url in the generated file.

@nbd168 is this possible with your suggestion to do it in the Makefile context. Does I see my mirror url if the file is already downloaded into the dl directory?

nbd168:
Ah, I guess I misunderstood your use case. I think a good way to test if offline builds are possible would be to write a script that does a HTTP HEAD request on your mirror for any file in the generated download metadata. That way you don't have to clear your dl/ folder and re-download everything to be able to test your mirror. Does that make sense?

feckert:
@nbd168 My use case is that I am running a build bot for Openwrt/lede and if I make a check in, then a new fresh build will start. When the build is finished, then the "dl" dir will be push to the mirror and "bin" dir will be pushed on to a the ftp server. After the build i can see where the package come from. 

When the build is finished i make a distclean and start with an absolutely fresh environment. No dl dir.

champtar:
@feckert you are wasting bandwidth, why not just force checking the hash?

nbd168:
If you only care about that use case, then I'd suggest simply adding an appropriate log message to download.pl and grepping for that in the build log.

In my opinion, your use case is just too specific and tied to your workflow that it doesn't make sense to even write a separate file to it, especially not to bin/.

feckert:
Thanks for your time. I will review your suggestions.

PR #142, opened 2016-06-21
ipq806x: enable ieee80211 phy hotplug and patch macaddress

ianchi:
Calibration data for QCA99x0 in this device has bogus macaddress.
The data cannot be modified directly, as it breaks checksum control.
Instead change the macaddress from phy add hotplug event.

### DIFF
--- a/target/linux/ipq806x/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
+++ b/target/linux/ipq806x/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
@@ -1,0 +1,21 @@
+#!/bin/ash
+
+[ "$ACTION" == "add" ] || exit 0
+
+PHYNBR=${DEVPATH##*/phy}
+
+[ -n $PHYNBR ] || exit 0
+
+. /lib/ipq806x.sh
+. /lib/functions/system.sh
+
+board=$(ipq806x_board_name)
+
+case "$board" in
+	ea8500)
+		echo $(macaddr_add $(mtd_get_mac_ascii devinfo hw_mac_addr) $(($PHYNBR + 1)) ) > /sys${DEVPATH}/macaddress
+		;;
+	
+	*)
+		;;
+esac


blogic:
why are you adding this file ->
target/linux/ipq806x/base-files/etc/hotplug.json

ianchi:
Sorry, it was a leftover from my debugging/testing of the hotplugs events not being always logged.
It shouldn't have been in this branch. I updated the commit without it.

heinzek:
@ianchi can You add C2600 device to your script?
```
        c2600)
                echo $(macaddr_add $(mtd_get_mac_binary default-mac 8)  $(($PHYNBR - 1)) ) > /sys${DEVPATH}/macaddress
                ;;
```

It works very well

blogic:
merged into my staging tree after adding the c2600 case

bendavid:
Thanks, confirm this works as expected for c2600.

ianchi:
@bendavid / @heinzek , if you want the mac address for wlan to be lower that the reference one, I think you should substract $PHYNBR, as you send it now, phy1 will get the same address as the reference (and the same as used in DT for eth).
I believe the desired effect is to have phy0 with an offset of -1 and phy1 with an offset of -2, and so change to:
```
echo $(macaddr_add $(mtd_get_mac_binary default-mac 8)  $((-$PHYNBR - 1)) ) > /sys${DEVPATH}/macaddress
```

bendavid:
So actually, how it is done now is consistent with how the MAC addresses are defined in the stock firmware, where indeed 2.4GHz wireless and eth1 (LAN) both have the reference mac address and therefore have the same MAC address.

This is not very nice, but I'm also not sure it will really cause any problems, given that packets should ~never be sent directly to/from the wireless station MAC address as far as I understand.  We could change this behaviour, though this implies using additional/different MAC addresses with respect to the stock firmware.

ianchi:
OK, I just found a bit weird to change only one of the wlan and leave the other one with "shared" mac address

PR #143, opened 2016-06-22
include/image.mk: Fix kernel filename with extra image name

dfc-d:
Changes to imaging building code resulted in failing to correctly generate
the kernel filename to use when generating images when EXTRA_IMAGE_NAME
is not empty.  This fixes that.

Signed-off-by: Daniel Dickinson <Daniel Dickinson lede@daniel.thecshore.com>


blogic:
the SoB line seems to be wrong. there is a duplicate name in it.

PR #144, opened 2016-06-22
include/image.mk: Fix kernel filename with extra image name

dfc-d:
Changes to imaging building code resulted in failing to correctly generate
the kernel filename to use when generating images when EXTRA_IMAGE_NAME
is not empty.  This fixes that.

Fixed signed-of-by: BTW there is not need to close PR's that need updating: a new forced update push can replace the bad PR with corrected version, and preserve history of discussion, and is generally accepted practice for GitHub PR's.


jow-:
Can you provide a before/after example? Just to better understand it

dfc-d:
On Tue, 2016-06-28 at 10:12 -0700, Jo-Philipp Wich wrote:
> Can you provide a before/after example? Just to better understand it

I think this was a result of a patch that got reverted.  I will let you
know once I've tested without this patch to see if it's still present.
 And of course output.

Regards,

Daniel

dfc-d:
Ok, probably because I was using John's staging tree to test a fix he asked me to look at the (the px5g mbedtls one), there is the ar71xx device profile patch also in the tree I built with and I get the following error:

make[4]: *** No rule to make target '/home/daniel/Build/openwrt/imagebuild/cshorewrt-imagebuilder-r6f92c25-ar71xx-generic.Linux-x86_64/build_dir/target-mips_34kc_musl-1.1.14/linux-ar71xx_generic/cshorewrt-r6f92c25-testrouter-ar71xx-generic-wndr3800-kernel.bin', needed by '/home/daniel/Build/openwrt/imagebuild/cshorewrt-imagebuilder-r6f92c25-ar71xx-generic.Linux-x86_64/build_dir/target-mips_34kc_musl-1.1.14/linux-ar71xx_generic/tmp/cshorewrt-r6f92c25-testrouter-ar71xx-generic-wndr3800-squashfs-sysupgrade.bin'.  Stop.

With the patch in this PR the image build occurs successfully because I have distinguished between the kernel image name and overall image name (kernel image name does not get the EXTRA_IMAGE_NAME variable added to the name while the whole image name does).

nbd168:
pushed a better fix to my staging tree ("build: remove image prefix from kernel files in KDIR"). It should fix your issue as well, while also ensuring that EXTRA_IMAGE_NAME still makes it into kernel images in bin/
Please check if this works for you as well.

PR #145, opened 2016-06-22
Add config option to make POSIX ACL support the default

dfc-d:
This patch adds an option to make POSIX ACL support the default where it makes sense (i.e. for filesystem likely to be on external storage but not flash and tmpfs).  By default this option is turned off as that is current LEDE situation and ACL will increase at least affected filesystem module size.

This patch set also makes XATTR the default where missed before (it's on by default already for most filesystems in LEDE).


dfc-d:
Forgot to mention: compiled tested on all three kernels, image tested on ar71xx (WNDR3800)

blogic:
2/2 says 
"Conflicts:
	target/linux/generic/config-4.3"

in the commit description and has a trailing whitespace in config/Config-kernel.in

dfc-d:
Fixed

blogic:
merged into my staging tree

dfc-d:
On Fri, 2016-06-24 at 03:14 -0700, John Crispin wrote:
> merged into my staging tree

FYI, I got looking at the menuconfig and the current ACL patch is kind
of ugly in that it just ends with all the options in the main
configuration option page.  It's cosmetic, but it'd probably be better
it were to be implemented as a submenu instead of in the primary
selection.

Does it matter to you?

Regards,

Daniel 
>

blogic:
its still in my staging tree so i can update it if you want. submenus will make it cleaner

dfc-d:
On Tue, 2016-06-28 at 00:33 -0700, John Crispin wrote:
> its still in my staging tree so i can update it if you want. submenus
> will make it cleaner
> 

I obviously didn't get to it before you merged, so I've issued a new PR
with the fix instead.

Regards,

Daniel

blogic:
i actually accidentally pushed it with some other stuff. will merge this in a sec

PR #148, opened 2016-06-23
[URFC] Add global setting for default SSL provider

dfc-d:
This patch set adds a global config option that combined with patches to SSL-using packages allows to select the default SSL provider (e.g. to allow to make openssl the default instead of polarssl) with option to allow or disallow fallback to openssl (that if default provider is not openssl, making packages which require openssl to default to n (not building)).

Also has a "No SSL" option that defaults all SSL and dependent package variants to default n.

### DIFF
--- a/config/Config-build.in
+++ b/config/Config-build.in
@@ -74,6 +74,62 @@ menu "Global build settings"
 
 	source "config/Config-kernel.in"
 
+	comment "Default Crypto Options"
+
+	choice
+		prompt "Extent to prefer crypto"
+		default CRYPTO_DEFAULT_PACKAGE
+		help
+		  How strongly to insist on using cryptography
+		  when available (for package builds).
+		config CRYPTO_DEFAULT_PACKAGE
+			bool "According to package and SSL provider defaults"
+		config CRYPTO_DEFAULT_ALL
+			bool "Default to always enabled"
+		config CRYPTO_DEFAULT_NONE
+			bool "Default to not building crypto except internal wireless"
+	endchoice
+
+	choice
+		prompt "Default SSL provider"
+			default SSL_DEFAULT_SELECT_POLARSSL if !CRYPTO_DEFAULT_NONE
+			default SSL_NOSSL if CRYPTO_DEFAULT_NONE
+			help
+			  Set the SSL library to enable by default when there is a
+			  choice if libraries (e.g. to prefer openssl when both
+			  polarssl and openssl are options).  For "No SSL" option
+			  selects no SSL when that is an option, but doesn't
+			  prevent building packages requiring SSL (i.e. it's a
+			  default not a hard setting).
+			config SSL_DEFAULT_POLARSSL
+				bool "PolarSSL"
+				select PACKAGE_libpolarssl
+			config SSL_DEFAULT_OPENSSL
+				bool "OpenSSL"
+				select PACKAGE_libopenssl
+			config SSL_DEFAULT_MBEDTLS
+				bool "mbedtls"
+				select PACKAGE_libmbedtls
+			config SSL_DEFAULT_NOSSL
+				bool "No SSL"
+	endchoice
+
+	config SSL_DEFAULT_FALLBACK_OPENSSL
+		bool "Allow fallback to OpenSSL if default not available"
+		depends on !SSL_DEFAULT_OPENSSL
+		default y if !SSL_DEFAULT_NOSSL
+		help
+		 For non-openssl as the default SSL library, allow
+		  to build packages against openssl if the default
+		  library is not avaiable or not an option.
+		  NB: Saying n here will make packages that can
+		  only be built against openssl default to n.
+
+	config SSL_DEFAULT_NOOPENSSL
+		bool
+		default y
+		depends on ( !SSL_DEFAULT_OPENSSL && !SSL_DEFAULT_FALLBACK_OPENSSL )
+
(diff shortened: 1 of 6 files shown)


blogic:
let us know when it is tested so we can start looking at it.

dfc-d:
It now does the right thing in terms of menuconfig / proper settings in .config.  Given that it's just enabling / disabling options that are already well-tested, I'm not sure test images are necessary ; the real trick is to make sure all relevant packages set the right config (and a corresponding patch for the feeds will also be relevant).

blogic:
i find this approach with allow and forbid to overly ambitious and non intuitive. the forbids are a choice option and you can only forbid one if you want that at all.

i undersatnd the idea that you dont accidentally ship several libs in parallel but there should be one choice which is "any,openssl,mbedtls, ...." this would still be ambiguous for packages that dont provide variant support.

dfc-d:
On Thu, 2016-06-30 at 01:26 -0700, John Crispin wrote:
> i find this approach with allow and forbid to overly ambitious and
> non intuitive. the forbids are a choice option and you can only
> forbid one if you want that at all.

The reason I did menu selections vs. choice is that especially in the
packages feed (for which I have a patch set I haven't pushed yet,
perhaps it would help demonstrate), there are a number of packages
which make picking a single library impossible.

Some packages only support polarssl.  Others only openssl.  Other only
gnutls.  I don't think any support only cyassl/wolfssl, but if one
wants primary choice to be cyassl/wolfssl, with a packages one normally
doesn't install available with with other libraries, there are very few
packages that actually support cyassl/wolfssl.

Nothing supports only mbedtls, but it would be helpful to support
mbedtls as the default, with the ability to build against other
libraries for packages that don't support mbedtls.  In particular,
except openvpn, you can currently build all of the base system (and
with a LuCI patch I have, luci-ssl) against purely mbedtls.

This actually ended up being more complicated than my first cut at
this, because of the problem of there being a variety of ssl libraries
and no single library (even openssl) that works with all, but it would
still be beneficial be able to

a) Have a convenient default selection of first choice for library
b) Be able to filter out unwanted libraries.

> i undersatnd the idea that you dont accidentally ship several libs in
> parallel but there should be one choice which is
> "any,openssl,mbedtls, ...." this would still be ambiguous for
> packages that dont provide variant support.

It's not just about picking one library, it's about the fact that one
wants to be able to set a default, but that unless wants to
deliberately exclude some libraries, some packages can't be built
unless you allow other libraries than the default (or default + one
fallback).

The reality is the Linux SSL situation is kind of a mess, and LEDE
isn't in a position to fix every package it the distro to use a common
set of libraries.

Regards,

Daniel

dfc-d:
Oh, and I remembered something; my initial approach was to apply the options on a per-package basis but that is not really a maintainable solution because there are *lot* of packages to touch especially in the packages feeds.

Doing a default for the small number of packages that do have a choice, and doing a FORBID option (which allows one to have DEPENDS:=@!SSL_DEFAULT_FORBID_LIBRARY in the *library* which automatically blocks all dependencies) greatly reduces the the number of packages that need to be touched.

dfc-d:
Basics tested, but there remains things like curl to be tested.

dfc-d:
If you've had a chance to read and think about my response/explanation for why I did what I did and not the naive approach you suggested (which is what I first implemented), do you have new comments and or suggestions for means to achieve the desired goals?

blogic:
@nbd168 comments ?

nbd168:
I don't like the fact that these commits contain unrelated changes, some of which seem wrong.
The DEPENDS rework in the curl makefile makes no sense to me.
Disabling SIGNED_PACKAGES if CRYPTO_DEFAULT_NONE is set is just plain wrong.
Porting the SIGNED_PACKAGES symbol to the SDK should be in a separate commit.

I also still have doubts about the whole approach and about including so many SSL libraries in the selection. In my opinion, the only SSL libs that are even relevant as a default choice are openssl, polarssl and mbedtls. Once migration from polarssl to mbedtls is done for core packages, polarssl will be out as well. Additionally, advantages/disadvantages of individual SSL libraries often vary between packages, depending on the quality of the SSL library integration and this is not easily apparent from the global selection.

For things like gnutls or cyassl, a default selection does not make much sense, since they're only supported by a few packages, and I can't think of a good reason why anybody would ever want to make these the default.

You might even get most of the value of the default selection handling by reducing it to disabling mbedtls, polarssl or openssl and changing the defaults to select the next best alternative.

dfc-d:
Patchset reduced in scope and split up.  Also the changes that were done because of recursive dependencies have been reverted; will have to see if the rework eliminate the problematic dependency cycles that led to the parts like the curl rework (now gone).
We know only worry about polarssl, mbedtls, and openssl as defaults and don't worry about forbiding or filtering out packages with other libraries.

nbd168:
Some issues with that:

- The CRYPTO_DEFAULT_* stuff doesn't make much sense to me. From reading the descriptions it is completely non-obvious what packages will be affected and how. I don't see the point in even having a global option for this.
- The SSL_DEFAULT_* options should not select any SSL libraries, especially since they might not even get used if there's no package that requires them. While you're at it, you can also fix the indentation there.
- The DEFAULT override in nettle is completely useless, in fact I think the whole commit should be dumped entirely.
- Please split up the 'hostapd curl uhttpd' commit.

blogic:
no feedback in over 3 weeks, please reopen when you have time again.

PR #149, opened 2016-06-24
ipq806x: remove unused dsa patch

ianchi:
switch is managed via LEDE's own drivers. remove unused patch adding DSA
drivers for AR8337.


blogic:
it is not unused but pending the userland patch to make it work ;)

PR #150, opened 2016-06-24
Change /dev paths to be more like a desktop distro

neheb:
This changes the default permissions for /dev entries to be more similar to a desktop distro. Taken from the defaults of Arch Linux and Ubuntu. Also changed some that were nonsensical. For example, all 660 permissions on desktop distros were of the form root:x where x is something other than root. As such, 660 is useless for LEDE where the specific group is missing. audio seems to be the only group that isn't.

Signed-off by Rosen Penev <rosenp@gmail.com>


neheb:
I messed up.

PR #153, opened 2016-06-25
Adjust /dev entries to desktop distro defaults

neheb:
This changes the default permissions for /dev entries to be more similar to a desktop distro. Taken from the defaults of Arch Linux and Ubuntu. Also changed some that were nonsensical. For example, all 660 permissions on desktop distros were of the form root:x where x is something other than root. As such, 660 is useless for LEDE where the specific group is missing. audio seems to be the only group that isn't.

Signed-off by: Rosen Penev <rosenp@gmail.com>


blogic:
@jow- looks good but can you please have a close look aswell ...

jow-:
LGTM

blogic:
merged, and also added a patch that drops the special case for gpio and hvc

jow-:
I am not sure about the gpio change, it logically makes sense but not sure how many hobby applications will break when the gpio devices are suddenly not non-root writable anymore...

blogic:
this is legacy from gpiodev and we dont have that int he tree anymore

jow-:
Ahh, right. Nevermind then :)

neheb:
hmm where did this go?

blogic:
into my tsaging tree where everything that i merge stays for a few days for etsting

PR #155, opened 2016-06-27
[openvpn] fix missing cipher list for polarssl in v2.3.11

stsydow:
I Saw later the change in openvpn upstream just now:
http://article.gmane.org/gmane.network.openvpn.devel/11647
https://community.openvpn.net/openvpn/ticket/683

Has been fixed shall be removed when v2.3.12 is released.

### DIFF
--- a/package/network/services/openvpn/Makefile
+++ b/package/network/services/openvpn/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=openvpn
 
 PKG_VERSION:=2.3.11
-PKG_RELEASE:=1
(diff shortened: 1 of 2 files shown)


jow-:
Hi, I staged a fix in https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=f98f4601de762251c4644047786affd34d5fb10c

Since I had to rework both the commit message and the patch file I couldn't take the PR as-is, hope this is okay for you. I used a `git format-patch` export from the upstream OpenVPN Git repo to create the patch file, to retain references to the original bug report.

stsydow:
It's fine thanks!

PR #157, opened 2016-06-29
dnsmasq: Add broken realtime clock build switch in full variant

dedeckeh:
By default dnsmasq uses the time function; which returns the time since
Epoch; to retrieve the current time. On boards which have no realtime
clock this can lead to side effects when the time is synced via ntp
as the "time wrap" forces dhcp leases to be considered as expired.
By enabling the broken realtime clock build switch dnsmasq uses the
times utility which returns the number of clock ticks.

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.76
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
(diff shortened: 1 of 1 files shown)


jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=03c0338e1cc56be73ca090d2392b3d133af94936

jow-:
@dedeckeh - I did some slight changes to your commit; changed the symbol name to match the existing ones (`CONFIG_dnsmasq_...` -> `CONFIG_PACKAGE_dnsmasq_...`) and added the new setting to `PKG_CONFIG_DEPENDS`. See https://git.lede-project.org/?p=lede/jow/staging.git;a=commitdiff;h=422af211620e1e0ef85a9702dd3d9360ef099f81

dedeckeh:
@jow ack looks indeed better

PR #159, opened 2016-06-29
Switch cake internal algorithm to COBALT a codel/blue hybrid

kdarbyshirebryant:
Update iproute2/tc patch to include new COBALT related stats.
Update kmod-sched-cake to point to COBALT commit.


jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=9772d2278978a28811ec9b7e3fe307bd5bd0ca81

PR #160, opened 2016-06-29
Tidy kmod-sched-cake makefile

kdarbyshirebryant:
Update iproute2/tc patch to include new COBALT related stats.
Update kmod-sched-cake to point to COBALT commit
Tidy kmod-sched-cake makefile


jow-:
Squashed into #159: https://git.lede-project.org/?p=lede/jow/staging.git;a=commitdiff;h=77b0c116ed4db2dfd3b7673104db0c2d259847cd

PR #169, opened 2016-07-02
Set /dev/kmsg to 600

neheb:
Small cleanup. Rebooted my router but I see no impact from this change. dmsg still works.


blogic:
merged into my staging tree

PR #170, opened 2016-07-02
Update smb.conf.template

neheb:
netstat shows the bound interfaces properly now:

tcp        0      0 localhost:netbios-ssn   0.0.0.0:*               LISTEN      
tcp        0      0 lede.lan:netbios-ssn    0.0.0.0:*               LISTEN

wonder if localhost should actually be bound...

still don't get some things though. the generated smb.conf file has "Lede" for both the NetBIOS name and the server string even though the init script shows that it should say "Samba on Lede". No idea. Also the charset options are set to UTF-8 even though Samba defaults to that.


blogic:
merged into my staging tree

neheb:
just realized smb encrypt should not have been changed. minor point since it almost never gets negotiated but still.

blogic:
please update the PR, i just dropped the patch from my tree

neheb:
hrm i force pushed a new commit. usually it would show up here. i'll open a new one.

champtar:
Closed PR doesn't update

PR #171, opened 2016-07-02
Update smb.conf.template take 2

neheb:
also removed the smb encrypt option this time. encryption requires SMB3 while the max version specified is 2 meaning it never gets used.


blogic:
merged into my staging tree

PR #172, opened 2016-07-03
mwlwifi: update to 10.3.0.17-20160617

thagabe:
Updating the mwlwifi driver to the newest version. This version removes patches as most have been included into the upstream code. This code is now compatible with all linux kernels >4.X which for mvebu the default is 4.4.XX. Tested on 4.4.{13/14} and speeds have been solid with stable ping. 

Signed-off-by: Gabe Rodriguez \<lifehacksback@gmail.com\>


nbd168:
Pushed to my staging tree.

PR #173, opened 2016-07-04
kernel: Move POSIX ACL and attr support options into submenu

dfc-d:
Make global options menuconfig cleaner by moving POSIX ACL
and attr support options into a submenu.


blogic:
merged into my staging tree

PR #174, opened 2016-07-04
scripts/feeds: Prevent .config autocreation

dfc-d:
When using scripts/feeds upgrade the .config needs to be
updated but the code to do so was also autocreating a
.config if one didn't exist.  This is counter-productive
when you have not yet used menuconfig (or other config targets)
because things like selecting or deselecting CONFIG_ALL
(to build all package by default) only works if the
package selection has not already been done via an
existing .config selection.

### DIFF
--- a/scripts/feeds
+++ b/scripts/feeds
@@ -534,6 +534,11 @@ sub install_package {
 sub refresh_config {
 	my $default = shift;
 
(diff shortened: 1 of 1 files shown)


blogic:
merged, thanks !

PR #175, opened 2016-07-04
ustream-ssl: Fix recursive dependency

dfc-d:
Two variants incorrectly include themselves in
conditional depends on ssl libraries, which results
in a recursive dependency.


blogic:
merged, thanks

PR #176, opened 2016-07-04
network/services/uhttpd: Add Basic Auth config

dfc-d:

instance: Which uhttpd config section the auth is associated with

If instance is not specified auth section applies to
all instances of uhttpd; if instance is specified
the auth section only applies to the uhttpd config
section with the same name as specified with the
instance option.

Has been tested and contains changes suggested by Jo

### DIFF
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -53,6 +53,28 @@ generate_keys() {
 	}
 }
 
+create_httpauth() {
+	local cfg="$1"
+	local basecfg="$2"
+	local prefix username password
+
+	local instance
+	config_get instance "$cfg" instance
+	if [ -n "$instance" ] && [ "$instance" != "$basecfg" ]; then
+		return
+	fi
+
+	config_get prefix "$cfg" prefix
+	config_get username "$cfg" username
+	config_get password "$cfg" password
+
+	if [ -z "$prefix" ] || [ -z "$username" ] || [ -z "$password" ]; then
+		return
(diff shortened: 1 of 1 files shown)


blogic:
i am not sure if matching via config<-> instance is a good idea. sections of type httpauth should be named and then instances should have a 1:n relation to the httpauth sections.

dfc-d:
On Tue, 2016-07-05 at 23:21 -0700, John Crispin wrote:
> i am not sure if matching via config<-> instance is a good idea.
> sections of type httpauth should be named and then instances should
> have a 1:n relation to the httpauth sections.

Ok.  Will rework.

Regards,

Daniel

dfc-d:
Jo,

On Tue, 2016-07-05 at 23:21 -0700, John Crispin wrote:
> i am not sure if matching via config<-> instance is a good idea.
> sections of type httpauth should be named and then instances should
> have a 1:n relation to the httpauth sections.

For the luci app is there an easy way to get a list of named
configuration sections of httpauth for selecting from a list in the
main uhttpd section?

Regards,

Daniel

blogic:
@dfc-d you need to add proper syntax so that @jow- gets a notification

jow-:
@dfc-d - such a list can be easily assembled, e.g. like that:

    local authsections = { }
    cursor:foreach("uhttpd", "httpauth", function(s)
        if not s['.anonymous'] then
            authsections[#authsections+1] = s['.name']
        end
    end)

blogic:
3 weeks without feedback. please reopen the PR when you have time for feedback.

PR #178, opened 2016-07-06
Fix lede partition size for RT5350F-OLINUXINO-EVB

rtreffer:
The partition size is wrong, leading to out-of-disk-space even on no/moderate use.

Upstream fix from vendor: https://github.com/OLIMEX/openwrt/commit/2f25eb57edc79d33c4810d185c193be4293c434a
Suggested fix for openwrt: https://dev.openwrt.org/ticket/20321

### DIFF
--- a/target/linux/ramips/dts/RT5350F-OLINUXINO-EVB.dts
+++ b/target/linux/ramips/dts/RT5350F-OLINUXINO-EVB.dts
@@ -60,7 +60,7 @@
 
 		partition@50000 {
 			label = "firmware";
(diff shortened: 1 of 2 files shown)


jow-:
Can you provided a Signed-off-by line?

rtreffer:
Never done that, so give me some time to research that. But will do!

rtreffer:
@jow- like that?

blogic:
merged into my staging tree, thanks

next time also make sure the subject prefix is correct, i added "ramips: " while merging

PR #179, opened 2016-07-06
Fix lede partition size for RT5350F-OLINUXINO

rtreffer:
The partition size is wrong, leading to out-of-disk-space even on no/moderate use.

Upstream fix from vendor: https://github.com/OLIMEX/openwrt/commit/2f25eb57edc79d33c4810d185c193be4293c434a
Suggested fix for openwrt: https://dev.openwrt.org/ticket/20321


rtreffer:
put into PR-178

PR #180, opened 2016-07-06
add dlink dch-m225 support and update audio kernel module

igvtee:
* add dlink dch-m225 support
* fix pinmux for mdio as refclk. dch-m225 use this pin as refclk for audio codec
* add dma kernel module options need by i2s
* fix audio kernel module for mt7620. include simple audio module


target/linux/ramips/modules.mk
```diff
@@ -72,18 +72,58 @@ endef
...
+ Kernel modules for enable MediaTek hsdma engine.
+endef
 
+$(eval $(call KernelPackage,hsdma-mtk))
 
 define KernelPackage/sound-mt7620
   TITLE:=MT7620 PCM/I2S Alsa Driver
-  DEPENDS:=@TARGET_ramips_mt7620 +kmod-sound-soc-core +kmod-regmap @!TARGET_ramips_rt288x
+  DEPENDS:=@TARGET_ramips +kmod-sound-soc-core +kmod-regmap +kmod-dma-ralink @!TARGET_ramips_rt288x
   KCONFIG:= \
```
blogic: wont mt7621 require hsdma ? adding a flat dep on gdma while removing the mt7620 dependency seems wrong

igvtee: mt7621 has gdma and hsdma. gdma for slow devices like i2s. hsdma only support mem to mem copy. so mt7621 don't need hsdma for audio.

the gdma and i2s functions support on all ramips/mediatek targets except rt288x.
so i remote the mt7620 limit. because the mt7628 and mt7621 should support audio too.

blogic: ok, thanks for the explanation

blogic:
merged, thanks

PR #183, opened 2016-07-07
ar71xx: add software transmit timestamp support

sdwalker:
Add software transmit timestamp and ethtool (-T) timestamp support

Signed-off-by: Stephen Walker <stephendwalker+github@gmail.com>


lynxis:
hi @sdwalker do you did a benchmark? Does it do any bandwidth or cpu impact?

sdwalker:
No benchmarks were done.

nbd168:
What do you plan on using this for?

sdwalker:
Plan was software PTP using ptp4l from the linuxptp package.

blogic:
it would be good to get benchmarks. it looks like every skb will get cloned if you enable this. i would expect quite a performance hit.

nbd168:
I expect performace to still be okay. The clones are only done for packets where there's an attached socket and there's timestamping requested for the socket, so the only hit on normal operation would come from some simple checks. Routing performace will most likely remain unaffected.

nbd168:
Added to my staging tree

PR #184, opened 2016-07-07
[RFD] Replace DEVICE_TYPE with metapackages (collections)

dfc-d:
This patch series of three commits replaces

### DIFF
--- a/include/image.mk
+++ b/include/image.mk
@@ -425,7 +425,7 @@ endef
 define Device/DumpInfo
 Target-Profile: DEVICE_$(1)
 Target-Profile-Name: $(DEVICE_TITLE)
-Target-Profile-Packages: $(DEVICE_PACKAGES)
+Target-Profile-Packages: $(call DEFAULT_COLLECTION,$(DEVICE_PACKAGES))
 Target-Profile-Description:
 $(DEVICE_DESCRIPTION)
 @@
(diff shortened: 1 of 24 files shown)


dfc-d:
Sorry hit ENTER badly; This patch series of three commits replaces DEVICE_TYPE with adding an appropriate metapackage (collection) for the type of device, either in DEVICE_PACKAGES for new multi-profile devices, or in DEFAULT_PACKAGES for old-style devices.  The special 'collection-none' indicator is used to indicate that either no metapackage should be used, or (for new multi-profile) that DEVICE_PACKAGES metapackage should be used (that is it avoids a metapackage as part of generic DEFAULT_PACKAGES).
This is an RFD and is not yet tested, but I'd like feedback on the approach before spending time on this.

dfc-d:
Oh and obviously my other default packages patch is going to go away.

dfc-d:
I realized the collection-none special case is actually not needed; we can simply explicity specify collection-common-device which includes packages that should be on every device including developer boards and 'other'.  By having 'collection-%' package specified the macro skips adding collection-router, which is otherwise added as a default package (this should theoretically maintain current default behavior; the main devices which should notice a real difference are some nas devices and some APs which will have more appropriate package selections.

nbd168:
The main issue I have with this change at the moment is that it makes it harder to deselect packages that are selected by default. I don't know how to fix that properly yet, but I think it might be useful to have the system default-enable the dependencies of the default collection without enabling the collection itself.
I'll think about this some more.

blogic:
no feedback in 3 weeks and patch does no longer apply, please open a new PR when you have time for it.

PR #186, opened 2016-07-07
[RFD] Skip building images in core and allow IB configuration via menuconfig

dfc-d:
This patch series of three commits adds the option skip building images during the core built while allowing the user an easy way to set an appropriate configuration via menuconfig in imagebuilder.


nbd168:
Can you try adding an if !IB_NO_CORE_IMAGE block around the entire "Target Images" menu instead of putting it in every symbol one by one?

nbd168:
As for the second commit, I don't think it actually makes things better, I'd prefer to leave it out for now. Symbols like TARGET_IMAGE_IMAGES_PAD certainly look weird

dfc-d:
The second commit is so that convert-config.pl can skip CONFIG_IMAGE_ config symbols when creating Config-build.in for preserving user's configuration (since if you didn't build images in core it doesn't make sense to preserve the no image config in the imagebuilder).

nbd168:
I don't think it's actually necessary to skip these symbols. In case of a no-image config, no symbols will be copied over. In case of a config with images, the values will be used as defaults, while still allowing the user to configure them.

dfc-d:
Actually with a recent patch that preserves unset config options as well as set options (so that compilation gets the result in SDK as in core), that is no longer true.

nbd168:
It's still true even with that patch, since unavailable options are not emitted to .config.

dfc-d:
Ah, missed that.  Will rework based on that.

blogic:
4 weeks with no feedback. please reopen when you have time again.

PR #188, opened 2016-07-08
ar71xx: Fix legacy multi-device builds

dfc-d:
ar71xx legacy device handling was incompletely converted
to the new handling for legacy devices resulting in the
fallback designed for unconverted archs being triggered
for kernel image generation, which mean that *every*
legacy kernel image was getting generated when creating
images.  This patch renames the macro that triggered
this behaviour and uses the new legacy hooks for per-profile
handling to build only the desired images.


nbd168:
Image/BuildKernel contains a lot of stuff that should only run once, not once per legacy device.
It seems to me that you should probably only move the buildkernel call at the end of it, instead of renaming the whole thing.

nbd168:
Fixed properly in d6b3b44

dfc-d:
On Wed, 2016-07-13 at 05:00 -0700, nbd168 wrote:
> Fixed properly in d6b3b44

Sorry, got busy.  I actually had a fix (doing as you suggested and only
making the one call per-profile) that I hadn't finished testing before
getting completely swamped.

Regards,

Daniel

PR #190, opened 2016-07-08
brcm63xx: Fix image makefiles

danitool:
image block size and offset variables don't work outside the main
macro. As result of this some images could cause potential bricks,
and in fact it already did it as reported by some people.


KanjiMonster:
I fixed this in a slightly less intrusive way in eae812ddb6cc6edfd253711ebb2250c32c599692.

PR #196, opened 2016-07-13
[RFC] ipq806x: switch to cortex-a15 and optimize compiler flags

bendavid:
Main goal of this change is to use hardware integer division instructions for ipq806x.  These instructions are supported by cortex-a15 but not cortex-a9.  Getting gcc 53x to emit those requires using -mcpu=cortex-a15 (using -march=armv7-a -mtune=cortex-a15 does not work because armv7-a does not include those instructions as a mandatory feature).

I am not really happy patching the kernel makefile to do this, so open to other suggestions how to override the kernel build flags from the lede build system/makefiles.

Underlying motivation for this change is that the emulated integer division is actually a performance hotspot under wired network load, due primarily to the use of the modulo operator in the sttmac ethernet driver (eg here https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/drivers/net/ethernet/stmicro/stmmac/stmmac_main.c?h=v4.4.14#n2175).

This is likely anyway mitigated by driver changes in more recent kernels (see https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/commit/drivers/net/ethernet/stmicro?id=e3ad57c96715df2989ce6c18e58faf2913b305cb) so probably this is less important once updating to 4.6 or later.

@dissent1

### DIFF
--- a/include/target.mk
+++ b/include/target.mk
@@ -191,7 +191,7 @@ ifeq ($(DUMP),1)
     CPU_CFLAGS_cortex-a7 = -march=armv7-a -mtune=cortex-a7
     CPU_CFLAGS_cortex-a8 = -march=armv7-a -mtune=cortex-a8
     CPU_CFLAGS_cortex-a9 = -march=armv7-a -mtune=cortex-a9
-    CPU_CFLAGS_cortex-a15 = -march=armv7-a -mtune=cortex-a15
+    CPU_CFLAGS_cortex-a15 = -mcpu=cortex-a15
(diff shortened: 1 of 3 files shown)


bendavid:
It also may be possible to accomplish this with a different/more optimal set of compiler flags in more recent gcc versions.

ConorOG:
What improvement do you see with benchmarking?

nbd168:
Pushed a modified version of your changes with some extra cleanup into my staging tree

PR #197, opened 2016-07-13
tools: add endian conversion functions for OS X

Noltari:
The endian.h file present in tools/include prevents a successful compilation of dosfstools, since it looks for endian.h presence and it prevents it from including libkern/OSByteOrder.h on OS X.
Should I add a hack to dosfstools or should I add the needed definitions to tools/dosfstools?

### DIFF
--- a/tools/include/endian.h
+++ b/tools/include/endian.h
@@ -7,9 +7,22 @@
 #elif defined(__APPLE__)
 #include <machine/endian.h>
 #include <machine/byte_order.h>
+#include <libkern/OSByteOrder.h>
 #define bswap_16(x) NXSwapShort(x)
 #define bswap_32(x) NXSwapInt(x)
 #define bswap_64(x) NXSwapLongLong(x)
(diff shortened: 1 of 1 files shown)


Noltari:
I've just realized that another possibility is to fix it with:
```
diff --git a/tools/dosfstools/Makefile b/tools/dosfstools/Makefile
index 56a9aa5..c5fbcdb 100644
--- a/tools/dosfstools/Makefile
+++ b/tools/dosfstools/Makefile
@@ -21,6 +21,7 @@ include $(INCLUDE_DIR)/host-build.mk
 
 ifeq ($(HOST_OS),Darwin)
 LDLIBS += -liconv
+HOST_CFLAGS += -UHAVE_ENDIAN_H
 endif
 
 define Host/Compile
```

jow-:
Use the `HOST_CFLAGS` fixup I'd say.

Noltari:
Done!

PR #198, opened 2016-07-14
plan9 filesystem support

teknoraver:
used to share files between guest KVM and host

### DIFF
--- a/include/target.mk
+++ b/include/target.mk
@@ -258,6 +258,9 @@ ifeq ($(DUMP),1)
     ifneq ($(CONFIG_RTC_CLASS),)
       FEATURES += rtc
     endif
+    ifneq ($(CONFIG_VIRTIO),)
+      FEATURES += virtio
+    endif
     FEATURES += $(foreach v,6 7,$(if $(CONFIG_CPU_V$(v)),arm_v$(v)))
 
     # remove duplicates
--- a/package/kernel/linux/modules/fs.mk
+++ b/package/kernel/linux/modules/fs.mk
@@ -26,6 +26,24 @@ endef
 
 $(eval $(call KernelPackage,fs-fscache))
 
+define KernelPackage/fs-9p
+  SUBMENU:=$(FS_MENU)
+  TITLE:=Plan 9 Resource Sharing Support
+  DEPENDS:=+kmod-9pnet
+  KCONFIG:=\
+	CONFIG_9P_FS \
+	CONFIG_9P_FS_POSIX_ACL=n \
+	CONFIG_9P_FS_SECURITY=n
+  FILES:=$(LINUX_DIR)/fs/9p/9p.ko
+  AUTOLOAD:=$(call AutoLoad,30,9p)
+endef
+
+define KernelPackage/fs-9p/description
+  Kernel module for Plan 9 Resource Sharing Support support
+endef
+
+$(eval $(call KernelPackage,fs-9p))
+
 define KernelPackage/fs-afs
   SUBMENU:=$(FS_MENU)
   TITLE:=Andrew FileSystem client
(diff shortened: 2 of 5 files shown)


package/kernel/linux/modules/other.mk
```diff
@@ -1055,3 +1055,39 @@ define KernelPackage/bmp085-spi/description
...
+define KernelPackage/virtio-pci
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=Virtio PCI support
+  KCONFIG:= CONFIG_VIRTIO CONFIG_VIRTIO_PCI
+  FILES:=\
+	$(LINUX_DIR)/drivers/virtio/virtio_pci.ko \
+	$(LINUX_DIR)/drivers/virtio/virtio.ko \
+	$(LINUX_DIR)/drivers/virtio/virtio_ring.ko
+  AUTOLOAD:=$(call AutoProbe,virtio virtio_ring virtio_pci)
+endef
```
blogic: Auto probe should only list the last dependency, in this case virtio_pci. the rest will be probed.

package/kernel/linux/modules/other.mk
```diff
@@ -1055,3 +1055,39 @@ define KernelPackage/bmp085-spi/description
...
+
+
+define KernelPackage/virtio-pci
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=Virtio PCI support
+  KCONFIG:= CONFIG_VIRTIO CONFIG_VIRTIO_PCI
+  FILES:=\
+	$(LINUX_DIR)/drivers/virtio/virtio_pci.ko \
+	$(LINUX_DIR)/drivers/virtio/virtio.ko \
+	$(LINUX_DIR)/drivers/virtio/virtio_ring.ko
```
blogic: virtio and virtio_ring should be in their own kmod. this is duplicated in  the mmio module

package/kernel/linux/modules/other.mk
```diff
@@ -1055,3 +1055,39 @@ define KernelPackage/bmp085-spi/description
...
+	$(LINUX_DIR)/drivers/virtio/virtio_mmio.ko
+  AUTOLOAD:=$(call AutoProbe,virtio virtio_ring virtio_mmio)
+endef
+define KernelPackage/virtio-mmio/description
+ This driver adds virtio MMIO support.
+endef
+
+$(eval $(call KernelPackage,virtio-mmio))
+
+
```
blogic: 2 trailing blank lines

blogic:
the first commit is missing the prefix in the subject and the SoB line

nbd168:
What target(s) do you intend to use this with?

teknoraver:
KVM virtual images, I share files from guest and host passing "-virtfs local,id=r,path=$HOME,security_model=none,mount_tag=home" to qemu cmdline

nbd168:
You should not add kmod packages for virtio then, the kernel config already includes it.

I think it would be a good idea to add a virtio feature flag, which is auto-detected based on the kernel config. You could add that as a dependency for 9p packages.

package/kernel/linux/modules/other.mk
```diff
@@ -1055,3 +1055,47 @@ define KernelPackage/bmp085-spi/description
...
 
 $(eval $(call KernelPackage,bmp085-spi))
+
+
+define KernelPackage/virtio
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=Virtio
+  KCONFIG:= CONFIG_VIRTIO
+  FILES:=$(LINUX_DIR)/drivers/virtio/virtio.ko
+  AUTOLOAD:=$(call AutoProbe,virtio)
```
blogic: this line is not required

nbd168:
Pulled into my staging tree, thanks.

PR #199, opened 2016-07-14
gccgo support, persistent /var

teknoraver:


blogic:
both commits are missing subject prefix, description and SoB

nbd168:
Why do you want /var to be on a persistent filesystem? Also, I think some things running on the device might also need to be audited/changed, because they expect /var to be fresh at boot time.
One thing that comes to mind is dnsmasq dhcp detection.
Also, why golang in the same pull request as persistent /var support?

teknoraver:
I'm adding a daemon written in go, it saves stuff in /var/lib and expects them to be persistent across reboots

nbd168:
I don't want to take a patch that simply makes persistent /var an option, because that option will end up being mostly untested and will probably accumulate some nasty side effects.
What I'd like to see instead is some rework that makes /var persistent by default, but adds appropriate symlinks to /tmp for subdirectories that are supposed to be in RAM and cleared on boot.

stintel:
I like the ability to have a persistent /var, and have been working on it myself but never finished it. I pushed my work: https://git.lede-project.org/?p=lede/stintel/staging.git;a=shortlog;h=refs/heads/var_persist

PR #201, opened 2016-07-15
Roll-up of proposed DGN3500 handing

danielg4:
This is a resubmit of #127, with @KanjiMonster's newer boot code, for @blogic.

### DIFF
--- a/target/linux/brcm63xx/patches-4.4/367-MIPS-BCM63XX-add-support-for-loading-DTB.patch
+++ b/target/linux/brcm63xx/patches-4.4/367-MIPS-BCM63XX-add-support-for-loading-DTB.patch
@@ -110,8 +110,8 @@ Signed-off-by: Jonas Gorski <jogo@openwrt.org>
  	u32 reg, mask;
  
 +#if CONFIG_OF
-+	if (fw_arg0 == -2)
-+		early_init_dt_verify((void *)fw_arg1);
++	if (fw_passed_dtb)
++		early_init_dt_verify((void *)fw_passed_dtb);
 +#endif
 +
  	bcm63xx_cpu_init();
--- a/target/linux/generic/patches-4.4/150-zboot_appended_dtb.patch
+++ b/target/linux/generic/patches-4.4/150-zboot_appended_dtb.patch
@@ -1,0 +1,137 @@
+From patchwork Mon Jun 20 09:27:36 2016
+Content-Type: text/plain; charset="utf-8"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+Subject: [v2,1/2] MIPS: ZBOOT: copy appended dtb to the end of the kernel
+From: Jonas Gorski <jogo@openwrt.org>
+X-Patchwork-Id: 13698
+Message-Id: <1466414857-30456-2-git-send-email-jogo@openwrt.org>
+To: Ralf Baechle <ralf@linux-mips.org>
+Cc: linux-mips@linux-mips.org, Kevin Cernekee <cernekee@gmail.com>,
+ Florian Fainelli <f.fainelli@gmail.com>, John Crispin <john@phrozen.org>,
+ Paul Burton <paul.burton@imgtec.com>,
+ James Hogan <james.hogan@imgtec.com>, Alban Bedel <albeu@free.fr>,
+ Daniel Gimpelevich <daniel@gimpelevich.san-francisco.ca.us>,
+ Antony Pavlov <antonynpavlov@gmail.com>
+Date: Mon, 20 Jun 2016 11:27:36 +0200
+
+Instead of rewriting the arguments, just move the appended dtb to where
+the decompressed kernel expects it. This eliminates the need for special
+casing vmlinuz.bin appended dtb files.
+
+Signed-off-by: Jonas Gorski <jogo@openwrt.org>
+---
+v1 -> v2:
+ * drop #ifdefs and use IS_ENABLED
+ * drop accidentally added empty line
+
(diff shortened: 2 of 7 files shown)


target/linux/generic/patches-4.4/150-device_tree_cmdline.patch
```diff
@@ -0,0 +1,233 @@
+--- a/arch/mips/boot/compressed/decompress.c
```
KanjiMonster: Please keep the patches intact (with full commit log etc). also you need to update brcm63xx (and any other non-upstream users if there are any) as well, else you break them.

danielg4:
@KanjiMonster Can you test brcm63xx?

KanjiMonster:
I can test on brcm63xx once you updated the commit.

danielg4:
Updated with just the patch header?

KanjiMonster:
When you have added the required updates/changes for brcm63xx as well I can test that it works. Currently I can see from the changes alone that brcm63xx would be broken because you didn't touch it.

danielg4:
What exactly would be required on the brcm63xx side?

KanjiMonster:
it needs to use fw_appended_dtb instead of checking fw_arg0 for -2.

target/linux/generic/patches-4.4/150-device_tree_cmdline.patch
```diff
@@ -0,0 +1,233 @@
...
+ 		__dt_setup_arch(__dtb_start);
+--- a/arch/mips/bmips/setup.c
++++ b/arch/mips/bmips/setup.c
+@@ -149,8 +149,8 @@ void __init plat_mem_setup(void)
+ 	/* intended to somewhat resemble ARM; see Documentation/arm/Booting */
+ 	if (fw_arg0 == 0 && fw_arg1 == 0xffffffff)
+ 		dtb = phys_to_virt(fw_arg2);
+-	else if (fw_arg0 == -2) /* UHI interface */
+-		dtb = (void *)fw_arg1;
++	else if (fw_passed_dtb) /* UHI interface */
```
danielg4: @KanjiMonster What about this?

KanjiMonster: brcm63xx is bcm63xx, not bmips.

danielg4: OK. There was nothing else in the patchset you submitted upstream. Do you have downstream patches that were putting it in?

KanjiMonster: There are local patches for brcm63xx that add partial device tree support, so yes. You need to check all targets in lede that are MIPS and use appended dtb, and ensure that they are either updated by the backported patch, or update local patch(es).
Also please keep patches in tact as far as possible and don't merge them into one, and especially don't add non-upstream (submitted) changes to the patches.

danielg4:
Ready for testing.

KanjiMonster:
Tested on brcm63xx, both generic and smp, and both still work.

What's the point of the new revert commit? Can't you just use `CONFIG_MIPS_CMDLINE_DTB_EXTEND`?

danielg4:
No.
1) That would put the DTB bootargs in front of the cmdline instead of after it, so they could not function as an override.
2) Even if that were to put them at the end, changing the kernel config that way would change the meaning of existing DTB's, which @blogic has already NAK'ed, so it would no longer be possible to have one kernel config per target with stuff like this selected in the DTB where it belongs. IMHO, upstream should never have introduced such a mechanism, and this kind of thing should be universally DTB-controlled.

blogic:
patch 2/3 is missing the description of what it does and why we need it. please dont post links to threads for people to look up
patch 3/3 is missing any kind of info of why we should revert 2b4e5d4 and 2b4e5d4 is not properly annotated

KanjiMonster:
After thinking a bit about it, patch 3 should be contained to lantiq (because it's the only user) and be submitted upstream so we don't have yet another patch bitrotting here.

danielg4:
Regarding 2/3, the description of what it does is contained within the kernel patch files it adds to target/generic. The reason it's needed isn't very notable: On some targets (e.g. lantiq) the kernel does not work correctly without additional parameters not given in factory firmware.

Regarding containing 3/3, it's not used in other targets probably due to its obscurity. I don't consider it clean enough to upstream as is, but the bitrot is an obvious concern, especially since it already broke from upstream changes before, while still cleanly applying.

Regarding missing info in 3/3, I'll add it in my next push, assuming the above concerns are addressed.

hauke:
these patches were accepted upstream, are you planing in updating this pull request?

danielg4:
Yes, but life intervened. I _may_ have more time to do this during the coming week.

jow-:
@hauke - care to link those patches? Not sure how to go on here.

danielg4:
Just so you know: I still plan to rebase this when I get a chance, which is now looking like next month at the earliest.

hauke:
Some of the changes from this pull request are in https://git.lede-project.org/?p=source.git;a=commitdiff;h=993ad293597ac3e46574e2ccd074432824305a91

@danielg4 Please rebase on current master branch.

Hauke

danielg4:
I intend to.

hauke:
Some of the changes from this pull request are in 
https://git.lede-project.org/?p=source.git;a=commitdiff;h=993ad293597ac3e46574e2ccd074432824305a91

@danielg4 Please rebase on current master branch.

Hauke

mkresin:
I will close this due to inactivity. Please reopen the PR if you have the time to fix the outlined issues.

PR #203, opened 2016-07-16
add patch to fix ath10k stack traces

bendavid:
Patch comes from the thread here
http://lists.infradead.org/pipermail/ath10k/2016-July/008005.html

Fixes stack traces with multiple QCA99x0 radios which appeared with recent mac80211 updates.


heinzek:
Hi @bendavid. Now i tested Your patch. WiFi on Archer C2600 work amazing when both radio is enabled.
Great Job!!!

nbd168:
merged with a proper description in 7cdb51e0465ed

PR #207, opened 2016-07-18
samba: Remove redundant options.

neheb:
Removed charset since samba defaults to UTF-8. Traditionally this was not the case I believe. guest ok and invalid users are option for a share and not global ones. The former gets applied when configured through LuCI. Also sorted by alphabet.

### DIFF
--- a/package/network/services/samba36/files/samba.init
+++ b/package/network/services/samba36/files/samba.init
@@ -25,20 +25,18 @@ smb_header() {
 		done
 	)
 
-	local name workgroup description charset
+	local name workgroup description
 	local hostname="$(uci_get system.@system[0].hostname)"
 
 	config_get name        $1 name        "${hostname:-Lede}"
 	config_get workgroup   $1 workgroup   "${hostname:-Lede}"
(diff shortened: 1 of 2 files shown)


blogic:
i dont understand why they are redundant. please add a patch description, explaining why this needs to be applied.

blogic:
@neheb ping ?

neheb:
Will force push the changes. also realized the invalid users line was a mistake.

blogic:
it looks like the support for setting a charset other than UTF-8 is removed. are we sure that no one ever wants to use a different charset ?

neheb:
Hmm? The charset was pre-defined to utf-8 in the init script. The relevant
entries can still be added to smb.conf manually.

ignisf:
But why should it not be possible to change the charset from UCI?

blogic:
removing the uci bindig seems odd. please remove that part from your patch

neheb:
so basically the whole init script?

blogic:
yes

blogic:
@neheb ping ?

blogic:
2 weeks with no feedback. please reopen when you have time to implement the changes

PR #208, opened 2016-07-20
ar71xx: add support for gl-mifi

zhao-gang:
This add initial support for gl-mifi.
Product page: http://www.gl-inet.com/mifi/


target/linux/ar71xx/base-files/etc/board.d/01_leds
```diff
@@ -275,6 +275,13 @@ gl-inet)
 	ucidef_set_led_wlan "wlan" "WLAN" "gl-connect:red:wlan" "phy0tpt"
 	;;
 
+gl-mifi)
```
blogic: this is not alphabetically ordered..

blogic:
merged, thanks !

kristrev:
I just tested this commit and noticed an error, causing the boot to fail. There is a ";;" missing in ar71xx.sh.

blogic:
pushed a fix just now, please retest

kristrev:
Looks good, thanks.

PR #211, opened 2016-07-21
ramips: dch-m225 don't have ethernet. default enable wifi

igvtee:
dch-m225 don't have Ethernet port. use uci-default to default enable wifi.

Signed-off-by: Michael Lee <igvtee@gmail.com>


thagabe:
Hello @igvtee could you also send a patch upstream for VoCore? Since the main interface for it is Wifi and only the ones with docks have an ethernet that is actually a Wan port

blogic:
merged, thanks !

PR #217, opened 2016-07-27
ar71xx: Add support for the GL-MiFi

kristrev:
This commit adds support for the GL-MiFi (http://www.gl-inet.com/mifi/), which
is built on the Atheros AR9331 SoC. The CPU is running at 400 MHz and the device
is equipped with 1 WAN port, 1 LAN port, 1 USB port, an internal Mini-PCIe slot
(used for modem), 1 SIM card slot and 2.4 GHz WIFI. The device also has 64MB
RAM, 16MB storage and a micro SD slot.

I have not been able to test the SD card slot, but everything else (WAN, WIFI,
USB, LEDs and internal modem) seems to be working. My work is based on the patch
provided by GL against OpenWRT CC (available here:
https://github.com/domino-team/openwrt-cc/blob/master/patches/007-gl-mifi.patch)

Signed-off-by: Kristian Evensen <kristian.evensen@gmail.com>


blogic:
merged, thanks !

PR #218, opened 2016-07-27
base-files: sysupgrade: fix pseudobridge upgrades

psyke83:
Treat 'relayd' as an essential service to avoid connection interruptions during sysupgrade on devices configured as a pseudobridge.


blogic:
merged, thanks!

PR #221, opened 2016-07-28
uqmi changes

feckert:
1. fix ipv6 config
2. add metric for dhcp

### DIFF
--- a/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
+++ b/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
@@ -31,11 +31,7 @@ proto_qmi_setup() {
 
 	ipv4=1
 
-	if [ "$ipv6" = 0 ]; then
-		ipv6=""
-	else
-		ipv6=1
-	fi
+	[ "$ipv6" = 1 ] || ipv6=""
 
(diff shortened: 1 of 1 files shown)


package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
```diff
@@ -31,7 +31,7 @@ proto_qmi_setup() {
 
 	ipv4=1
 
-	if [ "$ipv6" = 0 ]; then
+	if [ -z "$ipv6" -o "$ipv6" = 0 ]; then
 		ipv6=""
```
blogic: i think that

 [  "$ipv6" = 1 ] ||  ipv6=''

would be better ?

package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
```diff
@@ -37,6 +38,10 @@ proto_qmi_setup() {
 		ipv6=1
 	fi
 
+	if [ "$metric" = "" ]; then
+		metric="0"
+	fi
+
```
blogic: same here check for 1 else set to 0. that will catch all cases

blogic:
merged, thanks for folding the fixes !

PR #224, opened 2016-07-31
hwmon: add SCH5627 Super I/O chips ; x86: enable cpu powersave

lucize:
for x86, enable ondemand cpu governor and frequency scaling  so in idle state cpu will run on the lowest frequency


blogic:
the second commit is not very desriptive. the subject should summarise what the patch does and the description needs to explain why it needs to be done.

lucize:
I don't know how to split pull request, github merged the first with the latest, right now the cpu governor is set to performance, meaning that the cpu will be at the highest frequency all the time, with the config enabled it will run based on how much load is on cpu, so in a normal case the machine will run at the lowest cpu speed sawing watt usage and heat production

blogic:
you can simply call
 $ git commit --amend 
and then edit the commit message. once doen call
 $ git push -f

lucize:
done, thanks !

blogic:
merged, Thanks !

PR #225, opened 2016-07-31
ipq806x: fix MAC_POWER_SEL for Netgear R7800

dissent1:
Fixes instability/corruption on the ethernet interface connected to port0 on the switch on Netgear R7800 as well.

Signed-off-by: Josh Bendavid <joshbendavid@gmail.com>

### DIFF
--- a/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
+++ b/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
@@ -246,7 +246,7 @@
 					0x00004 0x7600000   /* PAD0_MODE */
 					0x00008 0x1000000   /* PAD5_MODE */
 					0x0000c 0x80        /* PAD6_MODE */
-					0x000e4 0xaa545     /* MAC_POWER_SEL */
+					0x000e4 0x6a545     /* MAC_POWER_SEL */
 					0x000e0 0xc74164de  /* SGMII_CTRL */
 					0x0007c 0x4e        /* PORT0_STATUS */
 					0x00094 0x4e        /* PORT6_STATUS */


blogic:
merged ..  we missed that when we fixed the other dts files, Thanks !

dissent1:
@blogic @bendavid @ianchi 
Could you revert this commit please
By a deeper research by dd-wrt team mac_power_sel should be as it was before 0xaa545. Seems that this register depends on specific hardware and not chipset. Maybe some other routers on ipq806x should be taken a deeper look on.

dissent1:
@blogic 
There is a fix for mac_power_sel for r7800 based on dd-wrt source, that doesn't lead to occasional hangs during switch initialisation https://github.com/dissent1/r7800/commit/271965697c8d4850e5e52cb990dd5177d0a89d16
Above commit also adds additional registers that I'm not aware of. Can I add SOB and PR to lede repo?
@bendavid

bendavid:
Is there a discussion on that in ddwrt forums/git/etc somewhere?

dissent1:
The boot hang on current mac_power_sel I experience myself also.
Not much discussion there on details as usual, only commits
http://svn.dd-wrt.com/changeset/30304/src/linux/universal/linux-3.18/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
http://svn.dd-wrt.com/changeset/30408/src/linux/universal/linux-3.18/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
http://svn.dd-wrt.com/changeset/30422/src/linux/universal/linux-3.18/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
http://svn.dd-wrt.com/changeset/30427/src/linux/universal/linux-3.18/arch/arm/boot/dts/qcom-ipq8064-r7800.dts
And couple of posts considering boot hang bug and reversing above register
http://www.dd-wrt.com/phpBB2/viewtopic.php?t=289788&postdays=0&postorder=asc&start=240
Look for my and BrainSlayers posts

bendavid:
Ok, the last commit I don't understand, since the switch registers apply globally to the switch, so it seems there are conflicting values being set now.

Have you or anyone actually been able to reproduce the original problem from https://github.com/lede-project/source/pull/134 with any combination of register settings?

bendavid:
(reproduce on the R7800 I mean)

dissent1:
Well I don't have an ability to produce high Ethernet load at the place where router is located, so I couldn't verify it. But in common home use I didn't have mentioned problem, maybe because of absence of high load profiles at all.
According to S. Gottschal aa545 is also provided by netgear gpl. Because of lack of knowledge I couldn't find it myself.
Considering that I can't understand it myself but it just works, so I started this conversation for blesser minds to take part in it, maybe it could be helpful for other devices as well :)

KanjiMonster:
Not on the R7800, but on the R7500, it's as easy as pingflooding the wan port (the larger the packets. the faster it happens). Eventually after a few minutes there will be a flood of duplicate packets, and then the phy and mac are out of sync. Noticable then with pings suddenly returning a multiple of 1 second.

blogic:
you are adding comments to a closed thread. please move the relevant info into a ne issue so that we can work on it. it will simply get lost here

PR #227, opened 2016-08-02
gcc: optionally build gccgo compiler

teknoraver:
Tested with eglibc on x86 and armv7 so far

Signed-off-by: Matteo Croce <matteo.croce@canonical.com>


nbd168:
Added to my staging tree

PR #230, opened 2016-08-04
samba: bind to all interfaces in LAN firewall zone

psyke83:
Instead of binding to "lan" interface, bind to all interfaces in the LAN firewall zone.

Fixes samba when router is operating as a pseudobridge (as 'bind interfaces only' option will prevent the server from being accessible on the bridged subnet).

### DIFF
--- a/package/network/services/samba36/files/samba.init
+++ b/package/network/services/samba36/files/samba.init
@@ -6,7 +6,7 @@ USE_PROCD=1
 
 smb_header() {
 	local interface
-	config_get interface $1 interface "loopback lan"
+	config_get interface $1 interface "loopback $(uci get firewall.@zone[0].network)"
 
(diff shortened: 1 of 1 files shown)


psyke83:
Sorry, this commit doesn't work properly yet.

Just to clarify: When configured as a pseudobridge, the wwan interface is in the LAN zone, so "$(uci get firewall.@zone[0].network)" resolves to "lan wwan", which results in this configuration in smb.conf:

interfaces = 127.0.0.1/8 lo 192.168.1.1/24 br-lan 192.168.192.3/24 wlan0

The line above is correct and fixes samba, but it doesn't write the proper configuration on first boot, presumably due to wwan not being brought up in time for the samba script. I'll do more checking to see if I can fix this non-intrusively.

PR #232, opened 2016-08-04
hostapd: Allow RADIUS accounting without 802.1x

ignisf:
RADIUS accounting can be used even when RADIUS authentication is not
used. Move the accounting configuration outside of the EAP-exclusive
sections.

Signed-off-by: Petko Bordjukov <bordjukov@gmail.com>


nbd168:
hostapd.sh and wpa_supplicant.sh are gone from the hostapd package now, please drop these changes from your commit. The rest looks good to me.

ignisf:
@nbd168, done and thanks!

nbd168:
Pulled into my staging tree, thanks.

PR #233, opened 2016-08-05
samba: add reload interface triggers

psyke83:
Instead of manually binding to "lan":
* Bind to all interfaces defined in LAN firewall zone
* Skip interfaces which are not up during init_config
* Add network reload triggers for LAN zone interfaces (as not
  all interfaces may be up during first start of service)
* Stop using HUP signal for service reload (as smbd fails to
  honour new interface bindings upon config reload)

Without these changes, Samba fails to work correctly across
a bridged connection due to the 'bind interfaces only' option.

### DIFF
--- a/package/network/services/samba36/files/samba.init
+++ b/package/network/services/samba36/files/samba.init
@@ -5,16 +5,16 @@ START=60
 USE_PROCD=1
 
 smb_header() {
-	local interface
-	config_get interface $1 interface "loopback lan"
+	config_get samba_iface $1 interface "loopback lan"
 
 	# resolve interfaces
 	local interfaces=$(
 		. /lib/functions/network.sh
 
 		local net
-		for net in $interface; do
+		for net in $samba_iface; do
 			local device
+			network_is_up $net || continue
 			network_get_device device "$net" && {
 				local subnet
 				network_get_subnet  subnet "$net" && echo -n "$subnet "
@@ -93,14 +93,13 @@ init_config() {
 	config_foreach smb_add_share sambashare
 }
 
-reload_service() {
-	init_config
-
-	killall -HUP smbd
-}
-
 service_triggers() {
 	procd_add_reload_trigger samba
+
+	local i
+	for i in $samba_iface; do
+		procd_add_reload_interface_trigger $i
+	done
 }
 
 start_service() {
@@ -109,10 +108,12 @@ start_service() {
 	procd_open_instance
 	procd_set_param command /usr/sbin/smbd -F
 	procd_set_param respawn
+	procd_set_param file /var/etc/smb.conf
 	procd_close_instance
 
 	procd_open_instance
 	procd_set_param command /usr/sbin/nmbd -F
 	procd_set_param respawn
+	procd_set_param file /var/etc/smb.conf
 	procd_close_instance
 }


package/network/services/samba36/files/samba.init
```diff
@@ -96,11 +98,16 @@ init_config() {
 reload_service() {
 	init_config
 
-	killall -HUP smbd
+	killall smbd
```
ignisf: You kill samba on reload?

psyke83:
More details on what this solves & potential issues:

My router is set up as a pseudobridge with default lan IP of 192.168.1.1 & wwan IP of 192.168.192.3. Without my changes, smb.conf generates this config:

`interfaces = 127.0.0.1/8 lo 192.168.1.1/24 br-lan
`

Due to the 'bind interfaces only' option which is now set as default, it's impossible to access the Samba share over the bridge via the 192.168.192.3 IP.

With this commit, the config will generate the above config upon first start (as wwan takes more time to come up), but as soon as the wwan interface is up, the network reload trigger will reload the service, giving the proper configuration:

`interfaces = 127.0.0.1/8 lo 192.168.1.1/24 br-lan 192.168.192.3/24 wlan0
`

RE: the HUP signal being removed: despite the correct configuration being written, sending the hangup signal to smbd doesn't trigger the updated "interfaces" config to be applied correctly, so the share continued to be inaccessible upon rebooting the router. Invoking a service reload or restart was ineffectual, only start/stop worked (i.e., fully killing and respawning the smbd process). This is most likely a bug in Samba itself, but I haven't yet looked at the Samba source to see if I can identify the problem.

RFC on some details:
* Should the current "lan_interfaces" code be replaced with a static invocation of $(uci get firewall.@zone[0].network)? I added the more complicated check as I'm not certain that the lan firewall zone is guaranteed to be the same for all default device configurations (as well as possible user customization).
* Should the HUP signal removal be reverted in favour of fixing the problem in Samba itself?

nbd168:
I think binding to the firewall zone instead of the interface names makes the whole thing somewhat inconsistent. What makes samba special, so that it should pick the firewall zone, while other services use the interface directly?

psyke83:
Because "bind interfaces only" disallows clients over a bridge that don't match the lan subnet. Luci/uhttpd etc. doesn't present this kind of issue. I can workaround samba by putting the lan ip in the same subnet as the bridge, but that's strongly  warned against in the pseudobridge recipes on the wiki.

nbd168:
I think in this case it makes more sense to explicitly put the extra network in /etc/config/samba instead of implicitly having the init script pick it up via the firewall zone.

psyke83:
OK, but the script depends on the interface being up to determine the subnet. So editing the config doesn't solve the issue of wwan not being up during first boot. Would need to manually stop/start (not reload, due to HUP signal issue) service after wwan came up.

nbd168:
remove the reload_service() function entirely, add this to the smbd/nmbd instances:
`procd_set_param file /tmp/etc/smb.conf`

psyke83:
Thanks, will test in a few hours when back at my PC. That still won't solve the wwan subnet not being written in smb.conf after first boot, I think.

nbd168:
Right, your trigger change is still required as well. And the check that you added for an interface being up should also be there.

psyke83:
nbd168,

Removing reload_service() entirely and adding the procd_set_param entries works as expected (processes respawn upon config change caused by network trigger).  However, if the proposed solution is to manually edit smb.conf.template (I assume you meant this file instead of /etc/config/samba) instead of enumerating the LAN zone interfaces, I don't think any of this patch makes sense.

For my particular configuration, I could just edit my template to:
`interfaces = |INTERFACES|192.168.192.3/24 wlan0`

Manually setting the wwan configuration here obviates the need for any of this patch, as the proper config will always be generated upon first boot regardless of wwan being up in time... but the purpose of this patch was to help Samba work automatically on a pseudobridge configuration (the simplest solution being to remove the "bind interfaces only" option, but at the cost of potentially reduced security).

Since you're not happy with the idea of utilizing the LAN zone interfaces, shall I abandon the pull request?

nbd168:
It should be configured in /etc/config/samba, not smb.conf.template.
No need to abandon the pull request, you can just update it with the changes I proposed.

psyke83:
Thanks, I wasn't considering the proper procedure to define the samba interfaces via the uci config. I've updated the patch with your proposed changes.

blogic:
the commit mainly consists of the introduction of the $CONFIGFILE variable. is there a specific need for this ? if not then please remove that part and add it in a separate commit. right now that change is cluttering the commit.

psyke83:
Apologies - that was a minor cleanup that should've been in a separate commit. I've removed the variable since it's unrelated to the real purpose of the commit.

package/network/services/samba36/files/samba.init
```diff
@@ -109,10 +108,12 @@ start_service() {
 	procd_open_instance
 	procd_set_param command /usr/sbin/smbd -F
 	procd_set_param respawn
+	procd_set_param file /var/etc/smb.conf
 	procd_close_instance
```
blogic: it seems this is also not related to the description of the patch. either split it into 2 patches or rework the subject/description

package/network/services/samba36/files/samba.init
```diff
@@ -15,6 +14,7 @@ smb_header() {
 		local net
 		for net in $interface; do
 			local device
+			[ "$(network_is_up $net && echo yes)" = "yes" ] || continue
 			network_get_device device "$net" && {
```
blogic: && echo yes seems bad. what does network_is_up return on true/false case ? you should test for this. it looks like you are testing the return code here.

package/network/services/samba36/files/samba.init
psyke83: This line is already in the description:
* Force service stop/start upon config change

As far as I understand, "procd_set_param file" triggers an explicit restart (stop/start) when reload is called only if the file is changed. I'd be happy to change the description to something better - you want me to mention the function?

package/network/services/samba36/files/samba.init
```diff
@@ -5,7 +5,6 @@ START=60
 USE_PROCD=1
 
 smb_header() {
-	local interface
 	config_get interface $1 interface "loopback lan"
```
blogic: this is also fragile. you are moving the interface variable to the global scope as you need id when settign up the triggers. this will work as long as procd.sh does not have a variable called interface. please rename the variable to samba_ifaces or something similar

package/network/services/samba36/files/samba.init
blogic: just remove the "interface" from the subject as you add more than 1 reload trigger

psyke83:
blogic,

I've updated the subject & desc to clarify the details you mentioned, renamed the variable that was moved into global scope, and simplified the network_is_up check (it gives exit code 0 on success, i.e. network is up). I'm not sure if there's a simpler way to check the exit code without adding unnecessary if/then statements, etc.

blogic:
merged, Thanks for handling this so promptly today!

psyke83:
My pleasure. I'm glad to see high standards being maintained!

PR #234, opened 2016-08-05
dnsmasq: fix initscript trigger definition

NebuKatWork:
Previously, the context initialization / close for the raw trigger was omitted, this adds it.


nbd168:
It was omitted intentionally, it is not necessary in current LEDE

PR #235, opened 2016-08-05
base-files: add pkill function to profile

koniu:
This patch adds a simple pkill() function to /etc/profile if a full
procps version is not present.


blogic:
i am not sure this is sane. pkill will do a lot more than what this emulated function does. it will only implement a very minimal subset of pkill. this subset it implement is essentially the same functionality already provided by killall.

NAK from me. i would close this PR unless anyone has a different opinion

PR #236, opened 2016-08-05
[URFC] hostapd: Introduce FST support

ignisf:
802.11ad Fast Session Transfer seems like a very interesting functionality. Has anybody tested it or does anybody know if it can be used in the typical LEDE configuration of separate PHYs with separate hostapd instances for the different bands?

Currently I get the following worrying log entries:

```
root@threemileisland:~# logread | grep FST
Fri Aug  5 18:18:05 2016 daemon.notice netifd: radio1 (3432): FST: lan: wlan1: Zero llt adjusted
Fri Aug  5 18:18:05 2016 daemon.notice netifd: radio1 (3432): FST: lan: wlan1: cannot add MB IE: no backup ifaces
Fri Aug  5 18:18:05 2016 daemon.notice netifd: radio1 (3432): FST: lan: wlan1: cannot create MB IE
Fri Aug  5 18:18:05 2016 daemon.notice netifd: radio1 (3432): FST-EVENT-IFACE attached ifname=wlan1 group=lan
Fri Aug  5 18:18:06 2016 daemon.notice netifd: radio0 (3431): FST: lan: wlan0: Zero llt adjusted
Fri Aug  5 18:18:06 2016 daemon.notice netifd: radio0 (3431): FST: lan: wlan0: cannot add MB IE: no backup ifaces
```

I've started a thread in the hostap mailing list in hope somebody could shed some light on the use cases their implementation has in mind.

### DIFF
--- a/package/network/services/hostapd/files/hostapd-full.config
+++ b/package/network/services/hostapd/files/hostapd-full.config
@@ -166,4 +166,10 @@ CONFIG_NO_DUMP_STATE=y
 CONFIG_WPS=y
 CONFIG_FULL_DYNAMIC_VLAN=y
 
+# Enable Fast Session Transfer (FST)
+CONFIG_FST=y
+
+# Enable CLI commands for FST testing
+CONFIG_FST_TEST=y
+
 CONFIG_UBUS=y
--- a/package/network/services/hostapd/files/netifd.sh
+++ b/package/network/services/hostapd/files/netifd.sh
@@ -57,6 +57,9 @@ hostapd_common_add_device_config() {
 	config_add_boolean country_ie doth
 	config_add_string require_mode
 
+	config_add_string fst_group_id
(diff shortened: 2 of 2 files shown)


ignisf:
More info on the implementation: http://lists.shmoo.com/pipermail/hostap/2014-February/029534.html

ignisf:
Having done some testing and research, the current implementation of FST in hostap seemingly suggests using a single hostapd instance per FST group (that includes interfaces of the same or different bands). This is somewhat contrary to the current approach of having a hostapd instance per interface. I will continue playing around with this.

ignisf:
It also does *not* add the MB IE on the secondary BSSes on a phy.

ignisf:
Closing for now, I have a few things left to try and might open a PR if something comes out of it

PR #241, opened 2016-08-08
hostapd: IAPP: Set SO_REUSEADDR on listening socket

ignisf:
This patch makes it possible to use the Inter-Access Point Protocol
implementation of hostapd in one of the most common scenarios -- when
having more than one interface or BSS.

Upstream tracking: https://patchwork.ozlabs.org/patch/656798/

Signed-off-by: Petko Bordjukov <bordjukov@gmail.com>


ignisf:
As a side note, the upstream patch comes as a follow-up to https://patchwork.ozlabs.org/patch/542556/, which wasn't merged because there were concerns about `SO_REUSEADDR`'s compatibility.

ignisf:
Wrong patch

PR #242, opened 2016-08-08
[v2] hostapd: IAPP: Set SO_REUSEADDR on listening socket

ignisf:
This patch makes it possible to use the Inter-Access Point Protocol
implementation of hostapd in one of the most common scenarios -- when
having more than one interface or BSS.

Upstream tracking: https://patchwork.ozlabs.org/patch/656818/

As a side note, the upstream patch comes as a follow-up to https://patchwork.ozlabs.org/patch/542556/, which wasn't merged because there were concerns about `SO_REUSE`*`PORT`*'s compatibility.


stintel:
LGTM. 1 minor nit, following https://wiki.openwrt.org/doc/devel/patches#naming_patches the patch name should start with 1xx.

ignisf:
Hadn't noticed that, thanks, will change it

ignisf:
Bump,

Merged upstream http://lists.infradead.org/pipermail/hostap/2016-August/036158.html

nbd168:
Pushed a hostapd update to my staging tree

PR #245, opened 2016-08-09
target/toolchain: Fix toolchain packaging without package build

bombadil:
If the toolchain is packaged for later use as external toolchain, the resulting
tarball is created in $BIN_DIR. But without building all packages first that
directory isn't created, hence 'make target/toolchain/compile' fails when
trying to create the toolchain tarball with error "Cannot open: No such file or
directory".

To fix that the $BIN_DIR is created before using it.

Signed-off-by: Micha Lenk <micha@lenk.info>

### DIFF
--- a/target/toolchain/Makefile
+++ b/target/toolchain/Makefile
@@ -58,6 +58,7 @@ $(BIN_DIR)/$(TOOLCHAIN_NAME).tar.bz2: clean
 	find $(TOOLCHAIN_BUILD_DIR) -name .git | $(XARGS) rm -rf
 	find $(TOOLCHAIN_BUILD_DIR) -name .svn | $(XARGS) rm -rf
 	find $(TOOLCHAIN_BUILD_DIR) -name CVS | $(XARGS) rm -rf
+	mkdir -p $(BIN_DIR)
 	(cd $(BUILD_DIR); \
 		tar cfj $@ $(TOOLCHAIN_NAME); \
 	)


target/toolchain/Makefile
```diff
@@ -58,6 +58,7 @@ $(BIN_DIR)/$(TOOLCHAIN_NAME).tar.bz2: clean
 	find $(TOOLCHAIN_BUILD_DIR) -name .git | $(XARGS) rm -rf
 	find $(TOOLCHAIN_BUILD_DIR) -name .svn | $(XARGS) rm -rf
 	find $(TOOLCHAIN_BUILD_DIR) -name CVS | $(XARGS) rm -rf
+	mkdir -p $(shell dirname $@)
```
blogic: would $(BIN_DIR) not work here ?

bombadil: It would work, but it isn't directly used by tar a few lines down either. That's why I preferred to use $@ instead.
If that is preferred, I would happily accept this being amended to
```shell
mkdir -p $(BIN_DIR)
```

blogic:
please use BIN_DIR it is easier to read

target/toolchain/Makefile
bombadil: Fixed in commit 821f707.

bombadil:
I've updated the pull request. @blogic any other feedback?

nbd168:
Applied, thanks

PR #252, opened 2016-08-11
ar71xx:Fix profile name of Mercury MW4530R

981213:
The mw4530r-v1 profile in tp-link.mk is for Mercury MW4530R.There is no such a device called TL-WDR4530. 
Also change MERCURY to Mercury in /lib/ar71xx.sh


blogic:
merged, Thanks

PR #255, opened 2016-08-11
Ramips: Add support for Eylike SS aka HC5661 clone(MT7620A SOC)

Ntemis:
With 16mb flash & 128MB ram and a 802.11b/g/n SoC 2.4 GHz radio
Board name is: Eylike HC5661-v2
Diffs:
1. MAC's are taken from factory partition as we dont have valid Vfac_mac
in our bdinfo partition.
2. Board has support for USB 2.0 and SD Card module(missing)

Signed-off-by: Demetris <ierokipides.dem@gmail.com>


blogic:
patch looks ok but the SoB line is bad. please fix it so we can merge the patch. while at it remove the capitals in the subject line aswell please

PR #256, opened 2016-08-11
dropbear: Add option to hide dropbear version

dedeckeh:
As security precaution and too limit the attack surface based on
the version reported by tools like nmap add a Makefile option
to hide the dropbear version being sent over the wire.

### DIFF
--- a/package/network/services/dropbear/Makefile
+++ b/package/network/services/dropbear/Makefile
@@ -105,6 +105,14 @@ define Build/Configure
 	  mv $(PKG_BUILD_DIR)/options.h.new $(PKG_BUILD_DIR)/options.h || exit 1; \
 	done
 
+	# Force version number as UNKNOWN
+	awk 'BEGIN { rc = 1 } \
+	     /'define[[:space:]]DROPBEAR_VERSION'/ { $$$$3="\"\""; rc = 0 } \
+	     { print } \
+	     END { exit(rc) }' $(PKG_BUILD_DIR)/sysoptions.h \
+	     >$(PKG_BUILD_DIR)/sysoptions.h.new && \
+	mv $(PKG_BUILD_DIR)/sysoptions.h.new $(PKG_BUILD_DIR)/sysoptions.h || exit 1;
+
 	# Enforce rebuild of svr-chansession.c
 	rm -f $(PKG_BUILD_DIR)/svr-chansession.o
 endef


blogic:
same as @nbd168 said fro the dnsmasq patch ... if there is no good reasont o announce the version, we should hide it by default

blogic:
@dedeckeh @nbd168 i would simply hide it and have no option to enable it. not sure if this is sane though.

nbd168:
I'd prefer hiding the version unconditionally

dedeckeh:
@blogic @nbd168 Will rework both dnsmasq and dropbear patches based on this concensus

ynezz:
If I put aside, that this is security by obscurity, this change is going to violate [RFC 4253](https://tools.ietf.org/html/rfc4253#page-4):

> 4.2.  Protocol Version Exchange
> 
>    When the connection has been established, both sides MUST send an
>    identification string.  This identification string MUST be
> 
>       SSH-protoversion-softwareversion SP comments CR LF

and it's going to affect DH key exchange also:

> The part of the identification string preceding the Carriage Return
>  and Line Feed is used in the Diffie-Hellman key exchange.

Exchange key details:

> The following steps are used to exchange a key.  In this, C is the
> client; S is the server; p is a large safe prime; g is a generator
> for a subgroup of GF(p); q is the order of the subgroup; V_S is S's
> identification string; V_C is C's identification string; K_S is S's
> public host key; I_C is C's SSH_MSG_KEXINIT message and I_S is S's
> SSH_MSG_KEXINIT message that have been exchanged before this part
> begins.

dedeckeh:
Agree this offers no security but the intention of the patch is not to expose the dropbear version in use as this most of the time is a bad thing.

The reported software version is dropbear_ without version suffix; unless I'm missing something I don't think this violates the rfc as it's just a string but without the version number

jow-:
I am not opposing the change per-se but how widespread is this form of concealing the version number?

I could imagine that LEDE would stand out in a typical nmap fingerprinting as "that OS without a dropbear version". Granted, you cannot easily infer the version but given that dropbear is rarely updated, that updates tend to come with changes in advertised ciphers and that hiding version numbers is uncommon in other OSes I wonder how big the anonymizing effect might be in the end.

Just something to think about, the actual changes are okay with me.

dedeckeh:
This patch was triggered by a security report advising to conceal the version numbers of the daemons in use to limit the attack surface in an evil environment.
As initial approach I thought to make it configurable and by default disabled as indeed concealing the version number is not widespread in use.
After discussion with @nbd @blogic it was decided to conceal the version number unconditionally unless there was a good reason not to do it.
@jow Would you rather prefer to keep it configurable so LEDE does not stand out as "that OS without dropbear version" ?

jow-:
@dedeckeh - I am undecided, I tend to follow @nbd's and @blogic's reasoning here.

kdarbyshirebryant:
If I were evil cracker in possession of a zero day vuln I'm really not sure I'd bother checking versions being reported...I'd just try my vulnerability.  It smells like 'security through obscurity' to me.  I'd be interested to know the security report in question.

dedeckeh:
@nbd @blogic Updated as discussed

nbd168:
Added to my staging tree, thanks

PR #257, opened 2016-08-11
ipq806x: Fixed spi unexpected irq bug

heinzek:
Fixed spi unexpected irq bug
Added usb-dwc3-of-simple for supporting DWC3 USB

Based on work: Ram Chandra Jangir <rjangi@codeaurora.org>


blogic:
i did the same about 2 hours ago and was just about to start testing it.

i also added DSA support to my update. if ok with you i will close this PR and push my tree tomorrow instead.

blogic:
will push my tree in an hour, please test that tree instead of these patches

heinzek:
spi unexpected irq bug is fixed, but now switch not work.

blogic:
ok, i will push the spi fixes to trunk. my tree has a semi ready switch patch, so it is expected to not work.

PR #258, opened 2016-08-11
dnsmasq: Add option to hide dnsmasq version in full variant

dedeckeh:
As security precaution and to limit the attack surface based on versions
reported by tools like nmap; add a Makefile option to hide the version
reported by dnsmasq in response to name queries for version.bind

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.76
-PKG_RELEASE:=3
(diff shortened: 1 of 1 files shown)


nbd168:
Is there any good reason to ever report the version in name queries? If not, I think it might be a good idea to make hiding the version unconditional, or at least the default.

dedeckeh:
@nbd168 Not aware of any good reason to report the version name in queries unless you want a tool like nmap to recognize explicitly the dnsmasq version in use.
Would you be in favor of hiding the dnsmasq version unconditionally for all three variants or make it unconditionally for the nodhcpv6, dhcpv6 variants while make it possible to disable it in the full variant ?

PR #260, opened 2016-08-12
ramips: Add support for Eylike SS aka HC5661 clone(MT7620A SOC)

Ntemis:
With 16mb flash & 128mb ram and a 802.11b/g/n SoC 2.4 GHz radio
Board name is: Eylike HC5661-v2
Diffs:
1. MAC's are taken from factory partition as we dont have valid Vfac_mac
in our bdinfo partition.
2. Board has support for USB 2.0 and SD Card module(missing)

Signed-off-by: Demetris Ierokipides <ierokipides.dem@gmail.com>

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -187,6 +187,7 @@ ramips_setup_interfaces()
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "6@eth0"
 		;;
 	hc5*61|\
+	hc5661-v2|\
 	y1s)
 		ucidef_add_switch "switch0" \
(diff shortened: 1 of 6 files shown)


target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -187,6 +187,7 @@ ramips_setup_interfaces()
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "6@eth0"
 		;;
 	hc5*61|\
+	hc5661-v2)
 	y1s)
```
blogic: this cant be right.

Ntemis: oops
Closing down, will sent a fix asap :)

PR #261, opened 2016-08-13
lantiq: add kmod-ledtrig-usbdev to device packages which have usb led

malaakso:
By default USB LEDs are associated with usbdev trigger. However,
not all devices which have USB LEDs have kmod-ledtrig-usbdev in their
list of default packages. Add it to the relevant devices.

Signed-off-by: Matti Laakso <matti.laakso@outlook.com>


blogic:
merged, thanks !

PR #264, opened 2016-08-13
config busybox/config: Add a meta-option to enable LXC support

cshoredaniel:
Supporting lxc requires a number of configuration options;
add a meta option that selects all options required for lxc
support


blogic:
NAK, this is a random symbol selecting other random symbols. if at all, KERNEL_LXC_MISC should select NS and cgroups

PR #266, opened 2016-08-13
network/services/uhttpd: Add Basic Auth config

cshoredaniel:

httpauth section names are given included as list
items to the instances to which they are to be applied.

Further any existing httpd.conf file (really whatever
is configured in the instane with a default of
/etc/httpd.conf) is appended to the per-instance httpd.conf

Signed-off-by: Daniel Dickinson lede@cshore.thecshore.com

This has been updated per comments on previous version of this PR and I have a corresponding LuCI patch.
This has been updated per comments on previous PR that was closed as I was away for a while.  I have a corresponding LuCI patch.

### DIFF
--- a/package/network/services/uhttpd/files/uhttpd.config
+++ b/package/network/services/uhttpd/files/uhttpd.config
@@ -103,6 +103,11 @@ config uhttpd main
 	# except for development and debug purposes!
 #	option no_ubusauth	0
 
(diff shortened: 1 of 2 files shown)


cshoredaniel:
Updated; any comments?

cshoredaniel:
Any comments?

jow-:
Merged into my staging tree: http://git.lede-project.org/736442f

PR #267, opened 2016-08-15
Update tools/gmp and libgmp to 6.1.1

hnyman:
Update both tools/gmp and libgmp from 6.1.0 to 6.1.1

Minor update, release notes at https://gmplib.org/gmp6.1.html


blogic:
merged, thanks !

PR #273, opened 2016-08-16
fix building with glibc

Artox:
There are a few occasions where stdint.h is not explictly included.
Apparently musl includes it by default, but glibc does not.

Signed-off-by: Josua Mayer <josua.mayer97@gmail.com>


blogic:
this needs to be split into 2 commits please

Artox:
> this needs to be split into 2 commits please

Done

blogic:
merged, Thanks !

PR #274, opened 2016-08-16
Introduce full support for the Clearfog board

Artox:
The easiest method for installing anything on the clearfog is writing an sdcard image to a real sdcard, and bnooting it.
This patchset adds the required pieces for building such images.


KanjiMonster:
which clearfog is this? base? pro?

target/linux/mvebu/config-4.4
```diff
@@ -403,3 +403,11 @@ CONFIG_ZBOOT_ROM_TEXT=0x0
...
 CONFIG_ZLIB_INFLATE=y
 CONFIG_ZONE_DMA_FLAG=0
+CONFIG_MMC_BLOCK=y
+CONFIG_MMC_SDHCI=y
+CONFIG_MMC_SDHCI_PLTFM=y
+CONFIG_MMC_SDHCI_PXAV3=y
+CONFIG_MMC_SDHCI_F_SDH30=n
+CONFIG_MMC_SDHCI_OF_ARASAN=n
+CONFIG_MMC_SDHCI_OF_AT91=n
+CONFIG_MMC_SDHCI_PCI=n
```
KanjiMonster: Please do a `make kernel_defconfig` to ensure the config symbols are alphabetically ordered.

Artox:
It is the Clearfog Pro. My colleague jsut got a base and is working on it. The startup procedure should however be exactly the same.

target/linux/mvebu/config-4.4
Artox: Woudl you mind pointing me at information on this make target?
(I really have no idea what kernel_defconfig is, sorry)

Artox: edit: I think i found it: make savedefconfi

KanjiMonster: I'm talking about the openwrt make target. You have the target `kernel_oldconfig`, `kernel_menuconfig` and `kernel_defconfig` available that work similar to the `oldconfig`, `menuconfig` and `defconfig` targets, except that they work on the kernel config, and update the config in `target/linux/<target>/config-<version>`.

KanjiMonster:
> It is the Clearfog Pro. 

Then this should be also called as such. Since base and pro have different hardware setup, we will need different dts files and consequently different firmware images.

Artox:
> Then this should be also called as such. Since base and pro have different hardware setup, we will need different dts files and consequently different firmware images.
I disagree. The hardware is extremely similar; it is a matter of switch or no switch mostly.
If a different DTB is needed, it can be added. I see no reason why there shouldn't be multiple DTBs present in one image.

target/linux/mvebu/Makefile
```diff
@@ -11,7 +11,7 @@ BOARD:=mvebu
 BOARDNAME:=Marvell Armada 37x/38x/XP
 FEATURES:=usb pci pcie gpio nand squashfs
 CPU_TYPE:=cortex-a9
-CPU_SUBTYPE:=vfpv3
+CPU_SUBTYPE:=neon
```
KanjiMonster: As far as I can tell this will break any Armada XP devices like the WRT1900ACv1, as they don't support Neon. E.g. http://www.marvell.com/embedded-processors/assets/Marvell_ARMADA_SOC_Prestera_Solutions_Brochure_4web.pdf  describes the Armada 38x as "Single or dual core ARM v7-Coretex-A9 CPU with NEON SIMD and Point Unit (FPU)", while Armada XP is described as "ARM v7-compliant Core Processor (CPU) that includes Marvell micro-architecture enhancements and a double-precision IEEE-compliant Floating Point Unit (FPU) per core" without any mention of NEON support.

Artox: Good point. So this should probably be seperated in the group of boards with neon, and without.

KanjiMonster:
> The hardware is extremely similar; it is a matter of switch or no switch mostly.

There are more differences than just the switch: The base has less PCIe ports, no mSATA port, only one Ethernet port (vs two on the Pro), no Audio/TDM support, no GPIO header and no fan. So we definitely need two different dts. Also from a userland perspective we need to be able to tell them apart for appropriate network config.

> If a different DTB is needed, it can be added. I see no reason why there shouldn't be multiple DTBs present in one image.

And is the decision done which one to load? Currently we would need two different u-boot setups which would mean two different images which means ... we could as well treat it as two different devices.

Artox:
> And is the decision done which one to load? Currently we would need two different u-boot setups which would mean two different images which means ... we could as well treat it as two different devices.
Once U-Boot isn't broken anymore, there should be the standard fdt_file variable (I have to check for the correct spelling). Ultimately that should be used by the boot script.
Unfortunately right now u-boot default environment is completely empty. No load addresses, no board names, nothing.
So this is the goal for the boot script: In the end the only hardcoded thing should be bootargs to define root and overlay device.

Artox:
> Also from a userland perspective we need to be able to tell them apart for appropriate network config.
This is true. maybe we can find a solution there. I am not a fan of building a new image just for a slightly different board.

> and no fan
The Pro is already fanless.

KanjiMonster:
> Once U-Boot isn't broken anymore, there should be the standard fdt_file variable (I have to check for the correct spelling). Ultimately that should be used by the boot script.

But where would that variable's contents come from? Wouldn't the uboot environment also be stored on the sd card?

Artox:
> But where would that variable's contents come from? Wouldn't the uboot environment also be stored on the sd card?

It will be defined in the U-Boot source code. There is a board_detect function somewhere, and it allows u-boot to set the right addresses for the board it is running on.
This runtime detection is of course limited to the boards supported by a single u-boot build.

KanjiMonster:
> It will be defined in the U-Boot source code. There is a board_detect function somewhere, and it allows u-boot to set the right addresses for the board it is running on.

This assumes there will be a way of telling them apart from the software side. I think until we know there is such a way we should keep this as as strictly "pro", and if we find out later there is a way we can still rename/merge it.

Artox:
> This assumes there will be a way of telling them apart from the software side. I think until we know there is such a way we should keep this as as strictly "pro", and if we find out later there is a way we can still rename/merge it.

I checked the schematics of both the Clearfog Base, and teh Clearfog Pro:
[](http://wiki.solid-run.com/lib/exe/fetch.php?media=a38x:carrierboard:docs:clearfog_base-schematics-1.1.pdf)
[](http://wiki.solid-run.com/lib/exe/fetch.php?media=a38x:carrierboard:docs:clearfog-schematics-2.1.pdf)

Base schematics page 2 has B2B_MPP50 describes as "Board identifier", that should be enough to tell it apart from the Pro.

nbd168:
Pulled missing commits into my staging tree and cleaned them up. Still quite a bit of work that needs to be done, but at least the image boots the kernel now.

PR #275, opened 2016-08-17
fix kernel localversion for external kernel pulled with git

Artox:
Kernel Makefiles introduce all sorts of localversion autogeneration.
However the lede build system always installes modules to /lib/modules/\<VERSION\>/ (e.g. 4.4.15).

This means a kernel built from a git repo is not able to load any modules, because the module loader will look in the wrong place (/lib/modules/$(uname -r)/)


jow-:
LGTM

blogic:
merged, Thanks !

PR #276, opened 2016-08-17
ar71xx: add support for Zbtlink ZBT-WE1526

pepe2k:
The following patch adds support for **Zbtlink ZBT-WE1526** (**QCA9531** based) under ar71xx target.


blogic:
merged, Thanks !

PR #277, opened 2016-08-18
Generate switch configuration for BTHOMEHUBV2B

shogsbro:
For BT Homehub V2B, the file `/etc/board.json` is not populated with switch information, leading to a [crash in LuCI](https://github.com/openwrt/luci/issues/779) when trying to view Network->Switch.

### DIFF
--- a/target/linux/lantiq/base-files/etc/board.d/02_network
+++ b/target/linux/lantiq/base-files/etc/board.d/02_network
@@ -122,6 +122,8 @@ BTHOMEHUBV2B|BTHOMEHUBV3A)
 	lan_mac=$(mtd_get_mac_ascii uboot_env ethaddr)
 	wan_mac=$(macaddr_add "$lan_mac" 1)
 	ucidef_set_interface_lan 'eth0'
+	ucidef_add_switch "switch0" \
+		"1:lan:1" "2:lan:2" "3:lan:3" "4:lan:4" "5t@eth0"
 	;;
 
 BTHOMEHUBV5A)


blogic:
please add a Signed-off-by: line to the commit

mkresin:
Would you please confirm that the bridge port to physical port mapping is correct:

"<switchport>:lan:<physicalport>"

Means, RJ45 port 1 might be connected to switch port 3. Use swconfig to see which switch port gets a link when plugging in a cable.

jow-:
This board has indeed no port 0 ? Also you can leave out the numeric suffix if it is the same as the port number or if it does not affect the order, e.g. the following cases are all equivalent:

`"1:lan:1" "2:lan:2" "3:lan:3" "4:lan:4" "5t@eth0"`
`"1:lan" "2:lan" "3:lan" "4:lan" "5t@eth0"`
`"0:lan" "1:lan" "2:lan" "3:lan" "5t@eth0"`

shogsbro:
There are 6 ports in output of `swconfig dev eth0 show`. Port 5 is always UP, so I assume that is the CPU. When I connect a UTP cable to each of the ports labelled 1-4 on the back, the same ports in the output show as UP.

jow-:
Ok, so port 0 dead (unpopulated), 1-4 correspond to "LAN1" .. "LAN4" on the case and 5, correct?

shogsbro:
@jow-, would you like me to simplify the patch according to your explanation?

target/linux/lantiq/base-files/etc/board.d/02_network
```diff
@@ -122,6 +122,8 @@ BTHOMEHUBV2B|BTHOMEHUBV3A)
 	lan_mac=$(mtd_get_mac_ascii uboot_env ethaddr)
 	wan_mac=$(macaddr_add "$lan_mac" 1)
 	ucidef_set_interface_lan 'eth0'
```
mkresin: I might be wrong, but he ucidef_set_interface_lan should be removed as soon as you have a vlan setup.

shogsbro: Hmm, let me try that on the device.

mkresin: add the change to your last commit: ```git commit --amend target/linux/lantiq/base-files/etc/board.d/02_network```
force push to your github repo: ```git push --force```

Your PR should update automatically. 

But never do this with a repo where more than one person is working with!

shogsbro: Ah, that would have been easier. Yes, I now about the downsides of force pushing for anything cloned by others 👍

PR #278, opened 2016-08-18
Generate switch configuration for BTHOMEHUBV2B

shogsbro:
Signed-off-by: Sigurd Hogsbro <sigurd@hogsbro.org>


shogsbro:
Updated PR with feedback from #277

jow-:
LGTM (if the V3A has indeed the same layout).

shogsbro:
I unfortunately do not have a V3A to test with. Hmm, looks like it has [less ports](https://en.wikipedia.org/wiki/BT_Home_Hub) 👎

PR #280, opened 2016-08-18
Generate switch configuration for BTHOMEHUBV2B.

shogsbro:
Split from HH3A as it appears to have a different switch.
https://en.wikipedia.org/wiki/BT_Home_Hub#Models_and_technical_specifications

Signed-off-by: Sigurd Hogsbro <sigurd@hogsbro.org>

Updated PR based on feedback in #277 , #278.


blogic:
merged, Thanks !

PR #284, opened 2016-08-19
[RFC] ipq806x: fix (v2) MAC_POWER_SEL for Netgear R7800 and extra registers

dissent1:
Fixes instability/corruption on the ethernet interface connected to port0 on the switch on Netgear R7800.

When defining MAC_POWER_SEL register as 0x6a545 under phy0 the router occasionally hangs during the boot sequence on switch initialisation stage if specific devices are connected (i.e. 100 mbit only).
Workaround by dd-wrt devs seems to solve the problem.

Also adding extra switch registers found by dd-wrt team as well.

I have no idea what extra registers are for and why MAC_POWER_SEL is defined twice, but it solves the mentioned problem.
@bendavid @KanjiMonster @ianchi


dissent1:
#134 #225

dissent1:
Here they are
`-/* Queue Management Registers */ 
-#define S17_PORT0_HOL_CTRL0             0x0970 
-#define S17_PORT0_HOL_CTRL1             0x0974 
-#define S17_PORT1_HOL_CTRL0             0x0978 
-#define S17_PORT1_HOL_CTRL1             0x097c 
-#define S17_PORT2_HOL_CTRL0             0x0980 
-#define S17_PORT2_HOL_CTRL1             0x0984 
-#define S17_PORT3_HOL_CTRL0             0x0988 
-#define S17_PORT3_HOL_CTRL1             0x098c 
-#define S17_PORT4_HOL_CTRL0             0x0990 
-#define S17_PORT4_HOL_CTRL1             0x0994 
-#define S17_PORT5_HOL_CTRL0             0x0998 
-#define S17_PORT5_HOL_CTRL1             0x099c 
-#define S17_PORT6_HOL_CTRL0             0x09a0 
-#define S17_PORT6_HOL_CTRL1             0x09a4 `

diizzyy:
Is this still needed after https://git.lede-project.org/?p=source.git;a=commit;h=d7e4b9babb6ce8bf66c4c2e721b78c30d09afdda ?

dissent1:
Well it's different.
Just got a hang during the boot again, so those changes do not fix it decisively.
It's not ok to push it further. Will see if there's another fix.

TeutonJon78:
You'd want to check with @bendavid, as he had done a commit that exactly opposite of what this one is: https://github.com/lede-project/source/commit/66ffd0ddf9d06a5a72ca72ed19636d20eab95c64

bendavid:
As already discussed elsewhere, these settings may be implementation dependent and the correct settings for the R7800 may indeed be different than the C2600.

At the time it was assumed that C2600 was close to the reference design and that most of the other ipq806x routers with 4+1 wired ethernet ports should be similar.

dissent1:
This doesn't solve anything. Closing the commit.

blogic:
well this commit actually syncs with the SDK code. it might not be all that is required to make it fully reliable, but its not wrong.

bendavid:
Can you point to where this was added in the sdk?

dissent1:
Updated gmac according to gpl

dissent1:
Lede images have started to trigger a boot loop for me even previous builds that were booting fine before. Stock and ddwrt boot fine though.
Any idea what could be the reason?

dissent1:
Weird, but solved by itself.
Any idea why these lines appear in dmesg?
`[   2.324394] of_get_named_gpiod_flags: can't parse 'link-gpios' property of node '/soc/ethernet@37200000/fixed-link[0]'
[    2.359492] of_get_named_gpiod_flags: can't parse 'link-gpios' property of node '/soc/ethernet@37400000/fixed-link[0]' `
Can't be that fixed link should be removed from gmac in dts. Can it?

dissent1:
@blogic I guess it's ok to push now, if everything's ok with above commits

dissent1:
Don't pull yet, seems there are some issues and I have something more pending

ahmadrasyid10:
Sorry for my bad english 😙 
I hope LEDE support for TP-LINK MR3420V3 


Thank's

PR #288, opened 2016-08-20
cyassl: make CyaSSL/WolfSSL more configurable

RoadRunnr:
The default configuration might not be suitable for
every use case. Add options to enable/disable additional
options.

Signed-off-by: Andreas Schultz <aschultz@tpip.net>


blogic:
merged into my staging tree. i slightly changed the defaults so that they are the same as prior to applying the patch

PR #290, opened 2016-08-22
dnsmasq: remove pid file after service has been stopped

dibdot:
no stale pid file after dnsmasq full stop

Signed-off-by: Dirk Brenken dev@brenken.org


kdarbyshirebryant:
The real problem here is that dnsmasq is running as dnsmasq:dnsmasq but /var/run/dnsmasq is owned by root.  Dnsmasq itself would remove the pid file itself if it had permission to do so...it starts as 'root', creates the pidfile (in a directory created by the startup script..as root) and then drops to dnsmasq:dnsmasq.  Adding 'chown dnsmasq:dnsmasq /var/run/dnsmasq' to 'dnsmasq.init' would be preferable in my opinion.

nbd168:
Merged #318 instead.

PR #291, opened 2016-08-22
ath10k-firmware: add QCA9887 firmware

zhao-gang:
QCA9887 is experimentally supported in compat-wireless-2016-06-20.


nbd168:
Added to my staging tree

PR #292, opened 2016-08-23
build: file_copy now dereferences symlinks so that USER_FILES and its…

kevle:
… contents may be symlinks


kevle:
In order to build for multiple targets we exchange the .config and create a symlink for "files" in our environment for each one.

nbd168:
I think following symlinks within files/ or USER_FILES/ is the wrong thing to do, because that makes it impossible to add symlinks to the rootfs. Having files or USER_FILES themselves be symlinks should already be supported, since the code uses files/. or USER_FILES/*

kevle:
You're probably right that this is not a common use case and having actual symlinks might be more desirable.

PR #189, opened 2016-07-08
ar71xx: add support for Wallys DR531 and jjPlus JWAP230

pepe2k:
The following series adds support for **Wallys DR531** (QCA9531 based) and **jjPlus JWAP230** (QCA9558+QCA8337 based) under ar71xx target.

PR #297, opened 2016-08-24
ar71xx: two trivial patches

pepe2k:
Two minor changes:

1. Mark U-Boot and radio calibration data mtd partitions as read-only, as it's the general convention.
2. Update kernel config symbols.

Second one was previously rejected/omitted, but without any specific reason and answer for my comments/doubts, please take a look: https://github.com/lede-project/source/pull/189#issuecomment-232297229


blogic:
merged into my staging tree, sorry for the delay

pepe2k:
Perfect, thanks!

PR #298, opened 2016-08-25
dropbear: mdns flag is a bool, not integer

karlp:
Effectively the same for most purposes, but more accurate.

Signed-off-by: Karl Palsson <karlp@etactica.com>


blogic:
merged, Thanks !

PR #304, opened 2016-08-26
Some trivial changes

nwf:
Enable a few new kernel features and some scattered compilation fixes.

### DIFF
--- a/config/Config-kernel.in
+++ b/config/Config-kernel.in
@@ -251,6 +251,14 @@ if KERNEL_DEVTMPFS
 
 endif
 
+config KERNEL_KEYS
+    bool
+    default n
+
+config KERNEL_KEYS_DEBUG_PROC_KEYS
+    bool
+    default n
+
 #
 # CGROUP support symbols
 #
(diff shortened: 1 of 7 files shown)


nbd168:
Many of these commits lack an explanation describing why you made the change.
Also, if you're removing code, really remove it instead of commenting it out.

nwf:
OK, I'll get rid of the line from the kafs patch.

Is the commentary for host-dep python clear enough as it stands?

I no longer have the build failure logs for the musl or udev changes, but can reproduce them if that'd be desired.

For 9999, kmod_fou, and keyring, is additional commentary necessary?

nbd168:
9999 and many other patches need a prefix, e.g. build:
keyring needs anexplanation why you added that, it's not obvious at all
I guess fou is fine
"Stop autoloading kafs" makes no sense to me.
The host-dep python patch makes no sense to me, since tools/ already requires python.

nwf:
OK, the python one might require some more thought.  In particular, the problem is that some packages build-dep on host/python.  Thus, during the build, the meaning of "python" can change from the host's natively-installed python to the host/python we just built.

I'll comment on keyring and add prefixes.

blogic:
All patches are missing the SoB line

PR #307, opened 2016-08-30
openssl: enable asm opts in openssl for ipq806x

dissent1:
As assembler optimisations are known to bring some issues on specific cpus, let's enable it for those boards that are ok with it.
These optimisations bring severe performance boost for ipq806x boards as neon fpu seem to thread crypto operations by itself this way.

Could those who have c2600 and ea8500 compare the performance boost as well, please.

### DIFF
--- a/package/libs/openssl/Makefile
+++ b/package/libs/openssl/Makefile
@@ -156,6 +156,8 @@ else
     OPENSSL_TARGET:=linux-mips-openwrt
 #  else ifeq ($(CONFIG_arm)$(CONFIG_armeb),y)
 #    OPENSSL_TARGET:=linux-armv4-openwrt
+  else ifneq ($(findstring cortex,$(CONFIG_CPU_TYPE)),)
+    OPENSSL_TARGET:=linux-armv4-openwrt
   else
     OPENSSL_TARGET:=linux-generic-openwrt
     OPENSSL_OPTIONS+=no-perlasm


package/libs/openssl/Makefile
```diff
@@ -156,6 +156,8 @@ else
     OPENSSL_TARGET:=linux-mips-openwrt
 #  else ifeq ($(CONFIG_arm)$(CONFIG_armeb),y)
 #    OPENSSL_TARGET:=linux-armv4-openwrt
+  else ifeq ($(BOARD),ipq806x)
+    OPENSSL_TARGET:=linux-armv4-openwrt
```
KanjiMonster: Can you make this a Kconfig option in Config.in that depends on arm || armeb and defaults to y for TARGET_ipq806x ? That way it's easier to test for others (and configurable).

Also please add `PKG_FLAGS:=nonshared` to the Makefile.

nbd168: I think this should not be based on the target, it should be based on the CPU type

KanjiMonster: > I think this should not be based on the target, it should be based on the CPU type

That might be even better, then we don't need the nonshared flag.

nbd168:
Why only a15? I think it makes more sense to use it at least on all cortex cpus, maybe on all ARMv7

dissent1:
Okay, no problem, you scared me that there are issues on other cpus, so I put in only the one that I know is ok. I'll adjust the commit to armv7

nbd168:
To be sure I just took a look at the OpenSSL assembly code. It seems to rely on NEON for all ARMv7 for some reason, so I guess you need to check the CPU subtype.

KanjiMonster:
> It seems to rely on NEON for all ARMv7 for some reason, so I guess you need to check the CPU subtype.

That's what I  thought at first, too, but it seems it does runtime detection of features, see [crypto/armcap.c](https://github.com/openssl/openssl/blob/master/crypto/armcap.c)

dissent1:
I'm not sure that neon is mandatory, it benefits from neon the most yes, but it's not a must afaik

nbd168:
I will run a quick test on a non-NEON ARMv7 device to get a conclusive answer.

nbd168:
Tracked down the original issue why ARM assembly was disabled, it was crashing on Banana Pi (sunxi target). I verified that the current version works just fine on that device with assembly enabled.
My test on another non-NEON device (mvebu) also went through fine, so I guess we can enable ARM assembly globally again.

dissent1:
Should I modify the commit or it's up to you to reenable it?

nbd168:
Pushed a commit for that into my staging tree

PR #308, opened 2016-08-31
Add config symbols for kernel keyring support

nwf:
Enable selection of the kernel key retention framework and some of its
additional facilities; see Documentation/security/keys.txt and
security/keys/Kconfig for details

Signed-off-by: Nathaniel Wesley Filardo <nwfilardo@gmail.com>

### DIFF
--- a/config/Config-kernel.in
+++ b/config/Config-kernel.in
@@ -251,6 +251,25 @@ if KERNEL_DEVTMPFS
 
 endif
 
+config KERNEL_KEYS
+    bool "Enable kernel access key retention support"
+    default n
+
+config KERNEL_PERSISTENT_KEYRINGS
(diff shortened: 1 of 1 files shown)


nbd168:
What do you plan on using this for?

nwf:
I have Kerberos set up at home, and have patched packages' utils/krb5 to build against keyutils to avail itself of keyrings rather than files, which I think is an improvement.  The keyring infrastructure is also useful for in-kernel NFS and CIFS clients.  I have, un-pushed to packages as yet, also, a build of OpenAFS, whose client uses the keyring infrastructure to manage sessions' credentials.

I'll prep those patches to go upstream as I get time; I thought this one was reasonably self-contained and un-objectionable.

nbd168:
Added to my staging tree

PR #310, opened 2016-09-02
dnsmasq: Don't report version number

dedeckeh:
Don't expose the dnsmasq version number (e.g in response to name queries
for the domain version.bind) to limit the attack surface


kdarbyshirebryant:
Hmm, it looks to me like this patches out the version number from dnsmasq entirely, so not only does it remove visibility of version externally but now we can't see it internally either (eg. console log system startup)

nbd168:
You can always get it via opkg list_installed

nbd168:
Pushed to my staging tree with some cleanup

PR #311, opened 2016-09-02
dnsmasq: Add match section support

dedeckeh:
Match sections allow to set a tag specified by the option networkid if the client
sends an option and optionally the option value specified by the match option.
The force option will convert the dhcp-option to force-dhcp-option if set to 1 in
the dnsmasq config if options are specified in the dhcp_option option.

config match
    option networkid tag
    option match 12,myhost
    option force 1
    list dhcp_option '3,192.168.1.1'

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.76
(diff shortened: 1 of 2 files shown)


dedeckeh:
Ping ?
Any feedback on the patch as it extends the uci flexibility to classify and assign individual options to any dhcp option besides the already existing uci support via mac/userclass/vendorclass/circuitid/remoteid/subscrid.

blogic:

PR #312, opened 2016-09-02
gre: add "tap" to the prefix of GRE-TAP IPv4/6 tunnel interfaces.

rogerpueyo:
This commit modifies the /lib/netifd/proto/gre.sh script so that, when
GRE-TAP tunnels are created, either IPv4 or IPv6, the prefix before the chosen
interface name contains the "tap" substring, to differentiate them from non-TAP
GRE tunnels.

Right now, both GRE and GRE-TAP tunnel (either IPv4 or IPv6) interfaces defined
in /etc/config/network are named "gre-"+$ifname upon creation. For instance,
the following tunnels:

```
    config interface 'tuna'
            option peeraddr '172.30.22.1'
            option proto 'gre'

    config interface 'tunb'
            option peeraddr '192.168.233.4'
            option proto 'gretap'

    config interface 'tunc'
            option peer6addr 'fdc5:7c9e:e93d:45af::1'
            option proto 'grev6'

    config interface 'tund'
            option peer6addr 'fdc0:6071:1348:31ff::2'
            option proto 'grev6tap'
```

are named, respectively, "gre-tuna", "gre-tunb", "grev6-tunc" and "grev6-tund".

The current change makes that each GRE tunnel interfaces of the four different
types available (gre, gretap, grev6 and grev6tap) gets the full protocol name.

This is coherent with other types of virtual interfaces (i.e. PPP, PPPoE, PPPoA)
where the whole protocol name is used. For instance, a PPPoA interface named
"p1" and a PPPoE interface named "p2" will respectively appear as "pppoa-pa"
and "pppoe-p2", not as "ppp-p1" and "ppp-p2").

### DIFF
--- a/package/network/config/gre/Makefile
+++ b/package/network/config/gre/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gre
 PKG_VERSION:=1
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 PKG_LICENSE:=GPL-2.0
 
 include $(INCLUDE_DIR)/package.mk
--- a/package/network/config/gre/files/gre.sh
+++ b/package/network/config/gre/files/gre.sh
@@ -71,7 +71,14 @@ gre_setup() {
 
 	[ -z "$df" ] && df="1"
 
-	gre_generic_setup $cfg $mode $ipaddr $peeraddr "gre-$cfg"
+	case "$mode" in
+		gretapip)
+			gre_generic_setup $cfg $mode $ipaddr $peeraddr "gre4t-$cfg"
+			;;
+		*)
+			gre_generic_setup $cfg $mode $ipaddr $peeraddr "gre4-$cfg"
+			;;
+	esac
 }
 
 proto_gre_setup() {
@@ -89,7 +96,7 @@ proto_gretap_setup() {
 	gre_setup $cfg "gretapip"
 
 	json_init
-	json_add_string name "gre-$cfg"
+	json_add_string name "gre4t-$cfg"
 	json_add_boolean link-ext 0
 	json_close_object
 
@@ -129,7 +136,14 @@ grev6_setup() {
 		fi
 	}
 
-	gre_generic_setup $cfg $mode $ip6addr $peer6addr "grev6-$cfg"
+	case "$mode" in
+		gretapip6)
+			gre_generic_setup $cfg $mode $ip6addr $peer6addr "gre6t-$cfg"
+			;;
+		*)
+			gre_generic_setup $cfg $mode $ip6addr $peer6addr "gre6-$cfg"
+			;;
+	esac
 }
 
 proto_grev6_setup() {
@@ -147,7 +161,7 @@ proto_grev6tap_setup() {
 	grev6_setup $cfg "gretapip6"
 
 	json_init
-	json_add_string name "grev6-$cfg"
+	json_add_string name "gre6t-$cfg"
 	json_add_boolean link-ext 0
 	json_close_object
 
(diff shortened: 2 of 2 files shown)


dedeckeh:
The initial implementation restricted the suffix to "gre-" as Linux device names are restricted to 15 characters (see IFNAMSIZE in if.h) ; meaning interfaces names up to 11 characters can be configured for gre interfaces taking into account the suffix "gre-".
Extending the suffix for gretap interfaces will limit the interface name length further for gretap interfaces; existing gretap configurations close to the boundary can possibly run into trouble if the name now exceeds 15 charcters.
Patches to fix the above described problem were already done in the past see https://github.com/lede-project/source/commit/0d1b5a1fd2359949cd7b3cddce76f6f248758ed9

rogerpueyo:
Dear @dedeckeh ,

Thanks for your quick reply. I see your point regarding the lenght of the interface name and I understand that, in the worst case (GRE-TAP on IPv6) leaving only 6 characters for the actual name is quite short. However, I think that using the same prefix (i.e. "gre") for two interfaces that are actually defined by two different protocols in /etc/config/network (say, "gre" and "gretap") can be confusing.

I am currently working on a LuCI package to support GRE tunnels (please check https://github.com/rogerpueyo/luci/tree/luci-proto-gre). If I am not wrong, I found that LuCI relies on its protocol to display an interface's name (see the attached picture), therefore I thought it might make sense the patch above.

Last, I wonder if a compromise solution like the following would make sense, in order to distinguish TAP tunnels from non-TAP ones, while keeping the length of the prefix acceptable (max. 5 + 1 hyphen, like currently with grev6):

| Protocol     	|  Prefix 	|
|--------------	|:-----:	|
| GRE IPv4     	| gre4-  	|
| GRE-TAP IPv4 	| gre4t- 	|
| GRE IPv6     	| gre6-  	|
| GRE-TAP IPv6 	| gre6t- 	|

luci

dedeckeh:
@rogerpueyo Compromise proposal makes sense only downside is it can still break existing configs as it extends the suffix for IPv4 gre. Don't think you can tweak the prefix "gre-" further in a meaningfull way; so let's wait for the feedback from Lede core developers on this compromise.

jow-:
For me the compromise is fine.

rogerpueyo:
Dear @jow- and @dedeckeh ,

Thanks for your comments. The current commit leaves prefixes like this:

| Protocol     	|  Prefix 	|
|--------------	|:-----:	|
| GRE IPv4     	| gre4-  	|
| GRE-TAP IPv4 	| gre4t- 	|
| GRE IPv6     	| gre6-  	|
| GRE-TAP IPv6 	| gre6t- 	|

so please consider them for upstream merging.

dedeckeh:
@rogerpueyo Please bump PKG_RELEASE number in the package Makefile

rogerpueyo:
@dedeckeh ,

Done. Thanks!

zorun:
@rogerpueyo when your change is merged, can you please update the documentation at https://wiki.lede-project.org/docs/user-guide/tunneling_interface_protocols#common_options_for_gre_protocols ?  You can for instance reproduce the table you provided above.

Thanks!

rogerpueyo:
@zorun Definitely!

rogerpueyo:
Dear all,

I wonder if the commit looks fine now or is there any other change you may want to propose, before the pull request sinks down in the stack.

jow-:
Added to my staging tree: https://git.lede-project.org/9775594

jow-:
@rogerpueyo - I added a Signed-off-by on your behalf, see https://git.lede-project.org/2f8b96c - please give me a short ACK so that I can merge it to master.

rogerpueyo:
@jow- ACK; sign-off is perfect.
Thanks for staging it! :)

zorun:
@rogerpueyo can you update the documentation at https://lede-project.org/docs/user-guide/tunneling_interface_protocols#common_options_for_gre_protocols ?

rogerpueyo:
@zorun Done (https://lede-project.org/docs/user-guide/tunneling_interface_protocols#common_options_for_gre_protocols)

PR #313, opened 2016-09-02
Fix ath10k-ct failing to load due to missing symbols if neither ath9k nor ath10k are built.

alexgrin:


nbd168:
Applied, thanks.

PR #318, opened 2016-09-05
dnsmasq: fix remove pidfile on shutdown regression

kdarbyshirebryant:
Regression introduced by 3481d0d dnsmasq: run as dedicated UID/GID

dnsmasq is unable to remove its own pidfile as /var/run/dnsmasq is owned
by root and now dnsmasq runs as dnsmasq:dnsmasq.  Change directory
ownership to match.

dnsmasq initially starts as root, creates the pidfile, then drops to
requested non-root user.  Until this fix dnsmasq had insufficient
privilege to remove its own pidfile.


nbd168:
Applied, thanks.

PR #319, opened 2016-09-05
mwlwifi: Expose the IEEE 802.11w support to hostapd

ignisf:
Add a dependency on DRIVER_11W_SUPPORT in order to enable the IEEE
802.11w functionality in hostapd.

Tested with an Android phone, the phone connects and `iw wlanX station dump` indicates MFP usage.


nbd168:
Applied, thanks

PR #320, opened 2016-09-07
dnsmasq: Don't expose *.bind data incl version

kdarbyshirebryant:
Don't expose dnsmasq version & other data to clients via the *.bind
pseudo domain.  This uses a new 'NO_ID' compile time option which has been
discussed and submitted upstream.

This is an alternate to replacing version with 'unknown' which affects
the version reported to syslog and 'dnsmasq --version'

Run time tested with & without NO_ID on Archer C7 v2


nbd168:
Added to my staging tree, thanks

PR #321, opened 2016-09-07
dropbear: hide dropbear version

kdarbyshirebryant:
As security precaution and to limit the attack surface based on
the version reported by tools like nmap mask out the dropbear
version so the version is not visible anymore by snooping on the
wire. Version is still visible by 'dropbear -V'

Based on a patch by Hans Dedecker <dedeckeh@gmail.com>


nbd168:
Added to my staging tree, thanks.

PR #322, opened 2016-09-09
dnsmasq: make NO_ID optional in full variant

kdarbyshirebryant:
Permit users of the full variant to disable the NO_ID *.bind pseudo
domain masking.

Defaulted 'on' in all variants.


nbd168:
Why?

kdarbyshirebryant:
I've received comments regarding the inconvenience of losing access to caching statistics via dns queries.  It didn't seem too harmful to allow a way of re-enabling those stats via menuconfig should someone be willing to build themselves a package.

nbd168:
Applied, thanks.

PR #325, opened 2016-09-14
openssl: Remove DTLS and J-PAKE

neheb:
Nothing uses these.

By the way, is no-whrlpool a typo?


champtar:
Hi @neheb,
Dtls can be used for some vpn, webrtc, ..., so maybe no packages use it yet but i would prefer to have a config option

neheb:
Just pushed

blogic:
@neheb please rebase this so that there is a patch to make both configurable. having patch 1 remove dtls to have patch 2 re-add it as configurable is not very nice/clean

neheb:
Hmm? I didn't make both configurable since J-PAKE is used by nothing. Same with DTLS but I had a request to make it so.

blogic:
ok, well patch 1 still removes dtls for patch to add it back

blogic:
@neheb can you rebase the patches  fixing the dtls quirk ?

neheb:
Yeah I'll do it when I get back from work.

neheb:
Force pushed some new ones. Hopefully good now.

blogic:
merged, Thanks !

PR #327, opened 2016-09-17
add support for Zyxel NBG-419N2 (WAP3205v2)

k-laus:
applied bb-final-ramips-add-zyxel-nbg-419n2.patch from
123serge123, found at https://yadi.sk/d/1ZV0lKJwbTE65;
see https://forum.openwrt.org/viewtopic.php?pid=246905#p246905,
modified slightly to fit to CC release and to new lede build
system: image/rt305x.mk include file is used now

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -234,6 +234,10 @@ mzk-ex750np)
 na930)
 	set_usb_led "$board:blue:status"
 	;;
+nbg-419n2)
+	set_usb_led "$board:green:usb"
+	set_wifi_led "rt2800pci-phy0::radio"
+	;;
 newifi-d1)
 	set_usb_led "$board:red:status"
 	;;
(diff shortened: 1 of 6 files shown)


blogic:
we changed the way that dts files are formatted. please reformat the dts file such that it has the same pattern as the other files inside the target

k-laus:
With 8267248.. i did change the dts format to fit more closely the format that NBG-419.dts uses, I hope that's what you meant with "same pattern". recompiled and tested, still works for me.

blogic:
looks good, now please ...

* fold booth patches into 1 patch
* add a proper subject prefix to the patch ("ramips:")
* add a valid Signed-off-by line

k-laus:
both patches folded into one, added ramips: prefix and Signed-off-by line in https://github.com/lede-project/source/pull/327/commits/7be04e4c427ff686803af55ff2896c3721695981

k-laus:
shall I merge those two 7be04e4 and a442ad5 also to one patch?

blogic:
please fold them into 1 single commit

k-laus:
done - folded into one commit

blogic:
merged, Thanks !

PR #333, opened 2016-09-23
libunwind: use url alias

diizzyy:
Use alias instead of hardcoded URL


blogic:
merged, Thanks !

PR #338, opened 2016-09-25
uml: basic target fixes

ianchi:
Target is not generating working image.
Booting crashes with kernel panic, and console access is not enabled.


blogic:
merged 1/2

2/2 is not required anymore, ath9k_hwsim was fixed,

PR #342, opened 2016-09-26
ar71xx: Add support to TP-Link EAP120

heinzek:
Signed-off-by: Henryk Heisig hyniu@o2.pl


blogic:
urghs ... gpio based mdio seems ugly. @nbd168 do you know what gpios the mdio is normally on ?

target/linux/ar71xx/files/arch/mips/ath79/Makefile
```diff
@@ -84,6 +84,7 @@ obj-$(CONFIG_ATH79_MACH_DLAN_PRO_500_WP)	+= mach-dlan-pro-500-wp.o
 obj-$(CONFIG_ATH79_MACH_DR344)			+= mach-dr344.o
 obj-$(CONFIG_ATH79_MACH_DR531)			+= mach-dr531.o
 obj-$(CONFIG_ATH79_MACH_DRAGINO2)		+= mach-dragino2.o
+obj-$(CONFIG_ATH79_MACH_EAP120)			+= mach-eap120.o
```
ignisf: Do you have a tab too many here?

heinzek: I insert 3 tabs. DR531 and DRAGINO2 also have 3 tabs.

blogic:
merged, Thanks !

communig8-public:
We are testing out the TP-Link EAP120 using the LEDE 17.01.2 image already created for the EAP120 and are having some issues with the Ethernet interface.

With the Interface setup as a DHCP Client, it’s only getting an IP address if we connect it to a switch with a 1000Mbps port, 10Mbps and 100Mbps ports just don’t seem to work.

Also, when connected to either a 10Mbps/FDPX or 100Mbps/FDPX switch port the message logged to the Kernel log shows “eth0: link-up (10Mbps/Half duplex)” or “eth0: link-up (100Mbps/Half duplex)” respectively.

Lastly, using iperf to test the throughput we cannot seem to get more than 10-15Mbps.

We have tested this with about 5 different units so I don’t think it’s duff hardware.

Anyone else observed these issues?

PR #345, opened 2016-09-28
ramips: Add support for Phicomm K2 PSG1218

Biercenter:
- CPU: MT7620A 580MHz
 - Flash: 8MB
 - RAM: 64MB
 - External PA+LNA on both WLAN2.4 and WLAN5
 - one ethernet port 10/100Mbps

Signed-off-by: Biercenter <xuefulin@gmail.com>

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -256,6 +256,9 @@ pbr-m1)
 psg1208)
 	set_wifi_led "$board:white:wlan2g"
 	;;
+psg1218)
+        set_wifi_led "$board:white:wlan2g"
+        ;;
 px-4885)
 	set_wifi_led "$board:orange:wifi"
 	set_usb_led "$board:blue:storage"
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -84,6 +84,7 @@ ramips_setup_interfaces()
 	oy-0001|\
 	pbr-m1|\
 	psg1208|\
+        psg1218|\
 	sap-g3200u3|\
 	sk-wb8|\
 	wf-2881|\
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -93,6 +93,9 @@ get_status_led() {
 	psg1208)
 		status_led="$board:white:wps"
 		;;
+        psg1218)
+                status_led="$board:white:wps"
+                ;;
 	cy-swr1100|\
 	w502u)
 		status_led="$board:blue:wps"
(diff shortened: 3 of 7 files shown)


bobafetthotmail:
you need to write your real name in the signed off line, the nickname isn't enough.

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -84,6 +84,7 @@ ramips_setup_interfaces()
 	oy-0001|\
 	pbr-m1|\
 	psg1208|\
+        psg1218|\
```
blogic: there seems to be a whitespace error here. please use tabs

target/linux/ramips/base-files/lib/ramips.sh
```diff
@@ -349,6 +349,9 @@ ramips_board_detect() {
 	*"PSG1208")
 		name="psg1208"
 		;;
+        *"PSG1218")
+                name="psg1218"
+                ;;
```
blogic: there are spaces instead of tabs here aswell

target/linux/ramips/base-files/lib/upgrade/platform.sh
```diff
@@ -103,6 +103,7 @@ platform_check_image() {
 	pbr-d1|\
 	pbr-m1|\
 	psg1208|\
+        psg1218|\
 	psr-680w|\
```
blogic: space vs tabs

mkresin:
Your commit message is wrong. The device has 4x LAN ethernet and 1x WAN ethernet.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -256,6 +256,9 @@ pbr-m1)
 psg1208)
 	set_wifi_led "$board:white:wlan2g"
 	;;
+psg1218)
+        set_wifi_led "$board:white:wlan2g"
```
mkresin: There is no psg1218:white:wlan2g LED defined in the DTS.

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -93,6 +93,9 @@ get_status_led() {
 	psg1208)
 		status_led="$board:white:wps"
 		;;
+        psg1218)
+                status_led="$board:white:wps"
```
mkresin: There is no psg1218:white:wps LED defined in the DTS.

target/linux/ramips/dts/PSG1218.dts
```diff
@@ -0,0 +1,121 @@
...
+	compatible = "PSG1218", "ralink,mt7620a-soc";
+	model = "Phicomm PSG1218";
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		red {
+			label = "phicomm:red:status";
+			gpios = <&gpio0 8 1>;
+		};
```
mkresin: The naming of **all** LEDs is wrong. It should be something like:

```
<function> {
  label = "<boardname>:<colour>:<function>";
  gpios = <&gpio0 8 1>;
};
```

Have a look at the PSG1208.dts file to see how how should be done.

Biercenter: @mkresin thanks! :-)

PR #348, opened 2016-09-29
utils/busybox: Add mkfs/fsck commands (scripts)

cshoredaniel:
Some packages depend on the mkfs -t fstype or fsck -t fstype
command, but they aren't included in busybox.  Add trivial
wrapper that allow these commands to work.

### DIFF
--- a/package/utils/busybox/Makefile
+++ b/package/utils/busybox/Makefile
@@ -116,6 +116,9 @@ define Package/busybox/install
 	$(INSTALL_BIN) ./files/sysntpd $(1)/etc/init.d/sysntpd
 	$(INSTALL_BIN) ./files/ntpd-hotplug $(1)/usr/sbin/ntpd-hotplug
 	-rm -rf $(1)/lib64
+	$(INSTALL_DIR) $(1)/sbin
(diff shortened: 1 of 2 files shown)


nbd168:
I think a better place for this might be base-files. mkfs.* usually does not come from busybox.

cshoredaniel:
On Thu, 29 Sep 2016 01:52:21 -0700
nbd168 <notifications@github.com> wrote:

> I think a better place for this might be base-files. mkfs.* usually
> does not come from busybox.
> 

My thought was that lack of mkfs was a results of use of busybox rather
than e.g. coreutils (I'm not sure where mkfs/fsck top-level packages
usually are built from on other systems), which is why I didn't choose
base-files.

Regards,

Daniel

nbd168:
At least x86 uses mkfs.ext4 from e2fsprogs and mkfs.f2fs from f2fstools

cshoredaniel:
Realized mkfs/fsck meta-commands are supplied by util-linux; will enable building there and have packages that need them depend on those packages instead of making all builds have normally not needed commands; even if small.

PR #349, opened 2016-09-29
target/{sdk,imagebuild}: Fix for symlink-tree

cshoredaniel:
With symlink tree some directories are just symlinked which
means IB and SDK end up with a symlink instead of an actual
directory; this fixes the missing files by dereferencesing
the directories instead of copying the symlinks.


nbd168:
LGTM

blogic:
merged, Thanks !

PR #350, opened 2016-09-29
linux/modules/fs: Fix missing vfat dependencies

cshoredaniel:
vfat filesystem fails to mount without cp437 and iso8859-1 nls modules,
therefore make these required dependencies for vfat kernel module.

### DIFF
--- a/package/kernel/linux/modules/fs.mk
+++ b/package/kernel/linux/modules/fs.mk
@@ -426,7 +426,7 @@ define KernelPackage/fs-vfat
 	$(LINUX_DIR)/fs/fat/fat.ko \
 	$(LINUX_DIR)/fs/fat/vfat.ko
   AUTOLOAD:=$(call AutoLoad,30,fat vfat)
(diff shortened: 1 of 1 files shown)


mkresin:
Hey Daniel,

have you had a look at the other filesystems as well? I'm quite sure that I had a similar problem with a missing kmod-nls-utf8. But I can not remember which filesystem it was.

diizzyy:
I usually go with all these (cp437, iso8859-1, utf8) when using "external" filesystems to avoid issues further down the line.

blogic:
@mkresin i might be wrong here but it depends on the codepage used during formatting. imho asian sdcards wont be using the european codepage by default but require utf8. imho we should add utf-8 aswell to be on the safe side. i agree with @diizzyy on that one

cshoredaniel:
Updated (still only vfat patch though).

blogic:
merged, Thanks !

PR #354, opened 2016-09-29
ar71xx: massive cleanups (round 1)

pepe2k:
The following series is first part of planned massive cleanups in **ar71xx** target, here mostly focused on base-files scripts. More difficult changes, like renaming LEDs for all boards (which means change most of mach-*.c files), to match general convention "_board:color:name_", are planned in some "near future" :smirk:

Some of the commits could be squashed as they touch the same file. I split them just for better understanding what have been changed - small changes in code with block movements at the same time are not handled very well in diffs.

Changes were tested on real hardware, just to confirm that I didn't break anything in scripts.
But, if I did... just let me know. :sweat: 

Small statistics about change in code size (raw files):

- /lib/upgrade/platform.sh: **-570 bytes**
- /etc/board.d/01_leds: **-4243 bytes**
- /etc/board.d/02_network: **-1967 bytes**
- /etc/diag.sh: **-883 bytes**

In total: **-7663 bytes**.


blogic:
lgtm @jow- @nbd168 please have a look aswell if you spot anything

jow-:
LGTM

blogic:
merged, Thanks !

PR #355, opened 2016-09-29
lantiq: Fix buttons for ARV752DPW

FooDeas:
- set correct GPIOs for buttons
- set correct input event codes
- include corresponding headers and use macros

The changes have been tested successfully.

Signed-off-by: Andreas Eberlein <foodeas@aeberlein.de>

### DIFF
--- a/target/linux/lantiq/dts/ARV752DPW.dts
+++ b/target/linux/lantiq/dts/ARV752DPW.dts
@@ -1,6 +1,9 @@
 /dts-v1/;
 
-/include/ "danube.dtsi"
+#include "danube.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
 
 / {
 	model = "ARV752DPW - Arcor 802";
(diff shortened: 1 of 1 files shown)


blogic:
lgtm, we should do the sanity thing globally for lantiq dts files

mkresin:
Yeah, I've assigned the ticket to me to do the sanity thingy for the remaining boards.

FooDeas:
Okay. And what happens with my PR now?

mkresin:
Your patch as accepted.

I've added your patches to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I've modified the remaining device tree source files already, but the device tree macros can only be used with boards already migrated to the new image build code. I'm working at the moment to bring the last two boards to the new build code.

Please do have a little patience until I'm finished. I will commit all changes at once to not touch the same files twice.

FooDeas:
Thanks for answer! I am pleased to see that improvements are actually taken into account here.
If I can support you, just let me know.

mkresin:
Finally merged. Thanks a lot!

PR #356, opened 2016-09-30
ramips: Add support for Phicomm K2 PSG1218

Biercenter:
- CPU: MT7620A 580MHz
- Flash: 8MB
- RAM: 64MB
- External PA+LNA on both WLAN2.4 and WLAN5
- 4x LAN ethernet and 1x WAN ethernet

Signed-off-by: xuefu lin <xuefulin@gmail.com>

### DIFF
--- a/target/linux/ramips/dts/PSG1218.dts
+++ b/target/linux/ramips/dts/PSG1218.dts
@@ -1,0 +1,122 @@
+/dts-v1/;
+
+#include "mt7620a.dtsi"
+
+/ {
+	compatible = "PSG1208", "ralink,mt7620a-soc";
+	model = "Phicomm PSG1208";
+
+	gpio-leds {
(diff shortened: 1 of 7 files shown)


target/linux/ramips/dts/PSG1218.dts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+&gpio1 {
+	status = "okay";
+};
+
+&gpio3 {
+	status = "okay";
+};
+
```
blogic: are gpio1 and gpio3 really used ?

Biercenter: I copy of PSG1208.dts, and the router works.

blogic:
the dts does not reference the device nodes. please remove them

PR #361, opened 2016-10-01
source: Switch to xz for packages and tools where possible

diizzyy:
Adds a slightly higher compression level to xz by default which roughly raises memory usage from 100MiB to about 200MiB during compression, about 10MiB for decompression. (Source: xz manpage)
Changes git packages to xz
Removes mirror checksums to avoid mismatches with tarballs as they differ due to checkout date and time of files.
Changes a few source tarballs to xz if available upstream.


diizzyy:
Not sure if PKG_RELEASE should be bumped or not.

ignisf:
https://github.com/openwrt/packages/issues/149

Incrementing `PKG_RELEASE` has been a bit hazy to me, too. Has the `PKG_RELEASE` convention been documented anywhere in the wiki? I've forgotten to bump the `PKG_RELEASE` a few times and my patches have been merged anyways. Is the linked convention adopted by LEDE?

NeoRaider:
Generally, PKG_RELEASE should be incremented when the resulting binary package will change (either the contained files or the metadata like dependencies), so opkg will notice the update.

For build fixes or changing to another equivalent source URL or tarball (like here) no increment should be done, as there is nothing for opkg to update.

Other comments:
* Switching downloaded tarballs to .xz seems reasonable to me
* The tarball for the package *xz* should be kept as bz2, so xz can be built when it is not available on the host system (bzip2 is checked as a prerequisite of LEDE, xz isn't; adding xz to the prereqs might be an alternative)

How much smaller do the tarballs get with `-7e`? The effect will only be noticable for huge tarballs, but in my experience it's not worth the increased time needed for the compression of big files.

diizzyy:
Most base packages see a slightly smaller size (it's usually within 1-2KiB) however it's more beneficial for external packages which can be quite a bit larger and I don't see much of a difference in CPU time for base packages.

diizzyy:
I've added xz-utils (xz) to make prereq

tools/make-ext4fs/Makefile
```diff
@@ -10,13 +10,13 @@ include $(TOPDIR)/rules.mk
...
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL=$(LEDE_GIT)/project/make_ext4fs.git
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_VERSION:=484903e4332be6c317f531b008cb2a841a18c506
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
-PKG_MIRROR_MD5SUM:=b80d6ddcd6f6925e2f5b80775bb96f3e
-PKG_CAT:=zcat
+PKG_MIRROR_MD5SUM:=
+# PKG_CAT:=zcat
```
blogic: there is a commented out line here

tools/yaffs2/Makefile
```diff
@@ -10,12 +10,12 @@ include $(TOPDIR)/rules.mk
...
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://android.googlesource.com/platform/external/yaffs2
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)
 PKG_SOURCE_VERSION:=d333fc232d7e5ae3370080d5d6f7d88ea9c6b3a1
-PKG_MIRROR_MD5SUM:=8145d4dd9b46face26121f27ad60cc1c
+PKG_MIRROR_MD5SUM:=
 HOST_BUILD_DIR=$(BUILD_DIR_HOST)/$(PKG_NAME)
```
blogic: remove the whole line if it is not needed and explain this change in the commit message. this is unrealated to changing the compression to xz

tools/mtd-utils/Makefile
```diff
@@ -10,12 +10,12 @@ PKG_NAME:=mtd-utils
...
 
 PKG_SOURCE_VERSION:=aea36417067dade75192bafa03af70b6eb2677b1
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-$(PKG_SOURCE_VERSION).tar.gz
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-$(PKG_SOURCE_VERSION).tar.xz
 PKG_SOURCE_URL:=git://git.infradead.org/mtd-utils.git
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
 PKG_MIRROR_MD5SUM:=
-PKG_CAT:=zcat
+# PKG_CAT:=zcat
```
blogic: why not delete the line ?

tools/b43-tools/Makefile
```diff
@@ -10,12 +10,12 @@ include $(TOPDIR)/rules.mk
...
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=https://github.com/mbuesch/b43-tools.git
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)
 PKG_SOURCE_VERSION:=8dce53297966b31b6c70a7a03c2433978dd9f288
-#PKG_MIRROR_MD5SUM:=50ca3c763ee21ee213addd17cf1c1b86
+PKG_MIRROR_MD5SUM:=
 HOST_BUILD_DIR=$(BUILD_DIR_HOST)/$(PKG_NAME)
```
blogic: remove it if it not needed

tools/make-ext4fs/Makefile
diizzyy: Couldn't find any documentation or figure out what it actually did, there are more occurrences of stray lines in package Makefiles so I'd rather do a tree-wide cleanup afterwards.

tools/mtd-utils/Makefile
diizzyy: See #1 comment

tools/yaffs2/Makefile
diizzyy: It's an issue because git doesn't date and time-stamp files during checkout so each package will be different unless you checkout at the exact same time and at the exact same speed. As someone who doesn't have access to the "fallback" download server it's impossible/doesn't make any sense to add a checksum for a package that's more than likely mismatch what's going to be uploaded if any at all. There are more Makefiles in the tree which have empty PKG_MIRROR_MD5SUM vars so I figured it would be more consistent to leave those empty rather than removing them.

tools/b43-tools/Makefile
diizzyy: See first comment

diizzyy:
Due to my silly mistake not using branches makes this much harder to fix. Continued (and fixed) in #375

PR #363, opened 2016-10-02
tools/ccache: Update ccache 3.3.2 and refresh patch

diizzyy:
Update ccache 3.3.2 and refresh patch

Preserving the original patch comments here by Karl Vogel:

"From 90762a9b8d9a50b6176f10bd6c2e2b9501117561 Mon Sep 17 00:00:00 2001
From: Karl Vogel <karl.vogel@gmail.com>
Date: Tue, 14 Jul 2015 11:05:33 +0200
Subject: [PATCH] Include environment variable GCC_HONOUR_COPTS in hash.

The OpenWRT patch, 910-mbsd_multi.patch, to GCC adds an extra
compilation flag, -fhonour-copts, which is influenced by an
environment variable called GCC_HONOUR_COPTS.

Include this environment var in the hash calculation as otherwise
the gcc stdout warning from a previous compilation might be shown
where, even when GCC_HONOUR_COPTS is in 's'ilent mode.

Signed-off-by: Karl Vogel <karl.vogel@gmail.com>"

### DIFF
--- a/tools/ccache/Makefile
+++ b/tools/ccache/Makefile
@@ -8,11 +8,11 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/target.mk
 
 PKG_NAME:=ccache
-PKG_VERSION:=3.1.11
+PKG_VERSION:=3.3.2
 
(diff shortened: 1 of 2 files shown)


blogic:
merged, Thanks !

PR #366, opened 2016-10-02
tools/libressl: Update to 2.5.0 and use mirrors

diizzyy:
Updates LibreSSL to 2.5.0 and switches from main site to mirrors


blogic:
are we sure that we want to use a mirror unconditionally ? maybe just add the mirrors in addition to the main site

diizzyy:
I can add main site as last resort, they're both listed in the official mirror list. https://www.openbsd.org/ftp.html

diizzyy:
Superseded by #376

PR #368, opened 2016-10-02
mvebu: fix sysupgrade

accwebs:
sysupgrade command fails due to missing U-Boot environment-processing binaries on sysupgrade ramdisk. The missing binaries result in the following output:
	Switching to ramdisk...
	Performing system upgrade...
	ash: /usr/sbin/fw_printenv: not found
	ash: fw_setenv: not found
	ash: touch: not found
	cannot find target partition

Fixes FS#197.

Signed-off-by: Aaron Curley <accwebs@gmail.com>

### DIFF
--- a/target/linux/mvebu/base-files/lib/upgrade/linksys.sh
+++ b/target/linux/mvebu/base-files/lib/upgrade/linksys.sh
@@ -89,3 +89,5 @@ linksys_preupgrade() {
 	esac
 }
 
+append sysupgrade_pre_upgrade linksys_preupgrade
+


mkresin:
Good job!

But please fix the empty ```Signed-off-by:``` line is in your commit message. It should be ```Signed-off-by: Real Name <mail@domain.tld>```

accwebs:
Sorry -- I must be slow to understand the process. My impression was that the project maintainers were the ones that need to sign off, and that getting them to do so was part of the pull request process.  

Can you either describe the correct process or point me at the documentation?  I reviewed the page at https://www.lede-project.org/development.html but "TBD" is listed.

mkresin:
> Sorry -- I must be slow to understand the process. My impression was that the project maintainers were the ones that need to sign off, and that getting them to do so was part of the pull request process.

Have a look at section 1.11. of https://www.kernel.org/doc/Documentation/SubmittingPatches for an explanation about the SoB line.

> Can you either describe the correct process or point me at the documentation? I reviewed the page at https://www.lede-project.org/development.html but "TBD" is listed.

Yeah sorry. No one had the time to write the instructions yet.

accwebs:
Hah.  I thought "Signed-off-by:" had something to do with quality-control and thus had to be done by a maintainer.  My bad.

Anyway, updated.  Is that better?

anomeome:
Is it known whether the author of the commit causing the issue intentionally deleted this line in support of getting the clearfog device working? As there appear to be other FS in play with the clearfog, it was about as clear as fog to me.

accwebs:
@anomeome good question.  Unfortunately, it wasn't clear to me either.  A few thoughts:

1. The linksys_preupgrade() function (that I have enabled in the proposed change) appears to check the "$(mvebu_board_name)" variable before making any changes so I *think* having linksys_preupgrade() called is harmless. Unfortunately, I don't have a clearfog device to test with to know for sure.

2. Point #1 being said, just because it should be harmless doesn't mean it's the right way to do it.  I'm open to alternate suggestions if you can see a more correct way of "plugging in" to the sysupgrade script.  One way that comes to mind might be implementing a platform_pre_upgrade() method in platform.sh that checks the board type and then calls the linksys.sh method.

3. I followed the approach currently in this pull request because that's how the linksys.sh scripts in the following other targets did things:
./target/linux/ipq806x/base-files/lib/upgrade/linksys.sh
./target/linux/kirkwood/base-files/lib/upgrade/linksys.sh
However, it appears that linksys devices might be the only targets for these other arches.

blogic:
merged, Thanks !

PR #371, opened 2016-10-02
tools/upx: Update to 3.91 and use new tarball url

diizzyy:
Updates UPX to version 3.91 and also updates tarball url


blogic:
merged, Thanks !

PR #377, opened 2016-10-04
uhttpd: support using OpenSSL for certificate generation

hnyman:
@nbd168  @jow- 

Support the usage of the OpenSSL command-line tool for generating the SSL certificate for uhttpd. Traditionally 'px5g' based on PolarSSL (or also mbedTLS in LEDE), has been used for the creation. 
This patch enables creating a truly OpenSSL-only version of LuCI without dependency to PolarSSL/mbedTLS based px5g.

uhttpd init script is enhanced by adding detection of an installed openssl command-line binary (provided by 'openssl-util' package), and if found, the tool is used for certificate generation.

Note: After this patch the script prefers to use the OpenSSL tool if both it and px5g are installed. If the preference for an installed px5g is to be maintained, then the evaluation order should be reversed.

Reference to discussion at https://github.com/openwrt/luci/pull/807

Note: To drop the px5g dependency from the new luci-ssl-openssl collection in LuCI requires either
* this/similar uhttpd patch getting accepted by both LEDE and Openwrt, or
* a reliable config symbol to identify if we are building for LEDE or Openwrt. If one is available, then I think that the LuCI collections' dependencies can be organised so that the dependencies can be set according to build environment requirements.


jow-:
LGTM

jow-:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/jow/staging.git;a=commit;h=d068a12e8ec81a280de4ebd66ae94e2b94990b84

PR #378, opened 2016-10-05
uhttpd: revert to preferring px5g for certificate creation, create self-signed certificates with unique subjects 

hnyman:
@jow- 

I noticed today that Firefox seems to somewhat dislike OpenSSL-created certificates. Chrome, IE and Edge seem to use them happily, but Firefox forms the initial connection slowly and viewing the certificate details in Firefox is slow.

So, to avoid trouble and a decreased user experience for users who have installed openssl-util in their build, change the preference to be for px5g, for now.

After this change, the old px5g is used for the cert creation unless the user has explicitly removed px5g from luci-ssl(-openssl) collection or has manually deleted it from the image. Normal users will create the certs with px5g just as before.

But testing for a pure OpenSSL-based build is possible. And I will try to figure out how Firefox can be made to like the OpenSSL certs. Likely some creation parameters need to be added or changed.

### DIFF
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -46,12 +46,13 @@ generate_keys() {
 
 	# Prefer px5g for certificate generation (existence evaluated last)
 	local GENKEY_CMD=""
+	local UNIQUEID=$(dd if=/dev/urandom bs=1 count=4 | hexdump -e '1/1 "%02x"')
 	[ -x "$OPENSSL_BIN" ] && GENKEY_CMD="$OPENSSL_BIN req -x509 -outform der -nodes"
 	[ -x "$PX5G_BIN" ] && GENKEY_CMD="$PX5G_BIN selfsigned -der"
 	[ -n "$GENKEY_CMD" ] && {
 		$GENKEY_CMD \
 			-days ${days:-730} -newkey rsa:${bits:-2048} -keyout "${UHTTPD_KEY}.new" -out "${UHTTPD_CERT}.new" \
-			-subj /C="${country:-DE}"/ST="${state:-Saxony}"/L="${location:-Leipzig}"/CN="${commonname:-Lede}"
+			-subj /C="${country:-DE}"/ST="${state:-Saxony}"/L="${location:-Leipzig}"/O="${commonname:-Lede}$UNIQUEID"/CN="${commonname:-Lede}"
 		sync
 		mv "${UHTTPD_KEY}.new" "${UHTTPD_KEY}"
 		mv "${UHTTPD_CERT}.new" "${UHTTPD_CERT}"
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -44,10 +44,10 @@ generate_keys() {
 	config_get location   "$cfg" location
 	config_get commonname "$cfg" commonname
 
-	# Prefer OpenSSL for certificate generation (existence evaluated last)
+	# Prefer px5g for certificate generation (existence evaluated last)
 	local GENKEY_CMD=""
-	[ -x "$PX5G_BIN" ] && GENKEY_CMD="$PX5G_BIN selfsigned -der"
 	[ -x "$OPENSSL_BIN" ] && GENKEY_CMD="$OPENSSL_BIN req -x509 -outform der -nodes"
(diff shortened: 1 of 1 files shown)


ignisf:
Wait what? Is there a bug filed for this? I've been using certs generated with OpenSSL for ages. 😲

hnyman:
Related to the change from yesterday that enabled the auto-creation of missing certificates for uhttpd with openssl, if found. See https://github.com/lede-project/source/commit/3c4858eeb2bbb3107f87bb3be07d5c172c8e0ef9

This pull request just reverts the preference in selecting the cert creation tool to the situation before yesterday.

hardfalcon:
Can't we just switch from RSA to ECDSA certificates altogether? In comparison with ECDSA, the current default of RSA 2048 is a lot slower, computationally more intensive and requires a lot more entropy upon certificate generation, while still offering an inferior level of cryptographic security in compraison to a secp256r1 or secp384r1 based signature.

Browser support shouldn't be an issue as all modern browsers support ECDSA certificates with at least both of the above mentioned curves:
https://support.globalsign.com/customer/portal/articles/1995283-ecc-compatibility
https://tools.ietf.org/html/rfc6460#section-4.1

hnyman:
I have compared the generated certificates with SSL certificate analysers and although certificates are otherwise similar, there is one difference: 
* px5g generates certificates that have the attribute "X509v3 Basic Constraints: CA:FALSE". px5g sources explicitly call the routine to set that.
* OpenSSL generates currently certs with attribute "X509v3 Basic Constraints: CA:TRUE". As far as I have found from the internet, the command used is the standard one-liner for creating self-signed certificates. But it seems to leave the "this is a CA certificate" attribute active. And that might be a partial reason for the peculiar behaviour with Firefox that I noticed.

Searching the web for Firefox problems with self-signed certificates revealed that apparently Firefox may get confused if there are new certificates & needed security exceptions (due to the self-signed status) for the same name than previous certificates. Likely it does not clean the old exceptions when a new one with identical name is defined, and thus the database grows, and Firefox will spend exponentially more time in looking for the correct authentication chain. I deleted the certificate database from my Firefox and that seems to have helped, so far. But the phenomenon has not been visible earlier with px5g generated CA:FALSE certificates, so the attribute CA:TRUE likely pays a role here. 

Links discussing this Firefox thing:
https://bugzilla.redhat.com/show_bug.cgi?id=1204670#c34
https://bugzilla.mozilla.org/show_bug.cgi?id=1056341
https://bugzilla.mozilla.org/show_bug.cgi?id=1147544
https://github.com/cockpit-project/cockpit/pull/2455/
The fix applied by the Fedora/cockpit people was to avoid identical names for self-signed certificates by adding a machineID as a new O= item to the certificate subject. 

All in all, the bug will apparently only hit Firefox users who repeatedly generate new self-signed certificates with identical ID info and CA:TRUE attribute. 

A solution to produce different IDs might be to add a random number as the "O=" item to the certificate subject when it is autogenerated.

hnyman:
I have rebased and pushed a second commit.

The first commit reverts back to the old preference for px5g if that is installed. OpenSSL is used only if px5g is not found.

The second commit adds a partially random O= item to the certificate subject in order to make the automatically generated certificates' subjects unique.

Firefox has problems when several self-signed certificates with CA:true attribute and identical subjects have been seen (and stored) by the browser. Reference to upstream bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1147544
https://bugzilla.mozilla.org/show_bug.cgi?id=1056341
https://bugzilla.redhat.com/show_bug.cgi?id=1204670#c34

Certificates created by the OpenSSL one-liner fall into that category.

Avoid identical certificate subjects by including a new 'O=' item with CommonName + a random part (8 chars). Example: ``` /CN=LEDE/O=LEDEb986be0b/L=Unknown/ST=Somewhere/C=ZZ ```

That ensures that the browser properly sees the accumulating certificates as separate items and does not spend time trying to form a trust chain from them.

ignisf:
Just as a side note, AFAIK it's not required the organisation name to be a part of the certificate

hnyman:
ping. 
any comments to this?

jow-:
LGTM

hnyman:
ping? could this be merged?

blogic:
merged, Thanks !

jow-:
Pulled into my staging tree: https://git.lede-project.org/30c24d3

PR #380, opened 2016-10-06
ar71xx:correct wifi mac address for TL-WDR4900 v2

981213:
Currently the mac address of 2.4G and 5G wifi are the same,which might cause some problems.

Signed-off-by: Chuanhong Guo gch981213@gmail.com

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c7.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c7.c
@@ -218,7 +218,7 @@ static void __init common_setup(bool pcie_slot)
 	if (pcie_slot) {
 		ath79_register_pci();
 	} else {
-		ath79_init_mac(tmpmac, mac, -1);
+		ath79_init_mac(tmpmac, mac, -2);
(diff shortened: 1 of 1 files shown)


981213:
Anything wrong with this PR?
the mac addresses should not be the same ,otherwise some devices could only display either 2.4G or 5G SSID(Such as my Android smartphone).
@blogic

jow-:
@981213 - the change looks okay to me. Is the factory firmware doing the same?

jow-:
So I did a check on my WDR-4900 v1 - this device only has *three* MAC addresses available and I assume its similar for the v2. We cannot simply squat a fourth one as we might then intersect with the mac range given to another unit.

On my device the primary mac (the one printed on the case) is `A0:F3:C1:F8:5B:24`. This mac is shared by LAN and the 5GHz radio. The 2.4GHz radio uses `...:23` (base MAC - 1) and the WAN interface uses `...:25` (base MAC + 1).

Can you confirm that layout for the v2 using vendor firmware? I think we should use the same assignment for LEDE.

981213:
OK I'll check it this weekend:-D

PR #381, opened 2016-10-06
ar71xx:Ignore firmware building errors of UBNT and CyberTAN devices

981213:
otherwise the build will fail if the rootfs is too big for them.


blogic:
merged, Thanks !

PR #382, opened 2016-10-06
build: handle 'src' folder in the Build/Prepare/Default & Host/Prepare/Default build rules

commodo:
Traditionally, the `patches` folder would get handled implicitly by one of these `*/Prepare/Default` build rules.
But I"ve started to notice (in various packages), that some patches introduce whole files (example is `hostapd` which introduces `ubus.c` & `ubus.h`).

The problem with such files, is when updating then, when you're patching patches to update something.
So, as a measure to address this, I'm proposing that the `src` folder be handled (also implicitly) in the `*/Prepare/Default` build rules as a final step, and if present.

The added benefit of this, is that core packages (which only have an `src` folder), can just use the `Build/Prepare/Default` rule.

Possible side effects of this: this could be abused to add too much source (in some packages). But I think that with patches it could have been the same, or worse.

I feel the `src` folder is a convention (not sure if formalized yet) for adding C sources to a package.
This patchset would formalize this.

I am fine with splitting this PR/patchset into smaller sets, or re-spinning, or removing some elements.
I mostly care about the first patch (https://github.com/lede-project/source/commit/daaa2bb56472a63d2e490855011e85bd817c9ca1) ; the rest of the patches serve a more demo/cleanup purpose.

I thought that sending this as a PR (rather than patchset) would not clutter the mailing list.


commodo:
`Note:` I only compiled this successfully for a few packages (libnl-tiny, hostapd, swconfig, libiconv). Right now I have it compiling with `CONFIG_ALL` ; but in the meantime I thought about spinning up a PR, to start the review process, while I switch this to a background thread to work on something else.

I don't see a reason for this failing for the other packages, but you never know.
Will come back with a comment, when compilation is done [hopefully not filling up my HDD again].

commodo:
Compilation finished successful with CONFIG_ALL, target ar71xx.

commodo:
@nbd168 @jow- @blogic

include/host-build.mk
```diff
@@ -44,6 +44,7 @@ ifneq ($(strip $(HOST_UNPACK)),)
   define Host/Prepare/Default
 	$(HOST_UNPACK)
 	$(Host/Patch)
+	[ ! -d ./src/ ] || $(CP) ./src/* $(HOST_BUILD_DIR)
```
hnyman: I haven't thought this through, but at the first glance I wonder why you evaluate that via negation?
Why not ``` [ -d ./src/ ] && $(CP)... ``` ?

commodo: somewhat, related to this ; i was wondering if i should put this before the `$(XXX/Patch)` rules;
that would allow for source to be added (in `src`), and then patches to be applied after that, which [to me] sounds wonky/weird;
the kernel seems to do it like that [ though, i'm not entirely sure about it right now ]
the only "disadvantage" of this order, is that the build won't fail if a patch adds a file, and is being overwritten by anything in `src` ; but [personally] I don't see that as a big problem ;

commodo: updated comment of first patch

nbd168:
LGTM

blogic:
this would mean that the packages feed would potentially double copy the files

commodo:
well, that depends ; 
do packages in the feeds have/define/use `src` folders ?
and if so, is this ok ?

or, are you referring to [or thinking about] any particular packages feed ?

nbd168:
I just noticed one more thing: Please change the order of Host/Patch vs copying ./src to copy src first. That way patches can change things from src/ as well. It's the same order that is also used for the kernel tree.

commodo:
yep ; was wondering about that in a previous comment of mine:
https://github.com/lede-project/source/pull/382#discussion_r82341379

[ to shortly re-iterate ; i would prefer this version, but i have no strong opinion towards one option or the other, as long as I get the `src` behavior  ]

commodo:
re-spinned

commodo:
re-spinned again to fix conflicts; 
any more objections to this ?

blogic:
merged, Thanks !

PR #383, opened 2016-10-06
dslite: Quote resolveip hostname argument

dedeckeh:
Quote resolveip hostname argument to avoid bad shell injections.
While at it fix pattern match logic in case multiple IPv6 addresses
are returned for a hostname as they're seperated by newline by
resolveip and not a white space


jow-:
Pulled into my staging tree: https://git.lede-project.org/3eecd89

PR #386, opened 2016-10-07
iftop: prevent cpu spin in initial rate selection

kdarbyshirebryant:
In presence of rates above 1000MB, iftop will fall off the end of a rate
lookup table and then go into a 100% cpu spin.  Although the initial
rate may be erroneously over 1000MB on occasion, iftop no longer spins.

### DIFF
--- a/package/network/utils/iftop/Makefile
+++ b/package/network/utils/iftop/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=iftop
 PKG_VERSION:=1.0pre4
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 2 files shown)


lynxis:
Hi @kdarbyshirebryant ,
do you sent it upstream?

kdarbyshirebryant:
No.  And I won't.

blogic:
@kdarbyshirebryant i think the patch works but a far better solution would be to simply add another line to the scale table declaration. 
`/* Barchart scales. */
static struct {
    int max, interval;
} scale[] = {
        {      64000,     10 },     /* 64 kbit/s */
        {     128000,     10 },
        {     256000,     10 },
        {    1000000,     10 },     /* 1 Mbit/s */
        {   10000000,     10 },
        {  100000000,    100 },
        { 1000000000,    100 }      /* 1 Gbit/s */
    };`

PR #387, opened 2016-10-07
hostapd: Fix for wpa-supplicant-mesh package not containing wpa_supplicant binary and missing openssl dependency.

alexgrin:


jow-:
Merged into my staging tree with http://git.lede-project.org/d296dec - thanks @alexgrin and sorry for the long delay.

PR #396, opened 2016-10-08
package/libs/libtool: Switch to xz tarball

diizzyy:
Switch to xz tarball, there's no point pulling two different tarballs of the same source code (tools/libtool uses xz).


blogic:
merged, Thanks !

PR #403, opened 2016-10-11
target/sdk: Switch to xz compression instead of bz2

diizzyy:
Switch to xz compression instead of using bz2.
This makes a considerable difference in size, ar71xx SDK from 60M to 35M.


diizzyy:
Compile tested on ar71xx, Debian 8.6 (AMD64)

nbd168:
Merged to my staging tree at https://git.lede-project.org/?p=lede/nbd/staging.git;a=summary

PR #405, opened 2016-10-11
tools/dosfstools improvements

Noltari:
Any objections?


diizzyy:
I think we should try to be as close as possible to upstream as possible so I think this is a good idea.

nbd168:
Added to my staging tree at https://git.lede-project.org/?p=lede/nbd/staging.git;a=summary
Thanks

PR #407, opened 2016-10-11
Rename all files in package/network/utils/wwan/data from xxxx:yyyy to…

btsimonh:
… xxxx-yyyy and

modify makefile to rename them back after copy.
This is to ensure that git can be cloned onto a windows drive without failing.

@blogic - for you....


btsimonh:
Put a hold on applying this at all...  I'm investigating and documenting GIT file permissions attributes & the use of windows GIT tools... (for ref see http://stackoverflow.com/questions/2517339/how-to-recover-the-file-permissions-to-what-git-thinks-the-file-should-be).

jow-:
LGTM

blogic:
merged, Thanks !

PR #408, opened 2016-10-12
dnsmasq: Multiple dnsmasq instances support

dedeckeh:
Adds support in uci for configuring multiple dnsmasq instances via
multiple dnsmasq sections.
The uci sections host, boot, mac, tag, vendorclass, userclass,
circuitid, ... will refer to a dnsmasq instance via the instance
parameter defined in the section; if the instance parameter is
not specified backwards compatibility is preserved.

Start/Stopping a dnsmasq instance can be achieved by passing the
dnsmasq instance name as argument to start/stop via the init script.

Multiple dnsmasq instances is usefull in scenarios where you want to
bind a dnsmasq instance to an interface in order to isolate networks.

This patch is a rework of a multiple dnsmasq instance patch by Daniel Dickinson

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -6,18 +6,15 @@ START=19
 USE_PROCD=1
 PROG=/usr/sbin/dnsmasq
 
-DNS_SERVERS=""
-DOMAIN=""
-
 ADD_LOCAL_DOMAIN=1
 ADD_LOCAL_HOSTNAME=1
 
-CONFIGFILE="/var/etc/dnsmasq.conf"
-CONFIGFILE_TMP="/var/etc/dnsmasq.conf.$$"
-HOSTFILE="/tmp/hosts/dhcp"
+BASECONFIGFILE="/var/etc/dnsmasq.conf"
+BASEHOSTFILE="/tmp/hosts/dhcp"
+BASETIMESTAMPFILE="/etc/dnsmasq.time"
 TRUSTANCHORSFILE="/usr/share/dnsmasq/trust-anchors.conf"
-TIMESTAMPFILE="/etc/dnsmasq.time"
 TIMEVALIDFILE="/var/state/dnsmasqsec"
+BASEDHCPSTAMPFILE="/var/run/dnsmasq"
 
 xappend() {
 	local value="$1"
@@ -59,7 +56,7 @@ dhcp_calc() {
 
 dhcp_check() {
 	local ifname="$1"
-	local stamp="/var/run/dnsmasq.$ifname.dhcp"
+	local stamp="${BASEDHCPSTAMPFILE_CFG}.${ifname}.dhcp"
 	local rv=0
 
 	[ -s "$stamp" ] && return $(cat "$stamp")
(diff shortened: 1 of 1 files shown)


dedeckeh:
Compile and run tested on brcm63xx and AR7xx (TP-LINK TL-WDR4300)

jow-:
In principle fine, I just think that `dnsmasq_config` is a rather bulky and misleading name, maybe we could just call it `instance` ?

dedeckeh:
@jow renamed dnsmasq_config into instance as suggested

kbabioch:
@dedeckeh Thank you very much for putting this much effort into this and finally pulling together various patches flying around for years. I very much appreciate this and hope it will finally be included, so I have no longer to maintain my own branches.

However this probably also requires some changes to LUCI in order to become usable. At the very least it would be nice if there was an overview of leases from all instances. For convenience having the ability to configure each instance (as well as the "default" one) would be nice, too. Being then able to assign single instances of dnsmasq to a particular interface would also be great. All in all, I'm afraid that this requires a lot of changes in all kinds of places and it probably needs someone with more experience with the codebase to have a go at it. @jow- are you interested in this :-).

kbabioch:
While these change generate mostly correct configuration files for dnsmasq I've noticed two problems.

- All files contain DHCP ranges for all interfaces, e.g. statements like `dhcp-range=`. How can one assign a particular dhcp configuration block to a particular dnsmasq instance. Given them the same name is not working for me, although it should when I understand the code correctly (filter_dnsmasq & dhcp_add). Ideally only the correct DHCP ranges should end up in the configuration file.

- How are you dealing with the interface bindings? The option `nonwildcard` maps to dnsmasq's option `bind-dynamic`. There is no control over `--bind-interfaces` and or `listen-address`.

dedeckeh:
@kbabioch Thx for spotting the unconditionally remove of the files; fixed meanwhile in the updated PR

Regarding the other remarks
- Separate dhcp-ranges per instance can be obtained by using the instance parameter in the dhcp section. Assume two dnsmasq (instance1/instance2) are configured and two interfaces are available (iface1/iface2) following config will create seperate dhcp-ranges per dnsmasq instance

config dhcp 'iface1'
        option instance 'instance1'
        option interface 'iface1'
        option start '100'
        option limit '150'
        option leasetime '12h'

config dhcp 'iface2'
        option instance 'instance2'
        option interface 'iface2'
        option start '100'
        option limit '150'
        option leasetime '12h'

- Interface bindings are achieved by the (not)interface and nonwildcard uci parameters; assume again we have two instances; instance1 is bound to iface1 and instance2 is bound to iface2 by following config

config dnsmasq 'instance1'
        option nonwildcard '1'
        list interface 'iface1'
        ...

config dnsmasq 'instance2'
       option nonwildcard '1'
        list interface 'iface2'
        list notinterface 'lo'
        ...

dedeckeh:
ping 
any other comments on this PR ?

cshoredaniel:
On Wed, 12 Oct 2016 01:58:19 -0700
Hans Dedecker <notifications@github.com> wrote:

> Adds support in uci for configuring multiple dnsmasq instances via
> multiple dnsmasq sections.

Does it actually work without jails?  Last time I tried to do this I
ended with something that works with procd-ujail but not without jails.

Regards,

Daniel

dedeckeh:
@cshoredaniel Yes it works without jails using the config described in one of my previous comments

jow-:
Merged into my staging tree: http://git.lede-project.org/e777eec

bobafetthotmail:
now that this is on its way to be merged, the paragraph "Multiple DHCP/DNS server/forwarder instances" in the wiki can be filled https://wiki.lede-project.org/docs/user-guide/dhcp_configuration 

I'd rather have @dedeckeh do it as he is the most likely one to do it right, if none does it within a few days I'm doing it using the comments above as a base.

dedeckeh:
Op 25 okt. 2016 20:16 schreef "Daniel Dickinson" <notifications@github.com>:
>
> On Wed, 12 Oct 2016 01:58:19 -0700
> Hans Dedecker <notifications@github.com> wrote:
>
> > Adds support in uci for configuring multiple dnsmasq instances via
> > multiple dnsmasq sections.
>
> Does it actually work without jails? Last time I tried to do this I
> ended with something that works with proces-ujail but not without jails.
Yes it works without jails using the config described in one of my previous
comments

Hans

dedeckeh:
@bobafetthotmail I will update the "Multiple DHCP/DNS server/forwarder instances" section in the wiki and the other topics which can refer to a DHCP/DNS server/forwarder instance; will be probably not next week due to other commitments

bobafetthotmail:
thanks, will wait a bit then.

PR #409, opened 2016-10-12
mbedtls: enable NIST curves optimisation.

kdarbyshirebryant:
luci using ustream-mbedtls is extremely slow vs ustream-polarssl.
polarssl alias mbedtls v1 is configured to use NIST prime speed
optimisation, so no longer disable the default optimisation for
mbedtls v2.

Compile & run tested: Archer C7v2


jow-:
Merged into my tree with http://git.lede-project.org/a5d3b08

PR #411, opened 2016-10-13
ustream-ssl: sanitize ssl library selection & building

kdarbyshirebryant:
Both polarssl & mbedtls were selected as the default build variant.
Only build the selected variant/s.

### DIFF
--- a/package/libs/ustream-ssl/Makefile
+++ b/package/libs/ustream-ssl/Makefile
@@ -47,7 +47,7 @@ endef
 define Package/libustream-polarssl
   $(Package/libustream/default)
   TITLE += (polarssl)
-  DEPENDS += +libpolarssl
+  DEPENDS += +PACKAGE_libustream-polarssl:libpolarssl
   VARIANT:=polarssl
   DEFAULT_VARIANT:=1
 endef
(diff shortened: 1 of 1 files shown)


hnyman:
Yep, I noticed the same when looking at luci-app-openssl. I thought to write a PR, but you got there first ;-)

Ps. but there is some work left for others, too: px5g has no default variant (and px5g-mbedtls may get picked wrongly instead of px5g-polarssl when the dependency is just for px5g). https://github.com/openwrt/luci/issues/814#issuecomment-251324248

kdarbyshirebryant:
I wonder if the following tweak https://github.com/kdarbyshirebryant/lede-source/commit/8f7b450fb0e4f647c53d252668052704e8e3f31b would be helpful.  I found that without it, px5g forced downloads of mbedtlsv1 & v2 which wasn't expecting.

hnyman:
I think that a cleaner solution would be to add a new parameter to the variant creation call, to be set for the DEFAULT_VARIANT variable in each variant. 0 for mbedtls and 1 for polarssl to set it as the default. I haven't tried that yet, but that seems like a logical solution. (I am not sure if that is enough, but I guess that it should be done in any case.)

Note that LuCI-ssl(-*) collections have just a dependency for plain px5g, which may cause trouble if px5g-polarssl is the default while you aim for mbedtls-only based system. I would prefer to have a clear config ID symbol to be used for "we are now in LEDE", which could then be used in LuCI (and elsewhere) to adjust dependencies a bit differently for Openwrt and LEDE (e.g. explicitly define px5g-mbedtls or px5g-polarssl dependency for LEDE, while leaving plain px5g for Openwrt (that only has the plain px5g without variants).

PR #415, opened 2016-10-14
ath10k-firmware: update qca9984 firmware

dissent1:
A new qca9984 ath10k firmware has been released.


jow-:
Merged into my staging tree http://git.lede-project.org/aa231b2

PR #416, opened 2016-10-14
elfutils: bump to 0.167

luizluca:
Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

@nbd168, you introduced the patch 110-elf_compressed_header_support.patch, which I'm removing as upstream now has this commit:

https://git.fedorahosted.org/cgit/elfutils.git/commit/?h=elfutils-0.167&id=344ca0775da729e1bfdd61bb88ba4c64befece07

The only difference is that you checked for ELFCOMPRESS_ZLIB while they check for SHF_COMPRESSED. Is the upstream code enough for you?


nbd168:
Added to my staging tree, thanks.

PR #418, opened 2016-10-16
xfsprogs: install path consistent with fs tools

bobafetthotmail:
changed install path from /sbin to /usr/sbin to be consistent with other filesystem tools 
ext2-3-4 and f2fs tools are in /usr/sbin, for example


blogic:
merged, Thanks !

PR #419, opened 2016-10-17
toolchain: build fix for Ubuntu 16.10

teknoraver:
Ubuntu 16.10 compiles with `-fpie' by default
This breaks toolchain cross compilation, add `-fno-pie' to override it

### DIFF
--- a/toolchain/gcc/common.mk
+++ b/toolchain/gcc/common.mk
@@ -92,8 +92,8 @@ endif
 GCC_CONFIGURE:= \
 	SHELL="$(BASH)" \
 	$(if $(shell gcc --version 2>&1 | grep LLVM), \
-		CFLAGS="-O2 -fbracket-depth=512 -pipe" \
-		CXXFLAGS="-O2 -fbracket-depth=512 -pipe" \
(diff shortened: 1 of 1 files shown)


blogic:
@teknoraver guess what i will point out ? same as the last 4 times ... Signed-off-by is missing ... 

please take a piece of paper and write on it "i shall never post patches without my SoB ever again" and hang it on your wall so you see it every day ;)

blogic:
this is a llvm option added to the gcc paramters. is your ubuntu using llvm or did you incorrectly rebase this patch ?

teknoraver:
That's a GCC option, according to the manpage:
```
-fpie
-fPIE
	These options are similar to -fpic and -fPIC, but generated position independent code can be only linked into executables. Usually these options are used when -pie GCC option is used during linking.

	-fpie and -fPIE both define the macros "__pie__" and "__PIE__". The macros have the value 1 for -fpie and 2 for -fPIE.
```

jow-:
Yeah, but you're adding it to a place that is protected by `gcc --version 2>&1 | grep LLVM`

teknoraver:
can't reproduce anymore, maybe it was something unrelated, or a bug fixed by Ubuntu

PR #420, opened 2016-10-17
build: scripts/package-metadata.pl: conditionalize all source depende…

nwf:
…ncies

Make all source dependencies apply only when packages have been
selected.  Most of these changes are no-ops, as typically a source
package has only one package; thus

$(curdir)/feeds/packages/foo/compile += ...

becomes the equivalent

$(curdir)/feeds/packages/foo/compile += $(if $(CONFIG_PACKAGE_foo), ...)

However, when one source package has multiple build packages in it (e.g.
gnunet), it should not be the case that selecting any one package pulls
in all dependencies of any package that shares the same source package.
This change results in dependencies being broken out explicitly.  Rather
than

$(curdir)/feeds/routing/bar/compile += bar-deps... baz-deps...

we get

$(curdir)/feeds/routing/bar/compile += $(if $(CONFIG_PACKAGE_bar), bar-deps...)
$(curdir)/feeds/routing/bar/compile += $(if $(CONFIG_PACKAGE_baz), baz-deps...)

which is a more accurate assment of the situation and can dramatically
shave the dependency list.

Unfortunately, the resulting lists are quite redundant, but make should
make quick work of the duplicated targets.

Note that this diff is best viewed ignoring whitespace due to an
indentation change of otherwise unmodified code.

There may well be a better fix.

Signed-Off-By: Nathaniel Wesley Filardo nwfilardo@gmail.com

### DIFF
--- a/scripts/package-metadata.pl
+++ b/scripts/package-metadata.pl
@@ -406,7 +406,7 @@ sub gen_package_mk() {
 	foreach my $name (sort {uc($a) cmp uc($b)} keys %package) {
 		my $config;
 		my $pkg = $package{$name};
-		my @srcdeps;
+		my %srcdeps;
 
 		next if defined $pkg->{vdepends};
 
@@ -431,12 +431,14 @@ sub gen_package_mk() {
 		}
 
 		foreach my $spkg (@{$srcpackage{$pkg->{src}}}) {
+			my $pkgsrcdeps = [ ];
 			foreach my $dep (@{$spkg->{depends}}, @{$spkg->{builddepends}}) {
 				$dep =~ /@/ or do {
 					$dep =~ s/\+//g;
-					push @srcdeps, $dep;
+					push @$pkgsrcdeps, $dep;
 				};
 			}
+			$srcdeps{$spkg->{name}} = $pkgsrcdeps;
 		}
 		foreach my $type (@{$pkg->{buildtypes}}) {
 			my @extra_deps;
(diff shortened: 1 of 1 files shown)


jow-:
The idea makes sense but has the potential to break some multi-ipk source packages;
So far one could rely on the fact that a binary dependency implies an unconditional build dependency on the package in question and that a given source bundle build-depends on the union of all (unconditional) binary dependencies...

However I'd be willing to give it a spin and then fix the resulting fallout as it appears (assuming there even is some in the first place).

@lede-project/developer - opinions?

nwf:
It would be pretty easy to add the implied dependency explicitly if it turns out to have large breakage.  From my (completely unscientific) memory, though, most packages are good about having explicit dependencies (e.g. bar-cli DEPENDS +libbar).

nbd168:
The problem with dealing with fallout from this change is the fact that it's hard to detect. You need to clean and build each package individually, each time with a fresh staging_dir, and also test multiple variations of package selections for the same source package.

I think pulling in all dependencies is a reasonable default. Any package that properly deals with their build dependencies can explicitly make them conditional with the existing code, the way that many packages already do:

`define Package/foo  ...  DEPENDS:=+PACKAGE_foo:bar ... endef`

If that looks a bit convoluted, I would be okay with adding a flag to make this more explicit.

Either way, NACK from me on this change

nwf:
@nbd168 Can you expand on the explicit flag alternative, as the proposed workaround seems pretty gross when there are a lot of dependencies (though the motivating example of gnunet, which does have a lot of dependencies, already has them going through macros, so perhaps it'd be easy to adjust in one place).

PR #421, opened 2016-10-18
ramips: add support for MikroTik hEX v3 (RB750Gr3)

ndoo:
Documentation on Wiki is forthcoming; device has to be flashed through SPI programmer due to proprietary bootloader.
Signed-off-by: Andrew Yong me@ndoo.sg

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -134,6 +134,7 @@ ramips_setup_interfaces()
 	jhr-n825r|\
 	jhr-n926r|\
 	mzk-wdpr|\
+	rb750gr3|\
 	rt-n14u|\
 	ubnt-erx|\
 	ur-326n4g|\
--- a/target/linux/ramips/dts/RB750Gr3.dts
+++ b/target/linux/ramips/dts/RB750Gr3.dts
@@ -1,0 +1,121 @@
+/dts-v1/;
+
+#include "mt7621.dtsi"
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+
+/ {
+	compatible = "mediatek,mt7621-eval-board", "mediatek,mt7621-soc";
+	model = "MikroTik RB750Gr3";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x10000000>;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,57600";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		usr {
+			label = "rb750gr3:green:usr";
+			gpios = <&gpio0 0 GPIO_ACTIVE_HIGH>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		mode {
+			label = "mode";
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+		};
+
+		res {
+			label = "res";
+			gpios = <&gpio0 18 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	gpio_export {
+		compatible = "gpio-export";
+		#size-cells = <0>;
+
+		usb {
+			gpio-export,name = "usb";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 12 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		linux,modalias = "m25p80";
+		spi-max-frequency = <10000000>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x30000>;
+			read-only;
+		};
+
+		partition@30000 {
+			label = "u-boot-env";
+			reg = <0x30000 0x10000>;
+			read-only;
+		};
+
+		factory: partition@40000 {
+			label = "factory";
+			reg = <0x40000 0x10000>;
+			read-only;
+		};
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
+		};
+
+	};
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xe000>;
+	mtd-mac-address-increment = <1>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "i2c", "rgmii2", "sdhci";
+			ralink,function = "gpio";
+		};
+	};
+};
+
+&sdhci {
+	status = "okay";
+};
+
+&xhci {
+	status = "okay";
+};
+
+&pcie {
+	status = "disabled";
+};
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -190,6 +190,9 @@ get_status_led() {
 	wr512-3gn)
 		status_led="$board:green:wps"
 		;;
+	rb750gr3)
+		status_led="$board:green:usr"
+		;;
 	sap-g3200u3)
 		status_led="$board:green:usb"
 		;;
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -367,6 +367,9 @@ ramips_board_detect() {
 	*"Q7")
 		name="zte-q7"
 		;;
+	*"RB750Gr3")
+		name="rb750gr3"
+		;;
 	*"RE6500")
 		name="re6500"
 		;;
--- a/target/linux/ramips/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/base-files/lib/upgrade/platform.sh
@@ -105,6 +105,7 @@ platform_check_image() {
 	psg1218|\
 	psr-680w|\
 	px-4885|\
+	rb750gr3|\
 	re6500|\
 	rp-n53|\
 	rt5350f-olinuxino|\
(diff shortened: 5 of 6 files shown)


target/linux/ramips/dts/RB750Gr3.dts
```diff
@@ -0,0 +1,106 @@
...
+	status = "okay";
+};
+
+&xhci {
+	status = "okay";
+};
+
+&pcie {
+	status = "disabled";
+};
```
blogic: there seems to be a stray character at the end of the last line

ndoo:
@blogic did a fixup and pushed again, GitHub no longer complains about the missing \n at EOF now. Thanks for pointing that out.

ndoo:
Ok, after many squashed commits, I am now happy with the state of this PR and will not be making further changes.

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -337,6 +338,10 @@ ramips_setup_macs()
 		lan_mac=$(mtd_get_mac_binary factory 40)
 		wan_mac=$(mtd_get_mac_binary factory 46)
 		;;
+	rb750gr3)
+		wan_mac=$(cat /sys/class/net/eth0/address)
```
mkresin: Please use ```wan_mac=$(mtd_get_mac_binary factory 57344)``` here

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -337,6 +338,10 @@ ramips_setup_macs()
...
+		lan_mac=$(macaddr_add "$wan_mac" 1)
```
mkresin: Drop this line and add a ```mtd-mac-address-increment = <1>;``` to the ethernet node in your device tree file.

target/linux/ramips/dts/RB750Gr3.dts
```diff
@@ -0,0 +1,111 @@
...
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x10000000>;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,57600";
+	};
+
+	gpio-leds {
```
ndoo: It is not GPIO controllable

mkresin: might be a good idea to add a note here that the power led is not gpio controllable.

target/linux/ramips/base-files/etc/board.d/02_network
ndoo: It seems `mtd-mac-address-increment = <1>;` only works for incrementing WAN MAC address from LAN MAC address.

In the case of RB750Gr3, WLLLL layout, would `mtd-mac-address-increment` function as expected? Because only wan_mac was specified here.

mkresin: mtd-mac-address-increment shouldn't care about the switch vlan layout. The change should be applied to all vlans and you need to fix it for the wan vlan afterwards. But I'm not entirely sure since I don't have hardware to test.

Please give it a try and report back.

ndoo:
Changes made except request to have LAN MAC address increment in DTS with `mtd-mac-address-increment = <1>;` as it seems to only work for WAN MAC address increment (I could be wrong)

mkresin:
Andrew, would you please do me favour:

1. drop the mac address stuff from board.d/02_network
2. change the ethernet dts node to
```
&ethernet {
	mtd-mac-address = <&factory 0xe000>;
	mtd-mac-address-increment = <1>;
};
```
4. paste the mac address stored at offset 0xe000 of the factory partition
3. paste the output of ```ifconfig -a```

I would like to understand what is going on there.

ndoo:
Compiling now. Basically I had to erase the entire SPI and start from scratch with a custom U-Boot.

So the Factory partition was created by hand and I actually just put the MAC address at that offset (rest of partition is 0xFF). In such a case of "clean room" implementation it’s entirely possible to get rid of factory partition and use a uEnv variable to store MAC address.

I just kept the factory mtd similar to other MT7621 reference platforms because the reference SDK U-Boot expects the kernel to start at the offset after the factory mtd.

ndoo:
MAC address at `<&factory 0xe000>`: `3B6C656B3204`

`ifconfig -a`:

    br-lan    Link encap:Ethernet  HWaddr 6C:3B:6B:65:04:33  
              inet addr:192.168.1.1  Bcast:192.168.1.255  Mask:255.255.255.0
              inet6 addr: fe80::6e3b:6bff:fe65:433/64 Scope:Link
              inet6 addr: fd0a:6549:d2f9::1/60 Scope:Global
              UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
              RX packets:2860 errors:0 dropped:0 overruns:0 frame:0
              TX packets:2231 errors:0 dropped:0 overruns:0 carrier:0
              collisions:0 txqueuelen:1000 
              RX bytes:339778 (331.8 KiB)  TX bytes:1330431 (1.2 MiB)

    eth0      Link encap:Ethernet  HWaddr 6C:3B:6B:65:04:33  
              inet6 addr: fe80::6e3b:6bff:fe65:433/64 Scope:Link
              UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
              RX packets:6117 errors:0 dropped:0 overruns:0 frame:0
              TX packets:5262 errors:0 dropped:0 overruns:0 carrier:0
              collisions:0 txqueuelen:1000 
              RX bytes:1953579 (1.8 MiB)  TX bytes:1753265 (1.6 MiB)
              Interrupt:10 

    eth0.1    Link encap:Ethernet  HWaddr 6C:3B:6B:65:04:33  
              UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
              RX packets:2884 errors:0 dropped:3 overruns:0 frame:0
              TX packets:2224 errors:0 dropped:0 overruns:0 carrier:0
              collisions:0 txqueuelen:1000 
              RX bytes:343202 (335.1 KiB)  TX bytes:1329709 (1.2 MiB)

    eth0.2    Link encap:Ethernet  HWaddr 6C:3B:6B:65:04:34  
              inet addr:192.168.100.24  Bcast:192.168.100.255  Mask:255.255.255.0
              inet6 addr: fe80::6e3b:6bff:fe65:434/64 Scope:Link
              inet6 addr: <redacted>/64 Scope:Global
              UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
              RX packets:3038 errors:0 dropped:7 overruns:0 frame:0
              TX packets:2767 errors:0 dropped:0 overruns:0 carrier:0
              collisions:0 txqueuelen:1000 
              RX bytes:1490142 (1.4 MiB)  TX bytes:360565 (352.1 KiB)

    lo        Link encap:Local Loopback  
              inet addr:127.0.0.1  Mask:255.0.0.0
              inet6 addr: ::1/128 Scope:Host
              UP LOOPBACK RUNNING  MTU:65536  Metric:1
              RX packets:16 errors:0 dropped:0 overruns:0 frame:0
              TX packets:16 errors:0 dropped:0 overruns:0 carrier:0
              collisions:0 txqueuelen:1 
              RX bytes:1408 (1.3 KiB)  TX bytes:1408 (1.3 KiB)

ndoo:
I have gone ahead and pushed the changes to my branch to move the MAC address spec into DTS instead of 02_network.

It does not have any user-facing impact; MikroTik would have done :32-:36 if ports are broken out from switch on MikroTik RouterOS anyway, so no matter how they are assigned here they will not exceed the allocated MAC address space for the hardware.

mkresin:
>Expected Behaviour (to replicate what MikroTik RouterOS creates for home setup)
Ether1 (WAN) = 6C:3B:6B:65:04:32
Ether2-5 (LAN) = 6C:3B:6B:65:04:33

Use something like ````wan_mac=$(macaddr_add "$(mtd_get_mac_binary factory 57344)" -2)``` in the board.d network file to use the same macs as the MikroTik RouterOS. I've no idea who increments the wan mac address. Might be a (switch) driver feature but I don't have the time to look at the driver source.

> Compiling now. Basically I had to erase the entire SPI and start from scratch with a custom U-Boot.
> So the Factory partition was created by hand and I actually just put the MAC address at that offset (rest of partition is 0xFF). In such a case of "clean room" implementation it’s entirely possible to get rid of factory partition and use a uEnv variable to store MAC address.

Which means you whole commit does only work with the board if a lot of special preparations are done. Please add this kind of information to the commit message.

This sounds like a really hackish approach. What's wrong with the MikroTik u-boot? Why not reuse the MikroTik partitions (or at least the partition that has the mac address stored)? How does the partition layout looks like used by the MikroTik firmware? These are all information I expect to see in the commit message in your case.

ndoo:
I will document the flashing instructions in the OpenWrt Wiki for now until LEDE has somewhere to document it.

MikroTik ships it with proprietary bootloader out of the box (RouterBOOT) that is very different from MediaTek SDK (and definitely not U-Boot), and they disabled UART on the bootloader and OS. Thus it is not straightforward to implement an image that will be flashable, and bootable, by RouterBOOT.

Stock firmware uses proprietary flash structure and squashfs headers as well with no GPL sources provided for squashfs tools.

Most recent MikroTik are not able to support OpenWrt/LEDE by virtue of using proprietary, silent bootloader as well as using NAND flash, which is harder to flash over. As this model is the first in recent time to use SPI flash, this is **relatively** easy to get LEDE installed on.

I would put it in commit message but seems more appropriate to maintain it in Wiki. It does sound involved at first, but the essential result of all this work is to turn the RB750Gr3 hardware back into a MT7621 dev board equivalent where I have a U-Boot that works with minimal modification (https://github.com/ndoo/RB750Gr3-U-Boot) and try to adhere as much as possible to known offsets & DTS conventions in MediaTek SDK and LEDE.

mkresin:
The commit message should make clear why something is done. And your text above is perfect. 

You should add a note in the commit message that flashing has to be done through SPI programmer and that you created a complete new partition layout without using any of the former existing partitions.

How to setup the partitions and/or install LEDE can go into the wiki.

The point is if someone has issue with the board or need to change setting introduced with the patch he is literal lost without the extra informations you provided. There is no obvious indication why things are done this way. And to be honest, I expect to have these informations in the repository and not at $random_website which might not be available (anymore) or someone removed the informations from this site because the are not relevant for daily use.

ndoo:
Where would an appropriate place to document this be? I’m not sure I can fit all that explanation into a commit message.

mkresin:
Just put it in the commit message. To my knowledge there is no character limit.

ndoo:
Done.

blogic:
he meant the "put it into the patch" not "put it in the PR"

ndoo:
My bad. Fixup done - added description/motivation of this RB750Gr3 support patch into the commit

blogic:
merged, Thanks !

PR #428, opened 2016-10-24
px5g-standalone: move to Encryption submenu

bobafetthotmail:
Part of a wider housekeeping effort on the packages repository.


blogic:
the description does not describe what the patch does. please add a better description

PR #430, opened 2016-10-24
util-linux: move packages to other submenus

bobafetthotmail:
cal, dmesg, flock, getopt, hwclock, logger, look, losetup, mcookie, namei, prlimit, rename, whereis moved to Util-linux submenu

moved packages with "disk" submenu in Filesystem submenu as "disk" (actually mostly partitioning/mounting tools) and "Filesystem" are close enough and we can do without this redundancy.

moved partx-utils, mount-utils, uuidd, wipefs and uuidgen to Filesystem submenu.

Part of a wider housekeeping effort on the packages repository.

### DIFF
--- a/package/utils/util-linux/Makefile
+++ b/package/utils/util-linux/Makefile
@@ -123,7 +123,7 @@ endef
 define Package/blkdiscard
 $(call Package/util-linux/Default)
   TITLE:=discard sectors on a device
-  SUBMENU=disc
+  SUBMENU=Filesystem
 endef
 
 define Package/blkdiscard/description
@@ -136,7 +136,7 @@ define Package/blkid
 $(call Package/util-linux/Default)
   TITLE:=locate and print block device attributes
   DEPENDS:= +libblkid +libuuid
-  SUBMENU=disc
+  SUBMENU=Filesystem
 endef
 
 define Package/blkid/description
@@ -148,6 +148,7 @@ define Package/cal
 $(call Package/util-linux/Default)
   TITLE:=display a calendar
   DEPENDS:= +libncurses
+  SUBMENU=Util-linux
 endef
 
 define Package/cal/description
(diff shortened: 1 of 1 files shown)


package/utils/util-linux/Makefile
```diff
@@ -123,7 +123,7 @@ endef
 define Package/blkdiscard
 $(call Package/util-linux/Default)
   TITLE:=discard sectors on a device
-  SUBMENU=disc
+  SUBMENU=Filesystem
 endef
```
blogic: thiis disc/block related and not FS related

package/utils/util-linux/Makefile
```diff
@@ -148,6 +148,7 @@ define Package/cal
 $(call Package/util-linux/Default)
   TITLE:=display a calendar
   DEPENDS:= +libncurses
+  SUBMENU=Util-linux
 endef
```
blogic: calendar will most likely be harder to find under the util-linux submenu

package/utils/util-linux/Makefile
```diff
@@ -158,7 +159,7 @@ define Package/cfdisk
 $(call Package/util-linux/Default)
   TITLE:=display or manipulate disk partition table
   DEPENDS:= +libblkid +libncurses +libsmartcols +libfdisk +libmount
-  SUBMENU:=disc
+  SUBMENU=Filesystem
 endef
```
blogic: "manipulate disc partitions" how is that FS related ?

blogic:
NAK. i looked over the first 4 changes and consider them all wrong or bogus.

bobafetthotmail:
Ok, I will probably just capitalize the "disk" in "Disk" with next PR.

>calendar will most likely be harder to find under the util-linux submenu

As it is now, once you enable the package feeds it's dumped in a multi-page list of random utils, like most others. I think that moving them to a single submenu would make them easier to find than scrolling through like 150 other packages ordered by name, even if the name of the submenu is more generic like utils-linux.

blogic:
moving stuff for the sake of it does not improve anything. either move stuff to a place that is obvious and easy to find or dont move it. moving it to an equally bad place will result in the patch not getting merged.

bobafetthotmail:
Fair enough.

PR #431, opened 2016-10-24
fconfig: move to Boot Loaders submenu of Utilities

bobafetthotmail:
Boot Loaders submenu of Utilities is the most logical place to find fconfig and other bootloader tools.


blogic:
merged, Thanks !

PR #440, opened 2016-10-25
lantiq: vr9 fxs support: Feature vr9vpe

kochstefan:
This patchset is based on an earlier patchset from Eddi De Pieri.
It was committed to the OpenWRT BB branch in 11/2014.
For details see description of single commits.

The patchset allows using the FXS ports
(for analogue telephones) from Lantiq VR9 based routers.

If the VR9 based router provides FXS ports and they shoud enabled then the following must added to the kernel command line:
mem=[TOTALMEMSIZE-2M] vpe1_load_addr=ADDRESS vpe1_mem=2M maxvpes=1 maxtcs=1

To use FXS 2M of RAM are needed for the VPE firmware. The size is given by vpe1_mem.
The available RAM must be reduced by this size using the mem argument.
A correct load address (example 0x83e00000) for the firmware must given, too.

Furthermore the following changes are needed to use FXS:
- Command line must be added to DTS of device (see above)
- DTS of device needs vmmc part similar to commit 1379a2d39c43b88badf992866ee1df1c35e8064a


blogic:
@hauke @mkresin @xdarklight 
looks good. comments/vetos before i merge this ?

mkresin:
Looks good to me.

But I would like to wait for a reply of @hauke since he seam to have played with the telephony subsystem recently.

blogic:
merged, Thanks !

PR #441, opened 2016-10-25
Change the way SPI and HSSPI modules are compiled

pgid69:
Compile them as dynamic loadable modules instead of being built into the kernel.
That's because kernel module https://github.com/pgid69/bcm63xx-phone.git needs full control of the SPI bus so there's must be a way to rmmod bcm63xx-spi without recompiling the kernel.

Signed-off-by: Gilles Mazoyer <mazoyer.gilles@omega.ovh>


KanjiMonster:
NAK. These need to stay built-in, else the kernel won't be able to boot from SPI flash (as required on many devices).

PR #445, opened 2016-10-27
Split the ppp-up for the IPv6 part

pticon:
If the IPv6 link over a PPPOE is established before the IPv4, it results that the IPv4 DNS servers are overwritten. The "resolv.conf" file will have duplicated IPv4 DNS servers. The main issue is that clients are not able to resolve IPv6 domain anymore.

The idea behind this fix is to split the file "/lib/netifd/ppp-up" in two parts : one for IPv4 and one for IPv6. Indeed, the IPv6 DNS servers will be updated only for the IPv6 link and the IPv4 DNS servers will be also updated only with the IPv4 link.


dedeckeh:
@pticon The PR contains different commits which are not related to the PPP change

jow-:
The change and approach look sane to me, though I don't have the prerequisites avaialble to runtime test it.

PR #447, opened 2016-10-27
ppp: split the ppp-up for the IPv6 part

pticon:
Signed-off-by: Pierre Lebleu pme.lebleu@gmail.com
#445 This is the same commit based on the LEDE branch.

If the IPv6 link over a PPPOE is established before the IPv4, it results that the IPv4 DNS servers are overwritten. The "resolv.conf" file will have duplicated IPv4 DNS servers. The main issue is that clients are not able to resolve IPv6 domain anymore.

The idea behind this fix is to split the file "/lib/netifd/ppp-up" in two parts : one for IPv4 and one for IPv6. Indeed, the IPv6 DNS servers will be updated only for the IPv6 link and the IPv4 DNS servers will be also updated only with the IPv4 link.

@jow- 
I give you the tests that I ran :

network configuration

```
config interface 'wan'
        option proto 'pppoe'
        option ifname 'eth4.835'
        option demand '0'
        option keepalive '4,30'
        option graceful_restart '1'
        option peerdns '1'
        option reqopts '1 3 6 15 33 42 51 121 249'
        option keepalive_adaptive '0'
        option release '1'
        option username 'user@domain'
        option password 'password'
        option ipv6 '1'
```

before the fix

```
cat /tmp/resolv.conf.auto 
# Interface wan
nameserver x.x.x.x
nameserver y.y.y.y
nameserver x.x.x.x
nameserver y.y.y.y
# Interface wan6
nameserver 2001:dead:beef:aaaa::11:c001
nameserver 2001:dead:beef:aaaa::11:c002
```

after the fix

```
cat /tmp/resolv.conf.auto 
# Interface wan
nameserver x.x.x.x
nameserver y.y.y.y
# Interface wan6
nameserver 2001:dead:beef:aaaa::11:c001
nameserver 2001:dead:beef:aaaa::11:c002
```

I understand your worries about the commit and I can run extra tests if you want. Please, let me know.

Cheers

### DIFF
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -206,6 +206,7 @@ define Package/ppp/install
 	$(INSTALL_DIR) $(1)/lib/netifd/proto
 	$(INSTALL_BIN) ./files/ppp.sh $(1)/lib/netifd/proto/
 	$(INSTALL_BIN) ./files/lib/netifd/ppp-up $(1)/lib/netifd/
+	$(INSTALL_BIN) ./files/lib/netifd/ppp6-up $(1)/lib/netifd/
 	$(INSTALL_BIN) ./files/lib/netifd/ppp-down $(1)/lib/netifd/
 endef
 Package/ppp-multilink/install=$(Package/ppp/install)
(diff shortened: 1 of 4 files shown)


blogic:
patch is missing the prefix in this case "ppp:" and you also forgot to add a description text explaining what the patch does and why

pticon:
@blogic I added a description in the commit message

pticon:
@jow- @blogic I found an other the way to fix this issue. I'm going to do extensive tests on this to be sure not to break anything.

pticon:
@jow- 
After some extra tests, I can state that this patch is the right way to fix this issue. Indeed, I closed the #451 due to the fact that it could flush the IPv4 address and the IPv4 default route. So, the "proto_set_keep 1" is needed.

dedeckeh:
Tested by Hans Dedecker  <dedeckeh@gmail.com>

pticon:
@jow- @blogic 
Any news on this patch ?

nbd168:

PR #451, opened 2016-10-28
ppp: Remove the proto_set_keep in /lib/netifd/ppp-up

pticon:
It fixes the issue when the PPP IPv4 link is coming up before
the PPP IPv6 link on the same interface. Indeed, the IPv4
name servers were duplicated in the resolv.conf file.

Signed-off-by: Pierre Lebleu pme.lebleu@gmail.com

### DIFF
--- a/package/network/services/ppp/files/lib/netifd/ppp-up
+++ b/package/network/services/ppp/files/lib/netifd/ppp-up
@@ -3,7 +3,6 @@ PPP_IPPARAM="$6"
 
 . /lib/netifd/netifd-proto.sh
 proto_init_update "$IFNAME" 1 1
-proto_set_keep 1
 [ -n "$PPP_IPPARAM" ] && {
 	[ -n "$IPLOCAL" ] && proto_add_ipv4_address "$IPLOCAL" 32 "" "${IPREMOTE:-2.2.2.2}"
 	[ -n "$IPREMOTE" ] && proto_add_ipv4_route 0.0.0.0 0 "$IPREMOTE"


pticon:
@jow- It fixes the issue #447 in a different way.
I will run extensive tests this week to check if everything is fine.

jow-:
@pticon - any news? Shall I prefer this one over #447 ?

pticon:
@jow- Yes, this one seems to be better than #447.
Just running more tests on this to prevent any issue.

nbd168:
I think this will break configurations that use both IPv6 and IPv4 over PPP

pticon:
@nbd168 I made the same test as #447 with this patch and it works well. 
This 'proto_set_keep' is a old stuff and it does not seem to be needed anymore.

dedeckeh:
@nbd168 Initially I thought this would indeed break configurations using both IPv6 and IPv4 over PPP as I tracked down proto_set_keep as a fix for dual stack route handling.
However after doing some testing on my setup with proto_set_keep removed and the latest version of netifd I did not notice any routing issue in a dual stack scenario. Maybe the issue is not visible anymore as the IPv6 default route is now created by the dhcpv6 protocol which was not the case when proto_set_keep was added.

As the ppp-up script is called for both ipv4 and ipv6 up the DNS servers can be duplicated in a dual stack scenario due to proto_set_keep 1.
This is the case if ipv4 is first up; when ipv6 is up the PPP daemon passes again the env variables IPLOCAL, IPREMOTE, DNS1 and DNS2 to the ppp-up script together with LLLOCAL (PPP daemon does not unset the env variables after calling the up script for a given protocol).
This causes the IPv4 DNS servers being duplicated in netifd for the ppp interface meaning resolv.conf.auto contains the same DNS servers twice.
Problem can indeed be solved by removing proto_set_keep or by calling a separate script when ipv6 is up as in PR #447 which requires proto_set_keep to be kept.
Which of the above two solutions has your preference ?

pticon:
I close this pull request because it seems the 'proto_set_keep' is needed sometimes.
The PR #447 seems to work fine.

PR #452, opened 2016-10-28
tools/qemu-img: upgrade to version 2.7.0 (and all that comes with it)

commodo:
QEMU requires glib2, which then requires gettext, and libffi.
Template for Glib2 was taken from the packages feed.

For `libs/gettext-full` I'm disabling the HostBuild for x86 ; since it's installed as part of tools.
( I hope this is correct/fine )

libiconv seemed to be optional (for gettext), so I am installing it optionally (same as gettext-full does for the host build).

I tried to add the minimally required stuff for the Makefiles.

Will reference for review:
- @tripolar for libffi
- @jow- : for gettext
- @yousong : for qemu

Did not see any licensing info for tools/*
If I need to add something, please notify me.

Compile/Tested: LEDE (on Mac OS X)


tools/qemu/patches/001-disable-mmx-for-pixman-on-macos-x.patch
```diff
@@ -0,0 +1,17 @@
...
+index 50b4b3a..fecd35d 100644
+--- a/Makefile
++++ b/Makefile
+@@ -206,8 +206,11 @@ subdir-%:
+ subdir-pixman: pixman/Makefile
+ 	$(call quiet-command,$(MAKE) $(SUBDIR_MAKEFLAGS) -C pixman V="$(V)" all,)
+ 
++ifeq ($(shell uname -s),Darwin)
++PIXMAN_DISABLE_MMX:=--disable-mmx
++endif
```
yousong: Can we use configure option `--without-pixman` to replace this patch?

commodo:
@yousong re-spinned

commodo:
I also changed the Host/Configure/Default rule.
It built fine with that change.

yousong:
ack from me then, thanks.

jow-:
Hm, thats quite a lot new host tools just to support qemu-img :/ Is glib2 and its associated dependency chain a new requirement in Qemu?

commodo:
they are required ;
but asking the question raises an interesting point

maybe i could try to see if we could get rid of those deps [some of them, hopefully all of them] by finding out which configure options to disable in the QEMU build;
as it is now, QEMU requires quite a bit of stuff to build

and i'm also not a fan of adding bloat [like this] for something like this [if we can avoid it]

i'll close the PR for now ; and come back when i have an answer for getting rid of these deps

jow-:
Thanks @commodo - much appreciated. If there is no way around this we need to see how we can integrate that stuff properly (or we find an alternative for what qemu-img is doing)

commodo:
so, i've tried to remove glib2 from qemu ; seems it's a bit hard;
glib2 seems quite entrenched in qemu

i tried patching the configure script to remove glib2 deps & checks
i even tried to remove the `glib-compat.h`

got to this point in errors
```
In file included from qemu-img.c:36:
In file included from /Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qom/object_interfaces.h:4:
/Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qom/object.h:334:5: error: unknown type name 'gchar'; did you mean 'char'?
    gchar *name;
    ^~~~~
    char
/Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qom/object.h:335:5: error: unknown type name 'gchar'; did you mean 'char'?
    gchar *type;
    ^~~~~
    char
/Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qom/object.h:336:5: error: unknown type name 'gchar'; did you mean 'char'?
    gchar *description;
    ^~~~~
    char
/Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qom/object.h:373:5: error: unknown type name 'GSList'; did you mean 'QList'?
    GSList *interfaces;
    ^~~~~~
    QList
/Users/aardelean/work/sources-work/lede1/build_dir/host/qemu-2.7.0/include/qapi/qmp/qlist.h:27:3: note: 'QList' declared here
} QList;
```

i tried removing a few headers, just for kicks; 
but there are also `g_` and `G` prefixed funcs and types ; which look like they're from Glib2

this is how far I've tried to push this: https://github.com/commodo/lede/commit/346098ffd67a08e0e3d1541a77be2a14c9bda5c0
I guess, it's not worth going further anymore, since I'd end up refactoring stuff a bit

maybe, internally we'll move to keeping qemu-img packaged ; rather than keeping it in `tools/`
that would help avoid too much divergence from LEDE and our internal tree [ which i my main goal atm ]

jow-:
What was your exact use case for the updated qemu-img ?

commodo:
VHD images for OpenWrt/LEDE to be used in Azure.
AWS is fine with current qemu-img tool.

PR #453, opened 2016-10-28
[RFC] ipq806x: workaround memory issue in Netgear R7800

dissent1:
KERNEL BUG: BAD_PAGE_STATE in process appears here and there during intensive memory usage.

https://bugs.lede-project.org/index.php?do=details&task_id=175

Reducing available memory by 2 MiBs workarounds the issue until further investigation.

Without these changes that bug appears rock steady at the same stage of boot process every boot.
Reducing total memory by 1 Mib moves the appearance of the bug onto later stage steadily again.
Reducing by 2 MiBs seems to solve the bug completely, at least i haven't met it for 3 days.
Also tested by crossflashing builds with different parameters to check if it's just a coincidence, but the above commit works in 100% cases, while other versions keep producing the above bug. I have even asked a panda to flash the device.

I have no idea what's in the end of the memory region or it just shifts memory usage somewhere in the middle.

### DIFF
--- a/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8065-r7800.dts
+++ b/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8065-r7800.dts
@@ -7,8 +7,12 @@
 	compatible = "netgear,r7800", "qcom,ipq8065";
 
 	memory@0 {
-		reg = <0x42000000 0x1e000000>;
+		reg = <0x42000000 0x1de00000>;
 		device_type = "memory";
+		/*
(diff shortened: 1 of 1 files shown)


dissent1:
Yes, it's definitively fixes the issue.
Might be defining a reserved memory region with reusable property is a better way. Testing, seems to work as well.

Any ideas what could be in there?

dissent1:
Added a second way of workarounding the issue, not sure which one is more correct though. 
If second one looks better it can just be fixed up into 1st one.

blogic:
merged, Thanks !

PR #454, opened 2016-10-28
lldpd: fix reload

commodo:
I thought I upstreame these a while back.
I only found the first one:
https://patchwork.ozlabs.org/patch/563075/

In the meantime, we've been using them internally [since the initial upstream].

These patches address some issues that I did not consider when I did the initial reload logic.

### DIFF
--- a/package/network/services/lldpd/Makefile
+++ b/package/network/services/lldpd/Makefile
@@ -94,6 +94,7 @@ CONFIGURE_ARGS += \
 	--with-embedded-libevent=no \
 	--disable-hardening \
 	--without-xml \
+	--sysconfdir=/tmp \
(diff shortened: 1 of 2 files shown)


jow-:
Minor stylistic nitpick, afterwards Acked-by from me.

Please use the `<<-EOF` syntax to allow indenting the here-doc termination tag as well.

Example:

    foo() {
        cat <<-EOF
            foo
            bar
            baz
        EOF
    }

commodo:
re-spinned ; if you want i can test these changes ; but only on Monday ; 
not sure if i can make it in the weekend

jow-:
Merged into my staging tree: http://git.lede-project.org/7353741, http://git.lede-project.org/0c0e612

commodo:
i can confirm now that the style change works fine in our internal setup; [ran a few tests with the change ]
thanks

PR #455, opened 2016-10-28
hostapd support for VLANs through a file in addition to Radius.

pekon:
Signed-off-by: Petr Konecny pekon@google.com


jow-:
Pulled into my staging tree: http://git.lede-project.org/a813507

PR #457, opened 2016-10-29
ramips: Archer C50: dts fix

heinzek:
- setting read-only flag to important partitions
- enabling PA to improve 2.4 GHz signal strength
- add missing leds
- rename colour led
- add mac adress to 5GHz wlan interface
- included <dt-bindings/input/input.h> and <dt-bindings/gpio/gpio.h>

Signed-off-by: Henryk Heisig hyniu@o2.pl

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -85,11 +85,11 @@ c20i)
 	ucidef_set_led_wlan "wlan" "wlan" "$board:blue:wlan" "phy0radio"
 	;;
 c50)
-	ucidef_set_led_default "power" "power" "$board:blue:power" "0"
-	ucidef_set_led_netdev "lan" "lan" "$board:blue:lan" "eth0.2"
-	set_usb_led "$board:blue:usb"
-	ucidef_set_led_wlan "wlan2g" "wlan2g" "$board:blue:wlan2g" "phy1radio"
-	ucidef_set_led_wlan "wlan5g" "wlan5g" "$board:blue:wlan5g" "phy0radio"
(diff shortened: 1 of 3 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -85,11 +85,12 @@ c20i)
 	ucidef_set_led_wlan "wlan" "wlan" "$board:blue:wlan" "phy0radio"
 	;;
 c50)
-	ucidef_set_led_default "power" "power" "$board:blue:power" "0"
-	ucidef_set_led_netdev "lan" "lan" "$board:blue:lan" "eth0.2"
-	set_usb_led "$board:blue:usb"
-	ucidef_set_led_wlan "wlan2g" "wlan2g" "$board:blue:wlan2g" "phy1radio"
-	ucidef_set_led_wlan "wlan5g" "wlan5g" "$board:blue:wlan5g" "phy0radio"
+	ucidef_set_led_default "power" "power" "$board:green:power" "0"
```
mkresin: Why do you switch off the power led after boot?

blogic: its the same as before.

mkresin: Which doesn't mean it was correct before...

According to the change done by @heinzek a few seconds ago the power led is ACTIVE_LOW instead of the former ACTIVE_HIGH.

@heinzek please add a few words about the ACTIVE_HIGH => ACTIVE_LOW change to the commit message to make clear this change was intentional.

heinzek: Power led fixed.

blogic:
merged, Thanks !

PR #459, opened 2016-10-30
ar71xx: Add net config for MR12 & MR16

riptidewave93:
Both the MR12 and MR16 are single gigabit ethernet devices, similar to the MR18. This change gives them the correct network config on a fresh install.

Signed-off-by: Chris Blake chrisrblake93@gmail.com


jow-:
Merged into my staging tree: http://git.lede-project.org/fc0cbe5

PR #462, opened 2016-10-31
build: drop `trapret` function from non-Linux HOST_TAR variant

commodo:
Looks like this was meant to workaround some limitations with
non-GNU tar variants (like BSD-tar which are present on Mac os BSD hosts).

Though, I cannot find any use of that `+s` option that's mentioned
in the comment.

Last hash of this I found was 24faf55360271cd0bfc4751753384f9210d52f7f

In my case, it now this fails for `python-setuptools` on Mac OS X (the host-build with):
```
trapret 2 tar -C <home-dir>/work/sources-work/lede/build_dir/target-i386_pentium4_musl-1.1.15/python-setuptools-27.2.0 --strip-components=1 -xzf <home-dir>/work/sources-work/lede/dl/setuptools-27.2.0.tar.gz
bash: trapret: command not found
```

So, I was thinking maybe it's time to remove this workaround (9 years later).
I could also fix the `python-setuptools` host build. If that's more preferred.

[ Btw, I just recently transitioned to a Mac machine for dev-ing,
  so a lot of (this Mac) stuff I'm finding out is new to me too. ]

### DIFF
--- a/include/shell.sh
+++ b/include/shell.sh
@@ -14,21 +14,6 @@ isset() {
 	[ -n "$var" ]
 }
 
-trapret() {(
-	local retvals="$1"; shift
-	local cmd="$1"; shift
-	for retval in $(echo $retvals); do
-		local trap_$retval=1
-	done
-	"$cmd" "$@" || {
-		local retval="$?"
-		eval "trapped=\${trap_$retval}"
-		[ -n "$trapped" ] || {
-			return $retval
-		}
-	}
-)}
-
 md5s() {
 	cat "$@" | (
 		md5sum 2>/dev/null ||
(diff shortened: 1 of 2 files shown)


jow-:
Iirc we're building our own host tar anyway because we need some other GNU features for reproducible builds (like `--sort` or `--owner`) so it likely makes no sense to keep this compat code.

commodo:
ah ; thanks for the clarification on that one (the `+s` one)

jow-:
Had a quick chat with @nbd168 on this and we concluded that we should drop that `trapret()` and see if and what breaks, so LGTM.

jow-:
Pulled into my staging tree: https://git.lede-project.org/2e2af34

PR #463, opened 2016-10-31
opkg: build against libcurl if it's available

commodo:
If libcurl is available just use it instead of wget for downloading.

### DIFF
--- a/package/system/opkg/Makefile
+++ b/package/system/opkg/Makefile
@@ -46,6 +46,7 @@ define Package/opkg/Default
   TITLE:=opkg package manager
   DEPENDS:=+uclient-fetch
   URL:=http://wiki.openmoko.org/wiki/Opkg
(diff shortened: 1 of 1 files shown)


jow-:
NAK - this will cause the opkg packages in the repo to depend on libcurl for all users. The libcurl support needs to be an explicit build variant or optional menuconfig symbol to avoid increasing the opkg size by default.

commodo:
oh ; ok ; will re-spin

closing until then

diizzyy:
@commodo 
Why is this even an issue as "wget" is provided by busybox?

commodo:
wget does not have SSL support by default ; libcurl might be linked against SSL
in our case libcurl use SSL
so, the BUILD_VARIANT option also works for me

jow-:
Current LEDE head actually uses uclient-fetch which disguises itself as wget. If you install one of the ustream-ssl providers, it gains SSL (https://) support.

PR #464, opened 2016-11-01
ramips: Fix flash size for Newifi D1

hbl0307106015:
1. The flash size of Newifi-D1 is 32MB, I am sorry for my mistake.
2. Reboot will always hang on (mt7621 with 32MB spi-nor flash)  Newifi D1, this 0902 patch can fix the problem.

### DIFF
--- a/target/linux/brcm47xx/image/Makefile
+++ b/target/linux/brcm47xx/image/Makefile
@@ -138,6 +138,7 @@ define Device/Default
 	KERNEL := kernel-bin
 	IMAGE_NAME = $$(IMAGE_PREFIX)-$$(1).$$(2)
 	KERNEL_NAME = vmlinux.lzma
+	KERNEL_INITRAMFS_NAME = vmlinux-initramfs.lzma
 	FILESYSTEMS := $(FS_64K)
 	IMAGES := trx
 	IMAGE/trx := append-rootfs | trx-with-loader


toolchain/gcc/patches/5.4.0/931-fix-mips-use-FP-instruction-on-none-soft-float-targets.patch
```diff
@@ -0,0 +1,87 @@
+diff --git a/libffi/src/mips/o32.S b/libffi/src/mips/o32.S
```
hbl0307106015: For more details, please visit below link:

[libffi #264](https://github.com/libffi/libffi/pull/264/commits/b74761126a209e779d0d23a74fe15d09ea69d72a)

hbl0307106015:
below link is the detail for libffi change which is from libffi #264

https://github.com/libffi/libffi/pull/264/commits/b74761126a209e779d0d23a74fe15d09ea69d72a

hbl0307106015:
@blogic Hi John Crispin, excuses me,  is there any wrong with my commit ? Please let mt know if I did do well enough, thanks!

target/linux/ramips/dts/PBR-D1.dts
```diff
@@ -0,0 +1,159 @@
...
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		wps {
+			label = "reset";
+			gpios = <&gpio1 38 1>;
+			linux,code = <0x211>;
+		};
```
981213: Is this key a WPS or reset button?

hbl0307106015:
@981213 look, that's it? I 'git rebase' the repository and create another new PR.

mkresin:
Have you send the reboot hang fix upstream (see http://www.linux-mtd.infradead.org/source.html topic "Patch submission")? It should be fixed in the kernel and backported if accepted upstream.

Or is this a backport from a more recent kernel version? If so, please mention the kernel commit id in the commit message.

target/linux/ramips/patches-4.4/0902-fix-reboot-hang-with-32m-spi-flash.patch
```diff
@@ -0,0 +1,69 @@
...
+@@ -907,6 +911,21 @@ static const struct spi_device_id *jedec_probe(struct spi_device *spi)
+ 	return ERR_PTR(-ENODEV);
+ }
+ 
++static int m25p80_reboot(struct notifier_block *nb, unsigned long val,
++			       void *v)
++{
++	struct mtd_info *mtd;
++
++	mtd = container_of(nb, struct mtd_info, reboot_notifier);
```
blogic: this patch look technically correct, however could add a

if (this is a chip that needs the quirk) {

}

line so that the registers are only written if we really need to.

blogic:
merged the 1st patch.

hbl0307106015:
@blogic  thanks, 

@mkresin  The 2nd patch (reboot hang) is not yet sent upstream, but I will check again and do it later. 

thanks for your suggestion.

target/linux/ramips/patches-4.4/0902-fix-reboot-hang-with-32m-spi-flash.patch
hbl0307106015: OK, got it, I think I should considerate what kind of flash chip (i.e. more than 32MB) need the quirk.

mkresin:
I've the impression you are facing the issue that was discussed in https://patchwork.ozlabs.org/patch/541391/.

I'm the opinion that I saw either a LEDE patch or some kernel mtd code which already addresses the issue. But I couldn't find it so far...

mkresin:
FYI, your mtd patch fixes the reboot issue seen on Archer C2600 v1.1 [as well](https://github.com/lede-project/source/pull/494#issuecomment-259737057). The used flash chip is similar to the one of the D1: MX25L25635.

Means, this patch need to go to target/linux/generic.

hbl0307106015:
@mkresin  OK, got it !

hbl0307106015:
@mkresin @blogic  OK, I updated the patch file just now, please check it, thanks!  Note The dts file should replace the m25p's compatible with "mx25l25635e" .

mkresin:
I'm it doubt that this will be accepted upstream. You should really send this in a first step as a RFC upstream to get some feedback on what's the preferred way of applying the chip specific workaround.

It could be for example a new device tree parameter which sets a flag which will be evaluated later (similar to [fast-read](https://github.com/torvalds/linux/blob/27bcd37e0240bbe33f0efe244b5aad52104115b3/drivers/mtd/spi-nor/spi-nor.c#L1423)) in case this workaround needs only be applied for some socs/configs. Or maybe a [hardcoded flag for the mx25l25635e](https://github.com/torvalds/linux/blob/27bcd37e0240bbe33f0efe244b5aad52104115b3/drivers/mtd/spi-nor/spi-nor.c#L874).

Nevertheless you should implement something based on the know JEDEC id instead of introducing a new modalias. Not sure if the opcode is supported for all Macronix SPI flash chips. In that case you can implement something based on the [manufacture part of the JEDEC id](https://github.com/torvalds/linux/blob/27bcd37e0240bbe33f0efe244b5aad52104115b3/drivers/mtd/spi-nor/spi-nor.c#L1273).

To implement something chip specific based on the JEDEC id you need to compare the remaining [parts of the JEDEC id](https://github.com/torvalds/linux/blob/27bcd37e0240bbe33f0efe244b5aad52104115b3/drivers/mtd/spi-nor/spi-nor.c#L1034).

dmke:
FYI, the Digineo AC1200 Pro (mt7621 w/ 32M flash as well) has the same problem.

The patch I butchered together looks oddly familiar (https://gist.github.com/dmke/d389bc364b3f73f525076eaee0019dc1) :-)

I'll have a look into the RFC process and ask upstream for advice.

hbl0307106015:
@dmke  thanks,   
 Excuses me,  I have a question, I can't send patch or even email to linux-mtd@lists.infradead.org, strange!!!!

bobafetthotmail:
did you set the email program to send "plain text" emails? Many email programs use HTML in the email and that is rejected from the mailing list servers.

dmke:
FYI: http://lists.infradead.org/pipermail/linux-mtd/2016-November/070368.html

rotanid:
why was this closed? is it fixed, also for Digineo AC1200 Pro ? @dmke 
at least there was an answer on the linux-mtd list 6 days ago...

dmke:
There are patches on the way enabling stateless addressing mode, but they still need to be tested (I hope to find some time this week to do some testing):

> Subject: [[PATCH v4 0/8] mtd: spi-nor: parse SFDP tables to setup (Q)SPI memories](http://lists.infradead.org/pipermail/linux-mtd/2016-November/070414.html)
> From: Cyrille Pitchen
> Date: Mon Nov 21 06:15:39 PST 2016
>
> See thread here: http://lists.infradead.org/pipermail/linux-mtd/2016-November/thread.html#70414

Since this PR will become superfluous once the spi-nor patches are available upstream and backported to 4.4, holding it open doesn't make much sense, I think.

dmke:
This seems promising: 

> [PATCH v2 0/2] mtd: spi-nor: add a stateless method to support memory size above 128Mib
> http://lists.infradead.org/pipermail/linux-mtd/2016-December/070889.html

What is necessary to backport these patches to LEDE's linux-4.4 kernel, so that I can test it?

mkresin:
Not sure how up to date the guide is: https://wiki.openwrt.org/doc/devel/patches#adding_or_editing_kernel_patches

Basically you need to backport the patches by adding them to target/linux/generic/patches-4.4. This time generic is the right place for the patches (instead of target/linux/ramips/patches-4.4) since the issue is present in multiple targets (at least on ipq806x and ramips).

PR #466, opened 2016-11-02
ramips: Use MT7621 I2C driver for MT7628/MT7688

svenschwermer:
The i2c-ramips driver does not work for the MT7628/MT7688 SoCs, the
i2c-mt7621 driver does.

Signed-off-by: Sven Schwermer <sven.schwermer@arcor.de>


blogic:
merged, Thanks !

PR #468, opened 2016-11-02
package/kernel/modules: add scsi-fusion and update some crypto modules

commodo:
@uweber to approve

We've added these kernel modules to our tree a while.
The scsi-fusion kernel module to better support VMware.

And the crypto updates to help better support some Intel HW extensions for doing crypto with ipsec.

Not sure if the LEDE community would want them in the main tree.
Hence this PR.

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@riverbed.com>

### DIFF
--- a/package/kernel/linux/modules/block.mk
+++ b/package/kernel/linux/modules/block.mk
@@ -527,3 +527,26 @@ define KernelPackage/scsi-tape
 endef
 
 $(eval $(call KernelPackage,scsi-tape))
+
+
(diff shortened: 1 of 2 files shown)


commodo:
my bad ; i was cleaning up some branches

nbd168:
I would prefer simply enabling these in the kernel config for the relevant x86 subtargets

commodo:
ok ; will think about it internally 
thanks :)

PR #469, opened 2016-11-02
base-files: add readout function for uboot-env device serial number

feckert:
Some devices have a serial number in their uboot-env section. This function
can be used to read out the serial number during board detection code.

### DIFF
--- a/package/base-files/files/lib/functions/system.sh
+++ b/package/base-files/files/lib/functions/system.sh
@@ -8,6 +8,23 @@ find_mtd_chardev() {
 	echo "${INDEX:+$PREFIX$INDEX}"
 }
 
+mtd_get_serial_ascii()
+{
+	local mtdname="$1"
+	local key="$2"
+	local part
+	local serial
+
+	part=$(find_mtd_part "$mtdname")
+	if [ -z "$part" ]; then
+		echo "mtd_get_serial_ascii: partition $mtdname not found!" >&2
+		return
+	fi
+
+	serial=$(strings "$part" | sed -n 's/^'"$key"'=//p')
+	echo "$serial"
+}
+
 mtd_get_mac_ascii()
 {
 	local mtdname="$1"


mkresin:
Would you please give an real world example where this function might/will be used.

I'm not a big fan of adding a orphaned/unuse function to system.sh.

feckert:
We can use this function for example in lantiq_board_detect() to read out the serial number of the board from uboot-env. After readout the value could be dumped to /tmp/sysinfo/serial

> lantiq_board_detect()
> ...
> echo $name > /tmp/sysinfo/board
> 
> case $name
>   XXX)
>         serial=$(mtd_get_serial_ascii uboot-env serial)
>         [-z "$serial" ] && serial="unknown"
>         ;;
> esac
> 
> if [ "$serial" != "unknown" ]; then
>     echo "$serial" > /tmp/sysinfo/serial
> fi
> ....

Every board which has an serial number in their uboot-env could be added with the board name to the switch case statement

mkresin:
Yeah dumping the serial is the obvious. But I still fail to see the real world benefit of doing this, which is most likely because I'm to focused on my own use cases. Lets wait to see what the others think about it.

jow-:
If I see it correctly the function is not really serial specific, right? Since a potential caller has to pass in the key name it is interested in the function looks more like a generic value getter to me.

I'm not opposed per-se as having access to a serial number makes things like generating initial wifi passwords, seeding the PRNG or creating unique SSIDs possible without having to rely on remotely sniffable mac addresses.

@feckert - do you plan to follow up on this with more changes actually using it (apart from populating /tmp/sysinfo) ?

blogic:
technically it looks ok, i simply dont see the purpose of this. having a unique identifier on the cpe is bad enough, providing an API to get hold of it is even worse.

feckert:
@mkresin @jow  The serial number should be read out at first from luci and luci2 to show it on the frontpage of the webgui.

Another scenario could be:
If you use a Auto Configuration Server (ACS TR069) for your system in the field the serial number is also a good choice to make any configuration decisions with this system.

@jow yes the default password could also be generated from the serial which is much better than use the MAC.

I have no stock firmware for an router so the serial could also be written into the system flash in a own partition. My PR will only cover the read out of serial from the uboot-env. Other system will save the serial some where else.

feckert:
@blogic Yes you are right the sysem could be tracked but this is the same as for the MAC. Maybe we could random the serial if some one does not want to show it.

blogic:
closing this. there were doubts ad no one picked it up int he last 3 weeks. i see the chances of it being merged as slim to none.

PR #472, opened 2016-11-02
mdadm: move to Disc submenu

bobafetthotmail:
in the Disc submenu in the Openwrt packages feed there is lvm2, so I thought mdadm could go in Disc too.


blogic:
the Author field is set to "Author: Alberto Bursi <starshipeleven@outlook.com>"

please make it the same as the SoB line.

PR #474, opened 2016-11-02
uqmi: move to GSM/3G/4G submenu

bobafetthotmail:
New submenu for GSM/3G/4G network tools, comgt will be moved there too.


blogic:
NAK, if we want to move it then to the WWAN submenu please

PR #475, opened 2016-11-02
comgt: moved to GSM/3G/4G and fixed homepage link

bobafetthotmail:
Moved to the same category and submenu as uqmi, another tool to operate cell modems.

The link was dead so I changed it to point at comgt's Ubuntu manpage, as it is the most reputable manpage site for comgt I found by googling.


blogic:
NAK, if we want to move it then to the WWAN submenu please

PR #476, opened 2016-11-03
lantiq: cleanup dts files

mkresin:
- remove not existing properties
- remove properties having the same values as the included dtsi
- remove nodes which are disabled in the included dtsi and not enabled
  in dts
- replace the deprecated pinctrl-* compatible strings
- use the same labels for nodes as the included dtsi
- move common used vr9 pci properties to vr9.dtsi
- fix spaces vs. tabs and remove superfluous linebreaks


blogic:
works for me, feel free to push

mkresin:
@blogic do you see something obvious wrong? compile tested for all lantiq boards. runtime tested on BTHOMEHUB5A, ARV7506PW11 and VGV7510KW22.

PR #478, opened 2016-11-03
comgt: move to WWAN submenu, fixed link

bobafetthotmail:
moving comgt and its modules to WWAN submenu to join uqmi as both are tools for WWAN modems.

I replaced the link with comgt's ubuntu manpage because the old link isn't working anymore.


blogic:
merged, Thanks !

PR #479, opened 2016-11-03
yamonenv: move to Boot Loaders submenu

bobafetthotmail:
Boot Loaders submenu is the place where all other boot loaders are.


blogic:
merged, Thanks !

PR #480, opened 2016-11-03
mdadm: move to Disc submenu

bobafetthotmail:
As in Disc there is also lvm2 and other hard drive tools in the Openwrt package feeds.


blogic:
merged, Thanks !

PR #481, opened 2016-11-03
[patch - bug fix]layerscape: add ls-ff-pad-to for pad 0xff to rootfs

fsl-jyt:
The public interface pad-to will fill 0. It let layerscape SOC flash device can
not be timely and correct access by mount_root, and then lead to the failure of
writeabled "rootfs_data" overlay fs mechanism.

So create a special ls-ff-pad-to to replace pad-to for fill 0xff to rootfs in
layerscape target.

Signed-off-by: Yutang Jiang <yutang.jiang@nxp.com>

### DIFF
--- a/target/linux/layerscape/image/Makefile
+++ b/target/linux/layerscape/image/Makefile
@@ -25,6 +25,12 @@ define Build/append-ls-dtb
 	dd if=$(DTS_DIR)/$(1).dtb >> $@
 endef
 
+define Build/ls-ff-pad-to
+	dd if=/dev/zero bs=$(1) count=1 | sed 's/\x00/\xff/g' > $@.new
+	dd if=$@ of=$@.new conv=notrunc
+	mv $@.new $@
+endef
+
 define Build/append-ls-rootfs-ext4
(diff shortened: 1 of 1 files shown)


blogic:
merged, thanks !

nbd168:
Padding to 64M seems completely unnecessary to me. Why not just use pad-rootfs, which will add end-of-filesystem markers and thus produce a smaller image.

blogic:
ok, dropped from my tree again

fsl-jyt:
Thank you @nbd168 for your advice.
I will study how to use the pad-rootfs and try to optimal the fix solution.
Our team is discussing to build some multimedia demo project on ls1043ardb, which requires  audio/video libraries and other software packages, I worry that the reserved space is not enough, so left all of the rest space in current flash bank to the rootfs.

fsl-jyt:
Hi @blogic ,
I have learned to use pad-rootfs and reduced the pad size, also solved another potential conflict problem, still continue to optimize the code. I need more time to do some test work to ensure the quality of the code. After verification, I will submit again the new patch.

fsl-jyt:
Hi @blogic and @nbd168 ,
I have submit a new PR 501: https://github.com/lede-project/source/pull/501
Please close this 481, thank you.

PR #484, opened 2016-11-03
ipq806x: fix pcie reset gpios

dissent1:
Fix perst-gpios property in accordance to the driver, so it stops spamming that it can't parse it.


blogic:
merged, Thanks !

PR #486, opened 2016-11-03
dnsmasq: Support add-mac option

dedeckeh:
Adds the mac address of the DNS requestor to DNS queries which
are forwarded upstream and can be used to do filtering by the
upstream servers. This only works if the requestor is on the
same subnet as the dnsmasq server

The addmac parameter can hold the following values:
	0 : mac address is not added
	1 : mac address is added in binary format
	base64 : mac address is added base64 encoded
	text: : mac address is added in human readable format
		as hex and colons


blogic:
merged, Thanks !

PR #490, opened 2016-11-05
fix: audio not working due to unset SND_PCM_TIMER

f3flight:
CONFIG_SND_PCM_TIMER is required for ALSA to work
(at least for some audio devices)


mkresin:
Thanks for picking this up by yourself. I was in a meeting and haven't had the time to come up with an own patch.

Some remarks:

- add you signed of by line
- prefix you commit title and describe what you have done (something like "kernel: add SND_PCM_TIMER to kmod-sound-core"
- reference "FS#256" in the commit message
- please improve the commit message. describe why it is required. maybe provide an example of the error message that get fixed
- add a note that the kernel size does not increase only the size of the kernel module

PR #491, opened 2016-11-05
fix: audio not working due to unset SND_PCM_TIMER

f3flight:
CONFIG_SND_PCM_TIMER is required for ALSA to work
(at least for some audio devices)

Signed-off-by: f3flight <f3flight@gmail.com>


mkresin:
Thanks for picking this up by yourself. I was in a meeting and haven't had the time to come up with an own patch.

Some remarks:

- please use tabs instead of spaces (keep the used indentation style)
- use your full name in signed of by line (see https://www.lede-project.org/development.html#submitting_patches)
- prefix you commit title and describe what you have done (something like "kernel: add SND_PCM_TIMER to kmod-sound-core"
- reference "FS#256" in the commit message
- please improve the commit message. describe why it is required. maybe provide an example of the error message that get fixed
- add a note that the kernel size does not increase only the size of the kernel module

f3flight:
I've updated the commit (sorry for doubling the pull request, not too familiar with this, thought I have to redo it when I update the commit)

f3flight:
ok now I believe I've addressed all the remarks

mkresin:
merged, thanks!

> sorry for doubling the pull request, not too familiar with this, thought I have to redo it when I update the commit

Ah, no worries. I will guide you this time and the next time you provide a ready to merge patch. Deal? ;-)

PR #495, opened 2016-11-07
ar71xx: wpa8630: Fix board name and set properly kerner size

heinzek:
Tested and working.


blogic:
merged, Thanks !

nbd168:
The kernel length change is wrong, because it still uses a bogus hardcoded partition size.
The code needs to be changed to remove the kernel/rootfs partitions and use the dynamic parsing of the firmware partition

PR #497, opened 2016-11-08
[RFC] add powerpc64 support

florolf:
Add support for building a pure 64 bit powerpc kernel and userland.

I'm planning to add support for the T4 family of Freescale CPUs/the T4240RDB and having ppc64-support is a prerequisite for that.

This is based on glibc because musl only has working powerpc64le-support right now, which doesn't seem to have caught on in the embedded word yet; at least for the T4240RDB, it is impossible to support without a lot of low-level restructuring.

I've looked at the way other architectures were added and tried to cover the important bits (I can build a firmware for the T4240RDB using this), but I probably have missed something, thus flagging this as RFC.


blogic:
you should send this together with a target that uses the support.

florolf:
Okay, will do.

PR #498, opened 2016-11-08
gre,vti: add empty install rules

commodo:
@dedeckeh for gre
@nbd168 for vti  ; the code was submitted from the email list, so I could not track the submitter's github account

I mostly tested this with `gre` and applied the same pattern to `vti`.

### DIFF
--- a/package/network/config/gre/Makefile
+++ b/package/network/config/gre/Makefile
@@ -60,6 +60,14 @@ define Package/gre/install
 	$(INSTALL_BIN) ./files/gre.sh $(1)/lib/netifd/proto/gre.sh
 endef
 
+define Package/grev4/install
+	:
+endef
+
+define Package/grev6/install
+	:
+endef
+
 $(eval $(call BuildPackage,gre))
 $(eval $(call BuildPackage,grev4))
 $(eval $(call BuildPackage,grev6))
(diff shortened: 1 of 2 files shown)


dedeckeh:
@commodo Idea behind the @(PACKAGE_GRE) dependency was to show only the gre sub-items when gre package is selected by the user. Similar approach is used in the PPP package with sub-items ppp-mod-pppoe, ppp-mod-pppoa, etc ... I felt this would be more intuitive and a more condensed overview if GRE is not selected but this is a matter of personal taste I guess

commodo:
@dedeckeh I assumed it was something like that.
hmm ; so i guess i can leave it like that, and do some more config symbol dependency magic

to give a bit of background: when I try to pull in just `+grev4`, it does not get enabled until I also select `+gre` ; but I can select `+gre` from a "platform" meta-package, or target, and use `+PACKAGE_gre:grev4` in the package dependencies

commodo:
re-spinned without the dependency re-ordering commits ; just the empty install rules

commodo:
@jow- @blogic @dedeckeh 
any input on this ?

jow-:
@commodo - I'm okay with having empty install sections but can you re-spin the change without adding the new `PackageEmptyInstall` helper? Just put the empty install sections into the `gre` and `vti` Makefiles.

commodo:
sure;
fine from me ; will re-spin

jow-:
Merged into my tree - thanks!

PR #499, opened 2016-11-08
lantiq: Fix BAR11MASK when specifying mem= parameter in cmdline neede…

dpeddi:
lantiq: Fix BAR11MASK when specifying mem= parameter in cmdline needed by vr9 platform to support tapi/vmmc/vpe

As specified by [PATCH 0/8] lantiq: vr9 fxs support: Enable FXS support for VR9 based routers, f the VR9 based router provides FXS ports and they shoud enabled then the following must added to the kernel command line:
mem=[TOTALMEMSIZE-2M] vpe1_load_addr=ADDRESS vpe1_mem=2M

By adding mem= parameter a pci device stop working correctly.
pci-lantiq.c module use  get_num_physpages() to compute dinamically the memory amount of the board.
the mem= make the module to compute in the wrong way the BAR11MASK, so in this situation the mask is misaligned with the dma area that the hardware expects.

This patch is a port of what legacy ifxmips_pci.c does.

### DIFF
--- a/target/linux/lantiq/patches-4.4/0154-lantiq-pci-bar11mask-fix.patch
+++ b/target/linux/lantiq/patches-4.4/0154-lantiq-pci-bar11mask-fix.patch
@@ -1,0 +1,22 @@
+--- a/arch/mips/pci/pci-lantiq.c
++++ b/arch/mips/pci/pci-lantiq.c
+@@ -62,6 +62,8 @@
+ #define ltq_pci_cfg_w32(x, y)	ltq_w32((x), ltq_pci_mapped_cfg + (y))
+ #define ltq_pci_cfg_r32(x)	ltq_r32(ltq_pci_mapped_cfg + (x))
+ 
++extern u32 max_low_pfn;
++
+ __iomem void *ltq_pci_mapped_cfg;
+ static __iomem void *ltq_pci_membase;
+ 
+@@ -87,8 +89,8 @@
+ 	u32 mem, bar11mask;
+ 
+ 	/* BAR11MASK value depends on available memory on system. */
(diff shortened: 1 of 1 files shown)


kochstefan:
This patch works with the mem=62M setting and enabled lantiq FXS support.

I can confirm that this patch solves the issue described within https://bugzilla.kernel.org/show_bug.cgi?id=110091

mkresin:
I've reformatted the commit message, fixed a few typos and pulled the PR into my staging tree: https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary.

Please confirm that you are fine with the changed commit message!

kochstefan:
Please append "maxvpes=1 maxtcs=1" to the command line example.

mkresin:
It wasn't meant that I fix your commit message. The "maxvpes=1 maxtcs=1" stuff wasn't present before and I will not add it. Please update this PR if you want to improve the commit message.

I just want to have your confirmation that it's fine to push the patch with the commit messages as it is in my staging repo, to make sure that I haven't _removed_ something relevant.

dpeddi:
For me is ok. Next step is enable vmmc on our board by modifying dtsi.

After that  nobody will care about the commit message.

Do you think John will push this patch to linux-next?

Regards

kochstefan:
To commit it on linux-next I would recommend to remove the part from "As specified" up to "2M". Because it is not relevant there, and the commit hash is not resolveable there.

mkresin:
> Do you think John will push this patch to linux-next?

Maybe. I guess it depends on how busy he is.  It would be better if you can pick up that task.

PR #500, opened 2016-11-09
Add the virtual.mk file and corresponding code to support Hyper-V target

tedaz:
To support the Hyper-V target, re-add the virtual.mk file and added code to this file to support Hyper-V.
Also, corresponding code added to other files to support Hyper-V.

### DIFF
--- a/package/kernel/linux/modules/virtual.mk
+++ b/package/kernel/linux/modules/virtual.mk
@@ -1,0 +1,257 @@
+#
+# Copyright (C) 2010 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+VIRTUAL_MENU:=Virtualization Support
+
+define KernelPackage/virtio-balloon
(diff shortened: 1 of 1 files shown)


bobafetthotmail:
Please add "Signed-off-by: yourName yourSurname <yourEmail>" in the commit message (not just here on github pull request message).

nbd168:
There's a lot of wrong stuff in that patch. A lot of the virtio stuff that you're packaging is already included in the x86 64-bit subtarget. Also, the kvm_guest subtarget (mentioned in some dependencies) does not exist anymore.
Do you still need 32-bit support? If not, I would recommend sending a patch that simply enables the hyper-v driversi n the kernel config

tedaz:
@bobafetthotmail @nbd168 Thanks for your replies. I re-created a new pull request, https://github.com/lede-project/source/pull/503, this patch only add/enable the Hyper-V support in the kernel.
Based on some previous discussion threads on the OpenWRT forum, it seems that 64-bit, SMP and big memory (>~800MB) wouldn't give to much help to x86 router. Is there any update on this point?

PR #501, opened 2016-11-09
[patch] layerscape: optimize IMAGE/xx.bin variables composition

fsl-jyt:
[PATCH 1/2]layerscape: ls1043ardb: add pad-rootfs to reduce the size of firmware.bin
[PATCH 2/2]layerscape: ls1012ardb: only reserve ext4 fs as default firmware.bin


blogic:
merged, Thanks !

PR #503, opened 2016-11-10
Add/enable Hyper-V support in the kernel

tedaz:
Add/enable Hyper-V support in the kernel. 
Signed-off-by: Tedaz <tedaz99999@hotmail.com>


nbd168:
Please don't enable it for the full x86 target, enable it for the 64-bit subtarget only. You can use: `make kernel_menuconfig CONFIG_TARGET=subtarget`
Also, the Signed-off-by needs to be in the commit message, not in the pull request description.
The commit title should be:
`x86/64: enable Hyper-V support in the kernel config`

tedaz:
@nbd168 Here is a new pull request #508 created, please review it.
I'm created by "make kernel_menuconfig CONFIG_TARGET=x86_64", but the changes still saved to the file "target/linux/x86/config-4.4", is this correct?

PR #507, opened 2016-11-10
Fixed multiple issues in F5D8235_V1.dts v2

NotTheEvilOne:
Resolves issues regarding the F5D8235_V1.dts as discussed in [1].

[1] [FS#244 - Belkin F5D8235-4 v1 does not boot with LEDE](https://bugs.lede-project.org/index.php?do=details&task_id=244)

### DIFF
--- a/target/linux/ramips/dts/F5D8235_V1.dts
+++ b/target/linux/ramips/dts/F5D8235_V1.dts
@@ -10,33 +10,31 @@
 
 	cfi@1f000000 {
 		compatible = "cfi-flash";
-		reg = <0x1f000000 0x800000>;
+		reg = <0xbc400000 0x800000>;
 		bank-width = <2>;
 		device-width = <2>;
 		#address-cells = <1>;
 		#size-cells = <1>;
 
-		partition@0 {
+		uboot: partition@0 {
 			label = "uboot";
-			reg = <0x0 0x30000>;
-			read-only;
-		};
-
-		partition@30000 {
-			label = "uboot-env";
-			reg = <0x30000 0x10000>;
-			read-only;
-		};
-
-		factory: partition@40000 {
-			label = "factory";
-			reg = <0x40000 0x10000>;
+			reg = <0x0 0x50000>;
 			read-only;
 		};
 
 		partition@50000 {
-			label = "linux";
-			reg = <0x50000 0x3b0000>;
+			label = "firmware";
+			reg = <0x50000 0x790000>;
+		};
+
+		partition@7e0000 {
+			label = "nvram";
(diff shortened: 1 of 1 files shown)


mkresin:
Hey Tobias,

thanks a lot for this PR. First of all I can confirm that your change of the flash reg param is correct. I found an [old thread](https://lists.openwrt.org/pipermail/openwrt-devel/2011-December/013271.html) which confirms your change. KSEG1ADDR(0x1c400000 ) = 0x1c400000 | 0xa0000000 = 0xbc400000. And it seams to me some of the other rt2880 based router as well as the F5D8235_V2 are having the same issue.

But I'm not sure about your changes to the flash layout. The ```ralink,mtd-eeprom = <&uboot 0x40000>;``` leads me to the conclusion that at 0x40000 is indeed the beginning of the factory partition.

The nvram partition shouldn't be required for LEDE. This partition seams to be used by the Belkin firmware to store the settings: https://wikidevi.com/wiki/Belkin_F5D8235-4_v1/nvram.

Can you give me a pointer why it is necessary to enable the i2c bus?

I've merged a modified version of your PR in my staging tree: https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary.

Would you please compile and test an image from my staging tree. I guess the best would be to test with an ramdisk image.

Would you please paste a (serial) bootlog. Does the wireless work? Would you please paste the output of ```ifconfig -a```. Does one of the mac addresses shown matches the ones on the label at the bottom of the routers case?

NotTheEvilOne:
So I decided for exactly the wrong partition layout (after looking at F5D8235_V2.dts). I haven't tested your staging tree yet but the mirrored one resulted in exactly the same issue as referenced in the Flyspray bug report linked above.

NotTheEvilOne:
Ok, here's another commit removing i2c again. As you may have already assumed it was unnecessary to get the "rtl8366s" switch in a working condition again. The change of "ralink,group" was the key here to resolve a GPIO pin mux error message.

Unfortunately I can't provide any boot logs or even more advanced information as it is still not booting with kernel 4.4 at all.

mkresin:
The F5D8235 V1 has multiple issues. Most of them are fixed by this patch.

But the upgrade to kernel 4.4 has introduced a (new regression)[https://bugs.lede-project.org/index.php?do=details&task_id=244]. The bugs fixed by this patch are ones which would show up later in the boot process.

The OP was able to boot the board with kernel 4.3. It doesn't make much sense to me to merge this PR as long as the new regression isn't fixed as well. I'll add this patch back to my staging tree if the board boots successfully.

PR #511, opened 2016-11-12
x86: add grub root config for xen

wklaebe:
Change grub's root device to xen/xvda,msdos1 for the x86-xen_domu
target so that it will boot without manual intervention.

Signed-off-by: Wilfried Klaebe <w+lede-project@chaos.in-kiel.de>


mkresin:
merged.

Accidentally the version from the flyspray ticket was committed instead of your PR.

PR #514, opened 2016-11-13
 arm64: boot-wrapper: Add mirror

diizzyy:
Adds Google's mirror as primary source and kernel.org as fallback.
Same as commit 0d4f02dfd650612aac6c11860139be03a0f54bee


blogic:
why should we not load this from kernel.org ?

diizzyy:
Several reasons,
* According to @jow- some buildbots have old SSL libs that doesn't interact well with TLS 1.1+, kernel.org enforces 1.2 while Google allows 1.0 and higher.
* Consistency
* Users have complained that kernel.org is slow/slower (main reason for 0d4f02dfd650612aac6c11860139be03a0f54bee )

blogic:
merged, Thanks !

PR #515, opened 2016-11-13
package/devel/trace-cmd: Add mirror

diizzyy:
Adds Google's mirror as primary source and kernel.org as fallback.
Same as commit 0d4f02dfd650612aac6c11860139be03a0f54bee


blogic:
merged, Thanks !

PR #516, opened 2016-11-13
package/firmware/fman-ucode: Use HTTPS

diizzyy:
Change URL to use HTTPS instead of HTTP


blogic:
merged, Thanks !

PR #517, opened 2016-11-14
odhcpd: Add reload support

dedeckeh:
odhcpd daemon has hitless config reload support by means of the
sighup signal; add reload_service function which uses sighup
signal to reload the config


nbd168:
Applied, thanks

PR #518, opened 2016-11-14
build: conditionalize all source dependencies upon request

nwf:
A package my assert PKG_DEPEND_CHECK_PACKAGE to cause all source dependencies
apply only when individual packages have been selected.  Most packages would
experience no change, as typically a source package has only one package; thus

$(curdir)/feeds/packages/foo/compile += ...

would become the equivalent

$(curdir)/feeds/packages/foo/compile += $(if $(CONFIG_PACKAGE_foo), ...)

However, when one source package has multiple build packages in it (e.g.
gnunet), it may not be the case that selecting any one package should pull in
in all dependencies of any package that shares the same source package.
Asserting PKG_DEPEND_CHECK_PACKAGE results in dependencies being broken out
explicitly; that is, rather than

$(curdir)/feeds/routing/bar/compile += bar-deps... baz-deps...

the dependency structure would be

$(curdir)/feeds/routing/bar/compile += $(if $(CONFIG_PACKAGE_bar), bar-deps...)
$(curdir)/feeds/routing/bar/compile += $(if $(CONFIG_PACKAGE_baz), baz-deps...)

which is a more accurate assment of the situation and can dramatically
shave the dependency list for complex source packages.

Note that this diff is best viewed ignoring whitespace due to an
indentation change of otherwise unmodified code.

Signed-Off-By: Nathaniel Wesley Filardo <nwfilardo@gmail.com>

This is take two of #420.


nbd168:
Similar issue as with the previous patch. Build dependencies should default to being unconditonal and should be marked explicitly only for those packages where PKG_CONFIG_DEPENDS, CONFIGURE_ARGS, etc. have been adjusted

PR #521, opened 2016-11-15
[patch]layerscape: uboot: using perl script:byte_swap.pl to replace tcl scri…

fsl-jyt:
The uboot-layerscape component uses a TCL script to process the final binary but tclsh is no build requirement of the LEDE buildroot.

Jo-Philipp Wich give me advice: Rewriting that byte_swap.tcl script in Perl.

Now I have studied TCL and Perl then rewriting it.


blogic:
merged, Thanks !

PR #522, opened 2016-11-15
[RFC] include/netfilter.mk: add optional 'netfilter-user.mk' inclusion

commodo:
I don't have a better idea at the moment with this.
Bottom line is, we'd want to extend 'include/netfilter.mk' without
patching it (internally).

And wrapping 'netfilter.mk' with a 'my-netfilter.mk' would work,
but it creates a risk of running out of sync between kernel modules
enabled, and what 'iptables' has configured.

The shortest/simplest fix is to add a simple '-include/netfilter-user.mk'
in 'netfilter.mk'.
And 'netfilter-user.mk' would live in an internal tree.
This makes it a bit easier when upgrading the OpenWrt/LEDE base.

Replacing with kernel config symbols is also a bit tricky, since
a lot of stuff in 'netfilter.mk' is interdependent.

I'm open to other ideas.

### DIFF
--- a/include/netfilter.mk
+++ b/include/netfilter.mk
@@ -373,4 +373,6 @@ IPT_BUILTIN += $(EBTABLES_IP4-y)
 IPT_BUILTIN += $(EBTABLES_IP6-y)
 IPT_BUILTIN += $(EBTABLES_WATCHERS-y)
 
+-include $(INCLUDE_DIR)/netfilter-user.mk
+
 endif # __inc_netfilter


commodo:
@nbd168 @blogic @jow-

nbd168:
Please provide some examples of how (and why) you want to patch this

commodo:
@nbd168 
Hey,
[ i knew before writing this, that i was gonna be verbose ; sorry about that ]

Our use case was:
* in `target/linux/generic/files` we have some netfilter kernel modules/extensions
* in `target/linux/generic/patche-x.x` we have some patches that enable building of that netfilter kernel code
* then we would patch (internally) `include/netfilter.mk` to provide/install our netfilter kernel modules (in the image) ; this would be convenient (on the short term), because netfilter.mk is included by `package/kernel/linux` & `package/network/utils/iptables`

Regarding `package/kernel/linux` (after a previous PR of mine here (this one: https://github.com/lede-project/source/pull/468 )), I've had an `aha` moment and decided to create a similar `package/kernel/linux` package, which just contains our own kernel modules.

That was an example of our use case.
The `why?` we (more me) want this (or some sort of common solution) is simply to try to contain/simplify future syncs (or merge hells) with LEDE/OpenWrt base.

Regarding netfilter kernel modules, it's a bit tricky because there's this `include/netfilter.mk`.
We could:
1. create an entire copy of it, and just put the modules we need there ; internally, this is not preferred, because it's a too-big branch from base
2. create `include/netfilter-wrapper.mk` which includes `include/netfilter.mk` ; then we'd have to be careful with syncs with base, so that nobody overrides this include in `iptables`
3. this solution

Atm, there's no escaping patching iptables ; and creating an iptables internal copy is also not desired ; so, for now (or until a better solution is found) it's fine.

I'm not strongly opinionated by any idea/approach on this ; because I don't know of any potential fallout/complications in the future.
And I'm fine if this proposed idea (in this PR) is not accepted.

Maybe after a while I'll get better idea and be able to come back and propose a LEDE/OpenWrt build-overlay mechanism that allows people to customize the build, while avoiding merge hells (with base) as much as possible, and would be fine for a lot of people (everybody is impossible kinda).

Thanks
Alex

nbd168:
What kind of netfilter patches are you using?
Also, is it really necessary to patch this into the kernel, or could you rework it to build it out of tree?

Either way, I'm not very fond of this PR, because I don't like adding more hacks to the build system that are only useful for weird out-of-tree patches.

commodo:
ok ; closing this

so, there could be an idea to rework this out-of-tree ; 
atm, i haven't thought of one ; but this is mostly due to the fact that we're upgrading from OpenWrt [from about a year ago] to LEDE ; 
and we've been porting stuff over to the LEDE tree ; for some stuff, I could easily figure an elegant non-invasive solution [to replace older patches], for others I couldn't [yet]

regarding your first question: we have some internal kernel modules that hook into netfilter and do various packet forwarding/routing/manipulation logic ; i did not write them ;
the way they were initially added into the OpenWrt tree, was not great ; so, i tried to think about untangling them somehow ; this is the best I could think of so far to do it with minimal effort for us ;
there is another solution, but I have to investigate it later [ after we've finalized the porting phase ]

thanks for your input :)

PR #523, opened 2016-11-15
Experiabox Vgv7519 voice support

dpeddi:
This series of patch enable tapi/vmmc on VGV7519.

- lantiq: disable VMMC_COEF for non FALCON device 
disable the menuconfig entry for VMMC_COEF since we use fixed coef (and is not actually used)

- lantiq: enable VMMC for VGV7519
add dts entry needed to use vmmc. We can add this by default since actually we fixed the pci/dma issue that messed up wifi on OpenwrtBB

- lantiq: add tapi/vmmc to VGV7519 defaults 
add tapi/vmmc driver to the firmware image by default

### DIFF
--- a/package/kernel/lantiq/ltq-vmmc/Config.in
+++ b/package/kernel/lantiq/ltq-vmmc/Config.in
@@ -24,7 +24,7 @@ choice
 endchoice
 
 choice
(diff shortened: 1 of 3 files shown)


blogic:
merged, Thanks !

PR #524, opened 2016-11-15
Vgv7510kw22 voice support

dpeddi:
This series of patch enable tapi/vmmc on Vgv7510kw22.

-lantiq: enable VMMC for Vgv7510kw22 
add dts entry needed to use vmmc. We can add this by default since actually we fixed the pci/dma issue that messed up wifi on OpenwrtBB

-lantiq: add tapi/vmmc to Vgv7510kw22 defaults
add tapi/vmmc driver to the firmware image by default

@kochstefan already confirmed to me that the configuration is identical to vgv7519


blogic:
merged, Thanks !

PR #527, opened 2016-11-17
kernel: remove pipetypes unused var warning

kdarbyshirebryant:
Update patch to remove pipetypes var declaration which was throwing
unused variable warning due to the original patch removing the only use.


blogic:
merged, Thanks !

PR #528, opened 2016-11-17
toolchain/kernel-headers: Fix toolchain kernel headers

KarlVogel:
The build system was picking up the host system's kernel configuration as a default to prepare the kernel headers. Additionally the build system was only applying generic files/patches, while some targets might have target specific patches/files which might also be relevant for user space headers.

This is an old patch that was send to the openwrt-devel mailing list in March, see https://lists.openwrt.org/pipermail/openwrt-devel/2016-March/040541.html -- but that was the timeframe when the fork happened, so the patch never made it in.

Had to rework the patch a bit for the recent  HOST_LOADLIBES addition to the KERNEL_MAKEOPTS, which uses STAGING_HOST_DIR, but in the toolchain steps, that variable points to another location and the actual one is kept in REAL_STAGING_HOST_DIR, therefor the KERNEL_MAKEOPTS addition in the Makefile. Another way to solve it would be to use:

```
STAGING_HOST_DIR:=$(REAL_STAGING_HOST_DIR)
```

instead, but that might have unwanted side effects, hence the choice to go with adding HOST_LOADLIBES again to the KERNEL_MAKEOPTS.

### DIFF
--- a/include/kernel-defaults.mk
+++ b/include/kernel-defaults.mk
@@ -79,6 +79,7 @@ else
 	ln -s $(CONFIG_EXTERNAL_KERNEL_TREE) $(LINUX_DIR)
   endef
 endif
+Kernel/Prepare?=$(Kernel/Prepare/Default)
 
 ifeq ($(CONFIG_TARGET_ROOTFS_INITRAMFS),y)
   ifeq ($(strip $(CONFIG_EXTERNAL_CPIO)),"")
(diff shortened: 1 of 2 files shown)


langerth:
I think the toolchain is per intention built without target specific patches. You can use the same toolchain for multiple targets, without the need to rebuild it each time.

KarlVogel:
To clarify, this stage installs the kernel headers, it doesn't generate toolchain binaries. It's still possible to use the compiler of another target, but you'll indeed need kernel headers matching your target's kernel if there's differences in headers due to platform patches. (for example the issue with the old layer7 patches)
Being able to use the same toolchain is already tricky, as it depends on .config settings if the toolchain is actually usable for the target (ie. depends on items like selected libc, hardening options like stack smash, ...)

hauke:
LEDE builds the packages not per target any more, but per architecture. For all the targets with a mips be 24K CPU the user space packages are only build once.
When you really need some target specific kernel headers include the kernel headers directly in your package.

nbd168:
@hauke is right, this is not a suitable approach to deal with target specific kernel header issues.

PR #494, opened 2016-11-06
ipq806x: add RPM cc, idle states and minor stuff

dissent1:
This PR adds support for RPM Clock Controller, RPM message RAM and cpu idle states.

Please take into account, that 179 rpm patch is a modified version of the one that has been submitted upstream. The original patch is for apq8064 board only, so I modified it to work on ipq806x only.
Sooner or later this patch will go upstream and we'll have to make another patch that adds support for ipq806x in addition to apq8064 but not replacing it.

003 upstream patch has also been modified- cxo/pxo clocks are left at 25kHz/25kHz in contrary to upstream changes as it's not completely stable now and requires more testing.

Also replacing existing patch for WDT Barktime with set of patches from upstream.

@heinzek

PR #525, opened 2016-11-16
kernel: correctly refresh patches for ipq806x

kdarbyshirebryant:
ipq806x patches were refreshed with incorrect quilt options & do not
conform to standard format.  This patch corrects those formatting
errors.

Signed-off-by: Kevin Darbyshire-Bryant <kevin@darbyshire-bryant.me.uk>

PR #529, opened 2016-11-17
ipq806x: refresh patches

hnyman:
This patch refreshes the ipq806x kernel patches.

There was a large PR for ipq806x in the queue when the kernel patches were refreshed for 4.4.32, so currently there is quite much fuzz for ipq806x.

Most of changes are pure offset corrections and cosmetics. 003-mv-cxo-pxo-xo-into-DT.patch is the exception, as there "refresh" wants to reorganise the patch a bit by keeping all deleted and added stuff as consecutive blocks.

EDIT: additionally some of the patches in #494 were backported from newer Linux kernels without refreshing the patches for 4.4.


kdarbyshirebryant:
Ahem https://github.com/lede-project/source/pull/525  but I like you commit message better :-)

At least I came up with the same changes.  I'll close my PR.

stintel:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/stintel/staging.git;a=commit;h=8459d85fa38e8882a9767c105fc01cb8e95e7d5b

PR #530, opened 2016-11-17
odhcpd: Fix dnsmasq rereading hostfile

dedeckeh:
Depending on the dhcp config pidof dnsmasq can return
multiple pids. Fix rereading of the hostfile by dnsmasq
in such case by sending SIGHUP signal to each of the
returned pids.


hnyman:
I wonder if it would make sense to update odhcpd upstream repo location at the same time.

I noticed a few days ago that currently the "old" upstream repos of sbyx for both odhcpd and odhcp6c say that they are now forks from github/openwrt. And there is actually a new commit for both odhcpd and odhcp6c in the new github/openwrt while the old repos by sbyx are stale, so apparently the development is moving there???? 

https://github.com/sbyx/odhcpd/commits/master
https://github.com/openwrt/odhcpd/commits/master

dedeckeh:
@hnyman Imo it makes sense to update the odhcpd upstream repo location as I opened a PR against the github/openwrt odhcpd repo for other patches. Will do this in a seperate PR

PR #533, opened 2016-11-17
ipq806x: add thermal sensors driver and small fixes

dissent1:
Allows us to enable embedded thermal sensors to check cpu temperature.

Also cherry-picking small fixes from qsdk repo.


hnyman:
I tested your thermal sensor patch in R7800 with IPQ8065 and it nicely makes the thermal data visible. Thanks.

target/linux/ipq806x/patches-4.4/309-2-tsens-Support-for-TSENS-driver.patch
```diff
@@ -0,0 +1,1611 @@
...
+ 			qcom,controller-type = "pmic-arbiter";
+ 		};
+ 
++		tsens: tsens-ipq806x {
++			compatible = "qcom,ipq806x-tsens";
++			reg = <0x900000 0x3678>, <0x700000 0x420>;
++			reg-names = "tsens_physical", "tsens_eeprom_physical";
++			interrupts = <0 178 0>;
++			qcom,sensors = <11>;
++			qcom,tsens_factor = <1000>;
```
hnyman: It is probably best to leave things as your have them. I looked at the source and the factor gets used in connection with "slope" etc., so changing factor would likely lead to other changes, too.

Like I said on the forum discussion https://forum.lede-project.org/t/supporting-thermal-sensors-on-ipq806x/205/17?u=hnyman , it is no big deal.   (Actually, having full normal Celsius degrees '56' there is much more userfriendly than having millicelsius like '56000')

bobafetthotmail: imho it would be better to keep it in millicelsius, if that's what upstream has set as standard then it will be what most scripts/packages will expect to read it (as shown by collectd-mod-thermal).

Most users won't be polling manually the temperature.

hnyman: Looks like backporting the newer OEM tsens driver will be a long effort. I have compiled and tried some test efforts from @dissent1 but so far no luck. There is not much documentation to go on...

This current version seems to work (and I already wrote LuCI support for it) , so I suggest that it is merged in until something better can be invented.

dissent1:
Well, for now there are only 2 working ways of enabling thermal sensors on ipq806x platform in lede: this one and that one https://gist.github.com/updateing/a3101c9d99463445b6fc2d41f63e21cd
This one provides more or less standalone patch but in full Celsius and removes OF-Thermal driver from kernel.
The other method is to cherry-pick the driver from the link above that provides info in millicelsius, but implements second qfprom driver along with the one that is in the kernel already, but embedded in nvmem driver.

Both ways are dirty and I'm not sure which one is better.

@blogic @mkresin Could you give an advise what to choose.

The problem also is that the version of the driver that went upstream seems doesn't support ipq806x. At least I couldn't backport it completely (though it builds correctly) to kernel 4.4 or it ain't working.

blogic:
not sure to be honest ... @nbd168

hnyman:
I have compiled and tested several of the alternative OEM driver code versions from dissent1. they all have failed me with R7800.

But when I now look more closely at the "full-celcius" approach in this pull request and which works for R7800, it seems to me that it would be rather straightforward to convert the driver to return millicelcius in a Linux standard way. 
* The first patch 309-1 actually adds ``` drivers/thermal/qcom-tsens.c ``` supporting millicelsius, function names are like ``` tsens_tz_code_to_mdegC ``` and defines are like ```TSENS_CAL_MDEGC  30000 ```
*  But the second patch 309-2 that claims to "port the driver and cleanup unused code" actually makes a large effort to replace millicelsius by celcius.  E.g. now there are ``` TSENS_CAL_DEGC  30 ```and ``` tsens_tz_code_to_degC ```. And this is the non-standard part.
* There are only a few places where the temp is actually assigned or used as input. I propose that we leave the internals and function names as they are, so that we do not need to change constant etc. But we add a "1000 *" multiplication to the three lines, where the read value is actually assigned to "temp" and "/ 1000" to the only place where "temp" is input (setting alert levels). 
  * in ``` tsens8960_get_temp ``` line 412 in patch: https://github.com/lede-project/source/pull/533/commits/c1ce64db7778e5969c3b2bdffc6507f8a65eeecf#diff-68010598af8ea82427b109203689efb0R412
  * in ``` tsens_tz_get_trip_temp ``` line 740 in patch: https://github.com/lede-project/source/pull/533/commits/c1ce64db7778e5969c3b2bdffc6507f8a65eeecf#diff-68010598af8ea82427b109203689efb0R740
  * in ``` tsens_scheduler_fn ``` line 934 in patch: https://github.com/lede-project/source/pull/533/commits/c1ce64db7778e5969c3b2bdffc6507f8a65eeecf#diff-68010598af8ea82427b109203689efb0R934
  * in  ``` tsens_tz_set_trip_temp ``` line 773 in patch it is used as  input value: https://github.com/lede-project/source/pull/533/commits/c1ce64db7778e5969c3b2bdffc6507f8a65eeecf#diff-68010598af8ea82427b109203689efb0R773 (and the same line is touched by patch 309-3). Afaik that is the only call to ``` tsens_tz_degC_to_code ```. Making sure that here 65000 gets passed as 65 would keep things straight.
Those line numbers are from the current patch set, not from the final file.

Alternative approach could be to add a new patch that changes the two functions ``` tsens_tz_code_to_degC ``` and ``` tsens_tz_degC_to_code ``` to be again "mdegC" and to add the "1000 *" and "/ 1000" there in the functions. That would then require changing the function calls, but there are only 3 and 1 calls.

I have not yet tried this, but it looks quite feasible. Likely more feasible than trying to get the undocumented other OEM driver to work.

hnyman:
My patch seems to work:
```
root@lede:~# cat /sys/class/thermal/thermal_zone2/temp
55000
```
image

dissent1:
We've managed to make upstream thermal driver work as expected. Had to make a custom patch (309) based on same patch for apq8064 board.

hnyman:
When I noticed that you started to finalize the commit for release, I compiled 9839404 from your sta14 branch (this commit before refreshing the other patches) and yes, it works nicely. R7800 shows 4 thermal sensors and reasonable values in milli-Celsius for all of them ;-)

dissent1:
Thanks again @hnyman for patient testing :)

Just FYI there are 4 thermal zones more, but these seem to be the most relevant and shouldn't confuse users with so many of them.

hnyman:
@blogic @mkresin 
just flagging that this should be ready for merge

dissent1:
Fixed merge conflicts

blogic:
merged, Thanks !

nbd168:
Does the "ipq806x: increase coherent dma pool size" patch actually fix anything for you?
I'm pretty sure they added it in the qsdk tree mostly to work around some ath10k stupidity that I fixed upstream months ago.

dissent1:
@nbd168 it seem to affect overall system behavior under heavy load: high network, CPU and USB activity at the same time. Also as an effect load average is lower on similar tasks.

PR #535, opened 2016-11-18
kexec-tools: fix compile error

pgid69:


mkresin:
Please use a proper commit title. Check the commit history of the package you're patching to see what I'm talking about. A proper commit title would be "kexec-tools: fix compile error".

Your current commit title should go into the commit message. Please add some more information about the conditions that trigger the error to allow one to reproduce the issue. For example which platform is affected.

At least our buildbots seam to build to package without issues. No faillogs regarding kexec-tools in https://downloads.lede-project.org/snapshots/faillogs/.

pgid69:
Hi

I change the title and the commit message.
I have the problem on two platform : mips/brcm63xx and mips/brcm47xx.
But i think the problem is independent of the platform because the type loff_t can't be defined if flag _GNU_SOURCE is not set
I attach the diffconfig i use for platform mips/brcm63xx to reproduce the error

[diffconfig.txt](https://github.com/lede-project/source/files/603003/diffconfig.txt)

nbd168:

PR #537, opened 2016-11-18
rtl8367b: add MDIO support, LED options, port control

13hakta:
Reformat tabs/spaces in code
Add direct mdio bus support
Add LED mode options - mode and blinkrate
Add green mode option - save energy when port is inactive
Add port disable option
Add ability to point cpu port
Fix RGMII initialization for ext2 port

Signed-off-by: Vitaly Chekryzhev 13hakta@gmail.com


target/linux/generic/files/drivers/net/phy/rtl8366_smi.c
```diff
@@ -919,6 +969,21 @@ static int rtl8366_smi_mii_init(struct rtl8366_smi *smi)
...
+	if (smi->mdio_enabled) {
+		np = of_parse_phandle(smi->parent->of_node, "mdio", 0);
+
+		if (np) {
+			bus = of_mdio_find_bus(np);
+
+			if (bus) {
+				smi->mii_bus = bus;
+				return 0;
+			} else return -ENODEV;
```
mkresin: I tested the patchset on a lantiq board, which is not yet supported because of the missing rtl8367b mdio support.

This return has to be -EPROBE_DEFER. In my case the mdio driver is loaded after the switch driver. Means, at the time the switch driver is loaded, there is no mdio bus yet. If you use -EPROBE_DEFER the switch driver is probed again at a later time.

The same used in the kernel in conjunction with of_mdio_find_bus as well.

mkresin:
Does it make sense do add device tree bindings for the led mode, blink rate and green mode? Or should they be controlled only via swconfig?

blogic:
there was a discussion upstream with @hauke regarding OF bindings of phy leds

target/linux/generic/files/drivers/net/phy/rtl8366_smi.c
13hakta: Good advice, will rewrite.

13hakta:
On my own I prefer to control leds and green mode by swconfig, as its handy to change options on the fly. Blinkrate could be binded, but its not that important option to put in devicetree. What are highlights of that discussion?

target/linux/generic/files/drivers/net/phy/rtl8366_smi.c
13hakta: I suppose that driver set for rtl8366/7 should be rewritten so common functions will be brought out to function library rather additional driver (rtl8366_smi). No other driver than mentioned uses that driver. What do you think?

13hakta:
Rebased to apply upstream changes, commits are the same, except that MDIO bus loader updated.

mkresin:
Due to the lack of a datasheet, I have no idea what the green mode does. If it might break the switch functionality, it isn't possible to login to the router and disable the green mode via swconfig. If we have a device tree option, we could prevent such situations. That is the whole point behind asking if a device tree option make sense.

13hakta:
Green mode is technology known as IEEE P802.3az (http://www.ieee802.org/3/az/). On my zyxel device green mode is enabled in stock firmware (and cannot be disabled by user).

mkresin:
The problem is that the RTL8367B switch driver is enabled for lantiq/xrx200, lantiq/xway, lantiq/xway_legacy, ramips/rt3883 and ramips/rt305x/wl-351. As long as we can not test all boards for possible issues because of a default enabled green mode, I refain from enabling it by default.

We do have exactly the same discussion at https://github.com/lede-project/source/pull/579#issuecomment-266585680. If no one comes up with a better solution I'm for using realtek,enable-powersaving to enable the green mode.

Once we manged to test the green mode on all boards having a RTL8367B (might never happen), I'm fine to enable the green mode by default and add a realtek,disable-powersaving device tree property instead.

> I suppose that driver set for rtl8366/7 should be rewritten so common functions will be brought out to function library rather additional driver (rtl8366_smi). No other driver than mentioned uses that driver. What do you think?

Improving/cleaning up stuff is always welcome. But it has to be tested with all rtl8366*/7* switches. I would suggest to do it as a follow up PR.

> On my own I prefer to control leds and green mode by swconfig, as its handy to change options on the fly. Blinkrate could be binded, but its not that important option to put in devicetree. What are highlights of that discussion?

The highlight are these two patches in the linux-netdev patchwork:

- [PHY: Add PHY LED control binding](https://patchwork.ozlabs.org/patch/630614/)
- [NET: PHY: Intel XWAY: add LED configuration support](https://patchwork.ozlabs.org/patch/630616/)

Unfortunately the patchwork status says changes requested and @hauke obviously stopped working on it.

As far as I can see, you just adding support for setting the LED function via swconfig without enforcing something/overwriting already existent settings (maybe configured by the bootloader). On the long run I prefer to have some device tree params to configure the switch LEDs. But I don't consider this a blocker for merging the patch.

13hakta:
I vote for controlling green mode by swconfig. In my PR green mode is NOT enabled by default.

mkresin:
> In my PR green mode is NOT enabled by default.

Well, you are right. I mixed it up with the rtl8366s PR (reviewing two similar PRs is really confusing) and I was fooled by my test device which has green mode enabled by default.

>  On my zyxel device green mode is enabled in stock firmware (and cannot be disabled by user).

Well that leads me to the next question. Why do we need this option if both of our devices have IEEE P802.3az enabled by default? I don't really see a reason to expose this setting to the userspace. Or missed I something that can be fixed by having this option? I mean, you wrote by yourself that this option can not be disabled with the stock firmware.

Don't get me wrong, you code is good. I'm just not that thrilled of adding functions/options to the switch driver that will be never used.

The same applies to the LED mode options and disable ports options. Similar exists for other switch as well. But they are exposed via device tree and used/required by some boards.

For example on lantiq, the phys have multiple pins for LEDs and depending on which led is attached to these pins, we need to setup the LED function register. It is done via the device tree property ```lantiq,led*``` instead of an userspace api, since this settings are hardware specific and it is a one time fix on switch init.

On ramips we have the device tree property ```mediatek,portdisable``` for some switch drivers. The property is used to disable unpopulated switch ports/phys to save power. Again, a setting that is hardware specific since the number of populated/unpopulated switch ports does not change during runtime. Albeit, one can argue that a userspace disable port option would allow to disable populated but not connected ports to save even more power. That is the reason why I did not mentioned the disable port option yet.

13hakta:
@mkresin `Well that leads me to the next question. Why do we need this option if both of our devices have IEEE P802.3az enabled by default?`
I was not exact in my words. This enabled by default in driver (from zyxel developers), other manufacturers may write another driver using API, as it is for custom firmware like Asus RT-N56U from Padavan:
https://bitbucket.org/padavan/rt-n56u/src/5aa2a755c4b0077eb4f24800cd09493fd9c4b182/trunk/linux-3.4.x/drivers/net/rtl8367/?at=master
So, this feature can/may/must by enabled by user himself by his own decision.
Disabling port feature is another ability, may be used not only for power consumption but also for administering purposes.

For LED case I remind that sometimes it can be useful (hm...so rare sometimes for SOHO routers) to set different LED modes with script to see if port is active or has specific port speed.

mkresin:
Let us postpone the discussion about the led, green mode and port disable options for now. I asked blogic for an opinion on this. He is way more familiar with what should be user space controlled and what should go into the device tree.

But I found another issue. The function ```rtl8366_smi_detect()``` is not unused. It is used in [target/linux/ar71xx/files/arch/mips/ath79/mach-wzr-hp-g300nh.c](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-wzr-hp-g300nh.c#L233). As soon as the function is removed the build of this boards will fail.

mkresin:
I had more closer look at your clean up patches. Most if not all spaces vs. tabs in rtl8366_smi.c  seam to intentional.

Old:
old

Changed:
new

I didn't checked all changes, but I guess this commit can be dropped.

The same applies to your spaces vs. tabs changes in commit "rtl8367b: code cleanup, remove unused macro". This commit is kinda hard to review. Would you please split this commit into one commit doing the cleanup and another commit removing the unused macros.

13hakta:
In my commit spaces turned to tabs and redundant spaces truncated to single space.
Removed commit with method removing - you're right, method call was hidden so far..
Squashed cleanup commits.

mkresin:
> In my commit spaces turned to tabs and redundant spaces truncated to single space.

... and your commit breaks the formatting/alignment intended by the author.  I've added the screenshots to show you the differences. With the old formatting the code is way better to read. Means, you are not fixing tab/space mistakes, you are making the code harder to read.

Instead of an commit doing the tab/space cleanup and one commit removing the unused macros, we now have a hugh commit which does both. That is exactly the opposite of what I've requested.

13hakta:
Well, I'm not against of dropping formatting commit as you supposed. And extract only removal part.

13hakta:
Splitted huge commit into 3 and extracted removal part.

13hakta:
Any more changes required?

jow-:
@mkresin @blogic - can I merge this?

mkresin:
@jow- No, please don't merge this patch yet. Found some more issues but I need some time to write a proper response. Hopefully I get to it during this week.

mkresin:
Hey Vitaly,

sorry for the late reply, I had some busy days.

The major problem with your pull request is that I wasn't able to find a datasheet for the rtl8367b and therefore I'm not able to validate your changes. All I have are your actually changes and your commit messages. And there the problem beginns. You do not have proper commit messages. The commit messages should explain the changes and should explain why the changes are required.

I'm not entirely sure but your change in "rtl8367b: add ability to point cpu port":

```diff
-	dev->cpu_port = RTL8367B_CPU_PORT_NUM;
+	dev->cpu_port = smi->cpu_port + 1;
```

looks like it will break existing boards. The old value was ```RTL8367B_CPU_PORT_NUM``` and the new one is in the end ```RTL8367B_CPU_PORT_NUM + 1```.

Furthermore, you have a lot of styling errors in your code. I used ./scripts/checkpatch.pl to find them.

Regarding the cleanup/code reformat. I obviously failed to point out that your reformatting makes things worse. I would like to keep the current formatting and fix only lines that do not match the way the code is formatted at the moment.

I've fixed most (but not all!) of the styling errors, the error mentioned above, some of the commit messages and a few other things a fail to remember at the moment in my staging repo at https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary.

Would you please test these patches, fix the remaining styling issues reported by checkpatch.pl and the commit messages!

At the moment I would like to focus on the essential required patches and defer the optional changes to a later point. Please reorder your commits so that the ones adding optional stuff are the last ones and I'm able to pick the first batch of commits.

13hakta:
@mkresin cpu_port is used in method rtl8366_init_vlan (rtl8366_smi.c) for initial VLAN configuration. It sets own VLAN for every port, but when system configures switch to take in account user preferences it resets to somewhat port 0-4 VLAN1 and port 5 VLAN2.
Problem with driver was that it didn't take in account different board may use different CPU port.

mkresin:
Either you missed what I tried to explain or I don't have the required knowledge of the code. The problem I might have seen is that you change the default CPU port from 5 to 6 in case the cpu_port device tree property is not used:

```diff
#define RTL8367B_CPU_PORT_NUM           5
-smi->cpu_port = RTL8367B_CPU_PORT_NUM;
-dev->cpu_port = RTL8367B_CPU_PORT_NUM;
+smi->cpu_port = RTL8367B_CPU_PORT_NUM;
+dev->cpu_port = smi->cpu_port + 1;
```

13hakta:
@mkresin yes, as I explained - it just insignificant cause its rewritten every time system boots.
As a reference I look to this driver, API is also present there https://bitbucket.org/padavan/rt-n56u/src/5aa2a755c4b0077eb4f24800cd09493fd9c4b182/trunk/linux-3.4.x/drivers/net/rtl8367/?at=master

mkresin:
Sorry, I still fail to understand why this change should not cause issues.

Just to make really sure we are talking about the same. It's not the change of ```smi->cpu_port``` I'm concerned about. It is the change of ```dev->cpu_port```. Your reference of ```rtl8366_init_vlan()``` is bogus to me since only ```smi->cpu_port```  and **not** ```dev->cpu_port``` is used in this function.

Without your patch the value of ```dev->cpu_port``` is 5. With your patch applied and without specifying anything in the device tree, the value of ```dev->cpu_port``` is **changed** to 6.

Either I missed it or you didn't provided an explanation why ```dev->cpu_port``` has to be RTL8367B_CPU_PORT_NUM + 1 instead of the actual used RTL8367B_CPU_PORT_NUM.

13hakta:
Some clarification required. At first you are right - `dev->cpu_port = smi->cpu_port + 1` - must be without increase. But it still may not be **5** by default.
Driver in title is announced to support _RTL8367R-VB_ but in fact it has code to work with _RTL8367R-VB_ and _RTL8367RB_.

Description page for [RTL8367R-VB](http://www.realtek.com/products/productsView.aspx?Langid=1&PNid=18&PFid=15&Level=5&Conn=4&ProdID=299) says it has 5+1 port. That means 5 wired and 1 extra port as CPU port - in this case **RTL8367B_CPU_PORT_NUM** = 5.

Description page for [RTL8367RB](http://www.realtek.com/products/productsView.aspx?Langid=1&PNid=18&PFid=15&Level=5&Conn=4&ProdID=300) says it has 5+2 port. That means 5 wired and 2 extra ports that act as CPU port - in this case **RTL8367B_CPU_PORT_NUM** is dependant on configuration and can be **6** or **7**.
Currently only one board (DIR-645) used this driver and it has RTL8367RB switch and it used default cpu_port value = 5 which is wrong.

After this investigation I see that commit may be rewritten in such way, that it wouldn't require manual configuration of cpu_port and will set it according to detected port and property "realtek,extifN" in devicetree.

mkresin:
Your reply raised even more question than it answered. I will get in contact with you via mail soon.

13hakta:
Patchset will be reworked.

hailfinger:
Was a reworked version of this patchset merged? I can't find anything related in the OpenWrt changelog.

mkresin:
Nope. Unfortunately a reworked version was never send.

13hakta:
Delayed, but still working

hailfinger:
Is there a subset of this pull request which can be merged as-is? I would appreciate having less external patches for the Vodafone Easybox 904xdsl.

13hakta:
@hailfinger yes, they are published at my source repo.

PR #545, opened 2016-11-21
base-files: add a wrapper for init scripts in profile 

bobafetthotmail:
"service" is a simple wrapper that will allow to call init scripts in /etc/init.d by writing "service" (the wrapper's name) followed by the init script's name and then by the command. Here an example.

current method:     #  /etc/init.d/network reload
with the wrapper:   #  service network reload

If the wrapper is called without arguments or with a wrong init script name, it will list the content of etc/init.d/ folder, showing what initscripts can be called.

This wrapper mimicks the main function "service" utility found in linux distros, see manual here https://linux.die.net/man/8/service that was born as a wrapper for initscripts and is now a wrapper for some systemctl (systemd) functions.

### DIFF
--- a/package/base-files/files/etc/profile
+++ b/package/base-files/files/etc/profile
@@ -41,3 +41,14 @@ in order to prevent unauthorized SSH logins.
 --------------------------------------------------
 EOF
 fi
+
+service() {
+if [ -f "/etc/init.d/$1" ]; then
+        "/etc/init.d/$@"
+else
+	printf '%s\n%s\n' "unknown service" "list of available services :"
+	ls "/etc/init.d"
+	return 1
+fi
+}
+


bobafetthotmail:
I've sent this from git and I checked that the file was also executable before committing, btw.

blogic:
i dont like this at all, also the package is wrong and the prefix aswell. the script does nothing but simple call delegation. its a NAK from me.

bobafetthotmail:
why is call delegation wrong? I find it's more convenient than having to write the full path each time, also having the script list all available services is more convenient than having to write again ls + full path.

Consider that this is meant for human interaction, not for scripts.

mkresin:
> why is call delegation wrong? I find it's more convenient than having to write the full path each time, also having the script list all available services is more convenient than having to write again ls + full path.
> 
Because is it just bloat. LEDE targets embedded systems and not enduser (desktop) systems.

Therefore the following argument is invalid:
> Consider that this is meant for human interaction, not for scripts.
> 
NAK from me as well.

bobafetthotmail:
> Because is it just bloat. 

It is 4 lines of script, let's not over-react.

>LEDE targets embedded systems and not enduser (desktop) systems.

And embedded systems don't need a decent CLI interface? 

You have a full-blown UCI CLI interface that does much more complex things parsing text config files and that's fine, but a simple wrapper to call init scripts to start/enable/restart services without writing a full path to it is wrong?

I mean, ok, you decide, but I don't think this reasoning is valid.

blogic:
so the main problems here are

1) service is a reserved namespace. we call the daemons managed by procd services. this wrapper is a init.d wrapper and not a service wrapper
2) if it were a service wrapper it would need to go into procd/ package
3) if it is an iitd wrapper it would need to be renamed to represent this
4) the subject prefix would need to be changed
5) this can all be handled inside the profile file using an alias

i like the idea of such a wrapper but it needs to be named differently and not be coded as a 10 liner script

bobafetthotmail:
>this can all be handled inside the profile file using an alias

Ok, you mean squash the script in a one-liner in the profile (i.e. preserving current functionality)?

Can do it like that, yes.

Name "service" is imho correct and was chosen because of the same reasons most other linux distros chose it.
The init script is just a middle man to start/stop/enable/disable/whatever a system service, so it is ok to call the wrapper/utility "service" as it allows you to start/stop/enable/disable/whatever system services.
See the manpage for the "service" utility this wrapper is emulating https://linux.die.net/man/8/service
(the same utility is now updated to work with systemd unit files too, and is still in use in major distros like Debian or OpenSUSE)

Anyway I can change the name to something else if you insist, the most important thing for me is to get the wrapper (with whatever name) in LEDE.

blogic:
ok, send a patch with the same feature set but as an alias inside the profile file please

karlp:
:-1: Adding alternative redundant paths to the same thing makes it harder to provide quality, simple documentation.  It's harder to assist people using multiple forms that all _should_ be doing the same thing.  This adds little, (there's tab completion of /etc/init.d/x<tab> but no tab completion of service <tab>) and muddies the waters for all.

bobafetthotmail:
updated, turned the script in a function in /etc/profile (an alias would not allow it to react in case of misspelled names), tested and works.

This has received mixed feedback here and in the mailing list (jow, blogic, lars karlsson like it, mkresin and karlp dislike) also the discussion there might not be over yet. http://lists.infradead.org/pipermail/lede-dev/2016-November/004186.html

package/base-files/files/etc/profile
```diff
@@ -41,3 +41,13 @@ in order to prevent unauthorized SSH logins.
...
 EOF
 fi
+
+service() {
+if [ -f "/etc/init.d/$1" ]; then
+        "/etc/init.d/$1" "$2"
+        return $?
+else
+        ls "/etc/init.d"
+fi
```
blogic: various issues here

* the implicit list is not obvious and jut a side effect of the code style. there is also no warning.
* what happens if $1 is null
* this is not an alias
* some initd commands take more than 1 option
* the return $? is duplicate as $? is already populated

bobafetthotmail:
updated as said in the last review thread.
Replaced "/etc/init.d/$1 $2" with "/etc/init.d/$@", removed the "return $?".

package/base-files/files/etc/profile
```diff
@@ -41,3 +41,12 @@ in order to prevent unauthorized SSH logins.
 --------------------------------------------------
 EOF
 fi
+
+service() {
+if [ -f "/etc/init.d/$1" ]; then
+        "/etc/init.d/$@"
+else
+        ls "/etc/init.d"
+fi
```
blogic: right now the list is implicit and wont even cause a bad error code. please drop the implicit list and return an error instead

blogic:
a simple "unknown service\n list of available services :\n ..... style echo and a return 1 would be sufficient i guess

bobafetthotmail:
updated with the error message and return 1.

bittorf:
i'am not convinced, that 'ls /etc/init.d' helps. the list is just too long. maybe we can only list items which define start() and stop()?

bobafetthotmail:
I'm not sure why you see this as a long list:

this is the result of a ls /etc/init.d in my test router with LEDE (default install + extroot):

```
root@lede:/# ls /etc/init.d
boot          firewall      network       sysntpd       urandom_seed
cron          fstab         odhcpd        system
dnsmasq       gpio_switch   rpcd          test
done          led           sysctl        uhttpd
dropbear      log           sysfixtime    umount
```

this is the result of a quick filtering for all initscripts that have a "start" in them (some name their start function "start_something()").

```
root@lede:/# for file in $( ls "/etc/init.d" ) ; do grep "start" "/etc/init.d/$file" 2>&1 >/dev/null ; if [ $? = 0 ] ; then echo $file ; fi ; done
cron
dnsmasq
dropbear
firewall
fstab
gpio_switch
led
log
network
odhcpd
rpcd
sysctl
sysfixtime
sysntpd
system
uhttpd
urandom_seed
root@lede:/# 
```
It's basically removing "boot" init script.

blogic:
merged, made various changes to the code  and changed the commit message.

PR #546, opened 2016-11-21
ramips: SamKnows SK-WB8 DTS cleanup

ndoo:
Use gpio.h definition of GPIO_ACTIVE_HIGH and GPIO_ACTIVE_LOW.
Remove unused backup partition to increase available JFFS space. As long as U-Boot env variable "bootcount" is < 3 (reset to 0 after boot by init script) SamKnow's U-Boot will not attempt to boot from the backup flash address (0xe30000).

### DIFF
--- a/target/linux/ramips/dts/SK-WB8.dts
+++ b/target/linux/ramips/dts/SK-WB8.dts
@@ -1,8 +1,8 @@
 /dts-v1/;
 
 #include "mt7621.dtsi"
-
 #include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
 
 / {
 	compatible = "mediatek,mt7621-eval-board", "mediatek,mt7621-soc";
@@ -22,12 +22,12 @@
 
 		wps {
 			label = "sk-wb8:green:wps";
-			gpios = <&gpio1 14 1>;
+			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
 		};
 
 		usb {
 			label = "sk-wb8:green:usb";
-			gpios = <&gpio1 15 1>;
+			gpios = <&gpio1 15 GPIO_ACTIVE_LOW>;
 		};
 	};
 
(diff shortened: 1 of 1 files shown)


blogic:
what happens if bootcount >=3 ?

ndoo:
U-Boot tries to boot from partition at 0xe30000.

If there is a valid kernel at 0xe30000 (only if SamKnows pushes an update causing this partition to have a copy of the old kernel?), U-Boot starts the kernel there.

If U-Boot fails to detect the kernel magic (possibly because SamKnows did not push a backup kernel to the partition), it drops to the U-Boot shell (which gives the appearance of a bricked router) until a TTL serial adapter is connected and `setenv bootcount 1`, `saveenv` and `reset` are issued.

`target/linux/ramips/base-files/etc/init.d/bootcount` ensures this does not happen as long as the boot succeeds up to S99 init scripts.

This is not elegant, but AFAIK LEDE does not support dual images.

ndoo:
@RedDwarf69 from SamKnows may be able to provide further insight into the recovery partition but as far as mainline LEDE is concerned I am not sure there is any advantage to keeping this partition.

Even a successful boot to recovery kernel would probably result in a brick to the end-user

mkresin:
FYI, dual images are used for [Linksys mvebu based boards](https://github.com/lede-project/source/blob/master/target/linux/mvebu/base-files/lib/upgrade/linksys.sh). It's not a not beauty but seams to work.

ndoo:
Ubifs does seem more amenable to multi partition setups - don't think the same level of (not beautiful) elegance can be achieved specifying offsets in DTS. Hmm.

RedDwarf69:
Feel free to nuke the recovery partition thing. It's there to allow us to do automated (kind of) safe firmware updates.
It's a last minute personal effort, meaning there is nothing actually written on it from factory (and to the day we still have not written anything to it). At some point we want to put in there something that downloads our latest firmware and writes it into the main partition. The idea is that:
a) If for some reason the main partition gets bricked, it will be restored after a few reboots.
b) When we want to deploy an update, we don't call mtd directly from the main partition. We just set bootcount to 3 and whenever the device is restarted it will boot into the recovery partition, which will update the main partition. This way when updating we don't reboot the device when it's inconvenient to the user, and if for any reason the update process fails, the recovery partition can just try again instead of getting the device bricked.

If somebody is using LEDE I guess they don't want the recovery partition to delete it and write the SamKnows firmware back again in its place. So I guess it makes sense to get rid of it.

ndoo:
@RedDwarf69 thanks for your insight. Yes, I was wondering why my unit couldn’t find kernel magic.

blogic:
merged, Thanks !

PR #547, opened 2016-11-21
ramips: fix wrong check for MT7621AT fix in sdhci

ndoo:
Use ralink_soc == MT762X_SOC_MT7621AT instead of CONFIG_SOC_MT7621 which is wrong and breaks builds on mt7620a-similar platforms (MT7621, MT7688)


ndoo:
@blogic FYI. I tested build on MT7621 target and it succeeded.

mkresin:
merged

PR #551, opened 2016-11-21
Vgv7519 uboot envtools

dpeddi:


package/boot/uboot-envtools/files/lantiq
```diff
@@ -26,6 +26,9 @@ BTHOMEHUBV3A)
 P2812HNUF1)
 	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x2000" "0x20000" "1"
 	;;
+VGV7519)
```
mkresin: why is it required to write the uboot env of the VGV7519 from userspace?

Side note: why is neither the u-boot nor the u-boot env partition of the VGV7519 write protected. Could you fix this as well.

dpeddi:
Not really needed, but nice to see the configured parameters.

About uboot write protection:
I think that is unprotected for historic reason development reason. I think
The_Lizard left this configuration to update u-boot without any connected
stuff (ttl2usb or jtag)

Consider that this board comes out with brnboot, so you really need at
least a ttl2usb adapter to switch to u-boot. So if you mess u-boot you can
always recover the board by connecting the ttl2usb adapter.

If you think that write protecting mtd is a best pratice I would protect
just the uboot partition, so I can still use uboot-envtool.

mkresin:
> Not really needed, but nice to see the configured parameters.

Please drop this commit from the PR. I see little to no use for anyone then you. strings /dev/mtd1 is your friend.

> 
> About uboot write protection:
> I think that is unprotected for historic reason development reason. I think
> The_Lizard left this configuration to update u-boot without any connected
> stuff (ttl2usb or jtag)
> 
> Consider that this board comes out with brnboot, so you really need at
> least a ttl2usb adapter to switch to u-boot. So if you mess u-boot you can
> always recover the board by connecting the ttl2usb adapter.
> 
> If you think that write protecting mtd is a best pratice I would protect
> just the uboot partition, so I can still use uboot-envtool.

Please protect both partitions. As you said, a serial adapter is required for this board anyway and therefore the env variables can be changed this way (which shouldn't be required at any time to my knowledge). And write access isn't required for your "see configured parameters" use case.

dpeddi:
done!

mkresin:
Well, the commit "uboot-envtools: add support for vgv7519 openwrt flash map" is still part of this PR...

dpeddi:
And so? We need to configure it if we want to use fw_printenv.

Other platform have a incredible list of devices in uboot helper file!

Il 22/nov/2016 12:48 PM, "Mathias Kresin" <notifications@github.com> ha
scritto:

Well, the commit "uboot-envtools: add support for vgv7519 openwrt flash
map" is still part of this PR...

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lede-project/source/pull/551#issuecomment-262221073>,
or mute the thread

dpeddi:
I've updated and splitted this PR. now it contains just the uboot-envtools for vgv7519, you can reject this and promote the others...

mkresin:
Still NAK from me.

Having u-boot envtools support for the board just to have a look at the values isn't simply not enough. Most of the boards in this file need access to the uboot-env because it is required for install or a mac address needs to be fixed.

pepe2k:
> Having u-boot envtools support for the board just to have a look at the values isn't simply not enough. Most of the boards in this file need access to the uboot-env because it is required for install or a mac address needs to be fixed.

@mkresin

What with users who would like to change some U-Boot config options, like 'bootdelay' which is sometimes set to huge values like 5-15 seconds or to 0 which simply disables U-Boot CLI access? Or add support for tftp update before starting the kernel (ex. change the 'bootcmd' variable)? Without serial console access 'u-boot envtools' is the only way to make it.

IMHO, if the board have U-Boot environment partition available, we should include support for R/W access to it by default.

Cheers

mkresin:
> What with users who would like to change some U-Boot config options, like
> 'bootdelay' which is sometimes set to huge values like 5-15 seconds or to 0
> which simply disables U-Boot CLI access? Or add support for tftp update
> before starting the kernel (ex. change the 'bootcmd' variable)? Without
> serial console access 'u-boot envtools' is the only way to make it.
>
> IMHO, if the board have U-Boot environment partition available, we should
> include support for R/W access to it by default.

I might miss the use case here. But I don't see that a casual user has
any need to manipulate the u-boot environment. Rather than that, we
should try to protect users from shooting them self into the foot.

At least for this board is literally no need to manipulate the env. It
is neither required to fix a mac address or to have write access for
any install hacks. I'm quite sure that "advanced users" will find
their way to disable the write protection and create an u-boot env
config to play with it.

pepe2k:
> I might miss the use case here. But I don't see that a casual user has any need to manipulate the u-boot environment. Rather than that, we should try to protect users from shooting them self into the foot.

For me this sounds like an assumption that all users don't know what they are doing and we should protect them from themselves. R/W access to U-Boot environment is just another feature which we should support by default - it's part of device configuration.

> At least for this board is literally no need to manipulate the env. It is neither required to fix a mac address or to have write access for any install hacks.

It's not only about some hacks or settings broken by vendor/design which need to be fixed, it's about giving users full control of the device. We already write-protect U-Boot and radio calibration data partitions (like ART on QC/A based devices), which is... lets call it "generally accepted rule". But on the other hand, sysupgrade -F ... allows to write any image even if it's dangerous and may lead to a bricked device.

So, maybe we should inform user that manipulating U-Boot env variables is dangerous, like we do with sysupgrade -F... and password-less SSH root access, instead of disabling this feature by default?

> I'm quite sure that "advanced users" will find their way to disable the write protection and create an u-boot env config to play with it.

Why not make it easy for everybody by default? IMHO, fully support for any particular device should also include access to U-Boot environment if it's available. I agree that it might be dangerous if someone doesn't know how to use it, but hey... changing firmware on the device is a synonym of "dangerous".

bobafetthotmail:
I also prefer having access to bootloader settings by default. The fact that no user really needs to access these settings also means that the risk of someone screwing them up by mistake is slim.

mkresin:
I guess we can stop here discussing the pros and cons of having access
to bootloader settings by default.

I will not merge the PR for the outlined reasons. But I do not
close/reject the ticket either. Other devs might have other opinions
on this.

blogic:
i agree with mkresin. imho all boards that do not have a requirement for ubootenc access should have this as a ro partition. for this reason i wont be the one to merge this either.

mkresin:
It's quite obvious that none of the other developers is interested in merging this. I'll close this now.

PR #553, opened 2016-11-22
libs: libnetfilter-queue: update to a newer version in git repo

commodo:
Last release of libnetfilter-queue was in 2012.
There don't seem to be any release tarballs since then.

This updates it to a more recent version, pointing to the git repo.

### DIFF
--- a/package/libs/libnetfilter-queue/Makefile
+++ b/package/libs/libnetfilter-queue/Makefile
@@ -8,14 +8,14 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libnetfilter_queue
-PKG_VERSION:=1.0.2
+PKG_VERSION:=2016-07-03
 PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
-PKG_SOURCE_URL:= \
-	http://www.netfilter.org/projects/libnetfilter_queue/files/ \
-	ftp://ftp.netfilter.org/pub/libnetfilter_queue/
-PKG_MD5SUM:=df09befac35cb215865b39a36c96a3fa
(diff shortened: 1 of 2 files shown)


package/libs/libnetfilter-queue/Makefile
```diff
@@ -9,13 +9,13 @@ include $(TOPDIR)/rules.mk
...
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
-PKG_SOURCE_URL:= \
-	http://www.netfilter.org/projects/libnetfilter_queue/files/ \
-	ftp://ftp.netfilter.org/pub/libnetfilter_queue/
-PKG_MD5SUM:=df09befac35cb215865b39a36c96a3fa
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=git://git.netfilter.org/libnetfilter_queue
```
diizzyy: You should use https:// instead here

diizzyy: Ahh, apparently you need a pretty new certificate repo for let's encrypt domains...
Oh well, http it is then.... still preferred over git.

package/libs/libnetfilter-queue/Makefile
```diff
@@ -9,13 +9,13 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libnetfilter_queue
 PKG_VERSION:=1.0.2
```
blogic: if you point it at  arev newer than 1.0.2 then the version string needs to change

commodo:
re-spinned ; added date as PKG_VERSION

nbd168:

PR #554, opened 2016-11-22
uboot-envtools: uboot-envtools have some stuff related to the cpu,  m…

dpeddi:
…ake  it not shared

This should fix following error on lantiq soc:

Downloading http://downloads.lede-project.org/snapshots/packages/mips_24kc/base/uboot-envtools_2015.10-1_mips_24kc.ipk.
Configuring uboot-envtools.
//usr/lib/opkg/info/uboot-envtools.postinst: .: line 10: can't open '/lib/ar71xx.sh'


mkresin:
merged, thanks!

PR #555, opened 2016-11-22
lantiq-vgv7519: write protect u-boot and u-boot-env partitions

dpeddi:


mkresin:
merged, thanks!

PR #556, opened 2016-11-22
ramips: RB750Gr3: Add power LED and buzzer to dts & cleanup dts

ndoo:
Expose unused pinmux pins as GPIOs, then export power LED and buzzer pins; configure switch port mapping as WLLLL.

### DIFF
--- a/target/linux/ramips/dts/RB750Gr3.dts
+++ b/target/linux/ramips/dts/RB750Gr3.dts
@@ -20,6 +20,11 @@
 	gpio-leds {
 		compatible = "gpio-leds";
 
+		pwr {
+			label = "rb750gr3:blue:pwr";
+			gpios = <&gpio0 16 GPIO_ACTIVE_HIGH>;
+		};
+
 		usr {
 			label = "rb750gr3:green:usr";
 			gpios = <&gpio0 0 GPIO_ACTIVE_HIGH>;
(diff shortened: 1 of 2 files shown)


target/linux/ramips/dts/RB750Gr3.dts
```diff
@@ -20,6 +20,12 @@
 	gpio-leds {
 		compatible = "gpio-leds";
 
+		pwr {
```
mkresin: please use this LED in target/linux/ramips/base-files/etc/diag.sh and remove the default-state = "on". If the led is enabled by the bootloader, use default-state = "keep".

target/linux/ramips/dts/RB750Gr3.dts
```diff
@@ -97,12 +109,13 @@
 &ethernet {
 	mtd-mac-address = <&factory 0xe000>;
 	mtd-mac-address-increment = <1>;
+	ralink,port-map = "wllll";
```
mkresin: Does this make the rb750gr3 setting in target/linux/ramips/base-files/etc/board.d/02_network obsolete? Means, are the vlans set by driver according to the portmap. Please test this with an clean image. Do not preserve the config when doing an sysupgrade.

ndoo:
Requested changes made

PR #558, opened 2016-11-23
ar71xx: move DomyWifi DW33D to nand subtarget

pepe2k:
Support for this device was included in 7ba9a3a504909ccf8f6d1aca9b1160443da7215b, but it should never go to generic target as it uses NAND FLASH and requires related kernel modules.

Plus, small fix: kmod-ledtrig-usbdev -> kmod-usb-ledtrig-usbport.


mkresin:
merged, thanks!

PR #559, opened 2016-11-23
ramips: use new usbport LED trigger for recently added devices

pepe2k:
Just a trivial fix.


mkresin:
I've just dropped the kmod-ledtrig-usbdev instead with ee6beb2e1cc735cbcb9e985f7ab18cad6fc218b8. The boards do not setup usb leds at all.

PR #563, opened 2016-11-26
ramips: RB750Gr3: Add pwr LED and buzzer to DTS, use pwr LED in diag.sh

ndoo:
Expose unused pinmux pins as GPIOs, then export power LED and buzzer pins
Use rb750gr3:blue:pwr LED in diag.sh for boot status instead of rb750gr3:green:usr

### DIFF
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -191,7 +191,7 @@ get_status_led() {
 		status_led="$board:green:wps"
 		;;
 	rb750gr3)
-		status_led="$board:green:usr"
+		status_led="$board:blue:pwr"
 		;;
 	sap-g3200u3)
 		status_led="$board:green:usb"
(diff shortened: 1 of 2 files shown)


target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -191,7 +191,7 @@ get_status_led() {
 		status_led="$board:green:wps"
 		;;
 	rb750gr3)
-		status_led="$board:green:usr"
+		status_led="$board:blue:pwr"
 		;;
```
blogic: please dont drop the vocals. use power instead

ndoo: Hi @blogic, the router explicitly labels it as "PWR" on the sticker, would it be preferred to use the label or call it "power"?

blogic: ah ok, in that case its fine

blogic:
merged, Thanks !

PR #566, opened 2016-11-28
docs: deleting docs because they are obsolete

bobafetthotmail:
the docs in /docs folder are pretty much obsolete and in a not very friendly format (latex, that requires to be "compiled"), leaving them here only causes confusion.
LEDE documentation's place is the wiki, or the site.

There was a discussion about them recently here https://forum.lede-project.org/t/latex-dev-docs-are-these-being-maintained/337/4
@jow- said: "I think we should nuke these docs"
 
If there is something you want to keep I will move it to the wiki.


diizzyy:
If everything is in the wiki I don't see the need for keeping it.

blogic:
merged, Thanks !

PR #567, opened 2016-11-29
ar8xxx switch driver fixes

dissent1:
@nbd168 I've ported some patches from chromium and qsdk repositories for ar8xxx switch driver.
Think it could be of any interest for you.

Run-tested on Netgear R7800 (ipq806x).


blogic:
merged, Thanks !

PR #568, opened 2016-11-29
dropbear: fix quoting of default path in makefile

NebuKatWork:
In 4d1c75c, i overlooked the incorrect quoting of the SED command as it was protected by #ifndefs (due to -DDEFAULT_PATH) so it doesn’t get triggered …


jow-:
@NebuKatWork - the used variable is the correct one but there are indeed some issues with the default path handling.

The `#define DEFAULT_PATH` is indeed improperly substituted but this problem usually does not surface because we're adding an explicit `-DDEFAULT_PATH` define to the compile flags.

Additionally the guards around the `#define DEFAULT_PATH` are introduced by one of our local patches which does not make much sense to me. In order to take the opportunity to get rid of a patch here I propse the following:

 1) drop `patches/500-set-default-path.patch`
 2) drop `TARGET_CFLAGS += -DDEFAULT_PATH=...`
 3) fix the sed substitution in `options.h`

But keep using `TARGET_INIT_PATH` instead of `CONFIG_TARGET_INIT_PATH`.

blogic:
merged, Thanks !

PR #570, opened 2016-11-29
uqmi: verify pin only when needed

obsy:
Pin should be verified only when SIM required this. Now if pin is set in config and SIM not required pin, connection fail.


blogic:
not sure about this one. if the user sets a pin and none is required this could indicate that he has the wrong sim in the slot. the config provided by the user should always be honoured. if a pin is givven but none is required then the config is bad and the software should not try to be smart and guess what is wrong. it should simply bog out and fail.

malaakso:
I've thought about the same thing earlier, and came to the same conclusion as @blogic on this one.

blogic:
@nbd168 @jow- i would tend to NAK this. could you tell is your thoughts please ?

PR #572, opened 2016-11-30
ramips: fix Lenovo Y1/Y1s led in init script

hbl0307106015:
This commit fix the y1 & y1s leds' definition.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -335,11 +335,18 @@ wndr3700v5)
 wt3020)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "0"
 	;;
-y1|\
-y1s)
-	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
+y1)
 	set_usb_led "$board:blue:usb"
-	set_wifi_led "$board:blue:wifi"
+	ucidef_set_led_netdev "wifi" "WIFI" "$board:blue:wifi" "wlan1"
+	ucidef_set_led_netdev "wifi5g" "WIFI5G" "$board:blue:wifi5g" "wlan0"
+	ucidef_set_led_netdev "lan" "LAN" "$board:blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_netdev "wan" "WAN" "$board:blue:internet" "eth0.2" "tx rx"
+	;;
+y1s)
+	set_usb_led "$board:blue:usb"
+	ucidef_set_led_netdev "wifi" "WIFI" "$board:yellow:wifi" "wlan1"
+	ucidef_set_led_netdev "wifi5g" "WIFI5G" "$board:blue:wifi" "wlan0"
+	ucidef_set_led_netdev "wan" "WAN" "$board:blue:internet" "eth0.2" "tx rx"
 	;;
 zbt-ape522ii)
 	ucidef_set_led_netdev "wlan2g4" "wlan1-link" "$board:green:wlan2g4" "wlan1"


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -335,11 +335,19 @@ wndr3700v5)
 wt3020)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "0"
 	;;
-y1|\
+y1)
+	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
```
mkresin: This shouldn't be required. The LED is already handled by target/linux/ramips/base-files/etc/diag.sh.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -335,11 +335,19 @@ wndr3700v5)
...
+	set_usb_led "$board:blue:usb"
+	set_wifi_led "$board:blue:wifi" "wlan1"
+	ucidef_set_led_netdev "wifi5g" "wifi5g" "$board:blue:wifi5g" "wlan0" "link tx rx"
+	ucidef_set_led_netdev "lan" "lan" "$board:blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_netdev "wan" "wan" "$board:blue:internet" "eth0.2" "tx rx"
+	;;
 y1s)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
```
mkresin: This shouldn't be required. The LED is already handled by target/linux/ramips/base-files/etc/diag.sh.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -335,11 +335,19 @@ wndr3700v5)
 wt3020)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "0"
 	;;
-y1|\
+y1)
+	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
+	set_usb_led "$board:blue:usb"
+	set_wifi_led "$board:blue:wifi" "wlan1"
```
mkresin: use ucidef_set_led_netdev for the 2.4 GHZ as well. Don't different functions for 2.4 and 5 GHz wifi. It is confusing.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -335,11 +335,19 @@ wndr3700v5)
...
+	ucidef_set_led_netdev "wifi5g" "wifi5g" "$board:blue:wifi5g" "wlan0" "link tx rx"
```
mkresin: drop the "link tx rx" part. It's the default if nothing is specified.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -335,11 +335,19 @@ wndr3700v5)
...
+	ucidef_set_led_netdev "lan" "lan" "$board:blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_netdev "wan" "wan" "$board:blue:internet" "eth0.2" "tx rx"
```
mkresin: any reason why the leds shouldn't show the link state?

mkresin:
Please use the full board name in the commit title.

Please add a commit message to your commit. It should describe why you have done something.

target/linux/ramips/base-files/etc/board.d/01_leds
hbl0307106015: For the LAN and WAN port, many of our customers think the LED should turn off after they pull out the cable."tx rx" state do the job and this really avoid confusing them.

target/linux/ramips/base-files/etc/board.d/01_leds
hbl0307106015: ok, I will use ucidef_set_led_netdev instead

target/linux/ramips/base-files/etc/board.d/01_leds
mkresin: Well, that sounds exactly what the ```link``` trigger should do.

hbl0307106015:
@mkresin  The PR has been changed, please check it again. Thanks!

target/linux/ramips/base-files/etc/board.d/01_leds
mkresin: Gave it a quick test on a VGV7510KW22 and it works for me with the link parameter.

If I unplug the network cable:

```
[  213.263560] eth0: port 3 lost link
[  213.266285] br-lan: port 1(eth0.1) entered disabled state
```

And the configured LED is switched off.

There must be something else wrong on the Y1/Y1s.

blogic:
merged, Thanks !

PR #573, opened 2016-11-30
ramips: fix PBR-D1 dts

hbl0307106015:
Fix pbr-d1 dts for button definition, it should be reset, not wps.

### DIFF
--- a/target/linux/ramips/dts/PBR-D1.dts
+++ b/target/linux/ramips/dts/PBR-D1.dts
@@ -43,7 +43,7 @@
 		#size-cells = <0>;
 		poll-interval = <20>;
 
-		wps {
+		reset {
 			label = "reset";
 			gpios = <&gpio1 38 1>;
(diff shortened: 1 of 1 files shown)


target/linux/ramips/dts/PBR-D1.dts
```diff
@@ -43,7 +43,7 @@
 		#size-cells = <0>;
 		poll-interval = <20>;
 
-		wps {
+		reset {
```
mkresin: Albeit the change of the node name is something pure cosmetical, the whole node still looks wrong.

If it's a reset button, the ```linux,code``` should be ```KEY_RESTART```.

If it's a WPS Button, the ```label``` should be ```wps``` or similar (depending on how it is named in the manual or what is printed on the case).

mkresin:
Please add a commit message to your commit. It should describe why you have done something.

blogic:
NAK, this PR is just plain broken. please reopen when you addressed the changes that mkresin requested

PR #574, opened 2016-11-30
ramips : tweak set_wifi_led and set_usb_led functions

hbl0307106015:
Since many routers may have two or more usb/wifi leds,
    the original set_wifi_led() and set_usb_led() can not meet
    our requirement, the commit can make it easay to define
    usb and wifi leds as below shows:

    set_usb_led "usb1" "$board:color:usb1" "1-1"
    set_usb_led "usb2" "$board:color:usb2" "2-1"

    and

    set_wifi_led "wifi_2ghz" "$board:color:wifi" "wlan0"
    set_wifi_led "wifi_5ghz" "$board:color:wifi5g" "wlan1"

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -4,11 +4,11 @@
 . /lib/ramips.sh
 
 set_usb_led() {
-	ucidef_set_led_usbdev "usb" "USB" "${1}" "${2:-1-1}"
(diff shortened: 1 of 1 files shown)


mkresin:
No really sure if we should do this. I guess these functions were meant for exactly these simple use cases where only one wifi and only one usb led exists.

Personally, I would prefer to use the "normal" uci helper functions in case there are more complex setups instead of extending the wrapper functions.

Any other opinions?

blogic:
NAK, call ucidef_set_led_usbdev directly if the mini wrapper is not sufficient.

PR #575, opened 2016-11-30
dnsmasq: reload config if host name is modified

feckert:
If the hostname in /etc/config/system is modified the dnsmasq will not
reread the update host file under /tmp/hosts/dhcp.$cfg.

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -785,7 +785,7 @@ dnsmasq_stop()
 
 service_triggers()
 {
-	procd_add_reload_trigger "dhcp"
+	procd_add_reload_trigger "dhcp" "system"
 	procd_add_raw_trigger "interface.*" 2000 /etc/init.d/dnsmasq reload
 }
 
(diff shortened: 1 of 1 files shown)


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -825,6 +825,7 @@ start_service() {
 
 reload_service() {
 	rc_procd start_service "$@"
+	killall -HUP dnsmasq
```
blogic: is this really required ? would the line above not restart the service ?

feckert:
@blogic 
I have tested it on my router. And the PID of dnsmasq Process is the same as before. This Signal only reloads some files see http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

> NOTES
> When it receives a SIGHUP, dnsmasq clears its cache and then re-loads /etc/hosts and /etc/ethers and any file given by --dhcp-hostsfile, --dhcp-hostsdir, --dhcp-optsfile, --dhcp-optsdir, --addn-hosts or --hostsdir. The dhcp lease change script is called for all existing DHCP leases. If --no-poll is set SIGHUP also re-reads /etc/resolv.conf. SIGHUP does NOT re-read the configuration file.

blogic:
merged, Tanks !

PR #577, opened 2016-11-30
Improve F5D8235 V1 support

NotTheEvilOne:
This PR contains all changes needed to get the device configured correctly in LEDE.


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -160,7 +160,9 @@ ex2700)
 	;;
 f5d8235-v1|\
 f5d8235-v2)
+	ucidef_set_led_default "power" "POWER" "$board:blue:router" "1"
```
mkresin: I see no reason to enable this LED by default. Please drop it.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -160,7 +160,9 @@ ex2700)
...
 	set_usb_led "$board:blue:storage"
+	set_wifi_led "$board:blue:wireless"
```
mkresin: The f5d8235-v2 has no blue wireless LED. Please split the f5d8235-v1 and f5d8235-v2 led config.

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -123,8 +123,9 @@ get_status_led() {
 	pbr-d1)
 		status_led="$board:orange:power"
 		;;
+	f5d8235-v1|\
 	f5d8235-v2)
-		status_led="$board:blue:router"
```
mkresin: Why is is the ":blue:router" behaviour unpredictable for failsafe indication? I don't get what you are trying to tell us.

The f5d8235-v2 has no blue wired LED. If this change is really required, please split the f5d8235-v1 and f5d8235-v2 led config.

target/linux/ramips/dts/F5D8235_V1.dts
```diff
@@ -68,15 +68,70 @@
 	gpio-leds {
 		compatible = "gpio-leds";
 
+		internet {
+			label = "f5d8235-v1:blue:internet";
+			gpios = <&gpio0 17 1>;
```
mkresin: Please include the dt-bindings/gpio/gpio.h header and use the GPIO_ACTIVE_LOW/GPIO_ACTIVE_HIGH macros. For reference: https://git.lede-project.org/a7de18718d4853e5387f51651573c79bc8a81169

target/linux/ramips/base-files/etc/board.d/01_leds
NotTheEvilOne: Please see comment below for diag.sh.

target/linux/ramips/base-files/etc/diag.sh
NotTheEvilOne: Issue with the blue:router is how it is controlled by GPIO. You can select one of two states as far as I can tell. It either blinks (low) or is on (high). Blinking does of course not look nice and it does not match how failsafe selection is indicated. As a result it looks wired.

Both v2 and v3 have the same LEDs as v1 but the DTS file for v2 is incomplete. I will split the configuration but would strongly recommend to not change the current selected indication of blue:router (see above) and blue:wired for failsafe indication as described.

mkresin: Man, you told me about it and I forgot it again :-(. Please add this detailed explanation to the commit message.

target/linux/ramips/base-files/etc/board.d/01_leds
mkresin: Mhh, it still don't explain why this LED need to be enabled after boot. To only reason I meant to understood so far, is because it looks nice. Just drop it!

I guess it would be a good idea to ucidef_set_led_default the blue:wired to 0. It should be only used temporary as a workaround for a missing power led. The same is done on other targets/boards in case there is no power led or the power led is not controllable via gpio.

NotTheEvilOne: Just to reconfirm. Dropping this line would mean that the router LED will blink indefinitely no matter in which state the device is.

The change to blue:wired sounds good even though it means that I need to split V1 and V2 again as the blue:wired is missing there as I learned minutes before ;)

blogic:
merged, Thanks !

blogic:
not merged, patch fails to apply to trunk

NotTheEvilOne:
As requested by @blogic I cherry picked the PCI/USB changes from #578 additionally to this PR as they change the DTS file again. Furthermore I merged all changes from lede/source to resolve the previous conflict issue.

mkresin:
Sorry for the confusion! It was a lack of communication between blogic and me.

mkresin:
merged into my staging tree. thanks!

PR #579, opened 2016-11-30
Extend rlt8366s to fix issues with only one port being usable

NotTheEvilOne:
Before this change only port 4 of F5D8235 V1 worked at 1000Mpbs. Comparing the current driver with the GPL_BELKIN_F5D8235-4_v1000 v1.01.24 sources showed that additional steps are required to set-up the rlt8366s switch correctly.


mkresin:
Hey Tobias,

Your commit prefix is wrong. You are not changing a ramips specific file.

Would you please split your commit to at least two commits. I'm not that familiar with switch drivers but you are changing at least two things within the driver (auto negotiation and green mode).

Would you please have a look at #537. Vitaly adds with this PR the green mode to the rtl8367b as well but additionally he provides an interface to enable/disable the green mode instead of enforcing a fixed value. The discussion how it should be done the best way is still ongoing.

As you might already noticed, I've problems to get what you are changing here (RTL8366S_PHY_* for example). Would you please extend your commit message(s) to describe in detail what you are changing and why this change is required. Only describing the old and the new behaviour is to high level for a switch driver (taking into account that the driver is used by several other boards and the change might cause problems on these).

mkresin:
I had a more closer look at the RTL8366S code. I'm far away from understanding how the switch really works but what I got so far is that the switch has an ASIC which is kind of controllable by the driver.

After comparing your chipDefault values with the one I found in the [vendor driver](https://github.com/mirror/dd-wrt/blob/master/src/linux/universal/linux-3.5/drivers/net/ethernet/ag7100/RTL8366S_DRIVER/rtl8366s_api.c#L1815), I'm really concerned about possible side effects on other boards.

The vendor driver writes two different chip default arrays to the ASIC, depending on an value read from the ASIC. With your patch you are not doing this distinction. I found no documentation for the initvals and therefore I'm not sure how board specific these values are. As far as I can see, no board is using the capability to provide the initvals via platform data.

Are these GREEN_ETHERNET, PHY_POWER_SAVING and chipDefault values really required to fix your issue?

---

The changes to the RTL8366S_PHY register are at least suspicious.

To my reading of the the Datasheet, bit 9 of the RTL8366S_PHY_1000BASET_CTRL register is 1 (Advertise PHY is 1000BASE-T full duplex capable) by default.

The same applies to the RTL8366S_PHY_AN_CTRL register. All values you are writing to these registers are the default values:

bit5: 1 (10Base-TX half duplex capable)
bit6: 1 (10Base-TX full duplex capable)
bit7: 1 (100Base-TX half duplex capable)
bit8: 1 (100Base-TX full duplex capable)
bit10: 1 (Advertises that the RTL8366S/SR has flow control capability)
bit11: 1 (Advertises that the RTL8366S/SR has asymmetric flow control capability.)

Finally the RTL8366S_PHY_CTRL register. The only change you are doing is to "Restart Auto-Negotiation process" by setting bit9 to 1. All other values are the default ones.

My first guess is that the bootloader writes bogus values to the chip register and that is the root cause for your issue. The switch chip has a reset pin. In best case this pin is connected to a GPIO line of the SoC and you can do a power-on reset to reset the register values to the default ones.

Bootloaders on other boards might change the default register values for a reason. Not sure if overwriting any set register values causes issues for these boards.

NotTheEvilOne:
Hi,

Thanks for the time reviewing so far. I'll try to concentrate on your concerns first and will change the PR later if an agreement has been found.

First about the chipDefault values. I first had the comparison included but later found out that a check is already done at [rtl8366_smi.c](https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/generic/files/drivers/net/phy/rtl8366_smi.c#l1292).

The GREEN_ETHERNET and PHY_POWER_SAVING changes are not needed to fix the switch for the board but an additional positive side effect. The chipDefault is crucial to fix the issue encountered.

Same is for the reset to default values (like RTL8366S_PHY_1000BASET_CTRL and others). As the issues are seen at u-boot as well I agree with your that u-boot somehow partly initializes the switch but leaves it in a bad state.

A more suitable solution would be to activate the platform specific initvals or use a GPIO reset pin. Unfortunately as far as I understood the code initvals come from mtd by default and that's not the case for this board as well. I have no idea how to even identify the GPIO reset pin even if it is connected and programmatically available.

If you or anyone else has a idea how to implement a board specific solution to get initvals reset I would highly appreciate to hear from it. I guess the initialization part with default values (like RTL8366S_PHY_1000BASET_CTRL) shouldn't create many issues as they are the defaults anyway but the chipDefault ones are from the board. That's why I added comments underlining that as I have not found any switch vendor specific documentation for it as well.

mkresin:
> Thanks for the time reviewing so far. I'll try to concentrate on your
> concerns first and will change the PR later if an agreement has been found.

Thats the way to go. No need to change code just because I had a brainfart.

> First about the chipDefault values. I first had the comparison included
> but later found out that a check is already done at rtl8366_smi.c
> <https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/generic/files/drivers/net/phy/rtl8366_smi.c#l1292>.

You are right.

> The GREEN_ETHERNET and PHY_POWER_SAVING changes are not needed to fix
> the switch for the board but an additional positive side effect.

I'm totally with you. Green ethernet and power saving sounds like a good 
thing to do.

What about introducing a new device tree property 
realtek,enable-powersaving and enable powersaving depending on the 
presence of this parameter? Or maybe realtek,disable-powersaving?

As already said, not sure if it's a good idea to enable this feature by 
default. Personally I prefer to enable the powersaving features if 
requested via device tree. Just to be on the safe side.

> The chipDefault is crucial to fix the issue encountered.
>
> Same is for the reset to default values (like
> RTL8366S_PHY_1000BASET_CTRL and others). As the issues are seen at
> u-boot as well I agree with your that u-boot somehow partly initializes
> the switch but leaves it in a bad state.

Thanks for clarifying what is required and what is nice to have.

> A more suitable solution would be to activate the platform specific
> initvals or use a GPIO reset pin. Unfortunately as far as I understood
> the code initvals come from mtd by default and that's not the case for
> this board as well. I have no idea how to even identify the GPIO reset
> pin even if it is connected and programmatically available.

You could add a device tree parameter for the initvals similar to 
qca,ar8327-initvals.

The ar8327-initvals have a different meaning (Link speed, CPU port, led 
mode etc.) but I don't see a reason why not doing it the same way with 
your magic values.

> If you or anyone else has a idea how to implement a board specific
> solution to get initvals reset I would highly appreciate to hear from
> it.

According to the datasheet, the RTL8366S/SR supports two software 
resets; a chip reset and a soft reset. When CHIP_RESET is set to 0b1 
(write and self clear), the chip will take the following steps:

4. Reset all registers to default values.
5. Restart the auto-negotiation process.

Seams to me it's what we are looking for. Similar to the powersaving 
features guarded by a device tree property and it should do the job.

> I guess the initialization part with default values (like
> RTL8366S_PHY_1000BASET_CTRL) shouldn't create many issues as they are
> the defaults anyway but the chipDefault ones are from the board. That's
> why I added comments underlining that as I have not found any switch
> vendor specific documentation for it as well.

I wouldn't bet that all boards are fine with setting default values like 
RTL8366S_PHY_1000BASET_CTRL. I've already seen so much unexpected 
breakage...

mkresin:
>>  First about the chipDefault values. I first had the comparison included but later found out that a check is already done at rtl8366_smi.c <https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/generic/files/drivers/net/phy/rtl8366_smi.c#l1292>.

> You are right.

Either I'm to dumb to read the code or you are wrong. The  function ```rtl8366_smi_detect()``` is only used in [target/linux/ar71xx/files/arch/mips/ath79/mach-wzr-hp-g300nh.c](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-wzr-hp-g300nh.c#L233). In your case the function will be never called and the problem persists.

mkresin:
Regarding the initvals. Would you please check which of the initvals are really required, by comparing the default register values and the one you are set.

I had a look at some other board files and they usually set only some values like in https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-dir-825-b1.c#L98

mkresin:
@NotTheEvilOne Hey Tobias, do you plan to work further on this PR? Have I requested to much changes or something that can not be implemented?

NotTheEvilOne:
Hi Mathias,

I just rewrote the PR to contain two individual changes.

1. Adds DTS initvals and power saving features. As I wasn't able to find a approach how to iterate over all PHY ports to call the set_power_green as suggested by the PR you mentioned it's still done in "rtl8366s_setup". The soft reset after applying initvals makes sure that we can get rid of resetting all ports individually.

2. This one adds the magic values found in the Belkin sources for this device. I did a comparision of the values being set and the ones read by "rtl8366_smi_read_reg()". Surprisingly all beside one are different. We either just leave it as it is or I need to check what values are predefined and if they are all zeros for example. It's strange ...

Let me know what you think about the changes so far.

mkresin:
Merged with minor changes to the commit messages into my staging tree. Thanks!

PR #580, opened 2016-11-30
Fix kernel 4.4 issues preventing F5D8235 V1 from booting

NotTheEvilOne:
Trying to get a F5D8235 V1 router into a working condition showed two bugs in kernel 4.4+.

First a hang can be observed if no valid DTB is found and of_scan_flat_dt() is called to find early memory. Reason for that is that initial_boot_params is not checked in this function before trying to read from it.

Additionally the rt288x suffered from a bug introduced during the development of 4.4 (and is present up to 4.9-rc7 and most likely newer versions). As a consequence the kernel basically panics after hitting "BUG: Bad page state in process swapper". This issue is also described in a [LEDE Flyspray issue report](https://bugs.lede-project.org/index.php?do=details&task_id=244).

### DIFF
--- a/target/linux/generic/patches-4.4/996-mm-fix_alloc_node_mem_map_calculation.patch
+++ b/target/linux/generic/patches-4.4/996-mm-fix_alloc_node_mem_map_calculation.patch
@@ -1,0 +1,11 @@
+--- a/mm/page_alloc.c
++++ b/mm/page_alloc.c
+@@ -5326,7 +5326,7 @@
+ 		mem_map = NODE_DATA(0)->node_mem_map;
+ #if defined(CONFIG_HAVE_MEMBLOCK_NODE_MAP) || defined(CONFIG_FLATMEM)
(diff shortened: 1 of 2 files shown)


nbd168:
Please send those kernel changes upstream, and we will cherry-pick them into our tree based on the upstream submission

NotTheEvilOne:
They are already submitted to the linux-mips mailing list and patchwork [1] [2]. As I expect no fast integration I hope that LEDE will use them already before as otherwise the named router will stuck broken for a longer period of time than needed.

[1] [generic: Add check to of_scan_flat_dt() before accessing initial_boot_params](https://patchwork.linux-mips.org/patch/14605/)
[2] [generic: Fix alloc_node_mem_map with ARCH_PFN_OFFSET calculation](https://patchwork.linux-mips.org/patch/14627/)

nbd168:
Imported the commits from patchwork with full description and merged them to my staging tree at https://git.lede-project.org/?p=lede/nbd/staging.git;a=summary

Thanks

PR #582, opened 2016-11-30
Introduce CONFIG_MIPS_APPENDED_RAW_DTB to arch/mips/ralink

NotTheEvilOne:
With this PR both the OWRTDTB approach but also upstream CONFIG_MIPS_APPENDED_RAW_DTB can be used to provide a board specific DTB to the kernel.


blogic:
this needs to be a single patch. breaking support in the first patch and then fixing it in the 2nd is a no-go

also the first patch only makes sense if you also make the build system use that feature.

NotTheEvilOne:
Changed as requested. FYI the original (breaking) patch came from my [upstream patch](https://patchwork.linux-mips.org/patch/14589/). That's why I needed to reintroduce the OWRTDTB hack later on once again.

blogic:
one change per patch please. the board changes need to go into a second patch

target/linux/ramips/image/rt288x.mk
```diff
@@ -33,10 +33,13 @@ endef
 TARGET_DEVICES += ar725w
 
 define Device/f5d8235v1
-  DTS := F5D8235_V1
   IMAGE_SIZE := 7744k
   DEVICE_TITLE := Belkin F5D8235 V1
   DEVICE_PACKAGES := kmod-switch-rtl8366s
+  DEVICE_DTS := F5D8235_V1
+  DEVICE_DTS_DIR := ../dts
```
blogic: i think this line is not required

target/linux/ramips/rt288x/config-4.4
```diff
@@ -119,8 +119,8 @@ CONFIG_MIPS_CMDLINE_FROM_DTB=y
 CONFIG_MIPS_L1_CACHE_SHIFT=4
 CONFIG_MIPS_L1_CACHE_SHIFT_4=y
 # CONFIG_MIPS_MACHINE is not set
-CONFIG_MIPS_NO_APPENDED_DTB=y
-# CONFIG_MIPS_RAW_APPENDED_DTB is not set
+# CONFIG_MIPS_NO_APPENDED_DTB is not set
+CONFIG_MIPS_RAW_APPENDED_DTB=y
```
blogic: thi should be enabled when all boards are converted to using this. i do not want some boards doing it one way and other another way. either we append or use the OWRTDTB hack but now both.

NotTheEvilOne: As far as I have seen it does not create any issues to use both concurrently. Removing "CONFIG_MIPS_RAW_APPENDED_DTB=y" results in no DTB to be found if append-dtb and DEVICE_DTS is configured. I'm not aware if it is possible to set kernel configuration settings per board.

target/linux/ramips/image/rt288x.mk
NotTheEvilOne: While testing a vanilla bug I have seen various issues getting CONFIG_MIPS_APPENDED_RAW_DTB working at all. Thanks to @mkresin to help me patch the Makefiles for it.

In fact IMHO both lines are required:
Only DEVICE_DTS is used in "Build/append-dtb" of include/image-commands.mk.
If DEVICE_DTS_DIR is not specified the default DTS_DIR pointing to the kernel is tried (which fails). Maybe I missed something where DEVICE_DTS is set to the value of DTS?

NotTheEvilOne:
Ok, I splitted the patch from the board changes. Please let me know how to handle DEVICE_DTS and the kernel configuration.

target/linux/ramips/patches-4.4/997-ralink-Introduce-fw_passed_dtb-to-arch-mips-ralink.patch
```diff
@@ -0,0 +1,27 @@
...
++	void *dtb;
++
+ 	set_io_port_base(KSEG1);
+ 
+ 	/*
+ 	 * Load the builtin devicetree. This causes the chosen node to be
+ 	 * parsed resulting in our memory appearing
+ 	 */
+-	__dt_setup_arch(&__image_dtb);
++	if (fw_passed_dtb) /* used by CONFIG_MIPS_APPENDED_RAW_DTB as well */
```
blogic: kernel coding style does not allow comments at the end of lines

target/linux/ramips/image/rt288x.mk
blogic: move it to the generic template in that case

target/linux/ramips/patches-4.4/997-ralink-Introduce-fw_passed_dtb-to-arch-mips-ralink.patch
```diff
@@ -0,0 +1,28 @@
...
+-	__dt_setup_arch(&__image_dtb);
++	if (fw_passed_dtb)
++		dtb = (void *)fw_passed_dtb;
++	else if (__dtb_start != __dtb_end)
++		dtb = (void *)__dtb_start;
++	else
++		dtb = &__image_dtb;
++
++	if (dtb)
++		__dt_setup_arch(dtb);
```
blogic: patch is starting to look good, last comment is ... why is there an if in line 24 ? dtb should not be null and if it is __dt_setup_arch() will panic which is expected so i think it is safe to drop this if statement.

blogic: ok, in that case please just do a if (!dtb) panic("missing dtb");

NotTheEvilOne: I guess that change would break at least one platform because as can be seen at [this](http://lxr.free-electrons.com/source/arch/mips/ralink/of.c?v=4.4#L78) source code line even ralink is intended to default to two different variants of other information from soc_info to initiate even without a DTB which worked for me for rt288x as well after fixing the above linked bug.

For platforms really needing DTB to work (even ralink ones) a check is part of "__dt_register_buses()" [[1](http://lxr.free-electrons.com/source/arch/mips/kernel/prom.c?v=4.4#L64)] that panics.

Based on the above check in __dt_setup_arch() I guess we can leave out the if condition all together as this function should be able to handle a null value for dtb.

NotTheEvilOne:
Dear blogic,

I just merged the latest changes to resolve a previously created conflict. Please let me know what you think about the revised patch. I would send the revised variant to the linux-mips mailing list otherwise as well.

blogic:
@NotTheEvilOne looking good but you have a 3rd merge conflict patch in the branch that we cannot merge.

blogic:
@NotTheEvilOne do you plan to provide a clean PR with no merge commits inside it so that we can merge it ?

NotTheEvilOne:
Cleaned up for merging.

jow-:
Rebased and merged into my staging tree. @blogic - please tell me if the staged commits in my tree still need changes.

stintel:
Does this PR fix a bug (and do we need to cherry-pick it into the 17.01 branch)?

PR #583, opened 2016-12-01
ipq806x: refactor rpm clock controller patches

dissent1:
RPM clock controller driver had made its way upstream and previous
approach of directly redoing a driver to support ipq806x is a no go anymore.

Thus reverting mentioned patches to upstream state and renaming
in correct patch numbering accordance.

To make the driver work on ipq806x boards we introduce a custom patch.


blogic:
merged, Thanks !

PR #586, opened 2016-12-02
test

houzhenggang:


blogic:
if it is only a test then close it please. if you want us to merge this then please fold the changes into 1 single patch.

thagabe:
@blogic Close it, Their repo is down already

PR #587, opened 2016-12-02
    ramips : fix PBR-D1 button definition in dts

hbl0307106015:
Due to the product specification,
    the button on PBR-D1 should be reset, not wps.


mkresin:
merged into my staging tree. Thanks!

PR #588, opened 2016-12-04
ar71xx: add support for YunCore SR3200 and XD3200

pepe2k:
The following two changes add support for **YunCore SR3200** and **XD3200** (both are **QCA9563** based) under ar71xx target.


noilly:
Ah! This is exactly what I need! Any idea on when this will be merged?

noilly:
I was looking at the change and I noticed that it mentioned 

> 6x ext. RP-SMA antennas (actually, only 4 are connected with radio chips)

I was wondering if you could tell me what that means? Does that mean two of the antennas are useless?

pepe2k:
> I was wondering if you could tell me what that means? Does that mean two of the antennas are useless?

Yes, middle (at least in my piece) antennas on both sides are "fake".

The QCA9882 is 2T2R only. QCA9563 is 3T3R, but only two chains are used (SMD components on the third chain are not populated). Take a look at the PCB:



Cheers

noilly:
Ah, thank you! My experience with electrical engineering is basically non-existent. When you say the "third chain" what do you mean?

pepe2k:
> Ah, thank you! My experience with electrical engineering is basically non-existent. When you say the "third chain" what do you mean?

Take a look at the PCB area near green "QC" sticker - there is ready place for SMD components for 3rd radio chain (from QCA9563, on the left you can see antenna cables going out from bottom side of PCB), but they are not populated, so radio works in 2T2R (MIMO 2x2).

Maybe this will also help you to understand it: http://goo.gl/IjA2bB

Cheers

noilly:
Wonderful, thank you for explaining and for the link! If I understand correctly, the area to the right of the sticker should have the SMD components for the third radio chain?

pepe2k:
> If I understand correctly, the area to the right of the sticker should have the SMD components for the third radio chain?

Not in this model, everything is fine here - it's AC1200 rated/advertised (300 Mbps on 2,4 GHz and 867 Mbps on 5 GHz, 2 streams per each band).

Anyway, I don't think this is the right place for such discussion, please use forum if you have more questions.

Cheers

blogic:
merger into my staging tree, Thanks !

PR #591, opened 2016-12-06
ramips: specify rom chip modalias for DIR-610 A1

shyba:
This should make the flash stable.

Signed-off-by: Victor Shyba <victor1984@riseup.net>

### DIFF
--- a/target/linux/ramips/dts/DIR-610-A1.dts
+++ b/target/linux/ramips/dts/DIR-610-A1.dts
@@ -50,7 +50,7 @@
 		#size-cells = <1>;
 		compatible = "jedec,spi-nor";
 		reg = <0>;
-		linux,modalias = "m25p80";
+		linux,modalias = "m25p80", "f25l32pa";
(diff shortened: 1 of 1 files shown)


blogic:
@shyba did you test if this is really required on v4.4 using LEDE

shyba:
As soon as I reboot I am loosing dhcp and kind of half-bricking it, but I think it's unrelated. I will test again tonight (gmt-3) without any configuration to check if persistence is working, but as far as I could read spi-nor.c on sources, it's ok without the modalias. It will use the jedec ID to probe.
I am 90% sure this can be closed. Will do a last test and confirm.

shyba:
So, dmesg is fine, everything get's erased and working. The boot issue is unrelated. Somehow it's going back to Emergency Room after reboot. I will use a serial adapter to hear the cry and try to fix, making another PR if it's on LEDE, but I think it's misconfiguration on my side.

Thank you for this amazing work!

PR #595, opened 2016-12-09
usb-serial: adding DWM-158 3g Modem

fid0did0:
This patch add the 3G modem DWM-158 to the usb-serial option driver.
The DWM-158 is a pcie 3G modem. It is embedded in the DWR-512 modem supportet by lede.
The patch has been submitted and merged in the upstream linux-next repository.


blogic:
normally this would go into the generic patches folder but as its a specific modem putting it into ramips folder does make sense.

blogic:
no further comments on this so i will just carry it in the ramips tree.

@fid0did0 has it already been accepted upstream ?

blogic:
merged, Thanks !

PR #597, opened 2016-12-10
libs/libpcap: update to 1.8.1

p-wassi:
Maintainer: @nbd168 
Compile tested: kirkwood, brcm47xx, ar71xx
Run tested: kirkwood, brcm47xx

Description:
Update libpcap to upstream release 1.8.1
To note:
-) MAJ version number in Makefile.in (via patch 100-...) is changed to 1.8
-) Patch 100- is reloaded from Debian's current libpcap (plus minor changes)
-) Is patch 103- still needed (adding the define for _BSD_SOURCE)?
    The thing compiled fine on Debian without 103.
-) In patch 201 some functions were moved from one C file to another.
    Since all the object files are linked in anyway, can we leave them, where they were
    originally placed?
-) Added a short explanation for patch 202 (Emergency Access Daemon)
-) Imported new patch 204 from Debian changing the path for USB devices


wongsyrone:
According to the official README [file](https://github.com/the-tcpdump-group/libpcap/blob/master/README):

> It sets the soname of the library to "libpcap.so.1"; this is what it
should be, *NOT* libpcap.so.1.x or libpcap.so.1.x.y or something such as
that.

package/libs/libpcap/patches/100-debian_shared_lib.patch
```diff
@@ -79,25 +79,25 @@ build a shared library.
...
- 	$(CC) $(FULL_CFLAGS) -Dyylval=pcap_lval -c grammar.c
+ grammar.c: $(srcdir)/grammar.y
+@@ -472,9 +496,16 @@ grammar.o: grammar.c
+ gencode.o: $(srcdir)/gencode.c grammar.h scanner.h
+ 	$(CC) $(FULL_CFLAGS) -c $(srcdir)/gencode.c
  
 +grammar_pic.o: grammar.c
 +	@rm -f $@
-+	$(CC) -fPIC $(FULL_CFLAGS) -Dyylval=pcap_lval -o $@ -c grammar.c
++	$(CC) -fPIC $(FULL_CFLAGS) -Dyylval=pcap_lval -o $@ -c grammar.c
```
wongsyrone: trailing white space

p-wassi:
Thanks Syrone for reviewing the PR.
The trailing whitespace is removed (it came from Debian), and the variables for
naming the lib are reworked. The result now looks like this:
'libpcap.so.1' is the library, additionally there's only one symlink from 'libpcap.so' to the lib.
Compile and run tested on kirkwood.

wongsyrone:
I think you should describe what you have changed in commit message. Any ideas?

p-wassi:
Did you mean more overall explanation, like this? Or the diff between my v1 and v2?

wongsyrone:
LGTM.

nbd168:
Merged to my staging tree, thanks!

By the way, while testing it, I found a parallel build bug. Since I've never seen that one before, I assume it's a regression and folded my fix for it into the update commit.

PR #598, opened 2016-12-10
net/utils/tcpdump: update to 4.8.1

p-wassi:
Maintainer: @nbd168 
Compile tested (full and mini): kirkwood, brcm47xx, ar71xx
Run tested (full and mini): kirkwood, brcm47xx

Description:
Update tcpdump to upstream release 4.8.1
To note:
-) Mirrors:
    ftp.gwdg.de brings a 404, therefore removed from the list
    www.br.tcpdump.org is NXDOMAIN, therefore removed from the list
-) Removing parser debug functions (patch 001):
    Parts of the debugging stuff were rewritten upstream, instead of removing
    the debug things in the source, this patch now removes these from
    the configure script, i.e. HAVE_xxx_DEBUG is not written to the confdefs.h
-) tcpdump mini (patch 100):
    Code has been re-organized, therefore additional C files are built for tcpdump mini.
    IPv6 related stuff is now also part of -mini
    -mini is about half the size of -full (48.7% - 49.7% depending on platform)

### DIFF
--- a/package/network/utils/tcpdump/Makefile
+++ b/package/network/utils/tcpdump/Makefile
@@ -8,15 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=tcpdump
-PKG_VERSION:=4.5.1
-PKG_RELEASE:=4
+PKG_VERSION:=4.8.1
+PKG_RELEASE:=1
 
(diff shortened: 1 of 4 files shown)


nbd168:
Merged to my staging tree along with some follow-up size reduction changes.

PR #601, opened 2016-12-10
libs/ncurses: update to 6.0

p-wassi:
Maintainer: noone in Makefile
Compile tested: kirkwood, brcm47xx, ar71xx
Run tested: kirkwood, brcm47xx

Description:
Update ncurses to upstream release 6.0

From the official release notes [1]:
> These notes are for ncurses 6.0, released August 8, 2015.
> This release is designed to be source-compatible with ncurses 5.0 through 5.9; providing a new
> application binary interface (ABI). Although the source can still be configured to support the
> ncurses 5 ABI, the intent of the release is to provide extensions which are generally useful, but
> binary-incompatible with ncurses 5

Binaries which have a dependency for ncurses need to be rebuilt.
This is successfully tested with nano, screen, mutt, lftp, noping, gpsd-clients

To note:
Patch 103 was dropped here, there were several changes upstream related to pkg-config, PKG_CONFIG_LIBDIR and pc-files, e.g. [2], [3]
Maybe @jow- (who commited 103) can have a look at this.
The path for PKG_CONFIG_LIBDIR is added to the CONFIGURE_ARGS in the Makefile.
Does this solve what 103 addressed?

[1]:
https://www.gnu.org/software/ncurses/#h2-release-notes

[2]:
http://invisible-island.net/ncurses/NEWS.html#index-t20110910

[3]:
http://invisible-island.net/ncurses/NEWS.html#index-t20151101


diizzyy:
Nice work, are patches 100, 101, 900 really needed?

p-wassi:
TBH: I have no idea what they were initially done for.
Just this: Debian still has a patch '02-debian-backspace' with the
description: Fix backspace key description in some terminfo entries.
Their patch does similar things like our 101.

Patch 100 goes back to 2009 and was introduced by @ffainelli ; the track stops at [1].
Trac does not reveal more on 4852. But the patches name's could be a reference to
ncurses' Changelog (100 = [2], 101 = [3]), however I don't see how the described changes
are related to the patches.

OT here, but maybe input for your ongoing discussion on contribution guideline for packages:
in Debian's package repo many of their patches have a short header explaining what they do.
It is quite helpful to not only have a patches name when going through the list to check if there's
something relevant.

[1]:
https://github.com/lede-project/source/commit/68289d2042eca8ff855c37eec10fdc601c7272f4

[2]:
http://invisible-island.net/ncurses/NEWS.html#index-t20080112

[3]:
http://invisible-island.net/ncurses/NEWS.html#index-t20080628

diizzyy:
Having a look at the archlinux repo, --without-debug --without-ada might be of interest?
--with-build-cppflags=-D_GNU_SOURCE isn't needed?

https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/ncurses
Not sure if the cxx bindings are built by default?

p-wassi:
Rebased on current master for PKG_HASH

@diizzyy without-debug and without-ada were already enabled.

diizzyy:
@p-wassi Ahh sorry, my bad.

ffainelli:
@p-wassi, sorry, I don't remember where I got the patch from, it may have been from the Debian packages back then.

blogic:
merged, Thanks !

PR #602, opened 2016-12-11
ipq806x: EA8500 fix sysupgrade over stock firmware

ianchi:
When running sysupgrade for the first time over the second partition
(that still had stock firmware) the rootfs wasn't flashed as there
wasn't enough space.
This happend because stock also uses UBI, but the volume name wasn't
recognised and wasn't deleted before flashing.


blogic:
merged, Thanks !

ianchi:
Hi @blogic 
I don't see this change actually merged into master yet, nor in your tree.
Did you find any problem with it?

blogic:
i had a few trees locally that i forgot to push. it'll be in my staging tree later today

PR #605, opened 2016-12-11
curl/libcurl: move from LEDE to openwrt/packages/net

chris5560:
@kaloz
cURL not LEDE base build relevant, so move to openwrt/packages/net
!!! Depends on corresponding pull at
- openwrt/openwrt#270
- openwrt/packages#3646


blogic:
@chris5560 i asked the owrt guys if we are allowed to process this. they said that you need to get an ack from kaloz. let us know when that ack has been given. i'll happily merge the patch in all 3 trees.

chris5560:
Ignore it!

PR #606, opened 2016-12-11
dma-buf: add missing submenu

bobafetthotmail:
this kernel module currently has not set submenu. 
Fix this by adding it to the "Others" submenu

### DIFF
--- a/package/kernel/linux/modules/other.mk
+++ b/package/kernel/linux/modules/other.mk
@@ -131,6 +131,7 @@ $(eval $(call KernelPackage,bluetooth-hci-h4p))
 
 
 define KernelPackage/dma-buf
+  SUBMENU:=$(OTHER_MENU)
   TITLE:=DMA shared buffer support
   HIDDEN:=1
   KCONFIG:=CONFIG_DMA_SHARED_BUFFER


blogic:
please dont put the actually patched line into the subject.

blogic:
description should wrap at 75 chars

bobafetthotmail:
done. 

It seems like github does not auto-wrap the text at 75 chars in the editor, while committing from command line and writing the commit message with nano (my editor of choice for git) wraps chars automatically.

blogic:
actually the patch adds the submenu and not you. please also change that.

bobafetthotmail:
changed.

blogic:
so, if this change was made using an online github editor, then you never tested it ?

bobafetthotmail:
>so, if this change was made using an online github editor, then you never tested it ?

I edited the file locally. It is a one-liner change after all.

blogic:
imho the text is way to prose. please cut the chatter and be precise

"this kernel module currently has not set submenu. fix this by adding it to the "Others" submenu"

bobafetthotmail:
done.

blogic:
merged, Thanks !

blogic:
ah yes, you had a wrong prefix. it should be kernel:. i fixed that for you. please pay attention to this in future.

PR #607, opened 2016-12-11
mikrotik: set kernel rw flag for rootfs

vladimir-zahradnik:
Hello,
after I flashed latest pre-built image for Mikrotik and booted it, I noticed the root filesystem was mounted as read-only. All changes made to the device therefore won't last after reboot. I'm using Mikrotik RouterBoard 450G.

Here is output:
root@lede:~# dmesg | grep yaffs
[    2.941747] yaffs: dev is 32505862 name is "mtdblock6" **ro**
[    3.072586] VFS: Mounted **root (yaffs filesystem) readonly** on device 31:6.

I checked arguments passed to the kernel and there was no explicit rw flag set. I know for sure that stable image Chaos Calmer worked without change, so maybe default values changed in newer kernel release.

After I passed the RW flag and built custom image, router works as expected:
[    2.930827] yaffs: dev is 32505862 name is "mtdblock6" **rw**
[    3.064720] VFS: Mounted **root (yaffs filesystem)** on device 31:6.

According to [this thread](https://forum.openwrt.org/viewtopic.php?id=64298) the issue affects only some devices. However I'm not able to verify this, as I have only one affected device.

Changes in my commit affect only mikrotik devices and I assume it is safe to have the flag set. If it's not acceptable for you, please let me know and I'll try to get more info on this issue.

Signed-off-by: Vladimir Zahradnik vladimir.zahradnik@gmail.com

### DIFF
--- a/target/linux/generic/files/fs/yaffs2/yaffs_mtdif.c
+++ b/target/linux/generic/files/fs/yaffs2/yaffs_mtdif.c
@@ -259,7 +259,7 @@ struct mtd_info * yaffs_get_mtd_device(dev_t sdev)
 		return NULL;	/* This isn't an mtd device */
 
 	/* Check it's NAND */
-	if (mtd->type != MTD_NANDFLASH) {
+	if (mtd->type != MTD_NANDFLASH && mtd->type != MTD_MLCNANDFLASH) {
 		yaffs_trace(YAFFS_TRACE_ALWAYS,
 			"yaffs: MTD device is not NAND it's type %d",
 			mtd->type);
--- a/target/linux/generic/files/fs/yaffs2/yaffs_vfs.c
+++ b/target/linux/generic/files/fs/yaffs2/yaffs_vfs.c
@@ -2598,7 +2598,7 @@ static int yaffs_remount_fs(struct super_block *sb, int *flags, char *data)
 	}
 
 	/* Check it's NAND */
-	if (mtd->type != MTD_NANDFLASH) {
+	if (mtd->type != MTD_NANDFLASH && mtd->type != MTD_MLCNANDFLASH) {
 		yaffs_trace(YAFFS_TRACE_ALWAYS,
 			"MTD device is not NAND it's type %d",
 			mtd->type);


blogic:
merged, Thanks !

vladimir-zahradnik:
Hi @blogic,
could you please look at this PR one more time? I don't see my patch among commits in master and GitHub reports this: 'Closed with unmerged commits'.

Thanks.

blogic:
it is in my staging tree and will be in master within 24 hours

vladimir-zahradnik:
Thanks for info.

nbd168:
I think this may be the wrong approach. fstools is usually responsible for re-mounting the root filesystem as writable. If this doesn't work, we need to figure out why

vladimir-zahradnik:
OK, I want this to be fixed. If it means more investigating, I am definitely in.

blogic:
ok, i have dropped the patch from my tree again

vladimir-zahradnik:
I will try to get more info and update the thread in such case. If by any chance some of you find a possible place where the issue may be, please update this thread. I will test changes on my device. Thanks.

blogic:
@vladimir-zahradnik the fstools package is responsible for the rw remount.
-> https://git.lede-project.org/?p=project/fstools.git;a=blob;f=mount_root.c
it is probably missing yaffs support.

vladimir-zahradnik:
@blogic Might be. But regardless I can confirm that Chaos Calmer stable worked without any modification. So if Openwrt 15 used yaffs already, it means that probably a bug was introduced in nightly and commit could be tracked. I will look at fstab anyway. Thanks.

blogic:
@vladimir-zahradnik correct, you could try to bisect it using only commit in

git log --oneline package/system/fstools/

vladimir-zahradnik:
@blogic OK, I'll try to do that. It will probably take me few days, as I don't have device with me now.

duvi:
According to the mentioned [openwrt forum thread](https://forum.openwrt.org/viewtopic.php?id=64298), this issue is already present in Openwrt trunk, but not present in Chaos Calmer branch, so I compared those two.
trunk: https://dev.openwrt.org/log/trunk/package/system/fstools
CC: https://dev.openwrt.org/log/branches/chaos_calmer/package/system/fstools

These are the additional changes to trunk, that supposedly resulted in Mikrotik read-only rootfs:
https://dev.openwrt.org/changeset/49223/
https://dev.openwrt.org/changeset/49335/
https://dev.openwrt.org/changeset/49336/
https://dev.openwrt.org/changeset/49337/

vladimir-zahradnik:
@duvi Thanks for the comparison, it helped me to get started. Anyway I managed to debug the code and find a probable cause. I compared booting of latest Chaos Calmer with nightly build of LEDE. The only difference was during yaffs remount.
CC: [   11.320000] mount_root: mounting /dev/root
LEDE: [   10.377908] mount_root: mounting /dev/root
[   10.382284] yaffs: MTD device is not NAND it's type 8

I tracked the issue into Yaffs code. When doing filesystem remount, there is a check if flash is of type NAND, but it fails to detect MLC (and probably TLC) NAND Flash.

Type 8 corresponds to MTD_MLCNANDFLASH as defined here:
https://github.com/torvalds/linux/blob/master/include/uapi/mtd/mtd-abi.h

@blogic Because I saw the Yaffs code is integral part of LEDE source tree and is not downloaded from other remote repo, I updated the code directly. But if necessary, I could also prepare a patch instead.

Change was successfuly tested on my device, please see my latest commit in this PR.

Thanks.

nbd168:

PR #592, opened 2016-12-06
github: adding a pull request template

bobafetthotmail:
This text is used by GitHub to remind important things to people posting PRs through the GitHub's web interface. See here for more information https://github.com/blog/2111-issue-and-pull-request-templates

Loosely inspired by OpenWRT package feed's own pull request template
https://github.com/openwrt/packages/blob/master/.github/pull_request_template

Signed-off-by: Alberto Bursi <alberto.bursi@outlook.it>

PR #609, opened 2016-12-12
ar71xx: build support and fixed partition map for WNR2000v1 board

htruong:
This is an old board with a small flash area that requires quite a bit of work to get LEDE working, but manually flashing the partitions after applying the patch has shown to be working alright (I have tested it on my board). Although, future considerations will be needed because the board has an extremely small amount of flash (4MB-). 

Here I add initial patch to support this build, and I will have to write an entry to a building and flashing guide later.  The claim that WNR2000v1 is supported with the AP81 branch on OpenWRT wiki is wrong and misleading. The reason is that AP81 has 8MB flash while WNR2000v1 has only 4MB flash. I plan to edit the wiki entry as soon as possible, too.


mkresin:
Your patch has formal issues. Please have a look at work in progress pull request guidelines the in #592.

- commit title is not prefixed and to long
- commit message is missing
- Signed off by is missing

Please take care to change the actual commit and not the title/description of the pull request only.

target/linux/ar71xx/image/legacy-devices.mk
```diff
@@ -434,6 +434,11 @@ define LegacyDevice/WPE72_16M
 endef
 LEGACY_DEVICES += WPE72_16M
 
+define LegacyDevice/WNR2000
```
mkresin: We have switched to a new image build code a while ago. The legacy*.mk files are only used for existing boards not yet converted to the new image build code.

Additions to this files (using the old image build code) are not accepted any longer. Please add your board to generic.mk and use the new image build code.

blogic: all the netgear routers are still on the old code. converting them is a different patch i guess

mkresin: Yeah, blogic is right. Missed that using the new build code means that the NetgearLzma recipe need to be converted to the new image build code. And that is indeed out of scope of this patch.

htruong:
I will take care of the formal issues and will try to resubmit it.

htruong:
Let me just close this one and create another pull request, due to not having the branch name correct as to the standard.

PR #610, opened 2016-12-12
ar71xx: Add support for TP-LINK TL-WR940N v4

kpanic23:
Signed-off-by: David Lutz <kpanic@hirnduenger.de>


NeoRaider:
A few minor comments:

* You don't touch `target/linux/ar71xx/base-files/etc/board.d/01_leds`. Does this board not have any LEDs that need triggers except for the status/power LED?
* You add a new config symbol `CONFIG_ATH79_MACH_TL_WR940N_V4` and enable it in `target/linux/ar71xx/config-4.4`. You should disable this symbol in `target/linux/ar71xx/mikrotik/config-default` and `target/linux/ar71xx/nand/config-default` as it is done for other boards.
* The last line in `target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr940n-v4.c` has excess whitespace.

kpanic23:
Of course the device has additional LEDs, I just couldn't find the LED file in LEDE, as the path seems to differ from OpenWRT :o)
Also, I have changed the kernel config for the mikrotik and nand targets.
And removed the nasty newline...

NeoRaider:
The newline should be there (C files should end with one), but without other whitespace after it.

Please also squash your changes into a single commit (`git rebase -i` will probably help you) and force-push the result over the PR branch.

kpanic23:
OK, commits are squashed.

NeoRaider:
Just tried it out; the profile definition is missing a DEVICE_TITLE, so the device is not selectable.

kpanic23:
My bad, I had just copied the changes from my gluon patch...
I have added a title line.

NeoRaider:
Thanks, pulled into my staging tree.

PR #611, opened 2016-12-12
ar71xx: Add support for Netgear WNR2000v1

htruong:
This patch adds supports for the WNR2000v1 board with 4MB flash, and produces device-specific factory, rootfs, and sysupgrade files for the WNR2000v1. This board is errorneously claimed as supported on the OpenWRT wiki as AP81, but AP81 image would not work because of APT81 image reuiring having 8MB of flash, while WNR2000v1 has only 4MB.

The image requires the u-boot bootloader to be modified to fuhry's bootloader first.

Short specification:

- CPU: Atheros AR9132
- 4x 10/100 Mbps Ethernet, 1x WAN 10/100 Mbps
- 4 MB of Flash
- 32 MB of RAM
- UART header (J1) on board
- 1x button

Factory/Initial flash instructions:

- Set up a TFTP server on your local machine.
- Download the uImage for ar71xx-generic and the rootfs image for ar71xx-generic-wnr2000 and save in the tftp server root.
- Gain serial access to the router via the UART port (telnetenable over the network only won't work!).
- Upgrade the u-boot bootloader to fuhry's version by running the script: http://fuhry.com/b/wnr2000/install-repart.sh
- When the router restarts, interrupt u-boot and gain access to u-boot command line.
- Repartititon the board and flash initial uImage and rootfs as follow.

Commands to type in u-boot:

	# tells u-boot that we have a tftp server on 192.168.1.10
	setenv serverip 192.168.1.10

	# tells u-boot that the router should take the address 192.168.1.1
	setenv ipaddr 192.168.1.1

	# erase the region from 0x050000-0x3f0000
	erase 0xbf050000 +0x3A0000

	# loads sqfs.bin on TFTP server, and put it to memory address 0x81000000
	tftpboot 0x81000000 sqfs.bin
	# it will tell you the length of sqfs.bin in hex, let's say ZZZZZZ
	# copy bit by bit 0xZZZZZZ bytes from offset 0x050000
	cp.b 0x81000000 0xbf050000 0xZZZZZZ

	# same to the uImage.bin, write it right next to sqfs.bin
	# again, 0xYYYYYY is the length that tftpboot reports
	tftpboot 0x81000000 uImage.bin
	cp.b 0x81000000 0xbf2a0000 0xYYYYYY

	# We need to tell the kernel what board it is booting into, and where to find the partitions
	setenv bootargs "board=WNR2000 console=ttyS0,115200 mtdparts=spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,3712k(firmware),64k(art)ro rootfstype=squashfs,jffs2 noinitrd"

	# Tell u-boot where to find the uImage
	setenv bootcmd "bootm 0xbf2a0000"

	# Tell u-boot to save parameters to the u-boot-env partitions
	saveenv

	# Reset the board
	reset

Tested on:

- WNR2000v1 board.
- Initial flash works.

Known bugs:

- I don't know why factory image doesn't work on initial flash on stock firmware in u-boot recovery mode while it should.
- Sysupgrade does not yet work, if you do -f it will mess up your installation (requiring a reinstall of sqfs and uImage).

Signed-off-by: Huan Truong <htruong@tnhh.net>


blogic:
merged, Thanks !

RaylynnKnight:
Since u-boot is supposed to be open source where is the source code located for the u-boot version that is loaded by the  http://fuhry.com/b/wnr2000/install-repart.sh script?

blogic:
you will need to contact the owner of that domain and ask him. we did not create nor distribute that script nor the files it uses.

bobafetthotmail:
the script links to a guy's site, that links to an openwrt forum thread where the guy posted, there you find a patch for uboot 1.1.4. https://forum.openwrt.org/viewtopic.php?pid=117117#p117117

bigoss:
connected via the UART port  can only see events on router can't write . How do i supposed to enable telnet / update u-boot ?

PR #613, opened 2016-12-13
Add support for TP-Link WR1043ND v4

Brother-Lal:
These commits will add support for the fourth version of the WR1043ND from TP-LINK.

Some commits from neoraider are squashed into my commit.


target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c
```diff
@@ -0,0 +1,191 @@
...
+	{
+		.bus_id = "ag71xx-mdio.0",
+		.phy_addr = 0,
+		.platform_data = &tl_wr1043nd_v4_ar8327_data,
+	},
+};
+
+static void __init tl_wr1043nd_v4_setup(void)
+{
+	u8 *mac = (u8 *) KSEG1ADDR(TL_WR1043_V4_FLASH_END - TL_WR1043_V4_MAC_OFFSET);
```
NeoRaider: Using a negative offset here is an interesting idea. It's a bit weird though that you use a negative offset for the MAC address, and a positive one for the calibration data.

target/linux/ar71xx/mikrotik/config-default
```diff
@@ -89,6 +89,7 @@ CONFIG_ATH79_MACH_RBSXTLITE=y
 # CONFIG_ATH79_MACH_TL_WR1041N_V2 is not set
 # CONFIG_ATH79_MACH_TL_WR1043ND is not set
 # CONFIG_ATH79_MACH_TL_WR1043ND_V2 is not set
+# CONFIG_ATH79_MACH_TL_WR1043ND_V4 is not set
```
NeoRaider: The same line should be added to `target/linux/ar71xx/nand/config-default `

tools/firmware-utils/src/tplink-safeloader.c
```diff
@@ -784,6 +854,8 @@ int main(int argc, char *argv[]) {
 		info = &eap120_info;
 	else if (strcmp(board, "ARCHERC9") == 0)
 		info = &e9_info;
+	else if (strcmp(board, "TLWR1043NDV4_EU") == 0)
```
NeoRaider: I think it would make sense to drop the `_EU` suffix here and in the image Makefile for now, as there are no other versions of the TL-WR1043v4 so far. If TP-Link makes proper use of their SupportList format, we might be able to support multiple versions with a single image (and if they don't we can split it later).

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c
NeoRaider: If you keep using a negative offset, it should be relative to 0x20000000 instead of 0x1fffffff, as 0x20000000 is the address just after the end of the flash and leads to more natural offsets.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -757,7 +757,25 @@ define Device/tl-wr1043nd-v3
...
-TARGET_DEVICES += tl-wr1043nd-v1 tl-wr1043nd-v2 tl-wr1043nd-v3
+
+define Device/tl-wr1043nd-v4
+    $(Device/tplink)
+    DEVICE_TITLE := TP-LINK TL-WR1043N/ND v4
+    DEVICE_PACKAGES := kmod-usb-core kmod-usb2 kmod-usb-ledtrig-usbport
+    BOARDNAME := TL-WR1043ND-v4
+    DEVICE_PROFILE := TLWR1043
+    TPLINK_HWID := 0
+    TPLINK_HWREV := 0
```
NeoRaider: I have one more request. Could you try setting TPLINK_HWID to 0x10430004 and remove the TPLINK_HWREV definition (so it defaults to 1), and check if the stock firmware still accepts the image?

The stock firmware images contain 0 here, but I suspect that the upgrader doesn't check it at all (as the mktplinkfw image is embedded into the safeloader image). Setting TPLINK_HWID to a unique value would allow the LEDE sysupgrader to check for the correct model by comparing the values in flash and in the image.

(this might become redundant soon with the new LEDE image metadata, but we don't have that for ar71xx yet)

target/linux/ar71xx/mikrotik/config-default
Brother-Lal: I noticed the 1043_v2 is missing in target/linux/ar71xx/nand/config-default . Is that an error or is there a difference?

pepe2k: AFAIK, these files shouldn't be changed manually. Instead, `make kernel_menuconfig CONGI_TARGET=subtarget` with save on exit should be used (after selecting _Generic devices with NAND flash_ in `make menuconfig`). That will pickup, refresh and update related kernel config-default.

Anyway, if seems that kernel configs for subtargets in ar71xx need general refresh and cleanups. Ex. there are many non-NAND boards enabled in ...nand/config-default.

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c
Brother-Lal: I used a negative offset because the mac is at the end of the flash, so it is the smaller offset.
And because i read somewhere that if the flash size changes, it might be possible that you can reuse the code without modification, as only the userdata partition infront of the config/art might change size. The calibration data was taken from dir 869, where its positive from eeproom start.
 I'll recompute the offset for 0x200000000 :+1:

NeoRaider: Following the same logic that it is at the end of the flash, TL_WR1043_V4_EEPROM_ADDR would be `(0x20000000 - 0x10000)`, so you could again just define the offset 0x10000.

(Note that you have a typo in the previous comment, the address is 0x20000000, not 0x200000000, which is not a 32bit number)

pepe2k: @Brother-Lal, please take a look how it's done ex. with ART partition, on other TP-Link devices like [TL-MR3x20](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-mr3x20.c#L81). In most of them, ART data resides in last 64 KB sector in FLASH, at offset _0x1000_ (in that sector).

As the SPI FLASH memory area in QC/A WiSoCs wraps around, you can just set the offset to `0x1fff1000` (it's exactly _0x20000000 - 0x10000 + 0x1000_ -> _SPI FLASH memory address + 16 MB - 64 KB + 4 KB_) and no matter what FLASH size (up to 16 MB) is used, this offset will always point to the _0x1000_ offset inside last 64 KB sector. TL-MR3220 has 4 MB FLASH chip by default.

IMHO, using negative offset here might be confusing for others. Personally, I would stick to the general convention like in many other ar71xx mach-*.c files.

Cheers!

Brother-Lal:
Added the requested changes, used a static address for the mac, as the other sample files.

target/linux/ar71xx/image/tp-link.mk
Brother-Lal: This works, added it into the commit.

NeoRaider:
Please squash the changes of your 4th commit into the other commits. The code itself should be okay now.

Brother-Lal:
Ok, done. Thought as the safeloader was touched, a second commit would clarify the change, but its shorter this way.

NeoRaider:
No, change my commit not to define the wrong string with "_EU" in the first place. It doesn't make sense to add and then modify something in the same pull request for no good reason.

Brother-Lal:
Ok, i didn't want to touch commit from somebody else.
Should be correct now?

NeoRaider:
Thanks, I found one more issue: while your signed-off has your correct name, the commit author is just "Brother-Lal". You can fix this by setting the correct name and mail address in the git config and running `git commit --amend --reset-author`. This will also allow you to add signed-off lines automatically using `git commit -s` in the future.

Brother-Lal:
I hope thats fixed? Thank you for your patience with a first timer :)

NeoRaider:
Thanks, pulled into my staging tree!

proudlytm:
Guys, just to ask, does that pull request mean that WR1043ND v4 is fully supported by LEDE now and it works? Has it been tested? Thanks!

Brother-Lal:
For me more or less everything works(lab,wlan,usb,leds) with this pr.
Tested on one device used to develop the support.

proudlytm:
That's a job well done then! Is the build going to get an official download link soon somewhere?

Brother-Lal:
As soon as the pr is merged and a snapshot compiled it should pop up here:
https://downloads.lede-project.org/snapshots/targets/ar71xx/generic/

proudlytm:
Nice! Thanks a lot again man! Will be waiting!

NeoRaider:
@Brother-Lal: I've rebased the image generation patches on top of some refactoring by @jow- I'd like to get merged first; you can find this in my staging tree at https://git.lede-project.org/?p=lede/neoraider/staging.git;a=summary . It would be great if you could test that this still generates working factory and sysupgrade images.

Brother-Lal:
@NeoRaider I compiled and tested it, everything seems to work.
Both factory and sysupgrade images tested, all fine.

proudlytm:
Guys, can I just ask - Do you have information on when the commit is going to be merged ?
@Brother-Lal @NeoRaider

proudlytm:
Edit: never mind, I just saw that the commits were pushed! That's great!

rotanid:
great, thanks @Brother-Lal and @NeoRaider !

proudlytm:
image

Couldn't wait for the snapshot to build! 😄 Works perfectly!

ramawp:
Hi thanks a lot for this, realy love it 🥇 . But how i can install luci from your package??

atomizer3:
Hey guys, great work :+1: 
Just wanted to ask if there were some special support added for 3g modems? Or it works as it was in main repo ?

ramawp:
Hey guys, that is good work. But i can download openvswitch, it show cannot satisfy the dependencies bla bla bla and it shows packages libpcap and zlib. How i can fix it?? Thank You

PR #615, opened 2016-12-14
ca-certificates: move to openwrt/packages/net

chris5560:
ca-certificates not base build relevant, so move to openwrt/packages/net
!!! Depends on corresponding pulls at
- openwrt/openwrt#273
- openwrt/packages#3656


blogic:
they should be present. wget and curl use them.

chris5560:
From my knowledge no hardware/base build doing a preselection.
- wget-ssl is already part of openwrt/packages
- move of curl to openwrt/packages is pending
- uclient-fetch also does not select any ssl library by default.

From my understanding lede-project wants to concentrate on the real base build,
so everything else should be moved out.

blogic:
On 14/12/2016 20:20, Christian Schoenebeck wrote:
> From my knowledge no hardware/base build doing a preselection.
> - wget-ssl is already part of openwrt/packages
> - move of curl to openwrt/packages is pending
> - uclient-fetch also does not select any ssl library by default.
> 
> From my understanding lede-project wants to concentrate on the real base
> build,
> so everything else should be moved out.
> 
> 

no one ever suggested that stuff needs to be moved out of the tree. i
really would like to keep certs in trunk, they are useful and should
stay there.

NAK from me on this patch.

nbd168:
ustream-ssl uses it, and uclient-fetch (which uses ustream-ssl) is even included in the base image.
There you have your base tree use of ca-certficates

PR #617, opened 2016-12-16
libs/cyassl: Enable multithreading

diizzyy:
More and more platforms are multicore SoCs, don't enforce singlethreading.


blogic:
merged, Thanks !

PR #618, opened 2016-12-16
include/package-ipkg.mk: allow to ignore default_postinst/_prerm

chris5560:
New IPKG_NO_DEFAULT
If set to IPKG_NO_DEFAULT=1 inside Makefile gives the
Developer/Packagemaintainer the chance to define prerm and postinst
without forced to commands given by default_postinst / default_prerm.

### DIFF
--- a/include/package-ipkg.mk
+++ b/include/package-ipkg.mk
@@ -197,19 +197,24 @@ $(_endef)
 			printf "Description: "; echo "$$$$DESCRIPTION" | sed -e 's,^[[:space:]]*, ,g'; \
 		) > control; \
 		chmod 644 control; \
-		( \
-			echo "#!/bin/sh"; \
-			echo "[ \"\$$$${IPKG_NO_SCRIPT}\" = \"1\" ] && exit 0"; \
-			echo ". \$$$${IPKG_INSTROOT}/lib/functions.sh"; \
-			echo "default_postinst \$$$$0 \$$$$@"; \
-		) > postinst; \
-		( \
-			echo "#!/bin/sh"; \
-			echo ". \$$$${IPKG_INSTROOT}/lib/functions.sh"; \
-			echo "default_prerm \$$$$0 \$$$$@"; \
-		) > prerm; \
-		chmod 0755 postinst prerm; \
 		$($(1)_COMMANDS) \
+		if [ "$${IPKG_NO_DEFAULT}" = "1" ]; then \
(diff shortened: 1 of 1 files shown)


blogic:
please provide some use cases for this feature

chris5560:
The real use case is to give back a package maintainer/developer the control what happen inside pre/post install/remove.

I personally hate it if a not configured service after fresh installation is enabled and started.
Samples for that are file servers, Proxy's, services like ddns-scripts and many more.
The only useful function I've seen inside default_postinst is the creation of user:group used to run the service/package.
All routers manufacturer firmware I know, came up with LAN/DHCP enabled and running http-server for UI and telnet/ssh access.
Everything else needs to be configured and enabled including WAN networks.
I already pointed that out when default_ functions where implemented but nobody want's to here it.

But there was implemented IPKG_NO_SCRIPT for TWO Makefiles looking like one very special use case without any documentation.
Problem with that is that it skip package postinst script completely.
The normal way should be to write a proper postinst doing the correct job for this special use case.
Ahh!!! there is the point above. There is no control.

It's currently a pain to build packages for lede-project and openwrt.
Different opkg versions, different functionality of basefile package and more.

blogic:
if your service auto starts with an insecure config then the package should get an uci option to not start it. moving the logic into the packages will imho lead to an unmaintainable mess.

nbd168:
I agree with blogic. Relying on /etc/init.d/service enable/disable for this stuff is too fragile and uci should be used instead.

blogic:
2 NAKs, closing the PR, sorry

chris5560:
Exactly the answer I expected.
If the development team needs something alias IPKG_NO SCRIPT it's implemented.
If you should implement an option to get back control, then NAK.
Nobody is forced to use my offer, but everyone is forced to use yours.

nbd168:
Maybe it helps if I provide a bit of background on this decision.

 We try our best to keep packages consistent with each other so that things work the way users expect them to.

Having disable-by-default be handled through bypassing the default postinst/prerm scripts has the side effect of not playing well with sysupgrade (violating user expected behavior), UI configuration, etc.. Also, If we add more common functionality to the default script, packages that choose to override this will miss out on that, and it's completely non-obvious to users of the package why that is the case.

You provided your reason why you wanted to make this change, and we gave you a different way to achieve the same outcome, while also taking things into account that you either did not consider, or did not care about.

Is there any other reason why this change matters to you?

PR #620, opened 2016-12-17
linux: Backport MTD patch to support memory size above 128Mib

corny:
Patch extracted from:
- http://lists.infradead.org/pipermail/linux-mtd/2016-December/070890.html
- http://lists.infradead.org/pipermail/linux-mtd/2016-December/070891.html

Related issue:
- https://github.com/freifunk-gluon/gluon/pull/962

### DIFF
--- a/target/linux/generic/patches-4.4/482-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch
+++ b/target/linux/generic/patches-4.4/482-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch
@@ -19,10 +19,10 @@ Tested-by: Vignesh R <vigneshr@ti.com>
 
 --- a/drivers/mtd/spi-nor/spi-nor.c
 +++ b/drivers/mtd/spi-nor/spi-nor.c
-@@ -68,6 +68,10 @@ struct flash_info {
- #define	SPI_NOR_DUAL_READ	0x20    /* Flash supports Dual Read */
+@@ -69,6 +69,10 @@ struct flash_info {
  #define	SPI_NOR_QUAD_READ	0x40    /* Flash supports Quad Read */
  #define	USE_FSR			0x80	/* use flag status register */
+ #define	SPI_NOR_HAS_LOCK	0x100	/* Flash supports lock/unlock via SR */
 +#define SPI_NOR_4B_OPCODES	BIT(11)	/*
 +					 * Use dedicated 4byte address op codes
 +					 * to support memory size above 128Mib.
@@ -30,16 +30,16 @@ Tested-by: Vignesh R <vigneshr@ti.com>
  };
  
  #define JEDEC_MFR(info)	((info)->id[0])
-@@ -181,6 +185,78 @@ static inline struct spi_nor *mtd_to_spi
+@@ -182,6 +186,78 @@ static inline struct spi_nor *mtd_to_spi
  	return mtd->priv;
  }
  
 +
-+static u8 spi_nor_convert_opcode(u8 opcode, const u8 table[][2])
++static u8 spi_nor_convert_opcode(u8 opcode, const u8 table[][2], size_t size)
 +{
 +	size_t i;
 +
-+	for (i = 0; table[i][0]; ++i)
++	for (i = 0; i < size; i++)
 +		if (table[i][0] == opcode)
 +			return table[i][1];
 +
@@ -56,10 +56,10 @@ Tested-by: Vignesh R <vigneshr@ti.com>
 +		{ SPINOR_OP_READ_1_2_2,	SPINOR_OP_READ_1_2_2_4B },
 +		{ SPINOR_OP_READ_1_1_4,	SPINOR_OP_READ_1_1_4_4B },
 +		{ SPINOR_OP_READ_1_4_4,	SPINOR_OP_READ_1_4_4_4B },
-+		{0, 0}
 +	};
 +
-+	return spi_nor_convert_opcode(opcode, spi_nor_3to4_read);
++	return spi_nor_convert_opcode(opcode, spi_nor_3to4_read,
++				      ARRAY_SIZE(spi_nor_3to4_read));
 +}
 +
 +static inline u8 spi_nor_convert_3to4_program(u8 opcode)
@@ -68,10 +68,10 @@ Tested-by: Vignesh R <vigneshr@ti.com>
 +		{ SPINOR_OP_PP,		SPINOR_OP_PP_4B },
 +		{ SPINOR_OP_PP_1_1_4,	SPINOR_OP_PP_1_1_4_4B },
 +		{ SPINOR_OP_PP_1_4_4,	SPINOR_OP_PP_1_4_4_4B },
-+		{0, 0}
 +	};
 +
-+	return spi_nor_convert_opcode(opcode, spi_nor_3to4_program);
++	return spi_nor_convert_opcode(opcode, spi_nor_3to4_program,
++				      ARRAY_SIZE(spi_nor_3to4_program));
 +}
 +
 +static inline u8 spi_nor_convert_3to4_erase(u8 opcode)
@@ -80,10 +80,10 @@ Tested-by: Vignesh R <vigneshr@ti.com>
 +		{ SPINOR_OP_BE_4K,	SPINOR_OP_BE_4K_4B },
 +		{ SPINOR_OP_BE_32K,	SPINOR_OP_BE_32K_4B },
 +		{ SPINOR_OP_SE,		SPINOR_OP_SE_4B },
-+		{0, 0}
 +	};
 +
-+	return spi_nor_convert_opcode(opcode, spi_nor_3to4_erase);
++	return spi_nor_convert_opcode(opcode, spi_nor_3to4_erase,
++				      ARRAY_SIZE(spi_nor_3to4_erase));
 +}
 +
 +static void spi_nor_set_4byte_opcodes(struct spi_nor *nor,
@@ -109,7 +109,7 @@ Tested-by: Vignesh R <vigneshr@ti.com>
  /* Enable/disable 4-byte addressing mode. */
  static inline int set_4byte(struct spi_nor *nor, const struct flash_info *info,
  			    int enable)
-@@ -1274,27 +1350,10 @@ int spi_nor_scan(struct spi_nor *nor, co
+@@ -1277,27 +1353,10 @@ int spi_nor_scan(struct spi_nor *nor, co
  	else if (mtd->size > 0x1000000) {
  		/* enable 4-byte addressing if the device exceeds 16MiB */
  		nor->addr_width = 4;
--- a/target/linux/generic/patches-4.4/478-mtd-spi-nor-rename-SPINOR-macros.patch
+++ b/target/linux/generic/patches-4.4/478-mtd-spi-nor-rename-SPINOR-macros.patch
@@ -10,14 +10,13 @@ This is a transitional patch and will help a later patch of spi-nor.c
 to automate the translation from the 3-byte address op codes into their
 4-byte address version.
 
-Signed-off-by: Cyrille Pitchen <cyrille.pitchen at atmel.com>
-Acked-by: Mark Brown <broonie at kernel.org>
-Acked-by: Marek Vasut <marek.vasut at gmail.com>
+Signed-off-by: Cyrille Pitchen <cyrille.pitchen@atmel.com>
+Acked-by: Mark Brown <broonie@kernel.org>
+Acked-by: Marek Vasut <marek.vasut@gmail.com>
 ---
  drivers/mtd/devices/serial_flash_cmds.h |  7 -------
  drivers/mtd/devices/st_spi_fsm.c        | 28 ++++++++++++++--------------
  drivers/mtd/spi-nor/spi-nor.c           |  8 ++++----
- drivers/spi/spi-bcm-qspi.c              |  6 +++---
  include/linux/mtd/spi-nor.h             | 22 ++++++++++++++++------
  5 files changed, 37 insertions(+), 34 deletions(-)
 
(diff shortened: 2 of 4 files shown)


NeoRaider:
I think it would make sense to move these patches to target/linux/generic/patches-4.4, where the other MTD patches are.

They could also need a refresh:
```
Applying /home/neoraider/Devel/LEDE/source/target/linux/ramips/patches-4.4/0400-mtd-spi-nor-rename-SPINOR-macros.patch using plaintext:
patching file drivers/mtd/devices/serial_flash_cmds.h
patching file drivers/mtd/devices/st_spi_fsm.c
patching file drivers/mtd/spi-nor/spi-nor.c
Hunk #1 succeeded at 1349 (offset -141 lines).
patching file include/linux/mtd/spi-nor.h
Hunk #1 succeeded at 42 (offset -1 lines).
Hunk #2 succeeded at 59 (offset -1 lines).

Applying /home/neoraider/Devel/LEDE/source/target/linux/ramips/patches-4.4/0401-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch using plaintext:
patching file drivers/mtd/spi-nor/spi-nor.c
Hunk #1 succeeded at 68 (offset -2 lines).
Hunk #2 succeeded at 185 (offset -7 lines).
Hunk #3 succeeded at 1398 (offset -141 lines).
```
See https://lede-project.org/docs/guide-developer/use-patches-with-buildsystem for a guide to the handling of patch files in LEDE.

corny:
ok, done

NeoRaider:
It seems one of the patches was marked as "Superseded" in upstream patchwork: http://patchwork.ozlabs.org/patch/704050/ , but I can't find a new version...

I don't feel qualified to review this myself, so I'd like to wait for some version the patches to hit http://git.infradead.org/linux-mtd.git before merging them (unless someone else from the LEDE team wants to step up and take them).

NeoRaider:
Another thing: The commit message is missing a rationale. It should explain what issues on what devices these patches fix.

dmke:
These patches are related to a soft reset issue, described in [this Gist](https://gist.github.com/dmke/d389bc364b3f73f525076eaee0019dc1) (see also #464).

It basically moves from a stateful MTD addressing mode to a stateless one.

Alternatives to this are either modifying the bootloader to try 3-byte addressing if it detects a failed 4-byte read, or resetting the memory just before the soft reset (depends on the actual memory chip, for Winbond W25Q256 this equals to writing 0x66 0x99).

dmke:
There has been an update:

- [PATCH v3 0/2] mtd: spi-nor: add a stateless method to support memory size above 128Mib [[link]](http://lists.infradead.org/pipermail/linux-mtd/2016-December/thread.html#71072)
  - https://patchwork.ozlabs.org/patch/707121/
  - https://patchwork.ozlabs.org/patch/707122/

corny:
I updated my pull request to use most recent patches.

corny:
Is still anything missing?

mkresin:
No there isn't really something missing. The problem is that the most recent patches (v3) are marked as superseded in patchwork but a v4 isn't send yet.

corny:
Ok, I agree to wait for the final version that gets merged.

mkresin:
The final versions of the patches are now accepted in the linux-mtd patchwork and part of https://github.com/spi-nor/linux/commits/next (which will be pulled into linux-mtd.git at some point):

- https://patchwork.ozlabs.org/patch/711144/
- https://patchwork.ozlabs.org/patch/711148/

Please take care to refresh all patches touching the same file. For each target the generic patches applied first and the target specific ones afterwards. That is why these patches could have an impact on targets patching the same file as well.

The usual way to refresh all patches is to select a target via ```make menuconfig``` and run ```make target/linux/{clean,refresh}```. Since it is a lot of work, KanjiMonster wrote a script to automate this: https://gist.github.com/KanjiMonster/dfcd1af3190aa7b2941f1bdcb119f25e.

Use the script the following  ```./refresh_all_targets_patches.sh 4.4 4.4.42``` and fixed the conflicts.

corny:
I updated the Patches, but they don't work for me. The device hangs instead of rebooting. Has anyone else affected devices that can be tested?

target/linux/generic/patches-4.4/482-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch
```diff
@@ -19,27 +19,27 @@ Tested-by: Vignesh R <vigneshr@ti.com>
...
- #define	SPI_NOR_DUAL_READ	0x20    /* Flash supports Dual Read */
+@@ -69,6 +69,10 @@ struct flash_info {
  #define	SPI_NOR_QUAD_READ	0x40    /* Flash supports Quad Read */
  #define	USE_FSR			0x80	/* use flag status register */
+ #define	SPI_NOR_HAS_LOCK	0x100	/* Flash supports lock/unlock via SR */
 +#define SPI_NOR_4B_OPCODES	BIT(11)	/*
 +					 * Use dedicated 4byte address op codes
 +					 * to support memory size above 128Mib.
 +					 */
  };
```
naobsd: this hunk and 2nd URI in commit log seem to be v3...

cstratton:
In the case of something like a W25Q256 on an MT7688, you can't fix it in the bootloader, as the actual issue is that the hardcoded (state machine or bootrom) boot behavior of the CPU will on reset try to read the chip in whichever 3 or 4 byte mode is selected by mode strappings.  If that doesn't match the current mode of the chip, the bootloader transfers U-Boot (or whatever) offset by one byte, and boot fails - typically you see endless SPI access with a scope, while a logic analyzer will show the address size mismatch.  

Essentially, whatever addressing mode the bootmode strapping selects (or whatever mode the flash chip's power-on-behavior register is programmed to) is the mode that chip has to _stay_ in (during both U-Boot and Linux), otherwise reset (software, hardware, or watchdog) will result in a hang if it occurs at _any_ time when the flash chip has been set to a mode other than the power on expectation.

target/linux/generic/patches-4.4/482-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch
corny: Yes, I put the wrong URI into my commit, but this hunk is equal for v3 and v4.

corny:
I tried with `SPI_NOR_4B_OPCODES = BIT(10)` and `SPI_NOR_4B_OPCODES = BIT(11)` without success.

target/linux/generic/patches-4.4/482-mtd-spi-nor-add-a-stateless-method-to-support-memory-size-above-128Mib.patch
naobsd: sure. sorry!

nbd168:
What's the status of this? Does the backport resolve these issues? If so, then please squash the update commits into the main one so it can be applied properly.
Please also port port this to 4.9

corny:
I received a mail from the developer but did not investigate further yet:

> Hi Julian,
>
> what is the exact part number of the SPI memory embedded on your ZBT-wg3526
> board? Which manufacturer? Spansion, Macronix, Micron, Winbond, SST, other?
> 
> Do you have the kernel boot logs? Is the SPI memory probed correctly?
> Did you put the SPI_NOR_4B_OPCODES info->flags in the relevant entry of the
> spi_nor_ids[] array? This is needed unless your board embeds a Spansion memory.
> 
> Without the SPI_NOR_4B_OPCODES flag, it should work just as before the
> patch: making the SPI nor memory enter its statefull 4-byte address mode.
> 
> With the flag, only the (Fast) Read, Page Program and Sector Erase op codes
> are substituted. So if the SPI memory fully supports the 4-byte address
> instruction set, there is no reason it doesn't work.
> 
> You may try to add some traces in the spi-nor.c driver to see what happens
> and where it blocks if spi_nor_scan() fails.
> 
> Best regards,
> 
> Cyrille

I'm going to take another look at it next week.

cstratton:
Stateless (or even better, only _ever_ operating the chip with one address
size) solutions are good where they work, but may require more precise chip
detection.

The remaining problem is chips like the MX25L25635E which unfortunately
seem to be in popular usage; these offer only software stateful mode
switch, they do not offer stateless 4-byte opcodes and they are not
configurable to 4-byte mode at power on (for that you have to buy a
different part number).

Thus using these as a boot source with an SoC like an MT7688 that needs a
predictable addressing mode to be found at reset is probably not a hardware
configuration that can be made safe by software; unfortunately it appears
such systems are being shipped (I say "appears" because SPI flash chip
identification printout especially in U-Boot will say this even when what
is actually installed is something like an MX25L25735F which has no support
for 3-byte addresses at all).

TeutonJon78:
The Archer C2600 v1.1 also has an affected NOR flash chip that is affected by the addressing size issue.

It has the mx25u25635f chip (so same family as the aforementined chip).

mkresin:
The best explanation for 3-byte and 4-byte thinghy I found so far can be read in the commit message introducing stateless 4-byte mode: http://lists.infradead.org/pipermail/linux-mtd/2013-March/046136.html

Summary:

* 3-byte address mode can only be used to address the first 16 MByte of the flash
* 4-byte address mode has to be used to address the flash behind this boundary
* there are two 4-byte modes:
    * stateless: flash is in 3-byte address mode and special (4-byte aware) op codes are used to read, program and erase the flash chip
    * statefull: 4-byte addresses are expected

The issue you are seeing is related to the "dumb boot sequence" performed by some SoCs, which expect the flash in 3-byte mode while reading the bootloader. For the SoC the 3-byte mode is fine since only the bootloader has to be read from flash and the bootloader is usually within the first 16MByte of the flash. But the kernel (and bootloader if kernel is behind the 16MB boundary) needs to enable the 4-byte mode to have access to the whole available flash.

Cyrilles patches only **allows to enable** stateless 4-byte on a per flash chip base. It is done by adding ```SPI_NOR_4B_OPCODES``` to the [spi_nor_ids array](https://github.com/torvalds/linux/blob/69973b830859bc6529a7a0468ba0d80ee5117826/drivers/mtd/spi-nor/spi-nor.c#L796) for flash chips supporting stateless 4-byte mode. As long as this flag isn't present, the stateless mode is only used for spansion flash chips by default. Long story short, add this flag to get it working.

The [MX25U25635F](http://www.macronix.com/Lists/Datasheet/Attachments/5174/MX25U25635F,%201.8V,%20256Mb,%20v1.5.pdf) used in the Archer C2600 v1.1 supports all required stateless 4-byte op codes according to page 18 - Read/Write Aray Commands (4 Byte Address Command Set). A patch for enabling stateless 4b support has been committed [upstream](https://github.com/spi-nor/linux/commit/b0fcb4b413028376894feaaaf62bcb09ab1b52f2) and was [backported to LEDE](https://github.com/lede-project/source/commit/226e194335c0d9a92402fbab772eac4f38786e45).

But ```SPI_NOR_4B_OPCODES``` will not work for all flash chips. The [Winbond W25Q256FV used in the Digineo AC1200 Pro/ZBT-WG3526 32MB](http://lists.infradead.org/pipermail/linux-mtd/2016-November/070368.html) for example, supports the stateless 4-byte op codes [only for reading and not for program or erase](http://lists.infradead.org/pipermail/lede-dev/2017-March/006752.html) (see [datasheet](https://www.winbond.com/resource-files/w25q256fv_revg1_120214_qpi_website_rev_g.pdf) page 25 - Instruction Set Table 1. For these chips we still need that clumsy "reset flash chip" or "switch chip to 3-byte mode" on shutdown. But even this would not fix the reboot bug entirely. If for example the watchdog resets the board or a reboot as a result of a kernel panic is done, the flash chip would be still in 4-byte statefull mode and the system hangs on reboot. The **only proper solution** for these flash chips is hardware design fix, by attaching the flash reset pin to the SoC reset pin or similar. For unknown reasons it isn't done by all OEM/ODM.

As I wrote in [PR#464](https://github.com/lede-project/source/pull/464#issuecomment-259898243) the/a chip reset should **not** be done unconditionally as it can be seen [here](https://lists.openwrt.org/pipermail/openwrt-devel/2015-July/034153.html), [here](http://patchwork.ozlabs.org/patch/721747/) and [here](https://patchwork.ozlabs.org/patch/541391/), since it isn't required for all flash chips. It should be done conditionally. One idea I had was a device tree property like ```reset-on-shutdown``` or a new [spi_nor_ids array](https://github.com/torvalds/linux/blob/69973b830859bc6529a7a0468ba0d80ee5117826/drivers/mtd/spi-nor/spi-nor.c#L796) flag. If possible the workaround should be accepted upstream or upstream should at least signalise it is the best we can do in software.

mkresin:
Beside the fact that there are still three commits instead of one, the PR as it is now causes build errors on bcm53xx, brcm63xx and layerscape. Cyrilles patches were backported to kernel 4.9 in LEDE with https://git.lede-project.org/174ce4c56d742712696e546eecee3d89e05c909a.

I would suggest to close this PR, since a lot targets got kernel 4.9 support in the meantime. And at least from my point of view it doesn't make much sense to backport these commits to the lede-17.01 branch and risk some kind of breakage.

cstratton:
For the Winbond W25Q256FV
<http://lists.infradead.org/pipermail/linux-mtd/2016-November/070368.html>the
chip itself can be persistently configured to start up in 4-byte mode.  If
the SoC can also be configured to expect four byte mode at boot (as for
example on the MT7688 with appropriate strapping) then simply leaving the
flash in 4-byte mode all the time works well; but any software (Linux or
U-Boot) that puts it back in 3 byte mode will indeed cause problems unless
a shared reset can be connected.

However, a shared reset is only valid in the case where an SoC's _internal_
watchdog and any other potential internal reset triggers will actually
_drive_ the external reset pin as an output or open-collector output.  A
few devices will do that, but many will not.

pepe2k:
@mkresin
> I would suggest to close this PR, since a lot targets got kernel 4.9 support in the meantime. And at least from my point of view it doesn't make much sense to backport these commits to the lede-17.01 branch and risk some kind of breakage.

If there are no more comments I also would like to see this closed.
[Your explanation](https://github.com/lede-project/source/pull/620#issuecomment-295401936) is perfect and, in my opinion, fully explains the problem.

Cheers,
Piotr

mkresin:
I'm going to close this one due to inactivity. Please reopen the PR if you have time to address the outlined issues.

corny:
There's a upcoming fix for LEDE: https://git.lede-project.org/?p=lede/nbd/staging.git;a=commit;h=c61c05e26b4bc63ef6044c7965c2d3a867584647

cstratton:
Sounds like another "fix" based on a fundamental misunderstanding of the
problem.

If an SoC requires the flash to be in one of 3 or 4 byte mode at boot, then
the flash _CAN NEVER EVER LEAVE THAT MODE_.

Otherwise you end up  with a system that silently hangs anytime there's an
unexpected reboot while the flash chip is in an unbootable mode.

The only way for a driver to know what mode is required at boot is to
either have that passed as a configuration option, detect that from SoC
specific properties, or detect it from chip specific properties or
persistent configuration registers.

zorun:
Thanks for the work, I have spent time to re-spin this for lede-17.01, see #1320

PR #621, opened 2016-12-17
pcmcia: new drivers and firmware

tmn505:
There are still some useful older VIA and Geode systems with PCMCIA bridges that can still be  repurposed for cheap router, with wan failover over cheap 3G modems for low bandwidth needs, hence this pull request. Small overview of the additions:

Sierra Wireless Aircards support
Option Fusion+ support
Cirrus Logic PD6729 bridge support
and small code rework.


blogic:
merged, Thanks !

PR #622, opened 2016-12-18
package/Makefile & ipkg-make-index.sh: add full package data list

bobafetthotmail:
The external script used to generate the package lists for the
LEDE wiki's table of packages [1] and package indexes [2] requires
a "Source:" field in the package lists to find package makefiles.
The package makefiles are used to read the package's Category and Submenu.

The "Source:" field was removed in commit
b4aa3c899cbf1f81cd8a1ea197481ac84a9d646e
to reduce package list sizes and lessen opkg issues in low ram devices.

Add a separate package list file with full data to be used by the wiki's script.
It's called Packages.manifest and isn't compressed as it's not necessary.

1. https://lede-project.org/packages/start
2. https://lede-project.org/packages/index/start

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -88,7 +88,8 @@ $(curdir)/index: FORCE
 	@for d in $(PACKAGE_SUBDIRS); do ( \
 		mkdir -p $$d; \
 		cd $$d || continue; \
-		$(SCRIPT_DIR)/ipkg-make-index.sh . 2>&1 > Packages && \
+		$(SCRIPT_DIR)/ipkg-make-index.sh . 2>&1 > Packages.manifest; \
+		grep -vE '^(Maintainer|LicenseFiles|Source|Require)' Packages.manifest > Packages && \
 			gzip -9nc Packages > Packages.gz; \
 	); done
 ifdef CONFIG_SIGNED_PACKAGES
--- a/scripts/ipkg-make-index.sh
+++ b/scripts/ipkg-make-index.sh
@@ -24,7 +24,7 @@ for pkg in `find $pkg_dir -name '*.ipk' | sort`; do
 	tar -xzOf $pkg ./control.tar.gz | tar xzOf - ./control | sed -e "s/^Description:/Filename: $sed_safe_pkg\\
 Size: $file_size\\
 SHA256sum: $sha256sum\\
-Description:/" | grep -vE '^(Maintainer|LicenseFiles|Source|Require)'
+Description:/"
 	echo ""
 done
 [ -n "$empty" ] && echo


bobafetthotmail:
I know the subject line is long but I could not shorten it more than that without dropping information.

package/Makefile
```diff
@@ -88,6 +88,7 @@ $(curdir)/index: FORCE
 	@for d in $(PACKAGE_SUBDIRS); do ( \
 		mkdir -p $$d; \
 		cd $$d || continue; \
+                $(SCRIPT_DIR)/ipkg-make-index.sh . -full 2>&1 > Packages-full ; \
 		$(SCRIPT_DIR)/ipkg-make-index.sh . 2>&1 > Packages && \
```
blogic: use tabs please. this change has not been tested. only send us tested changes.

scripts/ipkg-make-index.sh
```diff
@@ -21,11 +25,24 @@ for pkg in `find $pkg_dir -name '*.ipk' | sort`; do
 	sha256sum=$(openssl dgst -sha256 $pkg | awk '{print $2}')
 	# Take pains to make variable value sed-safe
 	sed_safe_pkg=`echo $pkg | sed -e 's/^\.\///g' -e 's/\\//\\\\\\//g'`
-	tar -xzOf $pkg ./control.tar.gz | tar xzOf - ./control | sed -e "s/^Description:/Filename: $sed_safe_pkg\\
+	if [ "$option" = "-full"  ]; then
+
+		tar -xzOf $pkg ./control.tar.gz | tar xzOf - ./control | sed -e "s/^Description:/Filename:
```
blogic: blank line

scripts/ipkg-make-index.sh
```diff
@@ -21,11 +25,24 @@ for pkg in `find $pkg_dir -name '*.ipk' | sort`; do
...
+	if [ "$option" = "-full"  ]; then
+
+		tar -xzOf $pkg ./control.tar.gz | tar xzOf - ./control | sed -e "s/^Description:/Filename: 
+$sed_safe_pkg\\
+Size: $file_size\\
+SHA256sum: $sha256sum\\
+Description:/"
+		echo ""
+
+	else
```
blogic: more blank lines

bobafetthotmail:
>the change is not tested. please only send us tested patches.

Umm, I tested it and it generates a package list called  Packages-full" with full info inside (i.e. it is working as intended), what makes you think it is not tested?

Will fix tabs and blank lines shortly.

blogic:
no you did not test the patch., you tested a similar patch on your laptop. you then used the brokrn github editor to try and recreate said patch and thus breaking it.

bobafetthotmail:
>no you did not test the patch., you tested a similar patch on your laptop. you then used the brokrn github editor to try and recreate said patch and thus breaking it.

FYI: Github webinterface allows only one modified file per commit (yes it's dumb) and does not allow me to squash commits without another commit nor edit commits in any way, and as you can see this is a single commit modifying 2 files.  
I had no other choice than making it locally from command line and push the commit to github.

I tested this by simply running a compile of default firmware for a random device using the same git branch I modified. 
It was done on my desktop with a Xeon E3-1275 V2 @ 3.50GHz processor, with a **make -j9**.
After it was done, I could see the new package list file in all the right places, and with non-filtered content (the folders with packages and the other folder with packages in /target).

Please say what makes this broken, it is formally wrong (and I will fix that as requested), but it ran fine in my test.

blogic:
adding spaces instead of tabs inside package/Makefile breaks it. please do not use the online editor in the future. the error rate is too high and I do not want to do your QA for you.

please send patches that you have previously tested and then send that exact patch file. if you try to make the changes base on that file using the web interface you will most likely end up sending us a broken patch such as this one.

bobafetthotmail:
That line in the Makefile is actually executed as shell, spaces don't break it so I couldn't notice that in my test.
There is a @ a few lines above (@ for d in ...) and at each endline since that symbol there is a \. 
That paragraph is actually a long "single" line executed as shell embedded in a makefile.

That said, fixed spaces and tabs, still works fine in my test.

blogic:
thanks,
@nbd168 can you comment on the actual patch please ?

jow-:
I'm okay with the idea but I'd like the file to be called `Packages.manifest` or something. Having a dash-suffix notation does not really fit with the rest of the files in that directory.

bobafetthotmail:
Updated. 

I changed the name of the generated list to to "Package.manifest" and also the "-full" CLI option to ipkg-make-index.sh has become "-manifest" as I like "manifest" more than "full" overall.

Tested and works.

nbd168:
Since generating the index takes a while, I'd prefer to generate the full index, then use grep to reduce it to a stripped down version, instead of running the script twice

bobafetthotmail:
ok, done. 

Now the script is reverted to before nbd168's patch. 
In the package/Makefile the script is called to generate a Package.manifest file with full information, 
then there is a grep that reads Package.manifest to generate the filtered package lists,
and from there it goes on as before.

Working fine in my tests.

nbd168:

PR #623, opened 2016-12-18
ath10k-ct: Enable DFS when "Enable DFS support" is set for kmod-ath

andrewstrohman:
This allows users of this package to configure DFS channels.
It mimics the behaviour of the ath10k module included in
package mac80211

Signed-off-by: Andy Strohman <andrew@andrewstrohman.com>


nbd168:

PR #624, opened 2016-12-19
update cake qdisc - again!

kdarbyshirebryant:
Track upstream changes...just when you thought they'd finished playing.

This time we've a simpler packet classifier mode 'diffserv3' using only 3 tins this time which is the new default.

'triple-isolate' has also been made the default flow/ip fairness mode.

Again, commits to both the kernel module (sch_cake) and iproute's tc are required to support the new 'diffserv3' mode.

Compile & run tested  AR71xx, Archer C7 v2


blogic:
merged, Thanks !

PR #625, opened 2016-12-19
libnl-tiny: define _GNU_SOURCE if not defined

commodo:
If _GNU_SOURCE was added as part of a package's TARGET_CFLAGS,
then compilation would fail for that module (especially if
warnings get treated as errors).


blogic:
merged, Thanks !

PR #627, opened 2016-12-19
procd: update procd.sh to support sending kill signal to a service

dedeckeh:
Add procd_send_signal which allows to send an optional specified
kill signal to one specified  or all instances of a given service.
By default SIGHUP is sent if no signal is specified


blogic:
merged, Thanks !

PR #629, opened 2016-12-20
ath10k-firmware: update qca9984 firmware

dissent1:
A new firmware that is available on code-aurora repository is newer than in Kvalo's repo.

Taking into account that firmwares that are in Kvalo's repo are considered to be tested by internal QCA team, this one seems to be more stable and fixes this bug:
ath10k_pci 0000:01:00.0: received unexpected tx_fetch_ind event: in push mode

At least i havent faced it for a while in contradiction to current version.

Thus switching firmware source for qca9984 until it or a newer version gets into Kvalo's repo.


nbd168:

PR #633, opened 2016-12-20
dnsmasq: add DHCP Unique Identifier for DHCPv6

arjendekorte:
Add DHCPv6 matching by DHCP Unique Identifier (RFC-3315) in addition to existing MAC-address (RFC-6939). The latter is not widely supported yet.

Signed-off-by: Arjen de Korte \<build+lede@de-korte.org\>

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -278,6 +278,7 @@ dhcp_host_add() {
 	config_get tag "$cfg" tag
 
 	if [ "$DHCPv6CAPABLE" -eq 1 ]; then
+		config_get duid "$cfg" duid
 		config_get hostid "$cfg" hostid
 		if [ -n "$hostid" ]; then
 			hex_to_hostid hostid "$hostid"
@@ -289,7 +290,7 @@ dhcp_host_add() {
 
 	config_get leasetime "$cfg" leasetime
 
-	xappend "--dhcp-host=$macs${networkid:+,net:$networkid}${broadcast:+,set:needs-broadcast}${tag:+,set:$tag}${ip:+,$ip${hostid:+,[::$hostid]}}${name:+,$name}${leasetime:+,$leasetime}"
+	xappend "--dhcp-host=$macs${duid:+,id:$duid}${networkid:+,net:$networkid}${broadcast:+,set:needs-broadcast}${tag:+,set:$tag}${ip:+,$ip${hostid:+,[::$hostid]}}${name:+,$name}${leasetime:+,$leasetime}"
 }
 
 dhcp_tag_add() {


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -289,7 +290,7 @@ dhcp_host_add() {
 
 	config_get leasetime "$cfg" leasetime
 
-	xappend "--dhcp-host=$macs${networkid:+,net:$networkid}${broadcast:+,set:needs-broadcast}${tag:+,set:$tag}${ip:+,$ip${hostid:+,[::$hostid]}}${name:+,$name}${leasetime:+,$leasetime}"
+	xappend "--dhcp-host=$macs${duid:+,id:$duid}${networkid:+,net:$networkid}${broadcast:+,set:needs-broadcast}${tag:+,set:$tag}${ip:+,$ip${hostid:+,[::$hostid]}}${name:+,$name}${leasetime:+,$leasetime}"
```
dedeckeh: According to the dnsmasq man page the id is both valid for both DHCPv4 and DHCPv6; for DHCPv4 it's mapped on the client identifier option and for DHCPv6 on the DUID option.
Therefore I propose to make the option id generic available for both DHCPv4 and DHCPv6

dedeckeh:
@arjendekorte I'm not sure the \ are allowed in the signed off line

bobafetthotmail:
@arjendekorte there is no need to escape the < and the > in the commit message's signed-off-by line.

package/network/services/dnsmasq/files/dnsmasq.init
arjendekorte: Great idea, but I don't think we already have something defined in uci to define the value of the DHCP client ID, so I have no idea what uci parameter we should use to fill this with. Maybe 'clientid' could be used, but without documenting there is probably little use to add this. The 'duid' on the other hand is already well documented. I don't see why this couldn't be merged.

dedeckeh: Why don't you call it id like dnsmasq is doing (or even clientid)?
It does not make sense to define a parameter in uci which is only valid for DHCPv6 while the underlying dnsmasq parameter is valid for both DHCPv4 and DHCPv6.

arjendekorte: Because 'duid' is already a well known parameter and documented. I didn't make this up, the name comes from https://wiki.openwrt.org/doc/uci/dhcp. Where 'duid' is specified, it is specific for DHCPv6 and should not be used for DHCPv4. So I repeat, if we want to use this, there needs to be a different uci parameter since for DHCPv4 and DHCPv6 this value has different meanings.

dedeckeh: If a uci parameter duid is introduced only valid for DHCPv6 and at a later point let's say clientid is introduced for DHCPv4 it can lead to config problems as both parameters map to the same dnsmasq option dhcp-host suboption id.
To which should we give precedence if both are specified ?
This is the problem I want to avoid by using two seperate options for DHCPv4 and DHCPv6

arjendekorte: There must be separate uci options for the DHCPv6 'duid' and DHCPv4 'clientid' because they are not the same, even dnsmasq (wrongly) treats them the same.
In a default LEDE/OpenWrt installation, DHCPv4 is handled by dnsmasq and DHCPv6 by odhcpd. So 'clientid' must in that case only be used by dnsmasq and 'duid' must end up only in the odhcpd configuration as they are not the same thing.

I'm not sure what to do in case dnsmasq is used for both DHCPv4 and DHCPv6. Maybe in that case we should opt for two dhcp-host lines, one for the DHCPv4 address using the 'clientid' and add another line for the DHCPv6 address with the 'duid'. I didn't check if dnsmasq allows listing the id: twice on the same dhcp-host line, in which case the solution would be trivial.

jow-: I'm with @arjendekorte here. Nothing in the dnsmasq man page indicates that multiple IDs are possible per `dhcp-host` so we should emit two distinct lines in case there is a conflict between `duid` and `clientid`.

jow-: @dedeckeh - since we do not implement `clientid` support yet and since it seems sensible to emit two different `dhcp-host` entries once there is support for `clientid` is there any other reason to not merge this PR?

dedeckeh: @jow- @arjendekorte I was planning to merge this PR as the odhcpd host implementation already accepts static IPv6 leases by also using duid as uci parameter

jow-: @dedeckeh - allright, I'll not interfere then :)

dedeckeh:
Pulled into my staging tree with corrected signed-off-by

PR #634, opened 2016-12-21
dnsmasq: use native 'interface-name' option for local host

EricLuehrsen:
- add_local_host UCI uses dnsmasq instance host file entry
- add_local_host implementation may drop some addresses (i.e. IP6)
- dnsmasq provides this function natively with 'interface-name'
- change add_local_host UCI to act during DHCP assignment
  - use 'interface-name' to assign host to all addresses on interface
  - assign '\<iface\>.\<host\>.\<domain\>' as the true name (PTR)
  - assign '\<host\>.\<domain\>' and '\<host\>' as convinience aliases (not tech. cnames)
- include new UCI add_local_gatename with similar function
  - most ISP don't rDNS (well), so logs/graphs nicer to read
  - default is off for those with dynamic DNS
  - infer WAN as DHCP 'ignore'

Signed-off-by: Eric Luehrsen <ericluehrsen@hotmail.com>

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -8,6 +8,7 @@ PROG=/usr/sbin/dnsmasq
 
 ADD_LOCAL_DOMAIN=1
 ADD_LOCAL_HOSTNAME=1
+ADD_WAN_HOSTNAME=0
 
 BASECONFIGFILE="/var/etc/dnsmasq.conf"
 BASEHOSTFILE="/tmp/hosts/dhcp"
@@ -293,6 +294,26 @@ dhcp_host_add() {
 	xappend "--dhcp-host=$macs${duid:+,id:$duid}${networkid:+,net:$networkid}${broadcast:+,set:needs-broadcast}${tag:+,set:$tag}${ip:+,$ip${hostid:+,[::$hostid]}}${name:+,$name}${leasetime:+,$leasetime}"
 }
 
+dhcp_this_host_add() {
+	# TODO: case-in do/not short-host or FQDN; UCI already intended ...
+	local ifname="$1"
+	local do_enable="$2"
+	local routerstub routername ifdashname
+
+
+	if [ "$do_enable" -gt 0 ] ; then
+		# All IP addresses discovered by dnsmasq will be labeled robustly (except fe80::)
+		ifdashname="${ifname//./-}"
+		routerstub="$( md5sum /etc/os-release )"
+		routerstub="router-${routerstub// */}"
+		routername="$( uci_get system @system[0] hostname $routerstub )"
+
+		xappend "--interface-name=$ifdashname.$routername.$DOMAIN,$ifname"
+		xappend "--interface-name=$routername.$DOMAIN,$ifname"
+		xappend "--interface-name=$routername,$ifname"
+	fi
+}
+
 dhcp_tag_add() {
 	local cfg="$1"
 
@@ -363,7 +384,11 @@ dhcp_add() {
 		DNS_SERVERS="$DNS_SERVERS $dnsserver"
 	}
 
-	append_bool "$cfg" ignore "--no-dhcp-interface=$ifname" && return 0
+	append_bool "$cfg" ignore "--no-dhcp-interface=$ifname" && {
+		# Many ISP do not have useful names for DHCP customers (your WAN).
+		dhcp_this_host_add $ifname $ADD_WAN_HOSTNAME
+		return 0
+	}
 
 	# Do not support non-static interfaces for now
 	[ static = "$proto" ] || return 0
@@ -380,6 +405,9 @@ dhcp_add() {
 	config_get leasetime "$cfg" leasetime
 	config_get options "$cfg" options
 	config_get_bool dynamicdhcp "$cfg" dynamicdhcp 1
+
+	# Put the router host name on this DHCP served interface address(es)
+	dhcp_this_host_add $ifname $ADD_LOCAL_HOSTNAME
 
 	leasetime="${leasetime:-12h}"
 	start="$(dhcp_calc "${start:-100}")"
@@ -605,6 +633,7 @@ dnsmasq_start()
 
 	config_get_bool ADD_LOCAL_DOMAIN "$cfg" add_local_domain 1
 	config_get_bool ADD_LOCAL_HOSTNAME "$cfg" add_local_hostname 1
+	config_get_bool ADD_WAN_HOSTNAME "$cfg" add_wan_hostname 0
 
 	config_get_bool readethers "$cfg" readethers
 	[ "$readethers" = "1" -a \! -e "/etc/ethers" ] && touch /etc/ethers
@@ -701,27 +730,6 @@ dnsmasq_start()
 	config_foreach filter_dnsmasq domain dhcp_domain_add "$cfg"
 	config_foreach filter_dnsmasq hostrecord dhcp_hostrecord_add "$cfg"
 	config_foreach filter_dnsmasq relay dhcp_relay_add "$cfg"
-
-	# add own hostname
-	[ $ADD_LOCAL_HOSTNAME -eq 1 ] && {
-		local lanaddr lanaddr6
-		local ulaprefix="$(uci_get network @globals[0] ula_prefix)"
-		local hostname="$(uci_get system @system[0] hostname Lede)"
-
-		network_get_ipaddr lanaddr "lan" && {
-			dhcp_domain_add "" "$hostname" "$lanaddr"
-		}
-
-		[ -n "$ulaprefix" ] && network_get_ipaddrs6 lanaddr6 "lan" && {
-			for lanaddr6 in $lanaddr6; do
-				case "$lanaddr6" in
-					"${ulaprefix%%:/*}"*)
-						dhcp_domain_add "" "$hostname" "$lanaddr6"
-					;;
-				esac
-			done
-		}
-	}
 
 	echo >> $CONFIGFILE_TMP
 	config_foreach filter_dnsmasq srvhost dhcp_srv_add "$cfg"
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -23,22 +23,22 @@ xappend() {
 }
 
 hex_to_hostid() {
-        local var="$1"
-        local hex="${2#0x}"     # strip optional "0x" prefix
+	local var="$1"
+	local hex="${2#0x}" # strip optional "0x" prefix
 
-        if [ -n "${hex//[0-9a-fA-F]/}" ]; then
-                # is invalid hex literal
-                return 1
-        fi
+	if [ -n "${hex//[0-9a-fA-F]/}" ]; then
+		# is invalid hex literal
+		return 1
+	fi
 
-        # convert into host id
-        export "$var=$(
-                printf "%0x:%0x"  \
-                        $(((0x$hex >> 16) % 65536)) \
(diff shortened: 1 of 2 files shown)


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -380,6 +394,16 @@ dhcp_add() {
 	config_get options "$cfg" options
 	config_get_bool dynamicdhcp "$cfg" dynamicdhcp 1
 
+	if [ $ADD_LOCAL_HOSTNAME -gt 0 ] ; then
+		local dashedname="$( echo $ifname | sed 's/\./-/g' )"
```
jow-: You can use `${ifname//./-}` here to avoid the need for spawning a `sed` subshell.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -380,6 +394,16 @@ dhcp_add() {
...
+		local routername="$( uci_get system @system[0] hostname OpenWrt )"
```
jow-: Hardcoding `OpenWrt` (or `LEDE`) here is not optimal. I suggest to source `/etc/os-release` and to use the `$ID` variable with a fall back to `OpenWrt`.

Example:

    local distribution="$( [ -f /etc/os-release ] && (. /etc/os-release; echo "$ID") || echo openwrt )"
    local routername="$( uci_get system @system[0] hostname "$distribution" )"

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -362,7 +363,20 @@ dhcp_add() {
 		DNS_SERVERS="$DNS_SERVERS $dnsserver"
 	}
 
-	append_bool "$cfg" ignore "--no-dhcp-interface=$ifname" && return 0
+	append_bool "$cfg" ignore "--no-dhcp-interface=$ifname" && {
+		if [ $ADD_LOCAL_GATENAME -gt 0 ] ; then
+			local dashedname="$( echo $ifname | sed 's/\./-/g' )"
```
jow-: You can use `${ifname//./-}` here to avoid the need for spawning a `sed` subshell.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -362,7 +363,20 @@ dhcp_add() {
...
+			local routername="$( uci_get system @system[0] hostname OpenWrt )"
```
jow-: Hardcoding `OpenWrt` (or `LEDE`) here is not optimal. I suggest to source `/etc/os-release` and to use the `$ID` variable with a fall back to `OpenWrt`.

Example:

    local distribution="$( [ -f /etc/os-release ] && (. /etc/os-release; echo "$ID") || echo openwrt )"
    local routername="$( uci_get system @system[0] hostname "$distribution" )"

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -23,22 +24,22 @@ xappend() {
...
+	# convert into host id
+	export "$var=$(
+		printf "%0x:%0x" \
+		$(((0x$hex >> 16) % 65536)) \
+		$(( 0x$hex        % 65536))
+		)"
 
-        return 0
+	return 0
 }
```
jow-: Please move those white space changes into a separate commit.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -8,6 +8,7 @@ PROG=/usr/sbin/dnsmasq
 
 ADD_LOCAL_DOMAIN=1
 ADD_LOCAL_HOSTNAME=1
+ADD_LOCAL_GATENAME=0
```
jow-: What is the meaning of `GATENAME`? Short hand for gateway hostname? If yes, why is it added for every non-dhcp interface?

Do we even need the distinction? You could make `ADD_LOCAL_HOSTNAME` control adding host names for both DHCP and no-DHCP interfaces.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -380,6 +394,16 @@ dhcp_add() {
...
+	if [ $ADD_LOCAL_HOSTNAME -gt 0 ] ; then
+		local dashedname="$( echo $ifname | sed 's/\./-/g' )"
+		local routername="$( uci_get system @system[0] hostname OpenWrt )"
+		# Put the router host name on this DHCP served interface address(es)
+		# All IP6 addresses discovered by dnsmasq will be labeled (except fe80::)
+		xappend "--interface-name=$dashedname.$routername.$DOMAIN,$ifname"
+		xappend "--interface-name=$routername.$DOMAIN,$ifname"
+		xappend "--interface-name=$routername,$ifname"
+	fi
+
```
jow-: This code block looks very similar to the "gatename" handling block above. Maybe you should consider moving it to a procedure?

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Sorry. Pushed a regex fix of my own mess.

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Yeah. Until recently my own test of the other was `gateway.out` but now that I have a sensible pattern...

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: -above-

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Its what was there. I didnt like it either, but it wasnt material to my goals.

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: -above-

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: I get frustrated with so many (non std) shell implementations, that I take the long way first to guarantee function.

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Yes, better explanation. I took `LOCAL_HOSTNAME` to mean your internal (business) network. If your exterior WAN however had a good name (or a few) in global DNS, then you may not want that sides name changed by this option. That would be for business class service from ISP. `LOCAL_GATENAME` allows you to split it. For most people with residential service on a consumer router, you will just set both true.

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Until recently my own test of the other was gateway.out but now that I have a sensible pattern... **DONE**

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: Its what was there. I didn't like it either, but it wasn't material to my goals. A fixed name or distro name has name collision potential still. Lets smash the MAC in there. **DONE**

package/network/services/dnsmasq/files/dnsmasq.init
EricLuehrsen: I get frustrated with so many (non std) shell implementations, that I take the long way first to guarantee function. DONE

EricLuehrsen:
@jow- All the requested changes have been fixed or explained. Please advise on next steps.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -604,6 +631,7 @@ dnsmasq_start()
 
 	config_get_bool ADD_LOCAL_DOMAIN "$cfg" add_local_domain 1
 	config_get_bool ADD_LOCAL_HOSTNAME "$cfg" add_local_hostname 1
+	config_get_bool ADD_LOCAL_GATENAME "$cfg" add_local_gatename 0
```
EricLuehrsen: Yes `add_wan_hostname` would be parallel. For 90%, **add_** is most meaningful, because ISP often dont bother with IP4 `c64-100-57-7.city.state.example.net` and never IP6 due to the wide complexity.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -292,6 +293,25 @@ dhcp_host_add() {
...
 }
 
+dhcp_this_host_add() {
+	# TODO: case-in do/not short-host or FQDN; UCI already intended ...
+	local ifdashname="$1"
+	local do_enable="$2"
+	local routerstub routername
+	
+	ifdashname="${ifdashname//./-}"
+	routerstub="$( cat /sys/class/net/br-lan/address )"
```
EricLuehrsen: oops... left my hard code for a quick test.

EricLuehrsen: Oops. That wasn't the best choice either. If no host name (what?), then just MD5 /etc/os-release.

dedeckeh:
@EricLuehrsen Patch fails to apply on latest trunk; can you rebase it ?
Could you make the commit comments of the second patch a bit more self explaining ?

EricLuehrsen:
**FIXED!**

EricLuehrsen:
**Fixed**

dedeckeh:
Pulled into my staging tree; thanks

stintel:
@EricLuehrsen this causes problems when a FQDN is configured in /etc/config/system. The domain name now appears twice in reverse DNS. Can you send a follow-up PR to fix this?

Next to that, there seems to be a bug in dnsmasq. From the manual page:
```
--interface-name=<name>,<interface>[/4|/6]
Return  a  DNS  record  associating  the  name  with  the primary address on the given interface. This flag specifies an A or AAAA record for the given name in the same way as an /etc/hosts line, except that the address is not constant, but taken from the given interface. The interface may be followed by "/4" or "/6" to specify  that  only  IPv4  or  IPv6 addresses of the interface should be used. If the interface is down, not configured or non-existent, an empty record is returned. The matching PTR record is also created, mapping the interface address to the name. More than one name may be associated with an interface address by repeating the flag; in that case the first instance is used for  the  reverse address-to-name mapping.
```
It does not just create an A/AAAA record for the primary address, it creates one for all addresses. And what is worse, it seems to actually resolve to the non-primary address first. This is quite annoying when you use floating IP addresses (e.g. VRRP), because when the floating IP is on the other device, SSH failes due to incorrect entry in the known hosts file.

I know that this is not a common setup, but it would be nice if there was an option to restore the previous behaviour, rather than just forcing this new feature on everybody.

EricLuehrsen:
For issue (1) there are few possible fixes and some more robust for wider array of users. I guess I never though of host name as FQDN; normally you let the domain controller append that for you. (A) if host name has dots, don't add anything to it (B) look for match from end, and squelch the overlap (C) use the UCI which does this also for DHCP and /etc/hosts ... (D) ?

For issue (2) I am not clear on the problem. The old way got the IP4 and list of IP6 addresses, ran a for-do, and labeled **everything** with no filters. dnsmasq does the same thing. It seems transparent.

EricLuehrsen:
I was editing my response when github went down. I have no idea what nonsense I left behind. I guess I will update later.



- Eric


-------- Original message --------

dedeckeh:
@EricLuehrsen As the commit introduces regression issues and since we have a release coming up the best way to go forward is to revert the commit for the moment.
After the release branch off a new PR can be integrated fixing those issues.

PR #635, opened 2016-12-21
Add support for different /etc/dnsmasq.conf file for each dnsmasq instance

HRomie:
the initial multiple instance support is already there but they all share the same /etc/dnsmasq.conf file which is not ideal. this simple patch would allow having different conf file if /etc/dnsmasq.conf.${cfg} exists. otherwise it falls back to the default file.

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -544,7 +544,7 @@ dnsmasq_start()
 	if [ ! -r "$dnsmasqconffile" ]; then
 		dnsmasqconffile=/etc/dnsmasq.conf
 	fi
-	[ -f $dnsmasqconffile ] && {
+	[ -r $dnsmasqconffile ] && {
 		xappend "--conf-file=$dnsmasqconffile"
 	}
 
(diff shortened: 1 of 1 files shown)


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -540,8 +540,12 @@ dnsmasq_start()
 	echo "# auto-generated config file from /etc/config/dhcp" > $HOSTFILE
 
 	# if we did this last, we could override auto-generated config
-	[ -f /etc/dnsmasq.conf ] && {
-		xappend "--conf-file=/etc/dnsmasq.conf"
+	local dnsmasqconffile="/etc/dnsmasq.conf.${cfg}"
```
HRomie: I am still unsure whether i should go with _/etc/dnsmasq.conf.${cfg}_ or _/etc/dnsmasq.${cfg}.conf_ . i decided to choose the former as it matches the rest of the vars assignments

EricLuehrsen:
Good form going forward would be `dnsmasq.<cfg>.conf`. I find that the former is a hack in dnsmadq.init because it was easy to append. I would not copy that. The challenge is that years of patching just-one-fix
have undone the sense of structure.

EricLuehrsen:
Note commit title form (and pull req), prefix with package and short description. Commit body is a bit more libral. Also squash (or fixup) your fix-up commits with interactive rebase.
`dnsmasq: add per instance custom conf files
- problem/desire/function
- add/delete/change`

HRomie:
@EricLuehrsen thanks for the notes. i'm really newbie on this. i'm gonna fixup the patch and also learn some basic git commands before requesting

PR #638, opened 2016-12-21
dnsmasq: add log facility option

dibdot:
add possibility to set the facility to which dnsmasq will send syslog entries, i.e. set it to '/dev/null' to mute dnsmasq output at all.

Signed-off-by: Dirk Brenken dev@brenken.org

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -573,6 +573,7 @@ dnsmasq_start()
 	append_bool "$cfg" allservers "--all-servers"
 	append_bool "$cfg" noping "--no-ping"
 
+	append_parm "$cfg" logfacility "--log-facility"
 	append_parm "$cfg" dhcpscript "--dhcp-script"
 	append_parm "$cfg" cachesize "--cache-size"
 	append_parm "$cfg" dnsforwardmax "--dns-forward-max"


EricLuehrsen:
I would suggest creating a combined UCI level-option to help simplify the user interface. dnsmasq has a number of log preferences including `--quiet-ra`, `--queit-dhcp`, various DNS query, and other things. `verbosity` or something where 0-9 and 0=/dev/null and 9 turns on the noisy RA log. I already have done something similar for myself, but my entire dnsmasq.init is so far off tree that its taking me time to piece it in.

General bias on my part, a UCI flag for every dnsmasq flag does not help simplify matters for the average -ish user. The conf file insert option provides highly refined options for advanced users.

dibdot:
Keep it simple.
A new, combined 'verbosity' or 'loglevel' flag sounds interesting, but this should preferably be implemented upstream, not in LEDE.

EricLuehrsen:
Simple for _who_ is the question ... upstream already provides all the options you could want and then yet more. Make this easier to deal with for the end user, LuCI design, and those helping others debug. The following is quite simple, and scalable.
```
case "$verbosity" in
0) 
  xappend "--log-facility=/dev/null"
  ;;
...
5) ... --quiet-dhcp ... --quiet-dhcp6 ... --quiet-ra ...
...
11)
  # "Eleven. Exactly. One louder. (TIST 1984)"
  xappend "--log-async"
  xappend "--log-queries=extra"
  xappend "--log-dhcp"
  xappend "--dnssec-debug"
  ;;
esac
```

dibdot:
Simple for the user: google for dnsmasq and you'll find almost every time the man page of the dnsmasq author, where you have a description of separate options and not a combined flag.

Just my two cents.

dedeckeh:
Introducing a verbose uci parameters is not an optimal solution as it would cause overlap as uci parameters for quietdhcp, logqueries are already in place.
One of the good things about dnsmasq is the man page documentation which currently allows you to map the uci parameters easily on the described options; adding a verbose uci option would require extra documentation in the Lede project.
Also take in mind the log-facility allows you to write to a file; use a verbose uci parameter possible takes this flexibility away.
If verbose abstraction is required this should be done in Luci rather than in uci

dedeckeh:
Pulled into my staging tree; thanks

PR #639, opened 2016-12-21
mac80211: Allow HT/VHT rates when running unencrypted mesh.

alexgrin:
When using mesh mode without encryption, the devices will come up in 20Mhz no-HT mode no matter what the radio settings are. This patch brings it up with correct settings.

Tested on Archer-C7 v2 (QCA9880, ath10k).
This patch was also tested on a forked branch of lede with Asus RT-ACRH13 (IPQ4018, ath10k)


jow-:
LGTM. @nbd168 ?

nbd168:
LGTM

jow-:
Pulled into my staging tree: https://git.lede-project.org/d007501

PR #640, opened 2016-12-22
ipq806x: fix pvs1_bin voltage in ipq8065 DT

dissent1:
Fixing pvs1 bin voltage as found in GPL tarball.

Also adjusting smb208_s2a/b minimum voltage to reflect lowest value in pvs table.

### DIFF
--- a/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8065.dtsi
+++ b/target/linux/ipq806x/files/arch/arm/boot/dts/qcom-ipq8065.dtsi
@@ -222,11 +222,11 @@
 			 < 600000000 1000000 >,
 			 < 384000000 975000 >;
 		qcom,speed0-pvs1-bin-v0 =
-			< 1725000000 1262500 >,
-			< 1400000000 1175000 >,
-			< 1000000000 1100000 >,
-			 < 800000000 1050000 >,
-			 < 600000000 1000000 >,
+			< 1725000000 1225000 >,
+			< 1400000000 1150000 >,
+			< 1000000000 1075000 >,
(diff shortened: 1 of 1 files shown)


blogic:
good to merge or is testing required ?

dissent1:
Sorry, yep, it's good, runtime tested on Netgear R7800.

Thanks!

alexgrin:
Does this improve any of the functionality? I'm curious because I just got back to testing with R7800s.

dissent1:
Nope, at least not now. It's just fixups of what I have found wrong. 
The thing is that Netgear R7800 chooses pvs bin n4, this number seems to be taken from eFuses and that has correct values in DT.
There may be a situation in future that some new device may choose pvs bin n1 or n6.  
Without this changes n1 is semi-correct and n6 is not available, because OPP driver disregards bins with voltage values that regulator cannot supply.

mkresin:
@dissent1 please add this info to the commit message:

> The thing is that Netgear R7800 chooses pvs bin n4,

> Without this changes n1 is semi-correct and n6 is not available, because OPP driver disregards bins with voltage values that regulator cannot supply.

The commit message should explain why you have done something. Please make clear that the change to regulator-min-microvolt is related to pvs bin n6.

I had exact the same question as alexgrin and only with your extra explanation this PR makes sense to me.

dissent1:
It's clearly stated that pvs bin n1 is fixed to reflect gpl tarball and smb regulator minimum voltage is adjusted to reflect lowest value in pvs table and as you can see the lowest value is in bin n6.

But ok.

blogic:
merged, Thanks !

PR #643, opened 2016-12-23
ath10k-firmware: update board data for qca9984

dissent1:
Current board-2.bin file for qca9984 in Kvalo's repo is from branch 10.4-3.2, while board-2.bin file in code-aurora repo is newer and corresponds to the branch 10.4-3.3, the same as recently updated firmware.

While updating qca9984 firmware I haven't noticed that board-2.bin has been updated in code-aurora several commits before. 
I haven't noticed any changes in my use cases, but I think it's better to update it to the same branch as firmware.

### DIFF
--- a/package/firmware/ath10k-firmware/Makefile
+++ b/package/firmware/ath10k-firmware/Makefile
@@ -177,9 +177,20 @@ define Download/qca99x0-board
 endef
 $(eval $(call Download,qca99x0-board))
 
+QCA9984_BOARD_REV:=deb1832c56c706d0f6cb539113e09f0daaa52b5f
+QCA9984_BOARD_FILE:=board-2.bin
+QCA9984_BOARD_FILE_DL:=$(QCA9984_BOARD_FILE).$(QCA9984_BOARD_REV)
 QCA9984_FIRMWARE_REV:=deb1832c56c706d0f6cb539113e09f0daaa52b5f
 QCA9984_FIRMWARE_FILE:=firmware-5.bin_10.4-3.3-00102
 QCA9984_FIRMWARE_FILE_DL:=$(QCA9984_FIRMWARE_FILE).$(QCA9984_FIRMWARE_REV)
+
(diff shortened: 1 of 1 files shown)


jow-:
LGTM. @nbd168 ?

stintel:
NAK - see http://lists.infradead.org/pipermail/lede-dev/2016-September/002836.html

dissent1:
The point is that with commit https://github.com/lede-project/source/commit/0ee964f28407aca0934455f81bb4b9546f8a3c11#diff-7767c6da016b31deb2fd133a9a1a4b59 firmware from code aurora has already been taken into lede because it has several fixes and seems to be as stable as in Kvalo's repo.
With this commit I was just getting the second part of the firmware - the board file - to get in line with firmware itself and its branch (10.4-3.3 instead of 10.4-3.2 used before). Also this one is 10 times bigger than the one used atm.
Just FYI.

stintel:
Then we should revert 0ee964f, IMO.
@jow- @nbd168 ?

dissent1:
Erm, why? In CA It's more stable than from Kvalo's and it fixes issue that occasionally drops throughout and spams dmesg, as said in commit message.

nbd168:
@stintel 
Regarding the email that you linked to: When that issue came up last, Kalle's releases were working made from a different branch than the codeaurora binaries.
In this case the codeaurora binary is from the same branch as Kalle's binary, just newer.
I would lean towards accepting this pull request as well

The ath10k firmware situation is basically one big incomprehensible mess, and it's never just one branch is the right one and the other one isn't. QCA has a truckload of different branches for the same things internally, and sometimes one happens to work better than the other.

I bet a lot more on test results by LEDE users than test results by one of the QCA internal test teams.

stintel:
@nbd168 fair enough. I'll reopen the PR. Will you take it?

nbd168:
Merged to my staging tree

PR #644, opened 2016-12-23
b53 switch driver: add get_port_stats()

p-wassi:
Maintainer: @KanjiMonster 
Compile tested: brcm63xx
Run tested: brcm63xx

Description:
Add get_port_stats() to the switches' ops. This lets switch LEDs blink on port traffic.
This function just reads the switches TxOctets and RxOctets registers (as seen in the MIBs)
for the corresponding port and returns these values.
Tested on a BCM63xx switch.

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/b53/b53_common.c
+++ b/target/linux/generic/files/drivers/net/phy/b53/b53_common.c
@@ -835,6 +835,28 @@ static int b53_port_set_link(struct switch_dev *sw_dev, int port,
 	return switch_generic_set_link(sw_dev, port, link);
 }
 
+static int
+b53_get_port_stats(struct switch_dev *dev,  int port,
+			struct switch_port_stats *stats)
+{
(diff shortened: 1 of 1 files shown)


p-wassi:
Hmm.... While making LEDs blink with above mentioned patch works,
is there any other need for having the get_port_stats() function?
I mean since it's already done and tested... or can I trash this code?

KanjiMonster:
I don't think it makes sense, unless we ever see a device actually needing that. Most SoCs with internal switches have the appropriate leds led out as gpios (sometimes two per port even).

danitool:
The HG556a, A, B and C has two LEDs on each port (BCM5325) wired to GPIOs. Also the Sagem F@ST2304 (still not supported, BCM6328) has the switch LEDs wired to GPIOs without HW pinmux available.

Right now the patch could be useful for the HG556a, but not sure if worth of it.

Regards.

PR #647, opened 2016-12-23
ramips: fix tp-link images

heinzek:
- Fix build error:
/etc/init.d/enablemodem: line 2: /lib/ramips.sh: No such file or directory

- fix tplink C20i, C50 and MR200 image building
- add SUPPORTED_DEVICES to fix sysupgrade firmware on MR200
based on @mkresin commit: https://git.lede-project.org/?p=lede/mkresin/staging.git;a=commitdiff;h=4d3360dad26a8bb69a3f925d0528232fe3de31db


blogic:
thanks for the fix, merged !

PR #648, opened 2016-12-23
brcm63xx: add support for Pirelli/ADB P.DG AV4202N

p-wassi:
This patch adds support for the ADB P.DG AV4202N aka Pirelli PRG AV4202N [1].
In order to get switch triggered LEDs working, SWCONFIG_LEDS is
enabled in kernel configuration. (Here, also [2] comes into play)

What does not work: xDSL part (Broadcom proprietary), 2x FXS port (but there's a solution to that)
What works: Ethernet + switch (VLAN), Wifi, GPIO LEDs+Btn, USB, sysupgrade
See more details on the device in [3], also short info on FXS is there.

Although we don't have DSL support, the device features 16MB Flash, 64MB RAM, 2x 400MHz MIPS32, 2x USB2.0 and therefore is worth taking a look at.

What I'm note sure about: should I now update the info in OpenWrt's wiki at [1] (support status, installation procedure, ...)?

[1]:
https://wiki.openwrt.org/toh/adb/p.dg_av4202n

[2]:
https://github.com/lede-project/source/pull/644

[3]:
https://pwassi.privatedns.org/lede/av4202/

### DIFF
--- a/target/linux/brcm63xx/base-files/etc/board.d/01_leds
+++ b/target/linux/brcm63xx/base-files/etc/board.d/01_leds
@@ -16,6 +16,9 @@ a4001n1)
 a4001n)
 	ucidef_set_led_usbdev "usb" "USB" "A4001N:green:usb" "1-1"
 	;;
+av4202n)
+	ucidef_set_led_netdev "wlan0" "WLAN" "AV4202N:blue:wifi" "wlan0"
+	;;
 dgnd3700v1_dgnd3800b)
 	ucidef_set_led_netdev "lan" "LAN" "DGND3700v1_3800B:green:lan" "eth0.1"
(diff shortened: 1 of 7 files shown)


danitool:
Hi p-wassi, there is a pending patch for fixing the switch LEDs problem on all BCM6368 boards.
https://patchwork.ozlabs.org/patch/702128/

With this patch the LEDs at the switch should be controled via hardware (assuming CFE previously configured the pinmux). Delete all switch LEDs related stuff, and test it.

I don't think using the kernel switch LED trigger is the way for making them work if the SoC provides itself the same functionality.

p-wassi:
Thanks danitool for pointing out that there is this patch.
I've successfully tested the mentioned patch on my device, therefore closed
my other PR.
Also, this PR is updated (LAN LEDs + surroundings removed)

KanjiMonster:
Added, thanks for contributing.

PR #649, opened 2016-12-25
FB7360SL: reset PHYs when link status changes

LipkeGu:
Due to a hardware bug of Atheros at803x phys, the driver need to reset
the phys on link state change. (see: [drivers/net/phy/at803x.c:337]
The bug occurs when the FIFO Buffer contains queued TX/RX frames.

* Use the correct compatible string for the at803x phys connected to
switch port 0 and 1.

* Fix the pinmux of the gpio lines connected to the reset pin of the phys
and define the reset-pins to let the driver do the fixups.

Reproduce case:
1. Start a Download (a large linux ISO image should be enough >= 2GB)
2. Unplug many times the cable while downloading the iso file.
3. After 3 seconds Plugin the cable.
-> After repeating 3x times the steps (2. and 3.) your NIC connection LED should indicate a "on-off-on-off.. state. 

Signed-off-by: Mathias Kresin <dev@kresin.me>
Tested-by: Guido Lipke <lipkegu@gmail.com>

### DIFF
--- a/target/linux/lantiq/dts/FRITZ7360SL.dts
+++ b/target/linux/lantiq/dts/FRITZ7360SL.dts
@@ -13,7 +13,6 @@
 		led-running = &power_green;
 
 		led-internet = &info_green;
-		led-dsl = &power_green;
 		led-wifi = &wifi;
 	};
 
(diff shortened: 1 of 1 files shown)


mkresin:
I've no idea why you are opening a PR with my patch. I've committed the changes > 24h ago:

* https://git.lede-project.org/7dd42d610dcd6f9271d5d44edff52caf633b8abf
* https://git.lede-project.org/ea9b71e5b1de8e40a7f92d10bfc1761a8955a2e8

PR #651, opened 2016-12-27
hostapd: expose wps_independent and ap_setup_locked as uci options

snh:
The first configuration option, `wps_independent`, determines if WPS operations, such as a WPS push button trigger event, should apply to all wireless networks within the same `hostapd` process ([the current `hostapd` default](http://lists.infradead.org/pipermail/hostap/2016-March/035301.html)), or only to the relevant wireless network (any non-zero value). 

The UCI option is named the same as the `hostapd` option, `wps_independent`, as this is already in line with the existing UCI naming for WPS options.

The default `wps_independent` setting of `0` prevents you from selectively enabling and using WPS push button configuration on a subset of wireless networks, such as a Guest network, whilst leaving WPS disabled on an Internal network. An OpenWrt [bug report](https://dev.openwrt.org/ticket/19485) and [forum topic](https://forum.openwrt.org/viewtopic.php?pid=272476) exists for this particular issue, which is resolved by setting `wps_independent` to a non-zero value.

In some circumstances, the default setting can also lead to enrollees being admitted to the wrong wireless network within the same `hostapd` process.

The second configuration option, `ap_setup_locked`, when set to `1`, will disallow new registrars. This is currently hardcoded to `0` in LEDE, disabling this locking. Alternatively, a value of `2` can be used to enable read only access to AP settings.

To group this with the other WPS options, the UCI option name of `wps_ap_setup_locked` was selected, which differs from the `hostapd` option name `ap_setup_locked`.

Whilst outside the scope of this pull request, changing the LEDE default for one or both of these options to `1` may be worth considering.

Signed-off-by: Steven Honson <steven@honson.id.au>

### DIFF
--- a/package/network/services/hostapd/files/netifd.sh
+++ b/package/network/services/hostapd/files/netifd.sh
@@ -150,6 +150,7 @@ hostapd_common_add_bss_config() {
 	config_add_string wpa_psk_file
 
 	config_add_boolean wps_pushbutton wps_label ext_registrar wps_pbc_in_m1
+	config_add_int wps_ap_setup_locked wps_independent
 	config_add_string wps_device_type wps_device_name wps_manufacturer wps_pin
 
 	config_add_boolean ieee80211r pmk_r1_push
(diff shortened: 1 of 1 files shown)


jow-:
Given the fact that we currently handle WPS settings as per-wifi-iface settings in UCI I'd strongly favor making `wps_independent` the default as anything else seems highly counter intuitive (and thus securtiy relevant).

Given that I wonder if we even want to have an option to *disable* `wps_independant` as it would somehow counteract our configuration structure.

LGTM on the `wps_ap_setup_locked` change. @nbd168 - thoughts?

nbd168:
I agree with what @jow- said. If `wps_independent` works the way you describe it, it should be on by default

snh:
@jow- @nbd168 Thanks for the feedback. I have changed the default as suggested.

jow-:
Pulled into my staging tree: https://git.lede-project.org/50adc0a

PR #652, opened 2016-12-29
elfutils: bump to 0.168

luizluca:
Compile tested on ramips.

Other changes:
- Project moved to sourceware.org
- musl patch where cleaned up and submitted upstream
- TEMP_FAILURE_RETRY macro fixed and submitted upstream

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>


jow-:
@luizluca - in the musl compat patch a lot of `<error.h>` includes are swapped with `<err.h>` ones and I wonder why this is needed. Was that done because musl provides no `error.h`?

Since musl `err.h` and Glibc `error.h` provide totally different functions I wonder if that patch actually just silences the missing include file error - in this case we could try something like `touch error.h` in the package build directory.

jow-:
@luizluca - ok I see that you're in contact with upstream about this already so LGTM.

jow-:
Pulled into my staging tree: https://git.lede-project.org/da2d9ae

luizluca:
Thanks for the review!

@jow- , they need error() from error.h in all those files. If they accept my suggestions, these includes would be centralized into the system.h and get my patch much smaller. error() isn't easily replaceable as it has a global variable to keep the number of times it was called. The libs use by OpenWRT/LEDE do not use this number of calls info but some of elfutils apps do. If I wish to get musl compatibility upstream, I need to do it the right way.

As you might have faced this kind of problem hundreds of times, maybe you could give me some tips. 😀
Should I inject the global variable into an existing (and linked by all) c file (which I don't know if it exists)? Or would it be better to create a new .c and conditionally include it when error.h is missing? Or even create an alternative static library for used error.h functions?

I'm trying to reduce the size and the number of patches as to much of them make maintenance harder, specially musl patch. It's now smaller but still big enough to always fail to patch automatically on every update. My goal is to have all patches converted into configure options or checks.

PR #653, opened 2016-12-30
curl: Switch default TLS library to mbedTLS

neheb:
luci-ssl has already made the switch since mainline support for PolarSSL is almost over (2016).


hnyman:
I think that you should also increase PKG_RELEASE in Makefile.

diizzyy:
mbedTLS 1.3 support should be dropped entirely, we don't want to have unmaintained code in the tree that's going to be a security issue.

neheb:
Fixed. Will also do a pull request on px5g.

diizzyy:
@neheb 
Already done :-)

diizzyy:
@neheb 
Please remove all references (options) to use PolarSSL (mbed TLS 1.3).
The rest of the work is in #655

neheb:
Adjusted commit message. I think that's it.

hnyman:
This needs to be rebased, as @hauke updated curl to 7.52.1 with 1436e1548881a8e196eeb25c7c807e2c3d65caf1

jow-:
Pulled into my staging tree: https://git.lede-project.org/4360320

jow-:
Rebased and line-wrapped: https://git.lede-project.org/e021ae6

hnyman:
> Rebased and line-wrapped: https://git.lede-project.org/e021ae6

@jow- 
minor thing, but the new patch likely needs to be refreshed due to the version bump
(based on line number difference in new/deleted patch: 712 vs. 724)

jow-:
@hnyman - ack, refreshed.

PR #655, opened 2016-12-30
Remove mbed TLS 1.3 from tree due to EOL

diizzyy:
Removes mbed TLS 1.3 as it's EOL at the end of 2016.
Also removes support for the ustream-ssl and px5g packages.
For consistency px5g is named px5g-mbedtls.


hnyman:
Removing polarssl totally right now (possibly just before a release branching) might be too drastic, as there are several packages in the packages feed still utilising it. 
https://github.com/openwrt/packages/search?l=Makefile&q=polarssl&utf8=%E2%9C%93

thess:
As maintainer of a couple of the packages mentioned, I think it premature to remove polarssl until these packages can be updated or fixed. At least one of them has been ported, by me, to use mbedTLS but is not available upstream.. The sources would have to come from a private repo or support several patches.

diizzyy:
That's easily fixed and given the conversion on IRC and partially on the mailinglists general consensus were that dropping a encryption library that's EoL is the way to go. There's neither interest or manpower to support mbed TLS 1.3 without no upstream support for several months. Unless you're willing to pick up that ball...

I can provide patches for the rest within a few days.

wongsyrone:
As long as I know, almost all symbols and functions are just renamed with "mbedtls_" prefix, only a little structures become opaque, the usage may get changed, which will introduce time cost.

hnyman:
@diizzyy 
I agree that EOLed SSL libraries need to phased out and be removed, eventually. 

But it might be easier to implement this in two phases:
* First replace polarssl as a default from packages in the core repo (px5g, curl). You have code here for px5g and #653 handles curl. This step could be done immediately.
* Then convert the packages in feed repos to use other SSL libraries, and only then remove the polarssl library package itself (and the libustream-polarssl variant of ustream-ssl). It is quite possible that some of the feed packages may require more work, so this step may take some time.

diizzyy:
@hnyman 
Whether we phase it out in LEDE repo first or not doesn't really matter in the end as very few external packages are affected and most a non crucial. IMHO it actually might be even beneficial to drop it in LEDE first since we'll see "hidden" breakage early on instead of having to do panic fixes just before release. It's already more or less decided that it should go before release so I think it's a good idea to merge it a soon as possible.

thess:
On Fri, 2016-12-30 at 18:02 -0800, Syrone Wong wrote:
> As long as I know, almost all symbols and functions are just renamed with
> "mbedtls_" prefix, only a little structures become opaque, the usage may get
> changed, which will introduce time cost.
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.

Having done a few of these conversions, I would say for the most part you are
correct. However, dealing with certificates (x509 functions) and encryption (rsa
functions) is not a one-to-one translation. That said, it is usually less that a
day's work to convert and test.

/ted

hnyman:
@nbd168 @jow- 
Any chance that the px5g and ustream-ssl changes could be cherry-picked from this PR already now?

* nothing selects libustream-polarssl currently, but it would be great to get it deprecated from users' configs.
* px5g has no default variant declared, so it would be better to just have the mbedtls version available.

Regarding the polarssl usage, most packages have been converted by now. To my knowledge, the last users are umurmur, bmx7, shairport. (additionally shadowsocks conversion is likely ready after enabling CFB. hopefully...)

diizzyy:
It's probably better to remove it completely as the only remaining port is shairport-sync which Ted is working on. The rest are just pull requests waiting to be approved and/or merged.

nbd168:
px5g-polarssl and ustream-polarssl are now removed from the tree

hnyman:
@nbd168 
I think that that one use of (BUILD_VARIANT) got left behind in your change:

https://github.com/lede-project/source/blob/master/package/utils/px5g/Makefile#L13
``` 
PKG_BUILD_DIR:=$(BUILD_DIR)/px5g-$(BUILD_VARIANT)
``` 
Looks harmless, but should maybe be tidied up.

nbd168:
@hnyman done, thx.

diizzyy:
@nbd168 @jow- 
As of a few seconds ago we now have no known users of PolarSSL/mbed TLS 1.3, can we remove the package now? :-)

diizzyy:
Closing this as it's been applied to master branch

Issue #162, opened 2016-06-30
[solved] Archer C7 v2 - no Wireless Intefaces


ds2k5:
Hello,
i build a Image for TP Link Archer C7 v2 with Luci ( see .config as config.txt )
But there a no wireless Interfaces.
What did i wrong ?
The Wireless Switch at the device is ON
Thanks
[archer-c7-v2_dmesg.txt](https://github.com/lede-project/source/files/340860/archer-c7-v2_dmesg.txt)
[archer-c7-v2_ifconfig.txt](https://github.com/lede-project/source/files/340861/archer-c7-v2_ifconfig.txt)
[archer-c7-v2_lsmod.txt](https://github.com/lede-project/source/files/340862/archer-c7-v2_lsmod.txt)
![archer-c7-v2_luci](https://cloud.githubusercontent.com/assets/11292286/16479751/9c9e9e62-3ea2-11e6-90f0-e8f4fc8fda85.png)
[config.txt](https://github.com/lede-project/source/files/340865/config.txt)

jow-:
Can you provide the output of `iw list` ?

jow-:
Ok so the wifi driver initialized at least. Can you paste the last 30 lines or so of `logread` after you issued a `wifi up` on the terminal?

ds2k5:
here it is
[archer-c7-v2_logread.txt](https://github.com/lede-project/source/files/341022/archer-c7-v2_logread.txt)

jow-:
The `INVALID_WPA_PSK` in your log means that your provided wifi password is shorter than 8 characters - can you try one which is at least 8 characters long?

ds2k5:
No, my password is: 64 charters long (lower, uppercase, numbers and special charters ( example !&$§ )

in /etc/config/wireless i can see the password
in the line: 	option password 
us WPA2-PSK

blogic:
does it work if you use 01234567 as a password ?

jow-:
64 characters is too long :) If your PSK is exactly 64 byte it must be in hexadecimal notation (e.g. as generated by the `wpa_passphrase` command.) Plaintext passwords must be between 8 and 63 chars.

ds2k5:
thanks jow and blogic

i found the issue. 2,4 GHZ is working now (with my long password)
I need to add the password at the line: option key
the WebGUI looks diffrent since reboot.
Don't know why. Before there are lines for raduis server etc. but i selected wpa2-psk

[...]

PR #657, opened 2017-01-01
mbedtls: Enable CFB, RC4 and Camellia

diizzyy:
Enable Cipher Feedback mode (CFB), RC4 and Camellia as these
are required by some packages such as upstream shadowsocks-libev.

These changes makes the IPK package about 6kbyte larger on ar71xx.

### DIFF
--- a/package/libs/mbedtls/Makefile
+++ b/package/libs/mbedtls/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mbedtls
 PKG_VERSION:=2.4.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_USE_MIPS16:=0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-gpl.tgz
--- a/package/libs/mbedtls/patches/200-config.patch
+++ b/package/libs/mbedtls/patches/200-config.patch
@@ -9,15 +9,6 @@
  
  /* \} name SECTION: System support */
  
-@@ -347,7 +347,7 @@
-  *
-  * Enable Cipher Feedback mode (CFB) for symmetric ciphers.
-  */
--#define MBEDTLS_CIPHER_MODE_CFB
-+//#define MBEDTLS_CIPHER_MODE_CFB
- 
- /**
-  * \def MBEDTLS_CIPHER_MODE_CTR
 @@ -441,17 +441,17 @@
   *
   * Comment macros to disable the curve and functions for it
@@ -174,24 +165,6 @@
  
  /**
   * \def MBEDTLS_X509_ALLOW_EXTENSIONS_NON_V3
-@@ -1501,7 +1501,7 @@
-  *      MBEDTLS_TLS_RSA_PSK_WITH_RC4_128_SHA
-  *      MBEDTLS_TLS_PSK_WITH_RC4_128_SHA
-  */
--#define MBEDTLS_ARC4_C
-+//#define MBEDTLS_ARC4_C
- 
- /**
-  * \def MBEDTLS_ASN1_PARSE_C
-@@ -1621,7 +1621,7 @@
-  *      MBEDTLS_TLS_PSK_WITH_CAMELLIA_128_GCM_SHA256
-  *      MBEDTLS_TLS_PSK_WITH_CAMELLIA_128_CBC_SHA256
-  */
--#define MBEDTLS_CAMELLIA_C
-+//#define MBEDTLS_CAMELLIA_C
- 
- /**
-  * \def MBEDTLS_CCM_C
 @@ -1635,7 +1635,7 @@
   * This module enables the AES-CCM ciphersuites, if other requisites are
   * enabled as well.


bobafetthotmail:
this commit looks weird on github https://github.com/lede-project/source/pull/657/commits/2c45de782795edb62a61ccc5464f4d4cf8e4ec42 , you are adding a line in a part that gets deleted?

diizzyy:
No, I'm just removing a few changes the patch applies. Works for me but it's perhaps the wrong way of doing it? :-)

bobafetthotmail:
ah, I got it, it's a patch patching a patch, I got confused by that.

nbd168:
I would prefer to patch out the use of these ciphers from shadowsocks-libev, especially insecure ones like RC4.

diizzyy:
Apparently RC4 and Camellia are use because they're lightweight. Removing those apparently will kill pretty much all compatibility. https://github.com/shadowsocks/shadowsocks/wiki/Encryption

nbd168:
I don't see anything about Camellia on that page. Since there were some recent advances in breaking RC4, I still think we should keep it disabled, regardless of compatibility issues. I'm also worried that enabling it in mbedtls will cause other users of it to start using it as well.

diizzyy:
Sorry, this should make it clearer...
https://github.com/shadowsocks/shadowsocks-libev/blob/master/src/encrypt.c#L104

I do understand your concern and agree however I think we're better off getting upstream to change preferences rather than doing non-standard changes. Given that shadowsocks-libev hasn't been updated for quite some time, most tutorials seems to run it on low-end hardware and most users seems to be chinese I wonder if it's an option to drop it.

wongsyrone:
As a contributor of shadowsocks-libev, I want to clarify shadowsocks's goal is to help people bypass censorship in China. It focuses on indistinguishability from random bytes instead of security. OpenVPN and other VPN implementations are secure, but almost all of them get blocked by GFW.

The mbedtls support is added by me as well.

wongsyrone:
If you want to learn more, I would recommend https://shadowsocks.org/en/spec/cipher.html and https://github.com/shadowsocks/shadowsocks-libev/blob/master/m4/mbedtls.m4

nbd168:
@wongsyrone, since you're familiar with the code, do you think it would be a problem to make camellia and rc4 optional (and disabled by default in OpenWrt/LEDE)?

wongsyrone:
For me, either way is okay since I'm familiar with both shadowsocks-libev and OpenWrt's buildroot.

For beginners, especially people who just want to get it compiled and don't want to know the detail, there would be much pain on passing configure stage. Downstream just [removes](https://github.com/shadowsocks/openwrt-shadowsocks/blob/master/README.md) patches done by OpenWrt/LEDE.

> rm -rf package/feeds/base/mbedtls/patches

Do you want me to make these cipher checks optional? The intention is to make maintainer's life easier. If not, we will get many questions asking why I cannot use it. We let users determine which one they want to use.

BTW, the most popular usage are rc4-md5, aes-256-cfb and chacha20.

wongsyrone:
Another option is to drop this package, it would better to ask @aa65535 first since he is the package submitter.

nbd168:
I think RC4 (and probably Camellia as well) should be made optional. I don't have any problem with enabling CFB again in mbedtls.

obsy:
ARC4 is also required by transmission (https://github.com/transmission/transmission/blob/master/libtransmission/crypto-utils-polarssl.c#L186)

nbd168:
Hm, that's bad. Maybe I'll change the code to enable RC4, but prevent it from being used with TLS

diizzyy:
Do we really have to make "odd" custom changes? I'm fine with slimming down mbed TLS but diverging from upstream isn't probably the  way to go.

jow-:
I agree that we shouldn't deviate form upstream here. Are you worried that RC4 is getting used in TLS client context?

nbd168:
I wanted to make sure that it doesn't get accidentally picked up by TLS client or server implementations using mbedtls

neheb:
+1

As an aside, this problem exists with Openssl. LEDE bug #162.

diizzyy:
If it happens unintentionally the client/server software should be fixed preferably upstream. The solution is not to hack the SSL library itself.

wongsyrone:
In my opinion, If mbedtls doesn't disable them by default, we can consider it isn't that severe to prevent usage. It's developer's responsibility to implement correctly and warn users if there are vulnerabilities.

wongsyrone:
Hi @nbd168 and @diizzyy 

I make these ciphers optional, you can update to latest upstream HEAD and try again. It will show warnings now.

diizzyy:
@wongsyrone 
Cool, thanks :-)

@nbd168 
Transmission is still going require RC4 (it's used by several other clients so it's a non optional requirement). OpenSSL on most other platforms/distros also supports Camellia so I don't know if it should go.

nbd168:
RC4 support is now enabled in LEDE master. It turns out that mbedtls has a separate config option (already set) which masks RC4 from SSL/TLS ciphersuites, so enabling it is safe

wongsyrone:
Any progress on re-enabling CFB?

diizzyy:
@nbd168 is still looking into it.

diizzyy:
I had a look about Camellia and the 128-bit version might be of interest as it is up to ~10% faster than AES 128-bit. 

https://www.jmest.org/wp-content/uploads/JMESTN42351298.pdf
https://forums.freebsd.org/threads/12955/
https://github.com/hex7c0/nodejs-cipher-performance
https://en.wikipedia.org/wiki/Camellia_(cipher)

If the tradeoff is very small like in this case it might actually be useful to have it available.

nbd168:
If it's not commonly used for any software that wants to use mbedtls, we should not look into Camellia at all. Bigger speedups are likely to be gained from using ChaCha20 instead, which has the advantage of being a state-of-the-art cipher and not just some exotic niche thing.

By the way, cipher performance is highly sensitive to the CPU architecture, so ~10% faster on one architecture might be ~50% slower on another.

nbd168:
CFB re-enabled in master

diizzyy:
Everything else is merged except Camellia, closing...

PR #660, opened 2017-01-02
toolchain/musl: update to 1.1.16

hnyman:
Update musl to 1.1.16

Remove patches that are in upstream:
* patches created by recent backport 2912f9f2a2e5997df069d38e20d85ff4cc51acef
* 400-fix_quoted_timezone.patch
* 010-fix-regression-in-tcsetattr-on-all-mips-archs.patch

I have compiled & flashed to ar71xx/WNDR3700 and ipq806x/R7800 without problems. 
Both routers seem to work normally ;-)

http://git.musl-libc.org/cgit/musl/log/

```
1.1.16 release notes

new features:
- s390x (64-bit S/390) port
- pthread_setname_np extension function
- limited pthread_setattr_default_np function to set stack size defaults
- header-level support for linux 4.7, 4.8, and 4.9 features
- confstr _CS_V6_ENV and _CS_V7_ENV items

compatibility:
- public prototypes for abi-compat *_unlocked symbols, etc.
- fflush_unlocked(NULL) now works
- resolv.h __RES version macro now matches supported APIs
- workaround for gdb bugs backtracing across signals on x86_64
- anchors ^ and $ are now accepted in BRE subexpressions
- building for thumb2-only arm isa levels is now possible

bugs fixed:
- integer overflows in regexec buffer allocation (CVE-2016-8859)
- failure of regexec to report matches at offsets past INT_MAX
- static-pie executables with initialized thread-local storage crashed
- printf failed to catch EOVERFLOW in some cases, wrongly produced it in others
- printf produced wrong output, result for float with precision near INT_MAX
- printf produced wrong results with alt-form octal, zero flag, & field width
- printf float rounding was wrong for some midpoint cases
- swprintf printed junk after internal (256-byte) buffer filled up
- strtod family rounded incorrectly in several corner cases
- getmntent failed to handle long records
- getopt_long_only wrongly treated "--" as an option
- asctime output wrongly varied by locale
- strftime %y specifier produced wrong output for negative tm_year
- time zone names quoted with <> were misparsed
- corner case integer overflow in tm_year for some date conversions
- failure to load shared libs whose names were prefixes of standard lib names
- wrong error codes for several failure cases in various functions
- various asymptomatic undefined behavior
- various minor namespace issues in headers

arch-specific bugs fixed:
- tcsetattr regression on mips (completely non-working)
- wrong pread/pwrite syscall calling convention on sh
- wrong preadv2/pwritev2 syscall numbers on x32
- mrand48/jrand48 produced wrong-signedness results on 64-bit archs
```


diizzyy:
Also submitted here: http://patchwork.ozlabs.org/patch/710318/

diizzyy:
The patch on patchwork also switches to git tarball download over the git protocol which is a step backwards.

hnyman:
Yep, and in addition of using git download, the patch submitted by @chunkeey via the mailing list also deletes ` 099-Add-format-attribute-to-some-function-declarations.patch ` that is a separate issue and afaik Openwrt/LEDE specific patch from @hauke instead of being a backported genuine upstream patch (although its name starts with 0).

(It was recently renamed from 040-xxx ot 099-xxx, but the file originates at https://github.com/lede-project/source/commit/be5fcf9fd76a30c8a28a17daf9c3ddcc9c3af0ba and I think that it should survive the version bump, or at least its need should be evaluated separately as it has nothing to do with 1.1.16 itself)

jow-:
I'll pick this PR and give it a few test rounds.

chunkeey:
@diizzyy what's git tarball download? And what makes you think this is it a step backwards?

neheb:
Firewalls block the git protocol

chunkeey:
@neheb actually, just take a moment and look at the download method for git. You will notice that it tries to pull a musl-1.1.16-$HASH.tar.xz from sources.lede-project.org server first.
If the file is there, it is downloaded, checked, extracted, .... the git address will no get involved there.

So please, everyone "verify" that your problems are actually problems or not. Thanks.

diizzyy:
@chunkeey 
My bad on the wording...
What happens is that the buildroot grabs the tree using GIT (GIT/HTTPS) and then compresses it to a reproducible tarball. There are several issues with this approach, first and foremost availability but other issues too. 

If our current source goes away, LEDE falls back to its mirror server. If you haven't added a tarball checksum the buildroot has no idea of integrity of the downloaded tarball. If both these sources are gone you'll have tree that wont build. If you use a release tarball you'll more than likely find lots of mirrors and in worst case download the tarball manually and place it in your download directory without touching the source code and integrity is ensured. There has been work done to provide availability to cause as little disruption as possible while trying to sure integrity and security. As far as firewall goes most corporate firewalls blocks git traffic which also is an issue. Using release tarballs also saves on processing power/time and in most cases also bandwidth. In the end we preferably don't want users to use LEDE's mirror server for fetching code and rather rely on mirrors etc wherever possible. Hope it clears things up.

nbd168:
@diizzyy 
We have checksums for reproducible tarballs in all places that use git download (at least in the core repo). Download tarballs are automatically uploaded by buildbot to a mirror used by download.pl. 

This means that as long as PKG_MIRROR_HASH is correct, and the first buildbot run has completed, builds will typically never even try to use the git protocol, since fetching mirror tarballs is preferred.
By the way, missing (or incorrect) PKG_MIRROR_HASH is reported by 'make check'.

I really don't think we need to strongly prefer release tarballs over git revisions. We should just have enough mirrors in the long run to make sure that there won't be any issue with using more mirror tarballs.

chunkeey:
In case this gets confusing for some that don't follow the mailing-list as well. 
I posted a v2 yesterday: https://patchwork.ozlabs.org/patch/710536/.
There I added the missing patch and the PKG_MIRROR_HASH so the
archives could be fetched from sources. The original v1 didn't have it
so it was a problem there.

currently, the following package/ get their sources from git://

libs/libnftnl/Makefile:PKG_SOURCE_URL:=git://git.netfilter.org/libnftnl
network/ipv6/odhcp6c/Makefile:PKG_SOURCE_URL:=git://git.lede-project.org/project/odhcp6c.git
network/services/odhcpd/Makefile:PKG_SOURCE_URL:=git://git.lede-project.org/project/odhcpd.git
network/utils/nftables/Makefile:PKG_SOURCE_URL:=git://git.netfilter.org/nftables
network/utils/arptables/Makefile:PKG_SOURCE_URL:=git://git.netfilter.org/arptables
utils/mtd-utils/Makefile:PKG_SOURCE_URL:=git://git.infradead.org/mtd-utils.git

So, I never thought that this is a problem. I've contacted Rick Felker, if he's
willing to setup a http:// https:// or github mirror for his sources. But I think
he's rather busy with the new release ATM.

diizzyy:
@nbd168 
I agree that there's no need to enforce release tarballs but they should always be preferred if possible.
I'm well aware of PKG_MIRROR_HASH, just mentioned what happens if you don't have it in the Makefile.

@chunkeey 
As far as repos on netfilter.org goes they're using "Let's encrypt" as provider which ca-certificates doesn't have a certificate for which means that downloads will fail hence git protocol. No tarball releases as far as I can see.

git repo downloads should preferably use https instead of git as long as downloads doesn't fail due to certificate issues.

chunkeey:
@diizzyy can you tell me your reason for "git repo downloads should preferably use https instead of git [...]"?

The git project has this on their page about the different protocols. https://git-scm.com/book/ch4-1.html

>The Pros
The Git protocol is often the fastest network transfer protocol available. If you’re serving a lot of traffic for a public project or serving a very large project that doesn’t require user authentication for read access, it’s likely that you’ll want to set up a Git daemon to serve your project. It uses the same data-transfer mechanism as the SSH protocol but without the encryption and authentication overhead.
>
>The Cons
The downside of the Git protocol is the lack of authentication. It’s generally undesirable for the Git protocol to be the only access to your project. Generally, you’ll pair it with SSH or HTTPS access for the few developers who have push (write) access and have everyone else use git:// for read-only access. It’s also probably the most difficult protocol to set up. It must run its own daemon, which requires xinetd configuration or the like, which isn’t always a walk in the park. It also requires firewall access to port 9418, which isn’t a standard port that corporate firewalls always allow. Behind big corporate firewalls, this obscure port is commonly blocked.

There's nothing there that would tell me what makes https superior to "plain" git in THIS case.
All around it's: "git protocol is often the fastest network transfer protocol available. "
The "big corporate" (or dorm/3g/great-wall/free-neighburs-honey-pot) firewall might aswell 
mess with HTTPS too (like China's Golden Shield/Great Wall/...).

diizzyy:
HTTPS uses porten 443 which is more commonly open than 9418. It's as simple as that.

chunkeey:
@diizzyy  Please, just take a moment and look at the download method for git [0].
You will notice that it tries to pull the packed archive from sources.lede-project.org server first.

The sources.lede-project.org server is accessible via HTTPS and HTTP. Specifying git://... as
the PKG_SOURCE_URL will not prevent people from building the packages. And in case of
musl, the git address is listed on the download page [2].

(Fun fact, you can git clone http://git.musl-libc.org/cgit/musl. But it takes ages).

[0] <https://git.lede-project.org/?p=source.git;a=blob;f=include/download.mk;#l161> (as of now)
[1] <http://sources.lede-project.org>
[2] <http://www.musl-libc.org/download.html>

nbd168:
Merged @chunkeey's patch to my staging tree

diizzyy:
@chunkeey 
I'd much rather see a tarball that easily obtainable but oh well...

@nbd168 
I guess we at least want to use HTTPS instead of GIT, at least that's what jow-'s commit defaults to for the sake of consistency?
https://github.com/lede-project/source/commit/3a085dcbe3917d762a96df35a8bd99322cf09f5f

hnyman:
@jow-  @nbd168 
Have you decided on which musl 1.1.16 will go to the release?

PR #672, opened 2017-01-04
mvebu: modernize fan_ctrl.sh deployment

cybrnook:
Simple goal of removing the "/sbin/fan_ctrl.sh" script, as well as the "*/5 * * * * /sbin/fan_ctrl.sh" cron job from ALL mvebu devices, when only mamba requires this.


hnyman:
I am not sure that I like the approach that you delete the script /sbin/fan_ctrl.sh, but still include the same script in all mvebu devices via uci-defaults (practically the same script wrapped with fire creation code).   If you look at the PR's "files changed" summary, Github thinks that this PR a file rename plus some editing...

I know that flash space is not critical for mvebu devices, but your approach forces the same script twice onto flash: first as the uci-defaults script and then after reboot it gets also written to flash as /sbin/fan_ctrl.sh

It would be simpler to 
* leave the actual /sbin/fan_ctrl.sh script in place, maybe add a comment line that it only targets mamba
* move the crontab creation into uci-defaults and make it mamba-specific (as you are doing now)

cybrnook:
Thank you for the constructive criticism @hnyman. I agree this could be accomplished in a simpler way. I am closing this, and will approach this from another angle, and submit a new PR for that change.

PR #673, opened 2017-01-05
build: add devel option to store build config in firmware

13hakta:
Store cleaned .config in /etc/build.config

This developer option simplifies to recompile custom image as it only required to extract config from firmware and build a new one.

Signed-off-by: Vitaly Chekryzhev <13hakta@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,9 +92,10 @@ diffconfig: FORCE
 	$(SCRIPT_DIR)/diffconfig.sh > $(BIN_DIR)/config.seed
 
 prepare: .config $(tools/stamp-compile) $(toolchain/stamp-compile)
+	$(_SINGLE)$(SUBMAKE) -r diffconfig
+
 world: prepare $(target/stamp-compile) $(package/stamp-compile) $(package/stamp-install) $(target/stamp-install) FORCE
 	$(_SINGLE)$(SUBMAKE) -r package/index
-	$(_SINGLE)$(SUBMAKE) -r diffconfig
 	$(_SINGLE)$(SUBMAKE) -r checksum
 
 .PHONY: clean dirclean prereq prepare world package/symlinks package/symlinks-install package/symlinks-clean
--- a/config/Config-build.in
+++ b/config/Config-build.in
@@ -67,6 +67,12 @@ menu "Global build settings"
 		  This removes all ipkg/opkg status data files from the target directory
 		  before building the root filesystem.
 
+	config INCLUDE_CONFIG
+		bool "Include build configuration in firmware" if DEVEL
+		default n
+		help
+		  If enabled, config.seed will be stored in /etc/build.config of firmware.
+
 	config COLLECT_KERNEL_DEBUG
 		bool
 		prompt "Collect kernel debug information"
--- a/package/base-files/Makefile
+++ b/package/base-files/Makefile
@@ -168,6 +168,10 @@ define Package/base-files/install
 				echo "$$$${conffile##$(1)}" >> $(1)/CONTROL/conffiles; \
 		fi \
 	done
+
+	$(if $(CONFIG_INCLUDE_CONFIG), \
+		echo -e "# Build configuration for board $(BOARD)/$(SUBTARGET)/$(PROFILE)\n" >$(1)/etc/build.config; \
+		cat $(BIN_DIR)/config.seed >>$(1)/etc/build.config)
 endef
 
 ifneq ($(DUMP),1)


diizzyy:
That's actually a very clever idea, the archive should probably be named something slightly more intuitive like build-cfg.gz or similar.

13hakta:
Agreed, may be compression is not necessary. And filename also may be like boardname.config

hnyman:
Related comment: I include in my own build the exact references to the git commits of all feeds used in the compilation. My build script generates this as `etc/Compile_info.txt` just before the make process starts:
```
LEDE Reboot r2750-18152e71d8 / 2017-01-04 11:08
---
main      2017-01-04 18152e71d8 brcm47xx: mips74k: specify DEVICE_PACK
luci      2017-01-03 a2dfd21c80 i18n: sync translations
packages  2017-01-04 dd2e073feb Merge pull request #3752 from diizzyy/
routing   2016-12-22 d848d49d24 Merge pull request #245 from christf/c
```
The files in generated by this script:

```
#!/bin/bash
#
# timestampVersion  -  Collect source version info and insert it into firmware

STATUSFILE=files/etc/Compile_info.txt
Nickname=`grep RELEASE: include/{version,toplevel}.mk | cut -d "=" -f 2`

echo LEDE $Nickname `scripts/getver.sh` / `date "+%F %H:%M"` > $STATUSFILE
echo "---" >> $STATUSFILE
echo "main      "`(git show --format="%cd %h %s" --abbrev=10 --date=short | head -n 1 | cut -b1-60)` >> $STATUSFILE
echo "luci      "`(cd feeds/luci && git show --format="%cd %h %s" --abbrev=10 --date=short | head -n 1 | cut -b1-60)` >> $STATUSFILE
echo "packages  "`(cd feeds/packages && git show --format="%cd %h %s" --abbrev=10 --date=short | head -n 1 | cut -b1-60)` >> $STATUSFILE
echo "routing   "`(cd feeds/routing && git show --format="%cd %h %s" --abbrev=10 --date=short | head -n 1 | cut -b1-60)` >> $STATUSFILE
git add $STATUSFILE

```

hnyman:
Yeah, it would be cleaner, but I think that the new diffconfig step is currently run after the image compilation has been done, so the output of that is not yet available using the base-files compilation.

stintel:
@jow- can we call the diffconfig target earlier, e.g.:
--- a/Makefile
+++ b/Makefile
@@ -91,9 +91,9 @@ diffconfig: FORCE
        $(SCRIPT_DIR)/diffconfig.sh > $(BIN_DIR)/config.seed

 prepare: .config $(tools/stamp-install) $(toolchain/stamp-install)
+       $(_SINGLE)$(SUBMAKE) -r diffconfig
 world: prepare $(target/stamp-compile) $(package/stamp-compile) $(package/stamp-install) $(target/stamp-install) FORCE
        $(_SINGLE)$(SUBMAKE) -r package/index
-       $(_SINGLE)$(SUBMAKE) -r diffconfig
        $(_SINGLE)$(SUBMAKE) -r checksum

13hakta:
@stintel In jow's commit diffconfig call is unconditional and not target specific. However if it would be accepted to repository, this commit can be modified.

jow-:
@stintel - that should be no problem, feel free to move it

stintel:
@13hakta so I suggest to move the call as I suggested above, and then based on INCLUDE_CONFIG do cat $(BIN_DIR)/config.seed >>$(1)/etc/build.config)?

13hakta:
@stintel sure cat is much faster, but I insist on conditional executing as few kilobytes in firmware may be precious.

hnyman:
@13hakta 
> sure cat is much faster, but I insist on conditional executing as few kilobytes in firmware may be precious.

Stintel just proposes that we avoid running diffconfig twice in the build process. His proposal does not change your conditions of including the file in the firmware image.

Currently diffconfig is run always at the end of the make process (after the firmware images have been created) to create the file config.seed in the same directory as the firmware images. He proposes that in the main Makefile the line making diffconfig target is moved from "world" to "prepare" (see above) and then you can use that file instead of re-running diffconfig for the second time.

Your own line should then be changed to something like 
` cat $(BIN_DIR)/config.seed >>$(1)/etc/build.config) `  
instead of 
` cd $(TOPDIR); ./scripts/diffconfig.sh >>$(1)/etc/build.config) `

It works. I tested.
(otherwise ok, but as SUBTARGET is apparently undefined, I see just // there
```
# Build configuration for board ipq806x//DEVICE_R7800

CONFIG_TARGET_ipq806x=y
CONFIG_TARGET_ipq806x_DEVICE_R7800=y
CONFIG_TARGET_BOARD="ipq806x"
```

hnyman:
> could not find if commit 2a72a91 is already in repo. 

The commit is there since yesterday. And it is visible in the commit log of your own fork repo's master branch:
https://github.com/13hakta/source/commits/master

But you have not rebased your "config" branch against "master", so it is not yet visible for your PR in that branch, yet:
https://github.com/13hakta/source/commits/config

(just do "git rebase master" in your feature branch to update with with new commits from master...)

> It could be even just: cat $(BIN_DIR)/config.seed >>$(1)/etc/config.seed) 

I though about that earlier, but did not like it. There is already the directory "config" in /etc, so filename completions in shell etc. could get more complicated if new file would start identically. Better to leave it as build.config

13hakta:
@hnyman yes, didn't notice in my outdated fork.
@jow- would you please update call sequence as @stintel proposed?

stintel:
@13hakta just move it yourself as part of this patch?

Makefile
```diff
@@ -83,6 +83,7 @@ prereq: $(target/stamp-prereq) tmp/.prereq_packages
 		echo '       Please provide a "$(INCLUDE_DIR)/site/$(ARCH)" file and restart the build.'; \
 		exit 1; \
 	fi
+	$(_SINGLE)$(SUBMAKE) -r diffconfig
```
hnyman: Wrong place, I think. Prereq is run separately and also much too early. (E.g. first time already during feed update)

It should be in "prepare", not "prereq". See Stintel's example in https://github.com/lede-project/source/pull/673#issuecomment-270682184

13hakta: ok

hnyman:
Hmm. That is still a different place than in Stintel's example above (that worked for me). Does it work ok for you?

But you may have stumbled into something that may interest the core developers.

-----
For core devs:

There are two "prepare" targets in the main Makefile, on lines 52 and 93:

https://github.com/lede-project/source/blob/master/Makefile#L52
https://github.com/lede-project/source/blob/master/Makefile#L93

As far as I can see, both targets are inside the same ifneq, so I don't understand having both of them. Both lines date to years 2007 and 2009 and were committed by @nbd168 , so this is rather ancient stuff.

They should likely be combined or one of them removed.

Key lines:
```
43:  $(target/stamp-compile): $(toolchain/stamp-install) $(tools/stamp-install) $(BUILD_DIR)/.prepared
52:  prepare: $(target/stamp-compile)
93:  prepare: .config $(tools/stamp-install) $(toolchain/stamp-install)
```
Based on those, the `$(target/stamp-compile)` will actually cause also both `$(tools/stamp-install)` and `$(toolchain/stamp-install)` , but  `.config` needs to be handled separately.

So, eliminating line 52 and making line 93 as   `prepare: .config $(target/stamp-compile)` should be enough and equal to the current situation ???  (I amk no Makefile specialist, but looks logical.)

13hakta:
@hnyman really, I found first prepare and moved line there, and after your comment noticed second and they look similar. I will move to second. And may be someone Makefile-specialist will clarify why it happened so and fix it. :)

hnyman:
I noticed one more problem: diffconfig is now run before BIN_DIR has been generated, so build fails if a make clean has been issued and there is no BIN_DIR. Currently ` make clean ; make diffconfig ` causes error, and it shouldn't because the build status should have no impact on running diffconfig.

I issued a separate PR #676 about that, as it is fundamentally a separate issue and should be fixed regardless of this PR.

13hakta:
So, currently we have to wait until that issue will be fixed?

EricLuehrsen:
One comment considering the small size of text in SquasFS, make the option a hard default ON, or even not an option at all. That way the firmware image always includes data to figure out what is what, regardless loss of access to locally configured build source (####! I dropped my laptop out the window).

hnyman:
@stintel 
Could you please evaluate this before the 17.01 branching.

#676 has been merged as c1b12aa838587952b269a75ad278e8805aa9a65f so all prerequisites should be in place.

stintel:
"I also think it makes more sense to put it in the Target Images or Global build settings menu."

13hakta:
"Global build settings" section is more suitable, but which section and place?

hnyman:
I propose "General build options" section (or maybe "Package build options" section)

13hakta:
So, I put it to "General build options" and disabled by default.

hnyman:
The PR should be refreshed, so that line 94 in Makefile (the "prepare:" deps) does not get unnecessarily changed. 
(Makefile has been changed by https://github.com/lede-project/source/commit/131db366a3785a72b6e742f1842d29c1ad200601 during the two months that this PR has been waiting.)

13hakta:
PR updated.

stintel:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/stintel/staging.git;a=commit;h=72aa9fcd7aaef1484b7b0980329cef56d38fe4e1

PR #675, opened 2017-01-06
mvebu: remove fan_ctrl.sh and replace with fan daemon

cybrnook:
Replace the current fan_ctrl.sh crontab and script that is coming pre-populated on all mvebu devices. Instead this will replace it with a new(er) fan daemon only for mamba.


target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
```diff
@@ -0,0 +1,120 @@
...
+case "$board" in
+armada-xp-linksys-mamba)
+
+	cat > /etc/init.d/fan_control << 'FC_EOF'
+#!/bin/sh /etc/rc.common
+
+# mambafan daemon control script (/etc/init.d/fan_control)
+# for start/stop of mamba fan monitor (/usr/sbin/fan_monitor)
+
+START=99
```
hnyman: I think that START value should be below 95 so that your script runs before S95done (that should run last as it runs /etc/rc.local that is claimed to be run after the system init finishes).

Fan control is not dependent on network or anything else that could start slowly, so there is no need to push the start very late. It could well be 45 or 55 or something.

(I know that there are some scripts running later, but that is bad practice and against the documentation concerning /etc/rc.local )

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
```diff
@@ -0,0 +1,120 @@
...
+start() {
+	start-stop-daemon -b -S -x /usr/sbin/fan_monitor -m -p $PIDFILE
+}
+
+stop() {
+	start-stop-daemon -K -p $PIDFILE
+}
+FC_EOF
+
+	cat > /usr/sbin/fan_monitor << 'FM_EOF'
```
hnyman: Same comment as to your previous PR: I do not like the approach where you create this script /usr/sbin/fan_monitor on the fly in uci-defaults, causing it to be placed on flash twice after the boot.

You could have this /usr/sbin/fan_monitor script always in /usr/sbin/
It would just be useless there in non-mamba devices, but it would cause no harm. That would also remvoe one permission setting command from the process.

I understand the uci-defaults approach concerning the init script /etc/init.d/fan_control that should only be created in mamba.

cybrnook:
Thanks hnynan, for the review. I will make those two corrections a little later this morning, lowering start/stop and pre-loading sbin.

cybrnook:
@hnyman , okay, it should be done now.

cybrnook:
@hnyman 

As to not keep adding,editing, and deleting. Perhaps I can add the revised, un-submitted changes for your approval before making it a PR. My goal is for my first commit to be a "clean" one :-) , I don't want it to need any amending.

If you would be so kind:
https://github.com/cybrnook/source/commits/mambafan

PR #682, opened 2017-01-08
ramips: Add support for VoCore2

ndoo:

### DIFF
--- a/target/linux/ramips/dts/VOCORE2.dts
+++ b/target/linux/ramips/dts/VOCORE2.dts
@@ -1,0 +1,123 @@
+/dts-v1/;
+
+#include "mt7628an.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "vocore,VoCore2", "mediatek,mt7628an-soc";
+	model = "VoCore2";
+
+	chosen {
+		bootargs = "console=ttyS2,115200";
+	};
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x4000000>;
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		status {
+			label = "vocore2:fuchsia:status";
+			gpios = <&gpio1 12 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "refclk", "gpio";
+			ralink,function = "gpio";
+		};
+
+		agpio {
+			ralink,group = "agpio";
+			ralink,function = "uart2";
+		};
+	};
+
+	uart1_pins: uart1 {
+		uart1 {
+			ralink,group = "uart1";
+			ralink,function = "uart1";
+		};
+	};
+
+	uart2_pins: uart2 {
+		uart2 {
+			ralink,group = "spis";
+			ralink,function = "pwm";
+		};
+	};
+};
+
+&wmac {
+	status = "okay";
+	ralink,mtd-eeprom = <&factory 0x4>;
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,chunked-io = <32>;
+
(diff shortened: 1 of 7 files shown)


target/linux/ramips/dts/VOCORE2.dts
```diff
@@ -0,0 +1,124 @@
...
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		linux,modalias = "m25p80";
```
mkresin: drop this line. linux,modalias was used for < kernel 4.4

target/linux/ramips/dts/VOCORE2.dts
```diff
@@ -0,0 +1,124 @@
...
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
+		};
+	};
+};
+
+&esw {
+	mediatek,portmap = <0x5>;
+	mediatek,portdisable = <0x3a>;
```
mkresin: Doesn't look correct to me. in the end you have port 0 and port 2 enabled but as far as I can see, this board has only one ethernet interface. The binary mask of you 0x3a portdisable is 111010, where port0 is the most right bit. If the bit is set for a port, the ports is disabled.

target/linux/ramips/dts/VOCORE2.dts
```diff
@@ -0,0 +1,124 @@
...
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
+		};
+	};
+};
+
+&esw {
+	mediatek,portmap = <0x5>;
```
mkresin: Doesn't look correct to me. According to your binary mask you are setting, port 0 is lan, port 1 is wan and port 2 is lan. But as far as I can see, this board has only one ethernet interface.

target/linux/ramips/dts/VOCORE2.dts
ndoo: Ack

target/linux/ramips/dts/VOCORE2.dts
ndoo: I would like to understand how to bitmask of "mediatek,portmap" bindings work but could not find documentation on this.

Port 0 is LAN, Port 2 is LAN, other ports are not exposed on PCB.

target/linux/ramips/dts/VOCORE2.dts
ndoo: Port 0:
Physically exposed on the dock

Port 2:
Physically exposed as pins on the PCB

Port 1,3,4,5:
Not exposed.

ndoo: http://vocore.io/v2.html

ndoo:
Removed `linux,modalias` line from DTS, for other changes requested see comments. esw port 2 was included in DTS as it is physically accessible from the PCB.

target/linux/ramips/dts/VOCORE2.dts
mkresin: > I would like to understand how to bitmask of "mediatek,portmap" bindings work but could not find documentation on this.

```git grep mediatek,portmap``` is your friend. Or grep your build_dir to find the corresponding code.

> Port 0 is LAN, Port 2 is LAN, other ports are not exposed on PCB.

Please use 0x7 as portmap (port 0-2 is lan, no wan).

target/linux/ramips/dts/VOCORE2.dts
mkresin: Please add this information to the commit message to make the review easier. The following PR is a really good example for a PR having all the required informations to make the review easy: https://github.com/lede-project/source/pull/611

ndoo:
Improvements made to the commit message

mkresin:
Hey Andrew,

perfect commit message!

as far as I can see, only the "wrong" mediatek,portmap is left. Do you agree that it should be changed to 0x7 (port 0 to 2 is lan, no wan).

ndoo:
You are right, I wanted to test flash it with the change to my VoCore2 first before pushing the commit. I have now tested it and here is the updated commit.

mkresin:
@ndoo you obviously have missed to push the commit. still no change here.

ndoo:
Done now sorry

mkresin:
merged, thanks.

PR #683, opened 2017-01-08
ramips: zyxel omni updates

13hakta:
* Update network switch identifier
* Update WAN MAC address
* Pad factory firmware to 64Kb as it required by zyxel web flasher

Signed-off-by: Vitaly Chekryzhev 13hakta@gmail.com


blogic:
series is looking good, could you add a description to the patches, specifically explaining why we need the 64k padding please.

blogic:
merged, Thanks !

PR #684, opened 2017-01-09
ramips: Fix VLAN limits for MT7621 GSW

akanouras:
Hopefully I'm doing this the right way...

Without this patch swconfig will only allow setting up a total of 16
VLANs, with VLAN ID range of 0-15.

Tested on ubnt-erx.

Signed-off-by: Antonis Kanouras <antonis@metadosis.eu>


diizzyy:
@blogic

blogic:
merged, Thanks !

akanouras:
Thanks for the fast response, and all your work!

PR #685, opened 2017-01-09
Fix dependency for hostapd

synnack:
Fix hostapd dependencies leading to a build failure when building eapol_test.

Signed-off-by: Wilco Baan Hofman <wilco@baanhofman.nl>


jow-:
Merged into my staging tree, thanks!

PR #687, opened 2017-01-09
ar71xx: fix network config for Mikrotik RB411U

obsy:
Mikrotik RB411U has only one ethernet port - eth0. This patch allows to create
correct config with one lan section.


nbd168:
Added to my staging tree at https://git.lede-project.org/?p=lede/nbd/staging.git;a=summary

Could you please try flashing the latest version from that tree using sysupgrade to verify that it still works after the UBI conversion?

Thanks

PR #689, opened 2017-01-10
linux: added LED Morse code trigger

greenbreakfast:
Added LED morse code trigger kernel module to buildroot:
- Addition of kernel package ledtrig-morse to led kernel modules
- Addition of kernel patch to add ledtrig-morse kernel module
  - Adds morse trigger to LED trigger kconfig file
  - Changes led drivers makefile to support compilation of morse code trigger code
  - Adds Gabor Juhos' ledtrig-morse driver implementation

Signed-off-by: Lazar Demin <lazar@onion.io>

### DIFF
--- a/package/kernel/linux/modules/leds.mk
+++ b/package/kernel/linux/modules/leds.mk
@@ -145,3 +145,18 @@ define KernelPackage/leds-pca963x/description
 endef
 
 $(eval $(call KernelPackage,leds-pca963x))
+
+
+define KernelPackage/ledtrig-morse
+  SUBMENU:=$(LEDS_MENU)
+  TITLE:=LED Morse Trigger
+  KCONFIG:=CONFIG_LEDS_TRIGGER_MORSE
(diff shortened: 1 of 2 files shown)


blogic:
@greenbreakfast Please add "linux :" as a prefix. We are trying to reduce the amount of patches to the kernel. we can merge the morse trigger but you would need to take care of sending it upstream for inclusion. history shows, that otherwise it will jut bit rot in our tree :/

jow-:
I think we can package this as external kmod, no need to patch it into the kernel.

KanjiMonster:
I want to point out that we had it until ~5 month ago and @nbd168 deleted it in 577f873dafdf3914c79ee21869288f20d419a39f - so this is currently actually a partial revert.

greenbreakfast:
Can we go ahead with the merge since it is a partial revert?

nbd168:
No, I don't want to add back kernel patches for this rarely used feature, if it can be avoided.

nbd168:
FYI, you can find a packaged version here: https://github.com/openwrt/openwrt/pull/331

PR #641, opened 2016-12-22
wwan: adding wwan support for the dwr-512 3G modem

fid0did0:
This PR allow the 3G modem embedded in the DWR-512 to be managed
by the wwan-ncm scripts. The modem will use the usb-option and
usb-cdc-ether drivers.
The DWR-512 DT is updated accordingly.

Signed-off-by: Giuseppe Lippolis <giu.lippolis@gmail.com>

PR #691, opened 2017-01-10
ramips: adding 3g led in dwr-512 uci-default

fid0did0:
The dwr-512 has an embedded 3g modem.
This patch enable by default the netdev trigger on the 3G front panel
led linked to the 3g network interface.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -163,6 +163,9 @@ vr500)
 dir-860l-b1)
 	ucidef_set_led_netdev "wan" "wan" "$board:green:net" "eth0.2"
 	;;
+dwr-512-b)
+	ucidef_set_led_netdev "wan3g" "wan3g" "$board:green:3g" "usb0"
+	;;
 ex2700|\
 wn3000rpv3)
 	ucidef_set_led_default "power_r" "POWER (red)" "$board:red:power" "0"


jow-:
Staged into my tree - thanks!

jow-:
Hm, reviewed it some more - is there indeed a "wan3g" interface by default on this device?
I cannot find any mention of "wan3g" or "dwr-512" in 02_network. Is this change depending on some other patches for this device which are not yet in master?

fid0did0:
Hi jow,
I explain you the context.
The dwr-512 have an embedded 3G modem with data connection over the usb (using the kmod_net_cdc_eth).
In order to establish a 3G connection the following section in /etc/config/network need to be added:
```
config interface 'wan3g'
        option proto 'wwan'
        option apn '***.***.**'
        option pincode '****'
        option auto '0'
```
This section is not added by default because the user need to populate the apn and pincode field.
I will describe how to setup the 3G modem in the LEDE/owrt device page.
In any case using this configuration the wwan subsystem will automatically generate an usb0 network interface:

```
root@lede:~# ifup wan3g
root@lede:~# ifconfig
usb0      Link encap:Ethernet  HWaddr 02:00:FF:AA:AA:AA
          inet6 addr: fe80::ffff:feaa:aaaa/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:1 errors:0 dropped:0 overruns:0 frame:0
          TX packets:1 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:150 (150.0 B)
```
This will be handled when the [(DWR-512: adding wwan support for the dwr-512 3G)](https://github.com/lede-project/source/pull/641) will be merged (it should be ready).

If I understood well the ``ucidef_set_led_netdev "wan3g" "wan3g" "$board:green:3g" "usb0"`` line use the wna3g identificator only to populate the "config led" and "name" field of the /etc/config/system.
```
config led 'led_wan3g'
        option name 'wan3g'
        option sysfs 'dwr-512-b:green:3g'
        option trigger 'netdev'
        option mode 'link tx rx'
        option dev 'usb0'
```
The important field is the "dev" field properly configurated with the usb0 interface.

Please let me know if I'm missing something.
Bye.

bobafetthotmail:
>This section is not added by default because the user need to populate the apn and pincode field.

imho it would be better to have it added with dummy values, like

```
config interface 'wan3g'
        option proto 'wwan'
        option apn 'my.apn.here'
        option pincode 'pincode'
        option auto '0'
```

so that anyone that writes **uci show network** or looks at this interface from luci will easily figure out what must be done there to enable the interface even without external instructions.

fid0did0:
Alberto,
Thanks for your comment.
I also though at this option, but at the end, I cannot see any similar example in the /etc/board.d/02_network and in the /lib/functions/uci-defaults.sh I cannot found any support function to setup the 3G interface.
Therefore I thought that the default behavior is to leave the user the 3G interface configuration.

mkresin:
IMHO out of scope of this PR but there are examples where some placeholder are added.

For pppoe connections (```ucidef_set_interface_wan "$interface_wan" "pppoe"```), [config_generate
](https://github.com/lede-project/source/blob/master/package/base-files/files/bin/config_generate#L127) will add a placeholder username and password.

But as written, it's a job for another PR.

fid0did0:
Hi mathias,
considering that the ucidef_set_interface_wan() is used to setup a wan and not a 3G wan with wwan protocol I should add a "ucidef_set_interface_wan3g()" function and modify the config_generate to support the wwan.
I agree, better do it after in a separate PR.

Bye.

fid0did0:
Mathias,  jow,
I would like to have this and the related [(DWR-512: adding wwan support for the dwr-512 3G)](https://github.com/lede-project/source/pull/641) PR in the close release branch.
Both the PR should be ready.
Do you think is it possible to merge before the release branch?

Thanks,
Bye.

blogic:
@bobafetthotmail adding a wrong pin/apn is always a bad idea as the device will come up and use them. if people want to write uci by hand they wont get around reading the docs. alternatively using Luci will help.

bobafetthotmail:
>adding a wrong pin/apn is always a bad idea as the device will come up and use them.

Isn't there a way to have it but keep it disabled like you do in fstab with "enabled=0" by default?

> alternatively using Luci will help. 

Luci will lack that interface anyway, just like uci. So they either read the docs or they read the docs to add that interface. 

I was just pointing out that it would make more sense to have a configuration for an integrated device already set, only thing missing are connection details. Aren't you doing something like that with Lantiq modems?

blogic:
On 15/01/2017 16:19, Alberto Bursi wrote:
>     adding a wrong pin/apn is always a bad idea as the device will come
>     up and use them.
> 
> Isn't there a way to have it but keep it disabled like you do in fstab
> with "enabled=0" by default?
> 
>     alternatively using Luci will help.
> 
> Luci will lack that interface anyway, just like uci. So they either read
> the docs or they read the docs to add that interface.
> 
> I was just pointing out that it would make more sense to have a
> configuration for an integrated device already set, only thing missing
> are connection details. Aren't you doing something like that with Lantiq
> modems?
> 

you did in fact advice people to add a wrong pin and apn. this is what i
commented on. no more not less. adding wrong configs is a no-go. the
stuff you describe above it something totally different than what you
wrote before. your previous comment was unrelated to the device being
populated by default.

bobafetthotmail:
>you did in fact advice people to add a wrong pin and apn.

I stated my opinion, I didn't request changes.

>the stuff you describe above it something totally different than what you wrote before.

No, I just adapted it to your feedback. I still find wrong requiring people to add a new interface manually (uci or luci) just to enable a modem in a device with integrated modem.

fid0did0:
So, is the PR ok or some change are needed?

pepe2k:
Hi @fid0did0,

> So, is the PR ok or some change are needed?

I see here two things which still need some explanation:

1. As you mentioned before, this PR in fact depends on your other one. I think it would be better to send it as one PR, with separate commits (this one as last one, as it depends on other changes which are not yet in repository).

2. This part:

> This section is not added by default because the user need to populate the apn and pincode field.
> I will describe how to setup the 3G modem in the LEDE/owrt device page.
> In any case using this configuration the wwan subsystem will automatically generate an usb0 network interface:
> [...]

What if the user doesn't use modem at all (thus, there won't be 'usb0' interface)? What with second LED, the "2G" one (as I can see on the device panel)? Can you describe how LEDs are handled in vendor firmware?

Also, what if the user selects different name in the network section, like 'modem' or whatever else, but different than 'wan3g'? I know that it's _just a name_, but IMHO it could be misleading and making too many assumptions about how the device will be used/configured is also a bad idea for me.

Cheers,
Piotr

fid0did0:
> As you mentioned before, this PR in fact depends on your other one. I think it would be better to send it as one PR, with separate commits (this one as last one, as it depends on other changes which are not yet in repository). Then we could focus on whole subject.

Hi Piotr, thanks for your comments. 
Merging this PR with the #641 (or close it until the other is meged) now means loose all the comment and explanation. It is better to be pragmatic and wait for the #641. (I really do not understand why the #641 is till pending).
In any case, if I will receive the same others similar advice, I will do it.

> What if the user doesn't use modem at all (thus, there won't be 'usb0' interface)? What with second LED, the "2G" one (as I can see on the device panel)? Can you describe how LEDs are handled in vendor firmware?

Actually I still didn't found the command to identify if the connection is in 2G or 3G mode, or to force the connection in 2G mode only. Therefore i'm driving only the 3G led (the 2G is currently available to be exported in /sys/gpio). IMHO having one led showing the connection status is better than no led at all.
After having the first PR merged I will try to improve this identification.

>Also, what if the user selects different name in the network section, like 'modem' or whatever else, but different than 'wan3g'? I know that it's just a name, but IMHO it could be misleading and making too many assumptions about how the device will be used/configured is also a bad idea for me.

Can I ask you why your comment cannot be applied to the "wan" name in the network section?
If the user will change the "wan" name in "modem", the led will stop to work as well. 
At the end IMHO it is important to document the standard behavior (as I do here [d-link_dwr-512_b](https://wiki.openwrt.org/toh/d-link/d-link_dwr-512_b), see the 3G modem section) . 
If the final user want change the name, IMHO it is supposed he know what he is doing!

bye,
Giuseppe.

blogic:

PR #692, opened 2017-01-10
include target: Unify and sanitize image and tool tarball filenames

cshoredaniel:
Currently tool tarballs (sdk, toolchain, and imagebuilder)
do not use the same name scheme as the firmware images.
This patch moves the filenaming logic into a common
location and updates the targets so that in future the
various files will all follow the same naming logic.


cshoredaniel:
Oops combined this with an update for per-device manifests when using per-device-rootfs.  Will split and resubmit.

PR #693, opened 2017-01-10
gettext-full: Add dependency on iconv-full

cshoredaniel:
gettext-full build fails without iconv-full.

### DIFF
--- a/package/libs/gettext-full/Makefile
+++ b/package/libs/gettext-full/Makefile
@@ -33,6 +33,7 @@ define Package/libintl-full
   CATEGORY:=Libraries
   TITLE:=GNU Internationalization library
   URL:=http://www.gnu.org/software/gettext/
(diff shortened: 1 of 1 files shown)


cshoredaniel:
The issue in question appears to be specific to brcm2708/bcm2708 - I haven't had the same issue with x86_64 or ar71xx/generic.

hnyman:
It may have been fixed by @jow-  with https://github.com/lede-project/source/commit/b95494baedbc46d088550e18370e3507af4f732e 

At least the host build brokenness was not specific to brcm2708/bcm2708:
I reported yesterday to jow that some buildbot buildslaves failed gettext-full host build for several platforms. And that broke glib2 build, which then broke lots of packages.

cshoredaniel:
Yes that looks like the actual problem and fix

PR #696, opened 2017-01-12
Fix switch port mapping

sotux:
Fix switch port mapping for below devices:
tl-wr841n-v7
netgear wnr2000-v3
netgear wndr3700 v1/v2
netgear wndr3800/wndr3800ch

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -271,7 +271,6 @@ ar71xx_setup_interfaces()
 	tl-wr941nd-v5|\
 	tl-wr941nd-v6|\
 	wnr1000-v2|\
-	wnr2000-v3|\
 	wnr2000-v4|\
 	wnr2200|\
 	wnr612-v2|\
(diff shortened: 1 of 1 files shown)


blogic:
patch 2/3 also touches gl-ar300 but does not mention it anywhere. there is also a 4th merge commit in this branch, please remove it

sotux:
The 4th merge commit has been removed.
Move gl-ar300 position is just for the alphabet order for this file, because I was told this file is alphabet ordered on my previous PR. But in fact it's not so strict.

target/linux/ar71xx/base-files/etc/board.d/02_network
```diff
@@ -421,6 +415,12 @@ ar71xx_setup_interfaces()
 		ucidef_add_switch_port_attr "switch0" 2 led 9
 		ucidef_add_switch_port_attr "switch0" 5 led 2
 		;;
+	gl-ar300|\
+	wnr2000-v3)
```
blogic: this reorders it to the incorrect alphabetical location

jow-:
Merged into my staging tree with fixed alphabetical order. Thank you!

PR #697, opened 2017-01-12
tools/qemu: use default host configure rule ; set appropriate vars & args

commodo:
Admittedly, this is my own OCD wanting to get rid of this.

Because I tried (a while back to upgrade QEMU to a newer version),
and (during that attempt) I tried to get rid of this.

Tested on Linux & Mac.


blogic:
merged, Thanks !

PR #700, opened 2017-01-12
dnsmasq: use ubus signalling in ntp hotplug script

kdarbyshirebryant:
Use ubus process signalling instead of 'kill pidof dnsmasq' for
SIGHUP signalling to dnsmasq when ntp says time is valid.

Compile & run tested: ar71xx Archer C7 v2

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasqsec.hotplug
+++ b/package/network/services/dnsmasq/files/dnsmasqsec.hotplug
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+. /lib/functions/procd.sh
+
 TIMEVALIDFILE="/var/state/dnsmasqsec"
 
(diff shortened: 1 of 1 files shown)


dedeckeh:
@kdarbyshirebryant you can use the wrapper function procd_send_signal taking as first argument the service; second argument the instance and last argument the signal.
By default procd sends a SIGHUP signal; so in this case there's no need to specify the signal. If you want to specify the signal it would be better to use  kill -l SIGHUP iso the hardcoded value 1

kdarbyshirebryant:
Updated to use procd_send_signal.  Run tested ar71xx

dedeckeh:
Pulled into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=ec91ec94e458220fb04d1513e614215d414749b8); thanks

PR #701, opened 2017-01-13
mvebu: set fan_ctrl.sh only on mamba

cybrnook:
Simple replacement of the default fan_ctrl.sh crontab for all mvebu devices (with or without fans). 
Instead we will create this cron job only on the mamba device.

Signed-off-by: Hans Geiblinger <cybrnook2002yahoo.com>

### DIFF
--- a/target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
+++ b/target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
@@ -1,0 +1,23 @@
+#!/bin/sh
+#
+# Copyright (C) 2017 LEDE-Project.org
+#
+
+. /lib/mvebu.sh
+
+board=$(mvebu_board_name)
+
+case "$board" in
+armada-xp-linksys-mamba)
+        # Set one execution of the fan script during every start (waiting for cron)
+        echo "/sbin/fan_ctrl.sh" > /etc/rc.local
+        # Set fan script execution in crontab
+        echo "# mamba fan script runs every 5 minutes" >> /etc/crontabs/root
+        echo "*/5 * * * * /sbin/fan_ctrl.sh" >> /etc/crontabs/root
+        # Execute one time after initial flash (instead of waiting 5 min for cron)
+        /sbin/fan_ctrl.sh
+    ;;
+*)
+esac
+
+exit 0
--- a/target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
+++ b/target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
@@ -10,10 +10,20 @@ board=$(mvebu_board_name)
 case "$board" in
 armada-xp-linksys-mamba)
         # Set one execution of the fan script during every start (waiting for cron)
-        echo "/sbin/fan_ctrl.sh" > /etc/rc.local
+        if grep -q fan_ctrl.sh /etc/rc.local
+                then
+                        exit 0
+                else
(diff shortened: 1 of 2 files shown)


nbd168:
Please change the code to check for existing entries before adding the fan_ctrl calls. uci-defaults may be re-run after sysupgrade and we don't want to add duplicates on every upgrade.

cybrnook:
Okay gentlemen, we should be okay now.

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
```diff
@@ -10,10 +10,20 @@ board=$(mvebu_board_name)
 case "$board" in
 armada-xp-linksys-mamba)
         # Set one execution of the fan script during every start (waiting for cron)
-        echo "/sbin/fan_ctrl.sh" > /etc/rc.local
+        if grep -q fan_ctrl.sh /etc/rc.local
+                then
+                        exit 0
+                else
+                        echo "/sbin/fan_ctrl.sh" > /etc/rc.local
```
jow-: The /etc/rc.local file is user property, with this code you would overwrite it on the first boot.

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
```diff
@@ -10,10 +10,20 @@ board=$(mvebu_board_name)
...
+        fi                        
         # Set fan script execution in crontab
-        echo "# mamba fan script runs every 5 minutes" >> /etc/crontabs/root
-        echo "*/5 * * * * /sbin/fan_ctrl.sh" >> /etc/crontabs/root
+        if grep -q fan_ctrl.sh /etc/crontabs/root
```
jow-: Use `grep -s -q ...` to suppress errors in case `/etc/crontabs/root` does not exist yet.

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
cybrnook: The thing with rc.local is that it comes by default with an ending "exit 0", so I can't "append" (>>) to rc.local as the files execution would have ended by the time anything after the "exit 0" comes. As well, I am sure you would not like the idea of me having to copy the contents to a temporary staging file to read out the contents of rc.local. Then I add my entry to the now empty rc.local, and then "append" the contents of the staged file back to rc.local, thus placing the fan execution at the top.

Instead I opted to just replace the file (but was only thinking of a clean install, not taking into account an "upgrade" scenario. I am open to suggestions of course if you think you have a cleaner way around this without me having to stage the contents to then write back.

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
cybrnook: Will do

target/linux/mvebu/base-files/etc/uci-defaults/04_mambafan
cybrnook: There is always of course the option of removing rc.local in total, but then user would have to wait until the first cron execution for the fan to spin down, which I don't like.

cybrnook:
Okay, I think I am going to make a new, clean PR for this. I think I am just going to drop the editing of rc.local altogether. Would be no different than it is today as far as how long it takes to kick in.

But the main goal is removing it from all the other devices. That's the direction I will take.

PR #702, opened 2017-01-13
mvebu: set fan_ctrl.sh only on mamba

cybrnook:
Simple replacement of the default fan_ctrl.sh crontab for all mvebu devices (with or without fans).
Instead we will create this cron job only on the mamba device.

Signed-off-by: Hans Geiblinger <cybrnook2002yahoo.com>

### DIFF
--- a/target/linux/mvebu/base-files/etc/crontabs/root
+++ b/target/linux/mvebu/base-files/etc/crontabs/root
@@ -1,1 +1,0 @@
-*/5 * * * * /sbin/fan_ctrl.sh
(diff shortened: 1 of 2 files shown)


blogic:
do only the mamba devices have a fan ?

nbd168:
Yes

cybrnook:
Is there anything else needed from my end to get this committed? I know everyone is busy...

nbd168:
Applied to my staging tree at https://git.lede-project.org/?p=lede/nbd/staging.git;a=summary with some cleanups.

cybrnook:
Got it, thanks.

anomeome:
Just a note that this addresses [FS#345](https://bugs.lede-project.org/index.php?do=details&task_id=345&order=id&sort=desc)

PR #703, opened 2017-01-13
ramips: simplify EX2700 network config

jclehner:
Don't create a vlan for the Netgear EX2700's only Ethernet port.


mkresin:
NAK from me.

We should add a vlan config by default in case the switch support vlans. A switch port can be a member of multiple vlans and in case a default/example config in /etc/config/network is missing, it's is unnecessary hard to write such a config.

blogic:
actually it is common practice to not run vlans on a single ethernet port device. was this changed ?

mkresin:
@blogic If you disagree, feel free to merge.

blogic:
merged, Thanks !

PR #704, opened 2017-01-14
dnsmasq: expand 'add_local_hostname' flexibility for self DNS

EricLuehrsen:
ref commit 612e227
ref commit ec63e3b

'option add_local_hostname' scripted implementation statically assigns
this host in auto generated host file at init. If IFUP or other signals
do not occur, then address changes are not tracked. The script doesn't
apply all the addresses at an interface. This may make logs obscure.
The script only puts the bare host name (maybe not FQDN) in host file,
but if '--exapandhosts' is enabled, then /etc/hosts entries will be
suffixed, and "127.0.0.1 localhost" becomes "localhost.lan".

dnsmasq provides an option to perform this function, but it is rather
greedy. '--interface-name=\<name\>,\<iface\>' will assign the name to all
IP on the specified interface (except link local). This is a useful
feature, but some setups depend on the original restrictive behavior.

'option add_local_fqdn' is added to enhance the feature set, but
if not entered or empty string, then it will default to original
option and behavior. This new option has a few settings. At each
increased setting the most detailed name becomes the PTR record:
0 - same as add_local_hostname 0 or disabled
1 - same as add_local_hostname 1
2 - assigns the bare host name to all IP w/ --dnsmasq-interface
3 - assigns the FQDN and host to all IP w/ --dnsmasq-interface
4 - assigns \<iface\>.\<host\>.\<domain\> and above w/ --dnsmasq-nterface

'option add_wan_fqdn' is added to run the same procedure on
inferred WAN intefaces. If an interface has 'config dhcp' and
'option ignore 1' set, then it is considered WAN. The original
option would only run on DHCP serving interfaces.

Reference (not dependent):
https://github.com/openwrt/luci/pull/959
https://github.com/openwrt/packages/pull/3873

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.76
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq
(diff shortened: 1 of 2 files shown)


jow-:
superseded by #674

PR #705, opened 2017-01-14
fstools: Only block mount on fstab restart

cshoredaniel:
On systems using external storage (not extroot) with
block-mount, bad behaviour occurs when the fstab LuCI
app commits a change because this calls
/etc/init.d/fstab restart which does a block umount,
which can result in problematic unmounts.

### DIFF
--- a/package/system/fstools/Makefile
+++ b/package/system/fstools/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=fstools
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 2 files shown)


cshoredaniel:
It appears that /etc/init.d/fstab reload ends up calling /etc/init.d/fstab restart - is that accurate?  If so overriding reload and preventing restart seems a better solution (start as start())

cshoredaniel:
To answer my own question; yes luci call reload which calls restart if no reload is present, which calls stop then start if no restart is present.  I am therefore updating this PR with a more appropriate (start and restart indicate deprecated file, reload and boot do block mount, stop does block umount.

cshoredaniel:
Actually this can be avoided by having LuCI exec 'block mount' instead of using init for applying changes, and requires no change to core.

PR #706, opened 2017-01-14
ar71xx: Chipidea USB device support

neykov:
Changes the platform to use the Chipidea driver instead of the
generic USB host driver which has support for both host and
device modes (selected on boot).

The changes in 930-chipidea-pullup.patch are already in mainline.
I'll upstream 920-usb-chipidea-AR933x-platform-support.patch once I
can test the changes with a newer kernel.

Signed-off-by: Svetoslav Neykov <svetoslav@neykov.name>

### DIFF
--- a/package/kernel/linux/modules/usb.mk
+++ b/package/kernel/linux/modules/usb.mk
@@ -1375,6 +1375,29 @@ endef
 $(eval $(call KernelPackage,usbip-server))
 
 
+define KernelPackage/usb-chipidea
+  TITLE:=Host and device support for Chipidea controllers
+  DEPENDS:=+USB_GADGET_SUPPORT:kmod-usb-gadget
+  KCONFIG:= \
+	CONFIG_NOP_USB_XCEIV=y \
+	CONFIG_EXTCON \
+	CONFIG_USB_CHIPIDEA \
+	CONFIG_USB_CHIPIDEA_HOST=y \
+	CONFIG_USB_CHIPIDEA_UDC=y \
+	CONFIG_USB_CHIPIDEA_DEBUG=y
+  FILES:= \
+	$(LINUX_DIR)/drivers/extcon/extcon.ko \
+	$(LINUX_DIR)/drivers/usb/chipidea/ci_hdrc.ko
+  AUTOLOAD:=$(call AutoLoad,51,ci_hdrc,0)
+  $(call AddDepends/usb)
+endef
+
+define KernelPackage/usb-chipidea/description
+ Kernel support for USB Chipidea controllers
+endef
+
+$(eval $(call KernelPackage,usb-chipidea))
+
 define KernelPackage/usbmon
   TITLE:=USB traffic monitor
   KCONFIG:=CONFIG_USB_MON
--- a/target/linux/ar71xx/Makefile
+++ b/target/linux/ar71xx/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 ARCH:=mips
 BOARD:=ar71xx
 BOARDNAME:=Atheros AR7xxx/AR9xxx
-FEATURES:=mips16
+FEATURES:=mips16 usbgadget
 CPU_TYPE:=24kc
 SUBTARGETS:=generic nand mikrotik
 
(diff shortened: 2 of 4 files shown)


target/linux/ar71xx/image/tp-link.mk
```diff
@@ -426,7 +426,7 @@ TARGET_DEVICES += tl-wpa8630
 define Device/tl-wr703n-v1
     $(Device/tplink-4mlzma)
     DEVICE_TITLE := TP-LINK TL-WR703N
-    DEVICE_PACKAGES := kmod-usb-core kmod-usb2
+    DEVICE_PACKAGES := kmod-usb-core kmod-usb-gadget kmod-usb-chipidea
     BOARDNAME := TL-WR703N
```
blogic: users will most likely expect usb support on the port and not gadget. can you drop this part from the patch please.

package/kernel/linux/modules/usb.mk
```diff
@@ -1375,6 +1376,30 @@ endef
...
+  DEPENDS:=+USB_GADGET_SUPPORT:kmod-usb-gadget
+  KCONFIG:=\
+	CONFIG_NOP_USB_XCEIV=y \
+	CONFIG_EXTCON \
+	CONFIG_USB_CHIPIDEA \
+	CONFIG_USB_CHIPIDEA_HOST=y \
+	CONFIG_USB_CHIPIDEA_UDC=y \
+	CONFIG_USB_CHIPIDEA_DEBUG=y \
+	CONFIG_USB_EHCI_HCD=n
+  FILES:=\
```
neykov: We can compile them, but not insert them together - this is meant to prevent the latter. Should I just remove it or is there a better way to express it? Can I do something similar for the lede menuconfig instead?

target/linux/ar71xx/image/tp-link.mk
neykov: Good point, updated.

package/kernel/linux/modules/usb.mk
blogic: this logic should not be represented inside the mk file. we can expect users to know what they are installing and not install both at the same time

neykov: Removed

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -426,7 +426,7 @@ TARGET_DEVICES += tl-wpa8630
 define Device/tl-wr703n-v1
     $(Device/tplink-4mlzma)
     DEVICE_TITLE := TP-LINK TL-WR703N
-    DEVICE_PACKAGES := kmod-usb-core kmod-usb2
+    DEVICE_PACKAGES := kmod-usb-core kmod-usb-chipidea
     BOARDNAME := TL-WR703N
```
blogic: please do not change the default package selection here. people will expect host mode as the default. happy to merge this once this bit has been fixed.

@neykov does usb gadget support dual role devices ? i would like to have a 703 i can use for mobile debugging, but that would require the usb device to expose an ethernet and console endpoint.

neykov: The chipidea driver supports host mode as well. Agree usb2 is more sensible as a default since chipidea hasn't seen too much testing on this platform. And using it as a device requires (minor) hardware modification of the boards so no big deal if the default builds are not chipidea based.

Yes it's possible to load several device modules at the same time. Tried it a long time ago and it was working just fine. Here is some documentation on the topic - https://www.kernel.org/doc/Documentation/usb/gadget_multi.txt.

I've been recommended another board based on the same platform which is much more compact - [Zsun WiFi Card Reader](https://wiki.hackerspace.pl/projects:zsun-wifi-card-reader). Might suit you better.

blogic:
merged, Thanks !

anotherfiz:
How would I merge this patch into an openwrt menuconfig?

neykov:
@anotherfiz  https://github.com/openwrt/openwrt/pull/327

anotherfiz:
Can you expand?

neykov:
@anotherfiz The same patch is filed against the openwrt source repository - you can directly checkout the PR branch or apply the changes on the latest master. You can download the patch from [1]. 

I strongly recommend to use the LEDE distribution though since OpenWRT development on this platform has stagnated (last time I checked). And obviously this patch is already available in LEDE.


[1] https://github.com/openwrt/openwrt/pull/327.patch

anotherfiz:
I see, I am trying to add this support for the 8devices rambutan module - they referenced this merge for USB device support here from this thread: http://www.8devices.com/community/viewtopic.php?f=22&t=985

I am wondering if there is a way I can do this manually with their git https://github.com/8devices/openwrt-8devices  , since 8devices seems to also have stagnated development for their own device

neykov:
1) download patch
2) `git clone openwrt-8devices`
3) `cd openwrt-8devices`
4) `git apply <path to patch>`

If you get any conflicts you'll have to modify the sources to accommodate the differences.

PR #707, opened 2017-01-14
ipq806x: refactor device tree, disable usb3 suspend, update eMMC stuff.

dissent1:
With this several commits we are:
1. Disabling usb3 phy suspend and adding tcsr control register into USB nodes in DT as described in Qualcomm SDK. According to QSDK usb3 phy suspend should be disabled for ipq806x.
This commit is runtime tested. I have no feedback, but it may help solving FS#177.

2. Enabling dma and bam for eMMC and SDCC3 and updating nodes in correspondence to Qualcomm SDK.
**This commit needs to be tested.**

3. Enabling hardware pseudo rng. This one is already present in ipq8065.dtsi so we're just adding it into ipq8064 as well.

And last, but not least:
4. Refactoring ipq8065 device tree to contain only specific for ipq8065 entries.
At the moment we have 2 seperate device tree sources: 1 for ipq8064 soc located in kernel source and 1 custom in LEDE for ipq8065.
ipq8064 and ipq8065 SoCs are completely identical except ipq8065 has higher cpu frequencies and adjusted corresponding power supply.

It makes easier to manage ipq806x device trees and instead of committing changes into 2 seperate equal DTS we are ending up with only 1.
It also enables ipq8065 devices to automatically receive changes made in upstream kernel.


blogic:
merged, great work !

dissent1:
Thanks!

PR #708, opened 2017-01-15
wwan: adding 3G rssi led indication

fid0did0:
Using this patch the system will provide the received
signal strength indication get from the the 3G modem,
driving properly the available leds.
The leds driving function is board dependent.
One example for the DWR-512 is added.

### DIFF
--- a/package/network/utils/wwan/files/wwan.sh
+++ b/package/network/utils/wwan/files/wwan.sh
@@ -99,6 +99,9 @@ proto_wwan_setup() {
 	comgt)		proto_3g_setup $@ ;;
 	*cdc_ncm)	proto_ncm_setup $@ ;;
 	esac
+
+	echo "Starting rssi_led_script"
+	nice -n10 set_3g_rssi_led.sh start &
 }
 
 proto_wwan_teardown() {
@@ -114,6 +117,12 @@ proto_wwan_teardown() {
 	comgt)		proto_3g_teardown $@ ;;
 	*cdc_ncm)	proto_ncm_teardown $@ ;;
 	esac
+
+	echo "Stopping rssi_led_script"
+	led_proc_id=$(ps | grep set_3g_rssi_led | awk '/start/ {print $1}')
+	echo "pid: $led_proc_id"
+	kill "$led_proc_id"
+	set_3g_rssi_led.sh stop
 }
 
 add_protocol wwan
--- a/package/network/utils/wwan/Makefile
+++ b/package/network/utils/wwan/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=wwan
-PKG_VERSION:=2014-07-17
-PKG_RELEASE=1
+PKG_VERSION:=2017-01-14
+PKG_RELEASE=1.1
 
 PKG_LICENSE:=GPL-2.0
 PKG_LICENSE_FILES:=
@@ -24,6 +24,8 @@ endef
 define Package/wwan/install
 	$(INSTALL_DIR) $(1)/lib/netifd/proto/
 	$(CP) ./files/wwan.sh $(1)/lib/netifd/proto/
+	$(INSTALL_DIR) $(1)/bin/
+	$(INSTALL_BIN) ./files/set_3g_rssi_led.sh $(1)/bin/
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/usb
 	$(INSTALL_BIN) ./files/wwan.usb $(1)/etc/hotplug.d/usb/00_wwan.sh
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/usbmisc
--- a/package/network/utils/wwan/files/set_3g_rssi_led.sh
+++ b/package/network/utils/wwan/files/set_3g_rssi_led.sh
@@ -1,0 +1,113 @@
+#!/bin/sh
+
+. /lib/ramips.sh
+
+set_led_rssi_off() {
+	case $board in
+	dwr-512-b)
+		echo '0' > /sys/class/leds/dwr-512-b\:green\:sigstrength/brightness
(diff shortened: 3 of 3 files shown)


fid0did0:
This shall be currently considered as a starting point to start a discussion about the framework.

package/network/utils/wwan/files/wwan.sh
```diff
@@ -114,6 +117,12 @@ proto_wwan_teardown() {
 	comgt)		proto_3g_teardown $@ ;;
 	*cdc_ncm)	proto_ncm_teardown $@ ;;
 	esac
+
+	echo "Stopping rssi_led_script"
+	led_proc_id=$(ps | grep set_3g_rssi_led | awk '/start/ {print $1}')
+	echo "pid: $led_proc_id"
+	kill "$led_proc_id"
+	set_3g_rssi_led.sh stop
```
fid0did0: Currently the script is unable to stop the set_3g_rssi_led script.

Manually I can get the proc id:
```
root@lede:/tmp# ps | grep set_3g_rssi_led | awk '/start/ {print $1}'
1165
```

but the script is unable as shown by logread:
```
Sun Jan 15 09:25:46 2017 daemon.notice netifd: wan3g (1336): Stopping rssi_led_script
Sun Jan 15 09:25:46 2017 daemon.notice netifd: wan3g (1336): pid:
Sun Jan 15 09:25:46 2017 daemon.notice netifd: Interface 'wan3g' is now down
```

Some idea to help me?

heinzek:
Hey @fid0did0 I check Your script and use it to Archer MR200. 
It look creepy but works good.
Check my changes:
http://tplink-forum.pl/pub/temp/set_rssi_3g_leds.sh

ps. MR200 need only gcom to run this script

fid0did0:
Hi heinzek,
Thanks for your report.
I merged your code and reported you in the commit message.

> It look creepy but works good.
I Know. I'm also not completely satisfied. 
Therefore if you have some advice to improve the framework we can discuss. I'm open to change.

On my board, the wwan.sh is unable to get the proc id and then the set_3g_rssi_led.sh cannot be stopped.
Is your board able to switch off the leds when you ifdown the interface?

> ps. MR200 need only gcom to run this script
@malaakso  shall we add the gcom  as a dependency?

mkresin:
Could it be that you are reinventing the wheel? We have:

- [a rssi led trigger](https://github.com/lede-project/source/blob/master/package/base-files/files/etc/init.d/led#L34)
- handled by the [rssileds package](https://github.com/lede-project/source/tree/master/package/network/utils/rssileds)
- the [ucidef_set_led_rssi](https://github.com/lede-project/source/blob/master/package/base-files/files/lib/functions/uci-defaults.sh#L565) and [ucidef_set_rssimon](https://github.com/lede-project/source/blob/master/package/base-files/files/lib/functions/uci-defaults.sh#L591) functions to add the leds via board.d file

Never used any of the above but it looks like somehow related. If you ```git grep "rssi"``` trough the LEDE source, you will find more rssi related stuff.

fid0did0:
Hi Mathias,
If I understood well the rssi led trigger is used for the wifi rssi.
This patch provides indication about the 3G modem rssi.

For sure it is possible to write a 3g rssi led trigger, but not very easy if you consider that the control modem interface is not very standardized. This is probably the reason why we have the wwan (3G modem handling) as a script as well.

Bye.

heinzek:
> I merged your code and reported you in the commit message.

I miss one part:
```
start)
	until [ 0 -gt 1 ]; do
		case $board in
		dwr-512-b)
			rssi3g=$(gcom -d /dev/ttyUSB0 -s /etc/gcom/getstrength.gcom|awk '/[CSQ]/ { print $2 }'|sed -e 's/,//')
			;;
		mr200)
			rssi3g=$(gcom -d /dev/ttyUSB1 sig |awk -F[\ ,] '/^\Signal/ {print $3}')
			sleep 2
			lte=$(gcom -d /dev/ttyUSB1 reg |awk -F[,] '/^\Registered/ {print $2}')
			;;
		esac

```
MR200 use ttyUSB1 to get signal and connection mode.

> On my board, the wwan.sh is unable to get the proc id and then the set_3g_rssi_led.sh cannot be stopped.
> Is your board able to switch off the leds when you ifdown the interface?

My board not using up/down interface. Modem work like Huaweii HiLink and have our webinterface. I need adb package to run it.
I run router and add rndis module to get working connection. If I need change APN or PIN i log in to webinterface and change this.

I think to create a separate package with this script, and add comgt as dependency.

Check 3ginfo package. This package can read signal strenght from a loots diffrent modems using one small config file. https://github.com/obsy/packages/tree/master/3ginfo

fid0did0:
Hi all,
Mathias I confirm you that the rssileds package support only the wifi links. The monitor uses the iwinfo library.
Nevertheless, I will explore if make sense to integrate there the 3G monitoring or similar alternative solution.
This activity will take a while. In the meantime, I will keep open this PR.

Heinzek,
> MR200 use ttyUSB1 to get signal and connection mode.

you are right I will modify the code to handle the IF as a script parameter.

>My board not using up/down interface. Modem work like Huaweii HiLink and have our webinterface. I need adb package to run it.
I run router and add rndis module to get working connection. If I need change APN or PIN i log in to webinterface and change this.

@malaakso do you think the wwan package can support this modem?

Bye.

malaakso:
I don't think that it makes sense to embed the signal quality checking in the led script. We'd first need a sane way to check signal quality periodically and make it available to ubus in a standardized way (I'm thinking proto_run_command in qmi.sh/ncm.sh/...). Then it would be possible to create device specific scripts to read ubus and toggle the leds.

Regarding MR200, if it works like HiLink with its own web interface, it's can't be supported by wwan.

fid0did0:
Matti,
I agree, what you are proposing is better.
I will close this PR and integrate the device led setting as soon as we have the signal quality check in the ubus.
Bye.

PR #711, opened 2017-01-16
image.mk: When using per-device-rootfs create per-rootfs manifest

cshoredaniel:
Prior to this the manifest in bin/... ended was whichever
rootfs was created last.  This patch creates a separate
per-device manifest when the rootfs is per-device; for
non-per-device-rootfs the behaviour remains unchanged.

### DIFF
--- a/include/image.mk
+++ b/include/image.mk
@@ -254,10 +254,10 @@ endef
 
 define Image/Manifest
 	$(STAGING_DIR_HOST)/bin/opkg \
-		--offline-root $(TARGET_DIR) \
+		--offline-root $(2) \
 		--add-arch all:100 \
 		--add-arch $(if $(ARCH_PACKAGES),$(ARCH_PACKAGES),$(BOARD)):200 list-installed > \
-		$(BIN_DIR)/$(IMG_PREFIX)$(if $(PROFILE_SANITIZED),-$(PROFILE_SANITIZED)).manifest
+		$(1).manifest
 endef
 
 ifdef CONFIG_TARGET_ROOTFS_TARGZ
@@ -318,6 +318,7 @@ define Device/Init
 
   IMAGES :=
   IMAGE_PREFIX := $(IMG_PREFIX)-$(1)
+  IMAGE_NAME_PREFIX = $$(IMAGE_PREFIX)
   IMAGE_NAME = $$(IMAGE_PREFIX)-$$(1)-$$(2)
   KERNEL_PREFIX = $$(IMAGE_PREFIX)
   KERNEL_SUFFIX := -kernel.bin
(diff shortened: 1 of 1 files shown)


jow-:
Looks okay on first sight - @nbd168 ?

include/image.mk
```diff
@@ -254,10 +254,10 @@ endef
 
 define Image/Manifest
 	$(STAGING_DIR_HOST)/bin/opkg \
-		--offline-root $(TARGET_DIR) \
+		--offline-root $(2) \
 		--add-arch all:100 \
 		--add-arch $(if $(ARCH_PACKAGES),$(ARCH_PACKAGES),$(BOARD)):200 list-installed > \
-		$(BIN_DIR)/$(IMG_PREFIX)$(if $(PROFILE_SANITIZED),-$(PROFILE_SANITIZED)).manifest
+		$(BIN_DIR)/$(1).manifest
```
nbd168: Please remove the $(BIN_DIR)/ prefix here

include/image.mk
```diff
@@ -464,6 +465,7 @@ define Device/Build/image
   .IGNORE: $(BIN_DIR)/$(call IMAGE_NAME,$(1),$(2))
   $(BIN_DIR)/$(call IMAGE_NAME,$(1),$(2)): $(KDIR)/tmp/$(call IMAGE_NAME,$(1),$(2))
 	cp $$^ $$@
+	$(if $(TARGET_PER_DEVICE_ROOTFS),$$(call Image/Manifest,$(call IMAGE_NAME_PREFIX),$(KDIR)/target-dir-$$(ROOTFS_ID/$(3))))
```
nbd168: Use $$(call Image/Manifest,$$@,...) here

include/image.mk
```diff
@@ -559,6 +561,6 @@ define BuildImage
 	$(MAKE) legacy-images
 
   install: install-images
-	$(call Image/Manifest)
+	$(call Image/Manifest,$(IMG_PREFIX)$(if $(PROFILE_SANITIZED),-$(PROFILE_SANITIZED)),$(TARGET_DIR))
```
nbd168: and put back the $(BIN_DIR)/ prefix here

cshoredaniel:
Will make the requested changes this weekend.  Basically using moving $(BIN_DIR) into the call instead of in the target macro.

cshoredaniel:
Updated.

cshoredaniel:
Actually, let me make sure I pushed the final branch of that; github doesn't look right (should be $^ I think).

cshoredaniel:
ah, no, that's right I had replace the wrong string

cshoredaniel:
Closing as have taken a break from LEDE/OpenWrt development for various reasons, so will reopen when this is back on the radar.

PR #712, opened 2017-01-16
include target: Unify and sanitize image and tool tarball filenames

cshoredaniel:
Currently tool tarballs (sdk, toolchain, and imagebuilder)
do not use the same name scheme as the firmware images.
This patch moves the filenaming logic into a common
location and updates the targets so that in future the
various files will all follow the same naming logic.


jow-:
Can you provide an example directory listing (`ls -l bin/targets/*/`) with the change applied?

cshoredaniel:
Do you want a straight LEDE build ls - or is one with a customized 'Image Options' .config options fine?

cshoredaniel:
Closing as have taken a break from LEDE/OpenWrt development for various reasons, so will reopen when this is back on the radar.

PR #714, opened 2017-01-17
packages/procd: add callbacks for additional scripts on reload_config

feckert:
- "/sbin/reload_config.service" is called for every service affected by
  a configuration change.
- "/sbin/reload_config.custom" is called once from reload_config.

This could be used to make related changes/informs on a config change.


nbd168:
This seems like the wrong place to add these kinds of hooks, because theoretically other places could emit config change events as well in the future. What are you trying to use this for?

feckert:
@nbd168 Yes you are right the hooks are only called if the "reload_config" comand is called.
I want to inform an ACS if the config has changed to start an inform.

nbd168:
You don't need any extra hooks for that. Add a service trigger for config.change to your init script and make it call a script for sending out the inform.

feckert:
@nbd168 I will try it.
Only for clarification. If the user saves a changed network config over luci then the "/etc/init.d/network" will be reloaded with the changed options. If this happens then my ACS should be informed that there is a change and will upload the changes. Does this work with your suggestion?

PR #718, opened 2017-01-18
linux: addition of patch to allow mtd m25p80 device to reboot

greenbreakfast:
Addition of mtd driver patch for m25p80 device:
* the existing m25p_remove() function is added to the spi driver as a shutdown function
* modified the m25p_remove() function to write SPI flash reset codes before performing a shutdown

Signed-off-by: Lazar Demin <lazar@onion.io>


mkresin:
NAK!

This patch was already send as Github PR and dropped in favour of backporting a patch that was send to linux-mtd. Have a look at https://github.com/lede-project/source/pull/620 for details.

For reference, the patch was already send in:

- https://github.com/lede-project/source/pull/464
- http://patchwork.ozlabs.org/patch/713750/

It was tried to get the very same patch upstream and that was the upstream response:

 - http://lists.infradead.org/pipermail/linux-mtd/2016-November/070368.html
 - http://lists.infradead.org/pipermail/linux-mtd/2016-November/070464.html

PR #719, opened 2017-01-18
ramips : fix mac address of miwifi-mini

hbl0307106015:
For the miwifi-mini, the offset of ethernet mac should be 0x28, which you can easyily dump from 'Factory' partition. And, add the case of miwifi-mini in ramips_setup_macs()  will make the allocation of MAC address in a more properly way.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -357,6 +357,10 @@ ramips_setup_macs()
 		lan_mac=$(mtd_get_mac_binary factory_info 13)
 		wan_mac=$(macaddr_add "$lan_mac" 1)
 		;;
+	miwifi-mini)
+		wan_mac=$(cat /sys/class/net/eth0/address)
+		lan_mac=$(macaddr_setbit_la "$wan_mac")
+		;;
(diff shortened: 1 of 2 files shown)


target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -357,6 +357,11 @@ ramips_setup_macs()
 		lan_mac=$(mtd_get_mac_binary factory_info 13)
 		wan_mac=$(macaddr_add "$lan_mac" 1)
 		;;
+	miwifi-mini)
+		wan_mac=$(cat /sys/class/net/eth0/address)
+		lan_mac=$(cat /sys/class/net/eth0/address)
+		lan_mac=$(macaddr_setbit_la "$lan_mac")
+		;;
```
blogic: you can do this in 2 lines

+		wan_mac=$(cat /sys/class/net/eth0/address)
+              lan_mac=$(macaddr_setbit_la "$wan_mac")

hbl0307106015: OK, the file has been changed, check it , thanks!

blogic:
merged, Thanks !

PR #721, opened 2017-01-18
Update to IPtables 1.6.0

Ansuel:
Removed unnecessary package and patches.
Tested and compiled on linksys wrt1900acs


diizzyy:
Nice! Any noticeable differences in general?

package/network/utils/iptables/Makefile
```diff
@@ -9,28 +9,26 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=iptables
-PKG_VERSION:=1.4.21
-PKG_RELEASE:=2
+PKG_VERSION:=1.6.0
+PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
 PKG_SOURCE_URL:=http://www.netfilter.org/projects/iptables/files \
```
diizzyy: Only http://www.netfilter.org/projects/iptables/files is a vaild source URL.
You can also add http://www.hu.netfilter.org and http://www.lt.netfilter.org as sources. 
Please place these before www.netfilter.org to keep main site offloaded.

Ansuel: Didn't notice that, now should be good. Right?

diizzyy: Yes, I would however recommend you to place the www.netfilter.org as last site. It's not written anywhere but we're trying to be nice to upstream by using mirrors provided whenever possible as a nice gesture.

nbd168:
Are you really sure that all of these patches are not relevant anymore?
I'm not convinced

package/network/utils/iptables/Makefile
```diff
@@ -13,9 +13,9 @@ PKG_VERSION:=1.6.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
-PKG_SOURCE_URL:=http://www.netfilter.org/projects/iptables/files \
-	http://www.lt.netfilter.org/projects/iptables/files/ \
-	http://www.hu.netfilter.org/projects/iptables/files/
+PKG_SOURCE_URL:= \ http://www.lt.netfilter.org/projects/iptables/files/ \
```
diizzyy: Typo... :/

Ansuel:
Well, the musl one is already included in the new version...
The 030 , the dependencie about that library is added to the makefile
Don't know for the other. My router is working fine for now with this commit included in a normal image and from the log i don't see any error

package/network/utils/iptables/Makefile
Ansuel: Just a little tired

diizzyy:
```
020 - 	insmod cmd exists in stock build, should be in?
030 -	not having libnfnetlink....
	No http://manpages.ubuntu.com/manpages/wily/man8/nfnl_osf.8.html
	Netfilter LOG over NFNETLINK interface
	Source: https://wiki.gentoo.org/wiki/Iptables
050 -	To save space? iptables-xml is used to convert the output of iptables-save to an XML format.
	Source. http://www.linuxfromscratch.org/blfs/view/cvs/postlfs/iptables.html
100 - 	Isn't installed as far as I can tell, doesn't matter?
200 -	Not sure what this fixes/does
300 -	If it compiles we don't need it? libxt_TCPOPTSTRIP stuff doesn't look like a musl fix?
600 -	To save space?
700 -	Same as above?
```

wongsyrone:
The dependency to libxtables can be disabled since we already have one in the codebase.

My modification is a little messy, if someone has interest, you can check https://github.com/wongsyrone/lede-1/commit/315c6f30ab6879b17d251b5c2b04214b54bbf60f and https://github.com/wongsyrone/lede-1/tree/master/package/network/utils/iptables

Ansuel:
I have added the corrected patch. Compiled and working fine on my router

mkresin:
Please squash your commits into one!

Please have a look at https://github.com/bobafetthotmail/source/blob/c7cfdff4fa40f53511b8c8a4bd2e60aa980dc6fd/.github/pull_request_template.

The commit subject is wrong,  the commit description and your Signed-off-by line is missing.

PR #722, opened 2017-01-19
[RFC] base-files: tell the kernel about ubox's modprobe if no /sbin/modprobe

commodo:
Not sure how valid this fix is.
Since ubox's /usr/sbin/modprobe points to /sbin/kmodloader
and the arguments it takes, differ from the default /sbin/modprobe.

The main gist of things is that the kernel's default is

```
root@(none):/# cat /proc/sys/kernel/modprobe
/sbin/modprobe
```

I got a case [ for `python-iptables` package ] not working
because it tries to read `/proc/sys/kernel/modprobe` and fail
when trying to run `/sbin/modprobe`.

My proposed fix [ well, up for discussion first ] is to
update `/proc/sys/kernel/modprobe` with `/usr/sbin/modprobe`
if it's available and no `/sbin/modprobe` is available by default.

Tested on LEDE trunk 1fb673ee12cad775acf8cb89fc8608e8c3c457a2
On x86_64.

### DIFF
--- a/package/base-files/files/etc/init.d/boot
+++ b/package/base-files/files/etc/init.d/boot
@@ -36,6 +36,9 @@ boot() {
 	grep -q debugfs /proc/filesystems && /bin/mount -o noatime -t debugfs debugfs /sys/kernel/debug
 	[ "$FAILSAFE" = "true" ] && touch /tmp/.failsafe
 
+	[ ! -x /sbin/modprobe -a -x /usr/sbin/modprobe ] && \
+		echo /usr/sbin/modprobe > /proc/sys/kernel/modprobe
+
 	/sbin/kmodloader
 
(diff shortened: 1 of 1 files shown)


zorun:
The same issue is fixed by one of @yousong 's proposed patches: https://patchwork.ozlabs.org/patch/713718/

It seems cleaner to avoid meddling with the kernel, and just use `/sbin/` instead, because that's what most other distributions seem to do.

commodo:
actually that does look better
the only thing to take into consideration there, is to avoid collisions between ubox's modprobe & busybox's

commodo:
Tried that patch and it worked fine too for my case.
@yousong any idea when/if that patch gets merged into the main tree ?

Will close this in the meantime.

Thanks
Alex

PR #723, opened 2017-01-19
ar7xxx: improve Mikrotik hAP Lite device support

samm-git:
This patches are to improve situation with rb941 support, see https://bugs.lede-project.org/index.php?do=details&task_id=407. CONFIG_MTD_SPI_NOR_USE_4K_SECTORS is removed because it makes a problem with this (and probably many other devices). 

Patches are based on work and conversations with Sergey Sergeev, author of the hAP lite support.


hauke:
1. Please rebase to remove ther merge
2. please add a Signed-off-by line
3. one patches removes something an an other commit adds this again later, please merge these commits.

samm-git:
Done, please review.

nbd168:
I think you should not disable CONFIG_MTD_SPI_NOR_USE_4K_SECTORS. Instead you should make the device use plain squashfs+jffs2 without UBI.

nbd168:
Sorry, I misread the filesystem and it's already using plain squashfs+jffs2. Will look into the 4k issues

nbd168:
Could you please describe in detail what issues you had with 4K erase blocks enabled?

samm-git:
Its not able to mount jffs in the reasonable time (so it boots from the tmpfs on timeout), also mtd erase was failing. Plus if you will take a look on the kernel description - it is sounds that feature itself is very buggy:


> 29           Please note that some tools/drivers/filesystems may not work with
 30           4096 B erase size (e.g. UBIFS requires 15 KiB as a minimum).

Also there are some discussions in the original message where this flag is considered buggy. I would not recommend to enable it unless you are 100% sure that it will work correctly, also Sergey Sergeev confirmed that its not needed for the older mikrotik devices as well.

nbd168:
What do you mean by 'not able to mound jffs in the reasonable time'? Do you have logs or something?

The feature itself is not buggy, it is simply incompatible with ubifs (which does not make sense on these kinds of NOR flash chips anyway).

The reason for using it is that jffs2 can make better use of the flash space, since it can erase and rewrite smaller chunks.

samm-git:
@nbd168 i can collect the logs, but you can check yourself as well. It was something like "found erase block on jffs2", and then - nothing in the logs and no mount, so i think it was some erase timeouts. Again, this was disabled with message "4k support is messy" + i found a lot of the reports in the net with a problems on jffs2 and this option enabled. I cant really get any objection why it should be enabled if we know that it  is causing a problems for this board. Do you know any mikrotil board which works correctly and faster with it?

nbd168:
I think disabling this option without knowing what's going on is a bad idea. 
There really isn't a lot of magic to this, so if it's breaking now, there might be a bug involved which could bite us later even with that option disabled.

samm-git:
I think that _enabling_ option known is buggy without strong objections to do so i s a very bad idea. Know lede snapshots for this device are just broken, this patch addressing that. If you want to fix 4k support on this specific device and enable it - thats great. But just not to accept fix for the things are broken sounds strange for me.

And again - do you know any mikrotik device which benefits from it or it just was enabled "because we can" ?

samm-git:
http://lists.infradead.org/pipermail/linux-mtd/2010-September/031876.html - please read that, as well as many other discussions on that issue

nbd168:
I don't have such a device, so I can't get logs for this issue myself. On other targets, e.g. ramips, it seems to be working.
Please note that I'm not opposed to disabling this option, I just want logs to see how exactly it is breaking in this case to make sure it's not an issue we will run into again later.

samm-git:
Ok, i will collect logs for you, no problems.

samm-git:
I did this experiment, from what i see - it actually works with 4k, but from 2 to 10 times slower compared to kernel without it. See https://gist.github.com/samm-git/6face53bc32d9b3ba58cba99ad3a3af8

samm-git:
Another test:
Another test (from kernel w/o 4k)
~~~
root@lede:/# time mtd erase /dev/mtd6
Unlocking /dev/mtd6 ...
Erasing /dev/mtd6 ...
real	0m 19.12s
user	0m 0.00s
sys	0m 12.93s
~~~
From kernel with MTD_SPI_NOR_USE_4K_SECTORS=y:

```
root@lede:~# time mtd erase /dev/mtd6
Unlocking /dev/mtd6 ...
Erasing /dev/mtd6 ...
real	2m 57.87s
user	0m 0.00s
sys	1m 59.72s
```

nbd168:
Thanks for the info, that makes more sense to me. By the way, I think you should only get these jffs2 init error messages when running sysupgrade between a 4k-enabled and a non-4k-enabled image.
Regarding this pull request, please change the title to something like:
"ar71xx: improve Mikrotik hAP Lite device support"
Also, please get rid of nand-utils on this device, it should not be necessary.
The 00-default.mk should be in a separate commit, since it's a fix that affects other devices as well.

samm-git:
Done, please review.

nbd168:
Please read my last comment again and update the commits accordingly:
- commit title
- splitting out the 00-defaults.mk change

Thanks

samm-git:
Thank you, i hope it is ok now.

samm-git:
@nbd168 any chance to get it merged? Also i have a patch to enable metadata generation for that device image, do you want me to do a separate PR or should i update this one?

nbd168:
Merged with some improvements to the commit messages.
Please make a separate PR for metadata generation

PR #724, opened 2017-01-20
ramips: update ZyXEL Keenetic boards

13hakta:
add: Export gpio to allow power management of USB port.
fix: For Keenetic Viva fix squashfs alignment as ZyXEL web-flasher requires to be aligned to 64kb.


jow-:
Pulled into my staging tree: https://git.lede-project.org/1c9a817

PR #726, opened 2017-01-20
ramips: cleanup mt7620 Netgear image code

jclehner:
1. Reuse functions defined in `image-commands.mk`
2. Rename fakeroot image


mkresin:
By removing the BLOCKSIZE parameter your are not disabling the 4K erase sector size for the images. It's just that pad-rootfs (called in the ```$$(sysupgrade_bin)``` build step) pads the image to a multiple of 64KByte instead of 4KByte. Therefore your commit message is at least bogus. You should point out which issue you are seeing!

According to the wiki, the ex2700 has a MX25L3205D flash chip. The datasheet of the MX25L3205D says "1024 Equal Sectors with 4K byte each; any Sector can be erased individually. The 4K support for this chip is enabled with target/linux/generic/patches-4.4/043-mtd-spi-nor-mx25l3205d-mx25l6405d-append-SECT_4K.patch. Means, ```cat /proc/mtd``` should show an erase size of 0x1000.

Not sure about the WN3000RPv3. The commit message from @f00b4r0 doesn't list which 8Mbyte NOR SPI chip is on the WN3000RPv3. But since the WN3000RPv3 has a 8 MByte flash chip, wasting up to 60KByte with padding doesn't hurt that much as with 4 MByte flash chips.

f00b4r0:
> Le 20 janv. 2017 à 19:49, Mathias Kresin <notifications@github.com> a écrit :
> Not sure about the WN3000RPv3. The commit message from @f00b4r0 <https://github.com/f00b4r0> doesn't list which 8Mbyte NOR SPI chip is on the WN3000RPv3. But since the WN3000RPv3 has a 8 MByte flash chip, wasting up to 60KByte with padding doesn't hurt that much as with 4 MByte flash chips.
> 
> 
The WN3000RPv3 has a Macronix MX25L6406EM2I-12G chip.

HTH

jclehner:
Okay, I was under the impression that `BLOCKSIZE` referred to the flash chip's erase*block* size, which *is* 64K for these chips ("64/128 Equal Blocks with 64K byte each"). 

The kernel padding of 64K on these devices is mandated by the bootloader, so I'll add a blocksize parameter to `netgear-kernel` instead.

jclehner:
I've updated the patch now, leaving `BLOCKSIZE` untouched, using `... | netgear-kernel 64k` instead.

target/linux/ramips/image/mt7620.mk
```diff
@@ -7,16 +7,13 @@ define Build/tplink-header
...
-define Build/netgear-header
-	$(STAGING_DIR_HOST)/bin/mkdniimg \
-		$(1) -v OpenWrt -i $@ \
-		-o $@.new && mv $@.new $@
+define Build/netgear-kernel
+	let \
+		bsize="$(subst k,*1024,$(1))" \
+		size="$$(stat -c%s '$@')" \
+		size="((1 + ((64 + size) / bsize)) * bsize) - 64"; \
+	dd if=$@ of=$@.new bs=$$size count=1 conv=sync \
```
mkresin: Can't we get rid of the whole block by using the existing [pad-to](https://github.com/lede-project/source/blob/master/include/image-commands.mk#L128) command?

If so, I would suggest to rename ```Build/netgear-kernel``` to ```Build/append-file```:

```
Build/append-file
  cat "$(1)" >> "$@"
endef
```

and change your KERNEL line to something like:

```
KERNEL := $(KERNEL_DTB) | uImage lzma | pad-to 64k | append-file netgear-fakeroot.uImage
```

jclehner: The second uImage header must be located in the last 64 bytes of the kernel image padded to 64k. The code above would add the uImage header *after* the padding.

This is the relevant code in the bootloader:

```
hdr = (((((ulong)hdr + sizeof(image_header_t) + ntohl(hdr->ih_size))
                / 0x10000) + 1 ) * 0x10000) - sizeof(image_header_t);
if (ntohl(hdr->ih_magic) != IH_MAGIC) {
    puts ("Second part Bad Magic Number\n");
    goto bad_img;
}

```

mkresin: Thanks a lot for the explanation. If I got it right, you need something like ```pad-to 65472```. But since it is kinda hard to read I would suggest to use ```pad-offset 64k 64``` which should create exactly the padding you need.

Now that I know that netgear-fakeroot.uImage  is just a fake uImage header, I would suggest to rename this file to something like netgear-fake-uImage-header. But that is really something optional and you might find even a better name.

jclehner: `pad-offset 64k 64` does exactly what I need, totally missed that for some reason, thanks!

The fakeroot file is simply the header of a zero-byte uImage, something which can't be created using `mkuimage`.

mkresin:
I'm really sorry for not noticing this earlier but we have already a ```append-file``` build step in ```target/linux/ipq806x/image/Makefile```.

Would you please create an extra commit, which moves the this build step from ```target/linux/ipq806x/image/Makefile``` to ```include/image-commands.mk```. This way you can reuse the build step in ```mt7620.mk``` without duplicating the same code again.

The commit moving the ```append-file``` build step should be the first one. The changes to the mt7620 netgear images should be the second commit.

Feel free to say you're sick of changing it once more. In that case I might change it before merging the PR. Anyway, would be really kind if you can do the extra step.

<hr>

Would you please update the title of the "ramips: update mt7620 Netgear image code" commit. Update could be all or nothing.

Please use "ramips: cleanup mt7620 Netgear image code" if the generated image is the same as before and you have just done a code cleanup.

Please use "ramips: fix mt7620 Netgear image code" in case the generated image is not the same as before. Please describe in the commit message what was wrong and what didn't worked in the end.

I've noticed that a fake uImage header is created in ```target/linux/ipq806x/image/Makefile``` as well. Maybe it's time to add a patch to ```tools/mkimage```, which allows to create an uImage header for zero-byte images instead of workaround this limitation over and over again. But that is for sure out of scope of this PR. If you are interested, contributions are welcome. I'm not a big fan of having binary files in LEDE.

jclehner:
Commits are now split.

> I've noticed that a fake uImage header is created in target/linux/ipq806x/image/Makefile as well. Maybe it's time to add a patch to tools/mkimage, which allows to create an uImage header for zero-byte images instead of workaround this limitation over and over again. But that is for sure out of scope of this PR. If you are interested, contributions are welcome. I'm not a big fan of having binary files in LEDE.

Will do that soon!

jclehner:
Commits are now split, mkimage changes are in PR #739.

mkresin:
Merged into my staging tree. Thanks a lot for taking the time to do the clean up work!

PR #728, opened 2017-01-22
uboot-kirkwood: add uboot for nsa310b

bobafetthotmail:
this commit allows to make a standalone u-boot for nsa310b.

While both first-stage and second-stage u-boot work fine if
installed to flash or loaded with kwboot,
I could not get stock u-boot nor bodhi's u-boot to chainload
any second stage u-boot (I also tried with dockstar's uboot
that works fine on this device if loaded with kwboot).

### DIFF
--- a/package/boot/uboot-kirkwood/Makefile
+++ b/package/boot/uboot-kirkwood/Makefile
@@ -58,6 +58,10 @@ define uboot/iconnect_second_stage
   TITLE:=second stage U-Boot for Iomega iConnect Wireless
 endef
 
+define uboot/nsa310
+  TITLE:=U-Boot for Zyxel NSA310
+endef
+
 define uboot/pogo_e02
   TITLE:=U-Boot for Cloud Engines Pogoplug E02
 endef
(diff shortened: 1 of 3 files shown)


nbd168:

bobafetthotmail:
please drop it from the staging tree, I need to alter default mtdparts due to feedback in other PRs.

kofec:
@bobafetthotmail 
I have bodhi's u-boot. Before changes provided by nbd168. I was able to run LEDE from usb. 
My questions:
1) Should this work with your changes and your uImage (start with bodhi's u-boot direcly from USB) - I didn't integrate dtb file - just point where it is located in env?
2) As I understand after your changes I will be possible to start from internal flash? So extroot should work ? So then it will be even better becuase we can use f2fs or ext4 on USB ?
3) Last question - Will sysupgrade work ?
thank again for your hard work

bobafetthotmail:
1. yes, it should work like that with bodhi's uboot, also with this one (as it is pretty much a clone of bodhi's), but I didn't test this as it's not needed for LEDE.

2. yes, I have LEDE running in the flash of my nsa310. extroot will work, but should not be needed as there is 128 MiB of flash in this NAS, uboot needs 1 MiB and the rest is for LEDE.

3. yes, sysupgrade is working. I actually tested this because it's necessary for LEDE.

PR #730, opened 2017-01-23
netifd: remove dead function in init script

dvlemplgk:
stop() is defined but never actually called since
it's overwritten by rc.common due to USE_PROCD=1.

This dead stop() function exists since commit 73179a188c6eed4db08ed5f9e109fae17db6575e but never worked as intended.

### DIFF
--- a/package/network/config/netifd/files/etc/init.d/network
+++ b/package/network/config/netifd/files/etc/init.d/network
@@ -31,11 +31,6 @@ reload_service() {
 	/sbin/wifi reload_legacy
 }
 
-stop() {
-	/sbin/wifi down
-	procd_kill network ''
-}
-
 service_running() {
 	ubus -t 30 wait_for network.interface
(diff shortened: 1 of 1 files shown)


dvlemplgk:
I'm not quite sure if it's correct to just remove the dead code or reverting the mentioned commit. It tried to fix something which was already fixed. 73179a188c6eed4db08ed5f9e109fae17db6575e tried to change the order of stop_service and procd_kill calls which was already done by 44e2de03a4a96a01a732b33bf14c059da1924304. 
In the end 73179a188c6eed4db08ed5f9e109fae17db6575e broke it because now _wifi down_ isn't called at all on a network stop.

nbd168:
Pushed a better fix here: https://git.lede-project.org/?p=lede/nbd/staging.git;a=commitdiff;h=0742f91a3a85091fb1897ac3610a6ec556faace1

PR #731, opened 2017-01-23
uboot-envtools: add nsa310b envs

bobafetthotmail:
accessing the u-boot's envs on this device is required to read the mac address.
These are the envs of the new u-boot, not of the stock one.

### DIFF
--- a/package/boot/uboot-envtools/files/kirkwood
+++ b/package/boot/uboot-envtools/files/kirkwood
@@ -18,6 +18,7 @@ dockstar|\
 guruplug-server-plus|\
 ib62x0|\
 linksys-viper|\
+nsa310b|\
 pogo_e02|\
 sheevaplug|\
 sheevaplug-esata)


mkresin:
For the mentioned use-case the uboot-envtools are not required. The uboot-env tools are usually required to write to an uboot-env.

To read the mac address from the uboot-env we have [mtd_get_mac_ascii](https://github.com/lede-project/source/blob/master/package/base-files/files/lib/functions/system.sh#L11). Have a look at https://github.com/lede-project/source/blob/master/target/linux/lantiq/base-files/etc/board.d/02_network for an example how this little helper can be used.

bobafetthotmail:
Yeah, I considered it, but on kirkwood uboot-envtools are installed by default anyway (see target/linux/kirkwood/Makefile), and as you can see in the file I modified, pretty much every other kirkwood device has its envs set up in it already.

I think it's because most kirkwood devices can run Debian ARM or Arch ARM too as long as there is a dtb for them, so I thought the general consensus was leaving uboot envs accessible. 
My NSA310 runs Debian since when I bought it (years ago), for example.

mkresin:
Yeah I see.  But I fail to see the relation of running debian/arch and having access to the uboot-env/uboot-env package installed. And I don't consider "because it is done for the other boards as well" as an explanation.

I had a brief look at the dts files and the env partitions are r/w as well. I prefer to have the env partitions write protected as long as no change to env variables are required. You know, these old run  "rm -f /" or "format c:" recommendations to fix a problem.

bobafetthotmail:
>But I fail to see the relation of running debian/arch and having access to the uboot-env/uboot-env package installed.

Setting uboot for multi-booting without need to use a serial, probably, maybe use lede as some sort of backup firmware for when the Debian/Arch installation fails after some tinkering.

>And I don't consider "because it is done for the other boards as well" as an explanation.

It's not "the other boards", generic. It's "all _kirkwood_ boards have uboot-envtools installed by default and most have envs exposed by default in it". If it's wrong someone should change that too.

>I had a brief look at the dts files and the env partitions are r/w as well.

No wait a sec. The envs of LEDE's uboot for nsa310b are technically in mtd0 just like bodhi's uboot I used as a base, they are currently inside the "uboot" partition, which is ro.

Which means I really need to change the mtdparts now as it makes no sense to have this PR for a ro partition. Sigh.

>You know, these old run "rm -f /" or "format c:" recommendations to fix a problem.

Ignoring for a moment the above (it's actually read-only already due to my mistake), with uboot envs erased the uboot will use the default envs set at compile time, that will still boot LEDE (also stock uboot was like this with stock firmware booting).

kofec:
@bobafetthotmail 
I have a questions. With the stock bootloader and firwmware/software when I shutdown nsa310 I'm able to wakeup on lan via mac address provided originally . But It was not working in LEDE (only wake via mac address 00:00:00:00:00:30). Will this work after your changes ?

bobafetthotmail:
@kofec please post questions in the right PR. This is about adding support for bootloader envs.

> Will this work after your changes ?

I don't know, I didn't test WoL as I never got it to work with Debian, I thought it was working only with stock firmware. 
Post in the other PR the instructions to do a test, so I can try it and answer your question.

bobafetthotmail:
ok, updated to work with envs in first flash block of mtd1 (to stay in sync with the other PR).

nbd168:

PR #732, opened 2017-01-23
uboot-kirkwood: add uboot for nsa310b

bobafetthotmail:
@nbd168 Sorry for the noise, but I had to respin the same commit to change mtdparts due to feedback.

this commit allows to make both a standalone and
a second-stage u-boot for nsa310b, but only
first-stage u-boot is added in the makefile and
available in the build system.

The reason is that while both work fine if installed to flash
or loaded with kwboot, I could not get stock u-boot nor
bodhi's u-boot to chainload any second stage u-boot
(I also tried with dockstar's uboot that works fine
on this device if loaded with kwboot).


nbd168:
Replaced the commit in my staging tree, thanks.

PR #735, opened 2017-01-24
kernel: kmod-crypto-iv dropped in Linux 4.8

Ansuel:


nbd168:
I already have a comparable change in my staging tree, as I'm working on Linux 4.9 support at the moment.

PR #736, opened 2017-01-25
mwlwifi: Fixes rewritten history hash and latest version

thagabe:
The author of the upstream mwlwifi edited the history of the previous commit.
This commit not only fixes the updated hash but also sends in the latest
commits he made to the code which are mainly testing.

Signed-off-by: Gabe Rodriguez <lifehacksback@gmail.com>


thagabe:
ran make package/mwlwifi/check FIXUP=1 and the hash stayed the same

jow-:
Pulled into my staging tree, thanks!

PR #737, opened 2017-01-25
tree : add tree new package

hbl0307106015:
Tree is a recursive directory listing command that produces a depth indented listing of files,
which is colorized ala dircolors if the LS_COLORS environment variable is set and output is to tty.

Tree has been ported and reported to work under the following operating systems:
Linux, FreeBSD, OS X, Solaris, HP/UX, Cygwin, HP Nonstop and OS/2.

root@lede:/# tree -L 1
.
├── bin
├── dev
├── etc
├── lib
├── mnt
├── overlay
├── proc
├── rom
├── root
├── sbin
├── sys
├── tmp
├── usr
├── var -> /tmp
└── www

15 directories, 0 files

For more details, please visit http://mama.indstate.edu/users/ice/tree/

### DIFF
--- a/package/utils/tree/Makefile
+++ b/package/utils/tree/Makefile
@@ -1,0 +1,42 @@
+#
+# Copyright (C) 2012 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=tree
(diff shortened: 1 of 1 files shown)


hnyman:
This should likely be in the community package tree instead of this LEDE core repo.

https://github.com/openwrt/packages

mkresin:
Yes, this should go into the (OpenWrt) community package tree which is used by LEDE as well.

PR #738, opened 2017-01-25
ar71xx: fix device title case for TP-Link TL-WR810N

markv:
The device title should match the same format as the other devices


jow-:
Staged into my tree - thank you!

jow-:
@markv - you need to Signoff your commit with your real name, otherwise I cannot pick it

markv:
Sorry, I didn't realize and I've already deleted the fork

PR #739, opened 2017-01-25
mkimage: allow empty input file

jclehner:
With this patch, `mkimage` can be used to create a header for a zero-length
image. This is useful for some Netgear routers, which expect a second
uImage header in the last 64 bytes of the `kernel` partition.

### DIFF
--- a/tools/mkimage/patches/220-allow-empty-input-file.patch
+++ b/tools/mkimage/patches/220-allow-empty-input-file.patch
@@ -1,0 +1,15 @@
+Index: u-boot-2014.10/tools/mkimage.c
+===================================================================
+--- u-boot-2014.10.orig/tools/mkimage.c	2017-01-23 14:55:13.455568945 +0100
++++ u-boot-2014.10/tools/mkimage.c	2017-01-23 17:34:46.496671478 +0100
+@@ -566,6 +566,10 @@
+ 		exit (EXIT_FAILURE);
+ 	}
+ 
++	if (!sbuf.st_size) {
++		return;
++	}
++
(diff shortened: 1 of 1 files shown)


jow-:
Hm, patch is not wrong per-se, just wonder if we can get around patching. Is it feasible to make an uimage out of a dummy file with known size, then truncate the resulting file?

jclehner:
The thing is this: when I added support for the EX2700 a year ago, I initially created the fake uImage header using `echo -e`. @blogic  then [suggested](https://patchwork.ozlabs.org/patch/583689/) that I add a static file instead [1].

While cleaning up the mt7620 image makefile a few days back (pull #726), @mkresin suggested that patching `mkimage` and creating the fake header on the fly would be a more elegant solution.

> Is it feasible to make an uimage out of a dummy file with known size, then truncate the resulting file?

The problem is this: the easy approach would be to create a small file using `dd if=/dev/zero ...`, and then create a uImage from that file. We can't truncate the resulting image though, since the bootloader verifies *all* checksums, including the image checksum, so we'd have to waste an additional 4 kilobytes, which I'd like to avoid on a device with only 4M of flash.

mkresin:
I don't see a way to prevent the patching. But I'm not that happy with the way it is done. I mean, disabling this legitimate check for all kind of input doesn't seam to me a good idea.

What do you think about introducing a new command line switch to disable the zero byte check in copy_file() or the call of copy_file() (if possible) at all?

jclehner:
I just noticed that `mkimage` has an undocumented `-s` switch (`skipcpy`) which does exactly what I need. Closing this PR.

PR #741, opened 2017-01-26
kmod-sched-cake & iproute2: support minimum packet length 'mpu' parameter

kdarbyshirebryant:
This patch gives cake the capability to "round up" the size of packets under a given minimum length in order to better match the characteristics of an upstream shaper. This is most helpful on "IPoE" links with low encapsulation overhead such as DOCSIS cable where small IP packets such as TCP ACKs can easily fall below the minimum ethernet frame size of 64 bytes in practice.

Without this capability, the 'upstream' shaper counts the padding added to these ethernet frames, while cake's shaper only counts the useful size of such packets. Unless a large enough margin has been allowed for by setting cake's rate some percentage below the 'upstream' shaper's rate, a number of small packets can result in cake's rate exceeding that of the 'upstream' shaper thus inducing bufferbloat.

See https://github.com/dtaht/sch_cake/pull/45 for the full gossip.


jow-:
Pulled into my staging tree: https://git.lede-project.org/f3cb82a

PR #744, opened 2017-01-27
ramips: split mt76 firmware from kmod package

jclehner:
This patch moves the mt76 firmware from `kmod-mt76` to `mt76-firmware`.
This is a metapackage, comprised of three subpackages for mt7603, mt7628
and mt7662.

By adding `mt76-firmware` as `DEFAULT_PACKAGE` for all mt76 targets, the
previous behaviour of including the firmware in all images is left
unchanged, unless explicity unselected in `DEVICE_PACKAGES`.

This is especially beneficial for devices with small flash chips, since
the mt76 firmware currently requires ~160K on squashfs after compression.

### DIFF
--- a/package/kernel/mt76/Makefile
+++ b/package/kernel/mt76/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mt76
-PKG_RELEASE=1
+PKG_RELEASE=2
 
 PKG_LICENSE:=GPLv2
 PKG_LICENSE_FILES:=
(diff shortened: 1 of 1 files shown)


nbd168:
First of all, I think adding a separate source package for the firmware does not make any sense as long as the firmware is coming from the same repo as the driver.

I also think the driver should be split instead of just the firmware. It could look like this:
kmod-mt76-core (common code shared by both drivers)
kmod-mt76x2 (mt76x2 driver + firmware)
kmod-mt7603 (mt7603 driver + firmware for mt7628 on ramips/mt76x8 subtargets or mt7603 for all other targets)
kmod-mt76 (metapackage selecting both kmod-mt76x2 and mt7603).

jclehner:
> First of all, I think adding a separate source package for the firmware does not make any sense as long as the firmware is coming from the same repo as the driver.

You're correct. I've now split the packages as suggested.

> kmod-mt7603 (mt7603 driver + firmware for mt7628 on ramips/mt76x8 subtargets or mt7603 for all other targets)

The code currently installs both mt7628 firmware *and* mt7603 firmware on mt76x8 subtargets. Is this the correct behaviour?

nbd168:
Looks good to me. I think you should also change the code to include only the mt7603 or the mt7628 firmware, depending on the target.

jclehner:
> I think you should also change the code to include only the mt7603 or the mt7628 firmware, depending on the target.

Done.

nbd168:
Thanks. I added a minor cleanup and pushed it to my staging tree.

PR #745, opened 2017-01-27
ramips: create mt7620 fakeroot image on the fly

jclehner:
This patch removes the static fakeroot uImage required by some
Netgear devices, using `mkimage -s` instead.


mkresin:
Would it be possible that you move ```Build/append-fakeroot``` to ```include/image-commands.mk``` and "fix" ```target/linux/ipq806x/image/Makefile``` as well. Have a look at ```Build/uImage``` in ```include/image-commands.mk``` for variables you can use.

The ```Build/append-file``` could be dropped afterwards.

jclehner:
Will do.

jclehner:
Done. Updated image code for `ipq806x`,`apm821xx` and `mt7620`. I built images with the old and new code, and they both yield the same result, but I only have an `mt7620` device to test this on.

mkresin:
Hey Joseph,

I've merged a modified version of your PR into my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Would you please review my changes (if possible runtime test) to make sure that i didn't broke anything!

I will try to find tester for at least one of the changed images in ```ipq806x``` and ```apm821xx```.

I had a look at ```ar71xx``` as well. The netgear images are already using the new build code. But it isn't a fakeroot what they are using. The whole Kernel is in that squashfs partition that has the uImage (filesystem) header.

mkresin:
I received the confirmation that everything is working fine with your patch + my changes on top for the apm821xx/WNDR4700 and ipq806x/R7800.

Only your confirmation for mt7620 is missing.

jclehner:
Confirmed working on `mt7620/EX2700`, sorry for the delay!

PR #746, opened 2017-01-27
ar71xx: small clean up in Kconfig.openwrt and subtargets kernel configs fix up

pepe2k:
5 short patches which fix up nand and mikrotik subtargets kernel configs, ATH79_NVRAM selection and code style in Kconfig.openwrt.

Also, drop help sections from Kconfig.openwrt. Only few boards have them, they don't seem to have any informational value and were created mostly by copy&paste.


blogic:
merged, Thanks !

PR #747, opened 2017-01-28
kirkwood: add Zyxel NSA325

bobafetthotmail:
Adding support for one of the bigger brothers of NSA310, 
with its own uboot, and adding its envs to uboot-envtools.
There is also a commit doing a cleanup for nsa310b mac address extraction that is a prerequisite for the nsa325 commit as nsa325 shares the same logic.

@mkresin  I switched mac address reading to **mtd_get_mac_ascii** helper as it seems the fw_env.config file is created way later than when network is set up, when I tested I still had that file included in the image through **/files** folder :/ 

### DIFF
--- a/target/linux/kirkwood/base-files/etc/board.d/02_network
+++ b/target/linux/kirkwood/base-files/etc/board.d/02_network
@@ -32,7 +32,7 @@ case "$board" in
 	;;
 "nsa310b")
 	ucidef_set_interface_lan "eth0" "dhcp"
-	ucidef_set_interface_macaddr "lan" $( fw_printenv ethaddr | awk -F"=" '{print $2}' )
+	ucidef_set_interface_macaddr "lan" $( mtd_get_mac_ascii uboot_env ethaddr )
 	;;
 *)
 	ucidef_set_interface_lan "eth0"
--- a/target/linux/kirkwood/base-files/etc/board.d/02_network
+++ b/target/linux/kirkwood/base-files/etc/board.d/02_network
@@ -30,7 +30,7 @@ case "$board" in
 	ucidef_add_switch "switch0" \
 		"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0" "6@eth1"
 	;;
-"nsa310b")
+"nsa310b"|"nsa325")
 	ucidef_set_interface_lan "eth0" "dhcp"
 	ucidef_set_interface_macaddr "lan" $( mtd_get_mac_ascii uboot_env ethaddr )
 	;;
--- a/target/linux/kirkwood/patches-4.4/194-nsa325.patch
+++ b/target/linux/kirkwood/patches-4.4/194-nsa325.patch
@@ -1,0 +1,257 @@
+--- a/arch/arm/boot/dts/Makefile
++++ b/arch/arm/boot/dts/Makefile
+@@ -205,6 +205,7 @@ dtb-$(CONFIG_MACH_KIRKWOOD) += \
+ 	kirkwood-nsa310a.dtb \
+ 	kirkwood-nsa310b.dtb \
+ 	kirkwood-nsa310s.dtb \
++	kirkwood-nsa325.dtb \
+ 	kirkwood-openblocks_a6.dtb \
+ 	kirkwood-openblocks_a7.dtb \
+ 	kirkwood-openrd-base.dtb \
+--- /dev/null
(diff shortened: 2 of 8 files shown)


mkresin:
Please split "uboot-kirkwood: add nsa325 and fix usb of nsa310 uboot". The changes to the nsa310 uboot should have to be an extra commit and you should explain what's the usb issue you are fixing.

target/linux/kirkwood/base-files/etc/board.d/02_network
```diff
@@ -30,9 +30,9 @@ case "$board" in
 	ucidef_add_switch "switch0" \
 		"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0" "6@eth1"
 	;;
-"nsa310b")
+"nsa310b"|"nsa325")
 	ucidef_set_interface_lan "eth0" "dhcp"
-	ucidef_set_interface_macaddr "lan" $( fw_printenv ethaddr | awk -F"=" '{print $2}' )
+	ucidef_set_interface_macaddr "lan" $( mtd_get_mac_ascii uboot_env ethaddr )
```
mkresin: This change has to go into an extra commit since it changes code related to the nsa310b as well. Maybe create a "kirkwood: cleanup nsa310b mac address extraction" commit before the commit adding support for the nsa325.

target/linux/kirkwood/patches-4.4/194-nsa325.patch
```diff
@@ -0,0 +1,257 @@
...
++			regulator-name = "USB Power";
++			regulator-min-microvolt = <5000000>;
++			regulator-max-microvolt = <5000000>;
++			regulator-always-on;
++			regulator-boot-on;
++			enable-active-high;
++			gpio = <&gpio0 21 GPIO_ACTIVE_HIGH>;
++		};
++
++/************
```
mkresin: Either enable it or drop it.

bobafetthotmail: Due to the fact that this is a dtb supposed to be upstreamed, I don't think it's good to remove the comments here. If people want to stop the watchdog from the kernel this is how it is done.

bobafetthotmail:
updated, also split the changes to nsa310b uboot in a separate PR as it's not related to NSA325 in any way. https://github.com/lede-project/source/pull/790

target/linux/kirkwood/patches-4.4/194-nsa325.patch
mkresin: Not sure if you got me wrong. It's not about that the DTS has comments. It's that you have dead code in your dts and we do not add dead code to LEDE.

Have a look at the dts files in LEDE and in the kernel. There shouldn't be any nodes which are commented out.

Btw. the comment style you are using is wrong. ```./scripts/checkpatch.pl```should warn about it.

nbd168:
I agree with @mkresin's comments about the dead code in DTS. Even after reading the description, it is completely unclear to me when/if you would want to uncomment that code.
I think it's better to remove that code for now and if you want to add it back, do so with a better explanation with clear instructions for users.

nbd168:
Merged to my staging tree with the suggested modification

bobafetthotmail:
thanks, I was very busy lately so I had not done that yet.

mkresin:
@bobafetthotmail According to the error in the bootlog posted in the [forum](https://forum.lede-project.org/t/nsa310-kernel-modules-not-loaded/1760):

```
[    3.014895] random: jshn: uninitialized urandom read (4 bytes read, 14 bits of entropy available)
[    3.090580] random: jshn: uninitialized urandom read (4 bytes read, 14 bits of entropy available)
/bin/board_detect: /etc/board.d/02_network: line 1: mtd_get_mac_ascii: not found[    3.123992] random: jshn: uninitialized urandom read (4 bytes read, 14 bits of entropy available)

[    3.158267] random: jshn: uninitialized urandom read (4 bytes read, 14 bits of entropy available)
[    3.186138] random: jshn: uninitialized urandom read (4 bytes read, 15 bits of entropy available)
```

an include of ```/lib/functions/system.sh``` is missing in ```/target/linux/kirkwood/base-files/etc/board.d/02_network```. Would you please fix this issue!

bobafetthotmail:
Oops. :( 
done https://github.com/lede-project/source/pull/872

PR #749, opened 2017-01-28
Bumpcake

dtaht:
This fixes the API breakage between mainline cake and lede cake. And lede cake, with itself.

Tested-by: dave taht <dave+github@taht.net>

### DIFF
--- a/package/kernel/kmod-sched-cake/Makefile
+++ b/package/kernel/kmod-sched-cake/Makefile
@@ -12,10 +12,10 @@ PKG_NAME:=sched-cake
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/kdarbyshirebryant/sch_cake.git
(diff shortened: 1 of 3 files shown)


diizzyy:
So it's broken in the 17.01 right now or just in master?
Also, please write in the commits what version you're bumping to. Make it much easier to backtrack.

dtaht:
So I take it this was not pulled as I thought it was, but the new sqm-scripts were?

@kdarbyshirebryant @tohojo ?

kdarbyshirebryant:
It has not yet been committed to LEDE master, no.  Patience.  Ideally this would have been done in lock step with the packages repo (where sqm-scripts resides) too, but that's even more fun to coordinate.  I didn't see jow in irc much yesterday...probably because I was mostly absent from it.  I'll ask him ever so nicely....

dtaht:
My issue was that I was seeing weirdness with the default cake install in head. I have plots.

jow-:
Merged into master.

PR #750, opened 2017-01-29
toolchain/binutils: Switch to 2.27 as default version

diizzyy:
Use 2.27 as default which is the current relese.
Makes it easier to get support and submit patches upstream if needed.

Tested on ar71xx, mt7621, kirkwood (arm)


Ansuel:
i use it for every build and i don't have any problem at all... good for me

nbd168:

PR #751, opened 2017-01-29
uboot-kirkwood: remove obsolete patches

bobafetthotmail:
all patches for CONFIG_SYS_GENERIC_BOARD are obsolete for
uboot 2016 sources.

Run-tested with the uboot of goflexnet, also the md5sum of
all other uboots is the same with or without these patches.


nbd168:

PR #752, opened 2017-01-29
uboot-kirkwood: fix goflexhome/net bootcommand

bobafetthotmail:
Goflexhome/net use uImage, and to boot an uImage the u-boot
must use bootm command, not bootz.

Fixes the "i cannot boot LEDE with this u-boot" issue that I
found out myself with my goflexnet.

@blogic Can this patch be sent also in 17.01 branch?

### DIFF
--- a/package/boot/uboot-kirkwood/patches/150-goflexhome.patch
+++ b/package/boot/uboot-kirkwood/patches/150-goflexhome.patch
@@ -10,7 +10,7 @@
 -	"bootm 0x800000"
 +	"ubi part ubi; " \
 +	"ubi read 0x800000 kernel; " \
-+	"bootz 0x800000"
++	"bootm 0x800000"
  
  #define CONFIG_MTDPARTS \
 -	"mtdparts=orion_nand:1m(uboot),6M(uImage),-(root)\0"


bobafetthotmail:
@stintel can this patch be sent also in 17.01?

mkresin:
Please improve your commit message. You know how it should look like. "bootz -> bootm" isn't really descriptive.

You are changing more than just switching from zImage to uImage. You are touching the mtdparts definition and your are switching from loading kernel from nand to loading kernel from an ubi volume. As long as it isn't mentioned in the commit it is consider as mistake.

bobafetthotmail:
Updated commit description.

> You are changing more than just switching from zImage to uImage. You are touching the mtdparts definition and your are switching from loading kernel from nand to loading kernel from an ubi volume. As long as it isn't mentioned in the commit it is consider as mistake.

No, I'm just modifying a patch that does that and github gets confused, see the file in master https://github.com/lede-project/source/blob/master/package/boot/uboot-kirkwood/patches/150-goflexhome.patch

EDIT: it seems that after my update github isn't confused anymore and shows only the line I changed.

nbd168:

stintel:
@bobafetthotmail please don't mention random people. I don't know anything about uboot.

PR #754, opened 2017-01-29
rt5350: added pcm interface in .dtsi

fid0did0:
Added the missing audio pcm interface in the .dtsi file for the
rt5350 device. The update has been verified from the data get
from the datasheet and is very similar to the mt7620a.dtsi


blogic:
merged, Thanks !

PR #755, opened 2017-01-30
iperf3: Update to 3.1.6

diizzyy:
Update to 3.1.6

### DIFF
--- a/package/network/utils/iperf3/Makefile
+++ b/package/network/utils/iperf3/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=iperf
-PKG_VERSION:=3.1.4
+PKG_VERSION:=3.1.6
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


diizzyy:
Preferred way instead of using git which is submitted here: http://patchwork.ozlabs.org/patch/715468/

mkresin:
Why do you create a PR for a patch that is already in patchwork? Is something wrong with the patch in patchwork?

> Preferred way instead of using git 

Preferred by whom?

diizzyy:
This has been discussed several times, release tarballs > SCM pulls unless it makes sense.
That patch switches from this behavior to downloading it off git repo instead which just adds overhead in this case.

mkresin:
Ah, I got your comment wrong. What I read "Preferred way instead of using git" == github PR prefered over git send-mail.

Anyway, wouldn't it make more sense to reply to the patch on the mailing list, so that author gets aware of the issue with his patch? I mean, that is one of the major disadvantages of github. It's really hard for people to learn from the mistakes made by others.

diizzyy:
Sure, I submitted it late last night and haven't got around to it yet. I'll do it ASAP.

bladeoner:
@diizzyy please update this commit to v3.1.6 since it has been released.

Thanks in advance.

nbd168:

PR #756, opened 2017-01-30
mvebu: add esata led definition

Ansuel:
Add ESATA led option for WRT linksys router

(with the help of Ralph Sennhauser <ralph.sennhauser@gmail.com>)


mkresin:
Either something is missing or your patch is wrong. There is no ```ucidef_set_led_disk``` in LEDE.

Btw, you need to provide your full name in the signed off by.

Ansuel:
@mkresin now should be good... you were right

Ansuel:
Working and tested on my router.
hard disk connected with esata trigger the led when some activity to the disk are done.
And with this, you can control the esata led that was disabled.

Ansuel:
Updated with new led.d

thagabe:
I have tested this as well, Will  build and test the new changes but it looks good and works well too.

thagabe:
https://github.com/lede-project/source/commit/726e45dfe956ba6e0814919f2943a6a9d2f529d3 
Someone might have beat you to the punch, but your naming is a bit more sane (USB3).

Ansuel:
@thagabe what you mean ? 
(anyway they call it usb1 cause on the router they are named usb1 and usb2 , they are both usb 3.0 if i remember correctly...)

thagabe:
USB2 is the only USB3, USB1 is USB2 because it is the combo port USB 2.0/esata

Ansuel:
Didn't know and notice that ahahah 
(updated the commit, i'm learning git... sorry for the mess HAHAAH)

thagabe:
1) Fret not, I am new to (project) programming in general haha

2) I think this is ready for commit... but can you squash all the commits into a single commit? I think @jow- likes it as a single commit. 

PS isn't usbport trigger the preferred way as it is upstream kernel?

Ansuel:
I squashed all the commits... (for now usbport don't work on my router, justtry to set leds with that and you will see that they don't work... this commit anyway is for another thing, it works with esata connected devices)

5 days with this pull i think it's enough for testing  

Ralph Sennhauser  is seragh....

thagabe:
Bad news

mkresin:
At first I was the opinion that all eSATA LED names beside the one of the mamba were wrong. But they are correct. The power and esata LEDs are connected to GPIOs on all boards and not to the pca9635 like all other LEDs.

But I'm undecided about two points.

1. If you have a look at the [device tree source files in the linux kernel]( http://lxr.free-electrons.com/source/arch/arm/boot/dts/kirkwood-topkick.dts?v=4.9#L132), the disk-activity trigger is set as default trigger. Maybe we should do it the same way. But that means you need to send your change for the [caiman](http://lxr.free-electrons.com/source/arch/arm/boot/dts/armada-385-linksys-caiman.dts), [cobra](http://lxr.free-electrons.com/source/arch/arm/boot/dts/armada-385-linksys-cobra.dts), [mamba](http://lxr.free-electrons.com/source/arch/arm/boot/dts/armada-xp-linksys-mamba.dts) to the kernel people for [upstream inclusion](https://www.youtube.com/watch?v=LLBrBBImJt4) first and backport that patch after it is accepted. The dts files for rango and and shelby are in ```target/linux/mvebu/files/arch/arm/boot/dts``` and are not yet submitted to the kernel.

2. I'm not sure whether this PR targets the 17.01 branch. 17.01 is already in release candidate phase, which means only bugfixes. But this is more an enhancement/new feature. The next kernel version in LEDE will be 4.9. So it might be better to wait till 4.9 support is added instead of backporting the disk-activity trigger.

Any opinions @Ansuel @thagabe @jow- ?

Ansuel:
About the point 2, change kernel version would only exclude the backport patch, we need that changes anyway cause for now the esata lad have no definition and stay disabled... I think it would be better to include this anyway, it's just a waste of time waiting the kernel update...

mkresin:
> About the point 2, change kernel version would only exclude the backport patch ... I think it would be better to include this anyway, it's just a waste of time waiting the kernel update.

Well, work is already underway to update. the kernel to 4.9 in master. And yes what I'm concerned about  the most is the backport port of the disk-activity trigger. Since it isn't a simple backport but rather than a replacement/rename of the ide-disk trigger.

```patch
-	ledtrig_ide_activity();
+	ledtrig_disk_activity();
```

Sicne you have added the patch to target/linux/generic/ all targets are affected by the change. I'm in doubt that you have tested all targets to make sure that it doesn't break anything.

>  we need that changes anyway cause for now the esata lad have no definition and stay disabled.

I'm well aware that the eSATA LED doesn't work. It might be annoying, but I don't consider it a bug or a release critical fix. It's just a missing feature.

After thinking more about it, the feature isn't worse to break boards in a *release candidate*. Lets wait for 4.9 instead of doing things twice.

thagabe:
I second @mkresin as it does seem like something that should be fixed upstream then we can downstream it, hopefully by the time of 17.01.1.

thagabe:
@Ansuel this doesn't work in Rango. We should just follow suit with @mkresin and wait for the upstream

thagabe:
When package hd-idle turns off the hdd the disklight will not come back up when the hdd is accesses but it will start showing disk activity then completely stop when no activity is happening.

Ansuel:
As we moved to 4.9 kernel we don't need the backport patch and we can include this with no risk of bugs with others device that could use ide_disk trigger

mkresin:
As I already told you in #lede-dev,  please send your dts changes upstream first and backport them to LEDe afterwards.

Ansuel:
Will do it now, just one question? What is the utility of doing this?

mkresin:
Superseded by #915.

PR #760, opened 2017-01-30
base-files: Wifi detect deprecation

DPini:
Since commit 5f8f8a366136a07df661e31decce2458357c167a wifi detect does not work anymore, and instead it exists wifi config to configure unconfigured wireless devices.
The help/usage of wifi was not modified and was still saying that you can use wifi detect, without saying anything about wifi config.
I have changed wifi command usage and also added a function called when executing wifi detect that shows a Warning so users know that it's deprecated and exits with an error.
It can be done that when running wifi detect, wifi config is ran, but it makes no sense, as the commit 5f8f8a366136a07df661e31decce2458357c167a explains: there are different things.

### DIFF
--- a/package/base-files/files/sbin/wifi
+++ b/package/base-files/files/sbin/wifi
@@ -6,8 +6,8 @@
 
 usage() {
 	cat <<EOF
-Usage: $0 [down|detect|reload|status]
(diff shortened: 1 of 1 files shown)


blogic:
merged, Thanks !

PR #762, opened 2017-02-01
ar71xx: base-files: cleanups and fixes in lib/ar71xx.sh

pepe2k:
The following series includes general cleanups and fixes in **lib/ar71xx.sh** script and 3 related **mach-\*.c** files (fix in machine names).

Last patch (**ar71xx: base-files: improve code readability in lib/ar71xx.sh**) contains rather a heavy changes thus marked as **RFC**.

Summary change in **lib/ar71xx.sh** size: +307 bytes


pepe2k:
Closing for now, I will rework it.

PR #763, opened 2017-02-01
ath10k-firmware: update qca9984 firmware and board data

dissent1:
Fixes firmware crash in rare cases and a bug
ath10k_pci 0001:01:00.0: received unexpected tx_fetch_ind event: in push mode
for those who kept experiencing it after previous firmware update. Seems that previous update has fixed this bug not for everyone.

https://forum.lede-project.org/t/netgear-r7800-exploration-ipq8065-qca9984/285/21
https://forum.lede-project.org/t/netgear-r7800-exploration-ipq8065-qca9984/285/38

It also fixes 80+80 vht mode (according to dd-wrt).

I have been running this firmware for a week without issues in my use cases.

### DIFF
--- a/package/firmware/ath10k-firmware/Makefile
+++ b/package/firmware/ath10k-firmware/Makefile
@@ -177,18 +177,18 @@ define Download/qca99x0-board
 endef
 $(eval $(call Download,qca99x0-board))
 
-QCA9984_BOARD_REV:=deb1832c56c706d0f6cb539113e09f0daaa52b5f
+QCA9984_BOARD_REV:=719c0127e52bd70559e71b85ab0331790e1bf66c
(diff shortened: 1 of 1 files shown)


nbd168:
Merged, thanks.

dissent1:
Thanks!
@nbd168 @mkresin 
Btw, there is a possibility provide calibration data through a device tree https://www.kernel.org/doc/Documentation/devicetree/bindings/net/wireless/qcom,ath10k.txt

Chromium has made a parser for this, that fills appropriate dt property with caldata blob taken from cbmem. https://groups.google.com/a/chromium.org/forum/m/#!msg/chromium-os-checkins/4Chpi6C1Sf4/6AodozUPtLoJ

Maybe there is a sense to utilize similar approach adjusted to mtd partition?

PR #765, opened 2017-02-01
ccache, samba36: fix samba.org download addresses to use https

hnyman:
samba.org has started to enforce https and currently plain http downloads with curl/wget fail, although the same links from browser redirect to https.

So, convert samba.org download links to use https.
Modernise download links at the same time.

Also convert samba.org links in URL fields to have https, although that is merely documentation isssue.

I already fixed the same thing on the packages repo a few days ago with https://github.com/openwrt/packages/commit/eb0f2f0021e471f8b19a77a80080ca60d9ce5ed0 , but then I noticed a few minutes ago from the failing Openwrt buildbot that this samba.org change also affects the ccache tool & samba36 package in the main repo.

### DIFF
--- a/config/Config-devel.in
+++ b/config/Config-devel.in
@@ -67,7 +67,7 @@ menuconfig DEVEL
 		bool "Use ccache" if DEVEL
 		default n
 		help
-		  Compiler cache; see http://ccache.samba.org/.
+		  Compiler cache; see https://ccache.samba.org/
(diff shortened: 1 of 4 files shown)


hnyman:
merged as https://github.com/lede-project/source/commit/eaf3fef94636e81772776754a45f1b8ba1886793

PR #766, opened 2017-02-02
dropbear: enable SHA256 HMACs

josephcsible:
The only HMACs currently available use MD5 and SHA1, both of which have known
weaknesses. We already compile in the SHA256 code since we use Curve25519
by default, so there's no significant size penalty to enabling this.

Signed-off-by: Joseph C. Sible <josephcsible@users.noreply.github.com>


nbd168:

PR #768, opened 2017-02-02
build: Pass -iremap gcc option as a single argument

wentasah:
Passing -iremap argument separately causes problems with projects that
use scons and its ParseFlags function. Consider this SConscript
example:

    env = Environment()
    d = env.ParseFlags("-iremap one:two")

ParseFlags will interpret one:two as a file name and the returned dict
d will contain only "-iremap". When the -iremap is passed to the
compiler without an argument, compilation obviously fails.

Signed-off-by: Michal Sojka <sojkam1@fel.cvut.cz>


nbd168:

PR #769, opened 2017-02-02
Added key_direction to append_params.

bdkoepke:
key_direction shows up as an openvpn option in the user-interface but does not end up in the /var/etc/openvpn*.conf file. Adding it to the list here fixed the issue for me.

Signed-off-by: Brandon Koepke <bdkoepke@fastmail.com>

### DIFF
--- a/package/network/services/openvpn/files/openvpn.init
+++ b/package/network/services/openvpn/files/openvpn.init
@@ -111,8 +111,8 @@ start_instance() {
 		connect_retry connect_timeout connect_retry_max crl_verify dev dev_node dev_type dh \
 		ecdh_curve echo engine explicit_exit_notify fragment group hand_window hash_size http_proxy \
 		http_proxy_option http_proxy_timeout ifconfig ifconfig_pool ifconfig_pool_persist ifconfig_push \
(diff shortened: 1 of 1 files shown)


dedeckeh:
Your PR is missing a S-o-B; also the commit subject needs to start with openvpn: adding ...
Have a look at https://github.com/lede-project/source/pull/592/commits/c7cfdff4fa40f53511b8c8a4bd2e60aa980dc6fd how to create a correct PR

bdkoepke:
Is this correct now? Apologies I wasn't familiar with the PR format used here. (Had to google S-o-B...)

mkresin:
Nope, the commit still has the no S-o-B and the prefix of the commit title is still missing. What you have updated is the PR message. But this message is only an envelope and will not be merged.

Please use ``git commit --amend`` to change the last commit. Use ```git push --force``` afterwards to overwrite your github repository. This PR gets updated automatically afterwards.

pepe2k:
@bdkoepke you should include SoB line inside commit message, not in PR text.

bdkoepke:
I believe this is fixed now. This is what happens when you don't get enough sleep. Let me know if this will work at this point.

blogic:
merged, Thanks !

PR #771, opened 2017-02-02
ar71xx: add support TP-LINK 886NV5

am2dk:
-CPU:TP9343
-Flash:8MB
-RAM:64MB
-10/100Mbps LAN
-450Mbps wifi 802.11

-The Device must be installed undead u-boot/breed,and firmware will be installed in undead u-boot/breed.  -http://www.right.com.cn/forum/thread-161906-1-1.html
-http://breed.hackpascal.net
-Signed-off-by: Xiang Wang <347006777@qq.com>


mkresin:
Okay, lets make it perfect. Please run ```git reset --hard HEAD~2``` to get rid of the revert commits. Run ```git commit --amend``` to fix the commit title and commit message:

- the prefix has to be "ar71xx:" take care of the ":" that is currently missing
- you need to provide a full name in the Signed-off-by. "am2dk" looks like a nick name.
- add angle brackets (<>)  around your mail address: ```Signed-off-by: Firstname Lastname <347006777@qq.com>```
- add a blank line before the Signed-off-by
- please add a few informations to the commit message on how LEDE can be installed on the TP-LINK 886NV5. have a look at https://github.com/lede-project/source/commit/fd62fa752bbff9ae5277098152e7960a14d241e1 for a good example

Finally, run ```git push --force``` to overwrite your github repository/branch. The PR will afterwards update automatically and should be ready to merge.

am2dk:
I have edit it just now.

mkresin:
- add a blank line before the Signed-off-by
- please add a few informations to the commit message on how LEDE can be installed on the TP-LINK 886NV5. have a look at https://github.com/lede-project/source/commit/fd62fa752bbff9ae5277098152e7960a14d241e1 for a good example

am2dk:
I deed the blank line,but it must be installed with u-boot.

pepe2k:
@am2dk, few things still need explanation and/or changes.

> it must be installed with undead u-boot/breed.

1. Could you please provide more details about this? Does it mean that the device you are adding support for is somehow locked down by the vendor (I mean, there is no way to change firmware in usual way, over GUI, or under bootloader with e.g. TFTP)?

2. What kind of steps are required to flash this device with LEDE image? That kind of information should be provided in commit message, so that the users know exactly how to install the image, especially if it's not the usual way. Does the user need to open the device and swap the bootloader with external FLASH programmer or is there some easier way to achieve this?

3. Could you also explain the purpose of the `tplink-wmac.c ` file, e.g. why it's needed for and what it does? As far as I can see it's not even compiled and used by your code. If it's there by accident and not needed at all, you should just drop it as it's not related with your submission.

4. Please fix the device name in the commit title, it should be `TL-WR886N/ND v5`, not just `886NV5`.

5. Please, change CONFIG_ATH79_MACH_TL_WR886N to CONFIG_ATH79_MACH_TL_WR886N_V5 (add _V5 suffix).

6. Please, remove also support for this device in `nand` and `mikrotik` subtargets kernel config (take a look how it was done for example here: 9c475eca3e1b004a7fd604dba1481ed2c09ffb35). Simply, add `# CONFIG_ATH79_MACH_TL_WR886N_v5 is not set` line in below files (just remember about alphabetical order!) is enough:

  - https://github.com/lede-project/source/blob/master/target/linux/ar71xx/nand/config-default
  - https://github.com/lede-project/source/blob/master/target/linux/ar71xx/mikrotik/config-default

Cheers,
Piotr

am2dk:
1.This Device is usually be modified with 16MB Flsah and 64MB RAM.So the factory machine is only 2MB flash and 8MB RAM.It can not be install LEDE or OpenWrt.
2. With the undead u-boot or breed,we can poke the hole behind the router befor poweron.Since the LED sparkles 3 times.The 192.168.1.1 can be a web for flashing.
3.tplink-wmac.c and tplink-wmac.h is needs for the m25p80.
4.Agree
5.Agree
6.The device do not suppt nand and mikrotik.

pepe2k:
> 1.This Device is usually be modified with 16MB Flsah and 64MB RAM.So the factory machine is only 2MB flash and 8MB RAM.It can not be install LEDE or OpenWrt.

@am2dk I'm afraid this device cannot be officially supported in LEDE. The minimum (and recommended) FLASH/RAM size is currently **4/32 MB**. As the device doesn't meet these requirements, it doesn't make sense to keep support for it in LEDE. It would be misleading for users who own a factory version, without the FLASH/RAM mod. This means a NAK from me.

mkresin:
NAK from me as well. We don't add support for each possible hardware modification of a board. LEDE needs at least 4 MB of flash and wouldn't start with 8 MB of RAM.

Please do not submit your other patches, in case LEDE would only work on these boards after changing RAM and/or flash.

PR #772, opened 2017-02-03
dnsmasq: honor quietdhcp option for DHCPv6

arjendekorte:
Do not spam the syslog with lease info if quietdhcp option is given. This already works for DHCPv4, make it work in the same way for DHCPv6.

Signed-off-by: Arjen de Korte <build+lede@de-korte.org>


dedeckeh:
@arjendekorte The commit message is missing S-o-B

PR #776, opened 2017-02-03
dnsmasq: honor quietdhcp option for DHCPv6

bobafetthotmail:
Do not spam the syslog with lease info if quietdhcp option is given. This already works for DHCPv4, make it work in the same way for DHCPv6.


mkresin:
Closed in favour of #779 which keeps and mentions the authorship.

PR #777, opened 2017-02-03
ipq806x: remove trailing whitespace from 10_fix_wifi_mac

hnyman:
There is a trailing TAB char after backslash in 10_fix_wifi_mac 
that prevents the assignment of the correct MACs for wifi devices.

The error has been caused by 71a39b869019c97f48bc85d9a8fabceead217d07

Removing that TAB makes the script to assign correct MACs again.

Reference to bug report FS#451
https://bugs.lede-project.org/index.php?do=details&task_id=451

@mkresin 
Please apply this also for the 17.01 branch


mkresin:
merged, thanks!

PR #782, opened 2017-02-04
kernel: bump to 4.4.47

anomeome:
Refresh patches for all supported targets
 - compile / test on ar71xx (c7v2)


anomeome:
@nbd, ha, are we good to move to 4.9 or do you want to keep 4.4 around

kdarbyshirebryant:
This really should be 2 commits - one to bump/refresh 4.4 and another to bump/refresh 4.9.
With both changes in one commit it makes it challenging if a revert is needed to fix a problem in one or the other.

Also it's unclear to me whether your 4.9 bump didn't have a refresh/doesn't need a refresh.

Borromini:
@anemome 4.4 will be kept for a while, took a while before 3.18 got the boot as well. Plus, 17.01 can still benefit from 4.4 bumps.

wongsyrone:
I agree with @kdarbyshirebryant , it should be split into two commits for 4.4 and 4.9 to make things clear.

anomeome:
Yep, my comment was made before I saw that it was a WIP, with limited target support, and was only meant as regards trunk not 17.01.

Borromini:
I have sent in a separate patch just to bump 4.4 for 17.01; it's in Patchwork.

anomeome:
@Borromini , this would need a refresh / regenerate after the commit that solicited my first comment, which you have apparently already done. Why don't you submit to ML for trunk in addition to the 17.01 submission; seems to be the preferred method anyways.

PR #783, opened 2017-02-04
dnsmasq: rfc6303 address exclusion

kdarbyshirebryant:
The essence of this has been lurking in my local tree for a long time as a simple modification to /etc/config/dhcp.  My coding and git knowledge has come along a little and I now offer this as a more flexible way of doing the same thing.  Fundamentally the idea is to make LEDE a good netizen by default but allow configuration flexibility should it be needed.

RFC6303 defines a number of arpa zones that should not be forwarded to
the wider internet and thus burden root DNS servers.  dnsmasq covers the
IPv4 zones itself by use of the '--localservice' option.  At the present
time it does not exclude IPv6 zones.

Similarly some top level domains should never be forwarded. e.g. example.

This change adds a list of domains that will be blocked when
'localservice' is used and augments that which is already blocked by
dnsmasq's notion of 'local service'.

To make this configurable rather than hard coded in dnsmasq's init
script, the new option of "list rfc6303 '/block_domain/'" in
/etc/config/dhcp, along with a default list that matches the RFC6303
recommendation.

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77test1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq/test-releases
@@ -152,6 +152,8 @@ define Package/dnsmasq/install
 	$(INSTALL_BIN) ./files/dnsmasq.init $(1)/etc/init.d/dnsmasq
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/ntp
 	$(INSTALL_DATA) ./files/dnsmasqsec.hotplug $(1)/etc/hotplug.d/ntp/25-dnsmasqsec
+	$(INSTALL_DIR) $(1)/usr/share/dnsmasq
+	$(INSTALL_DATA) ./files/rfc6303.conf $(1)/usr/share/dnsmasq
 endef
 
 Package/dnsmasq-dhcpv6/install = $(Package/dnsmasq/install)
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -17,6 +17,7 @@ BASETIMESTAMPFILE="/etc/dnsmasq.time"
 TRUSTANCHORSFILE="/usr/share/dnsmasq/trust-anchors.conf"
 TIMEVALIDFILE="/var/state/dnsmasqsec"
 BASEDHCPSTAMPFILE="/var/run/dnsmasq"
+RFC6303FILE="/usr/share/dnsmasq/rfc6303.conf"
 
 DNSMASQ_DHCP_VER=4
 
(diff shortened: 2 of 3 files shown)


dtaht:
dnsmasq also had some syntactic sugar for mapping real reverse ipv6 lookups to the right places (and also synthesizing results for reverse), but I'm using this patch as is so far and like it.

kdarbyshirebryant:
You're thinking of "--rev-server=<ip-address>/<prefix-len>,<ipaddr>[#<port>][@<source-ip>|<interface>[#<port>]]
    This is functionally the same as --server, but provides some syntactic sugar to make specifying address-to-name queries easier. For example --rev-server=1.2.3.0/24,192.168.0.1 is exactly equivalent to --server=/3.2.1.in-addr.arpa/192.168.0.1 "

Unfortunately with rev-server the ip address of the server to send the query is non-optional unlike --server where /168.192/ on its own means 'local' to dnsmasq.  Also the (admittedly ugly) reverse representation is exactly what you'll find in the RFC.

Also I wanted the rfc6303 list to be capable of holding both reverse and forward domains to treat as local (ie. not forward)

EricLuehrsen:
Technical Statement: Seems like the right thing to do, and it is executed fairly well as best as I can tell. There are some special addresses that fall on the border like CGN 100.64.0.0. If your ISP serves with CGN, you may want the names of the devices (dns-a.mytown.mystate.myisp.com). But CGN is either not used or is transparent strictly compliant with the respective RFC(s).

Process Statement: This might be one of those times where LEDE/OpenWrt team should put more encouragement or pressure as appropriate on upstream changes. dnsmasq should really handle this kind of thing on its own. Scripts should be used to join two programs or massage the UI. When we start scripting content that should be in a binary, we create a maintenance problem.

_deviating from topic: this is one area where a full recursive server like Unbound or Bind would do well as the primary entry DNS. Then down stream routers use dnsmasq. My work with Unbound+odhpcd appears ROM neutral with dnsmasq-dhcp6 (alone). Unbound is as easy as dnsmasq to configure (vs Bind, yikes), but is also comprehensive in its option set. Unbound does use a bit of RAM but it can be tuned and not much to worry about. Anyway, food for thought._

kdarbyshirebryant:
On 05/02/17 21:00, Eric Luehrsen wrote:
> Technical Statement: Seems like the right thing to do, and it is
> executed fairly well as best as I can tell. There are some special
> addresses that fall on the border like CGN 100.64.0.0. If your ISP
> serves with CGN, you may want the names of the devices
> (dns-a.mytown.mystate.myisp.com). But CGN is either not used or is
> transparent strictly compliant with the respective RFC(s).

It was those border cases that made me think it should be implemented as 
some sort of customisable block list rather than coding into the startup 
script.

>
> Process Statement: This might be one of those times where LEDE/OpenWrt
> team should put more encouragement or pressure as appropriate on
> upstream changes. dnsmasq should really handle this kind of thing on its
> own. Scripts should be used to join two programs or massage the UI. When
> we start scripting content that should be in a binary, we create a
> maintenance problem.

I agree with you.  In fact some the the IPv4 address blocking was 
updated as part of a patch I offered upstream.  Unfortunately there 
isn't the same infrastructure existing in dnsmasq for IPv6 and my 
c/understanding of dnsmasq internals is not up to the task of adding it.

However, we are on the cusp of a release and we have the opportunity to 
do the 'right thing' by the Internet, if not quite the right thing from 
a coding/upstreaming point of view.

>
> /deviating from topic: this is one area where a full recursive server
> like Unbound or Bind would do well as the primary entry DNS. Then down
> stream routers use dnsmasq. My work with Unbound+odhpcd appears ROM
> neutral with dnsmasq-dhcp6 (alone). Unbound is as easy as dnsmasq to
> configure (vs Bind, yikes), but is also comprehensive in its option set.
> Unbound does use a bit of RAM but it can be tuned and not much to worry
> about. Anyway, food for thought./

Coincidentally I started looking at unbound this very morning to see how 
it handled the above 'domain block' issue, because it *is* a fundamental 
issue no matter which resolver solution is adopted.  I was not put off 
and will try it when I 'get a moment'.  I have always favoured using 
dnsmasq as I like the 'one box does integrated dns, dhcpv4,v6 & RA'.  I 
perhaps need to break that personal bias/preference.

kdarbyshirebryant:
This has been reworked as a conditional include. No longer has so much static data sitting in uci.

PR #788, opened 2017-02-05
tools/e2fsprogs: update to 1.43.4

bladeoner:
e2fsprogs: update to 1.43.4

Tested with ar71xx.

Release notes:
Fix e2fsck to require that the system.data extended attribute is always present even for files smaller than 60 bytes, to be consistent with the kernel.

Fix a bug which was causing mke2fs -d and fuse2fs to improperly handle Posix ACL's.

Fix a bug which could cause mke2fs -d to fail if there is an zero-length file in source directory tree.

Clarify the default for the "proceed?" question printed by mke2fs and tune2fs. (Addresses Debian Bug: #852727)

Debugfs will now display project quota information.

Debugfs's do_set_xattr now uses C strings to parse extended attribute values, and will print the extended attributes using either C strings or hex bytes when using debugfs's do_get_xattr command. It will now allow developers to see the contents of the system.data extended attribute.

Fix a bug in mke2fs where I/O errors weren't getting properly reported to the user.

Fix multiblock writes in the unaligned Direct I/O fallback code. (Which is rarely needed on Linux systems, but which is required on Freebsd systems.)

Make sure the default mke2fs.conf file includes the uninit_bg feature flag. (Which was accidentally dropped in v1.43).

Fix a bug in resize2fs when operating on very large file systems which have a block size different from the VM page size.

If the reported device size is absurdly large, mke2fs will now report an error instead looping forever or crashing.

Fix various Coverity warnings and other memory leaks in fuse2fs and extended attribute manipulation functions.

Replace a test file but which had a "non-commercial use-only" copyright permission file with a newer version from the Cyrus imapd package which now has a 4-clause BSD license, which was making some lawyers nervous, even though the test file in question was only used in lib/et's regression testing and was never included in any compiled binary. (Addresses Debian Bug: #840733)

Stop installing mkfs.ext4dev and fsck.ext4dev

Update Chinese, Czech, Polish, Serbian, and Spanish translations and add the Finnish and Malay translation. (Addresses-Debian-Bug: #774379)

Update various man pages (Addresses-Debian-Bug: #852726)


diizzyy:
Are any of these updates tested?

bladeoner:
Yes I tested it with ar71xx.

nbd168:
Doesn't work for me, SHA256 hash doesn't match on download. Maybe the release tarball was changed in the mean time?

bladeoner:
@nbd168 Yesterday I experienced the opposite I checked the sha256 and it was different then in my commit. Tonight I will to check again, commit the other sha256 and let you know.

Could you please check my Cmake commit @hnyman confirmed it works but yesterday I had the same that the sha256 was different.

Thanks in advance.

diizzyy:
https://www.freshports.org/sysutils/e2fsprogs/
This is probably the reason why it fails.

diizzyy:
We should probably switch to this?
https://www.freshports.org/sysutils/e2fsprogs/

bladeoner:
@diizzyy Yesterday I tried to switch to .tar.xz but then the download didn't start.

I was also busy upgrading Glibc to version 2.25 but I was stuck with the same sha256 issue.
I also reinstalled my test environment to be sure that it didn't cause a problem

diizzyy:
@bladeoner 
http://paste.debian.net/913023/
Works fine, tested on LEDE trunk

diizzyy:
You need to fix your pull reqest

bladeoner:
@diizzyy I'm trying to merge the pull request at the moment.

bladeoner:
@nbd168 the SHA256 has been modified, can you please test again?

Thanks in advance.

bladeoner:
Something went wrong with this pull request, I opened a new one: 
tools/e2fsprogs: update to 1.43.4 #791

PR #790, opened 2017-02-06
uboot-kirkwood: fix usb of nsa310b u-boot

bobafetthotmail:
fixes issue "nsa 310b u-boot can initialize usb but cannot
use usb storage so it cannot load files from usb"

### DIFF
--- a/package/boot/uboot-kirkwood/patches/007-nsa310-uboot-generic.patch
+++ b/package/boot/uboot-kirkwood/patches/007-nsa310-uboot-generic.patch
@@ -1,3 +1,4 @@
+
 arm: kirkwood: add ZyXEL NSA310 device
 
 This patch add ZyXEL NSA310 1-Bay Media Server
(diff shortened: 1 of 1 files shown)


nbd168:
Why remove ext2 support here?

bobafetthotmail:
CONFIG_CMD_EXT2 is already set in include/configs/nsa310.h (found in the same patch), among other filesystems. see https://github.com/lede-project/source/blob/master/package/boot/uboot-kirkwood/patches/007-nsa310-uboot-generic.patch#L694
which made it redundant here.

I can add this to commit description if you think it is better.

mkresin:
Yes add it please to the commit message. Every change not mentioned in the commit message is assumed to be an error. And ext2 isn't in any point related to ext2.

Maybe it is better to either omit this change or to move it to an extra commit. One change (topic) per commit.

bobafetthotmail:
Ok, updated to change only the USB configuration option.

blogic:
merged, Thanks !

PR #793, opened 2017-02-06
utils/e2fsprogs: Update to 1.43.4

diizzyy:
* Update to 1.43.4
* Use xz tarball which saves about 2Mbyte in size

Changelog: http://e2fsprogs.sourceforge.net/e2fsprogs-release.html#1.43.4


diizzyy:
Briefly runtime tested on kirkwood

nbd168:
Added to my staging tree with the same modification as tools/e2fsprogs

PR #794, opened 2017-02-06
dnsmasq: bump to dnsmasq v2.77test2

kdarbyshirebryant:
New test release includes 2 LEDE patches that are upstream and may be
dropped, along with many spelling fixes.

Patches deleted:
010-localise-queries-apply-to-interface-names.patch
220-try-all-servers-on-fail.patch


zorun:
Looks good to me!  It's almost a no-op: test2 only has 3 new commits, 2 of which were already imported as patches, and the last one (spelling fixes) touching just a few bits of code to rename variables.

PR #795, opened 2017-02-07
uclibc++: patch bugfix erase() on derived __base_associative

scrpi:
This patch has also been sent upstream to uClibc++ maintainer.

Signed-off-by: Ben Kelly <ben@benjii.net>


nbd168:

PR #796, opened 2017-02-07
ppp: add pppoe-discovery to ppp-mod-pppoe package

feckert:

### DIFF
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -158,6 +158,17 @@ define Package/pppstats/description
 This package contains an utility to report PPP statistics.
 endef
 
+define Package/pppoe-discovery
+$(call Package/ppp/Default)
+  DEPENDS:=@(PACKAGE_ppp||PACKAGE_ppp-multilink) +ppp-mod-pppoe
+  TITLE:=Perform a pppoe-discovery process
+endef
+
+define Package/pppoe-discovery/description
+This tool performs the same discovery process as pppoe, but does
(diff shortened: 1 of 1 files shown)


jow-:
Whats the size of this extra executable?

nbd168:
Also, what's the reason for this change? The commit message is a 'little' vague

jow-:
Well I can see the use case (discovering PPPoE service announcements similar to how e.g. debian does it) but considering the fact that our current space situation already is severe and that ppp-mod-pppoe is a core package I tend to NAK this.

I have no issues with packaging this utility separately but the decision on whether to include it by default depends on the size it takes up...

Ansuel:
What are the benefits of having this?

feckert:
@nbd168 @Ansuel this tool is used to debug pppoe http://man.he.net/man8/pppoe-discovery
@jow- i would packaging it separately as you suggest. It think this is only selectable if ppp-mod-pppoe is selected?

jow-:
@feckert - no, packages will be selectable independently

feckert:
@jow-  Sorry for the misunderstanding. I would package pppoe-discovery like chat, pppdump or pppstats.

dedeckeh:
Pulled into my staging tree with a small grammatical change (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=a21e79487ee5fd670708acbd230da51386ac892f).
Including the pppoe-discovery package by default in the build makes no sense imho
Thx

PR #798, opened 2017-02-07
Ebtables: update to last commit

Ansuel:
Updated Ebtables and refreshed patch 


nbd168:
Please change the makefile to use PKG_SOURCE_DATE instead of PKG_VERSION  and drop PKG_SOURCE + PKG_SOURCE_SUBDIR

nbd168:
Also, please add PKG_MIRROR_HASH (e.g. using make package/ebtables/check FIXUP=1 after you've done the other rework).

Ansuel:
@nbd168 that command didn't give me hash output, anyway now the makefile should be good

nbd168:
that commit still doesn't contain PKG_MIRROR_HASH

Ansuel:
i don't know how to create it... need to put the hash of the download file? cause i think i don't have any mirror for a git repository

nbd168:
I just tested it, running `make package/ebtables/check FIXUP=1` adds the missing hash.
By the way, you also didn't refresh the patches, please run make package/ebtables/refresh as well.

Ansuel:
Ok done it... it was really easy

nbd168:
It's `PKG_MIRROR_HASH:=997a877da02d6e2141e6d31c5d4dd005737facecfdbea07308c0e1286db8591c`

Ansuel:
Yep, didn't know that if you add empty string and then do that command it autocomplete it

Ansuel:
It's good now

nbd168:
You forgot to remove PKG_SOURCE

Ansuel:
Done, sorry for all this mess

nbd168:
It seems that you still didn't refresh the patches

Ansuel:
@nbd168 looks like we don't need a refresh, they apply on my side

nbd168:
They apply with offsets (and fuzz), so they should be refreshed even though they apply

Ansuel:
I finally refreshed the patch and i notice a little typo in the makefile about ebtables-utils

nbd168:
Actually, that wasn't a typo. The code is fine the way it is, so please change it back.

Ansuel:
Really? then sorry can i ask an explaination?

Ansuel:
Done

nbd168:
DEPENDS += ebtables means append 'ebtables' to the existing DEPENDS variable (which is inherited from the ebtables package.
Technically, appending isn't necessary, since by depending on ebtables, it will inherit its dependencies as well.
Not using + is intentional, because that makes ebtables-utils visible only if ebtables is selected

nbd168:

PR #799, opened 2017-02-07
dnsmasq: workaround jail/procd/sysntp hotplug SIGHUP issue

kdarbyshirebryant:
Commit c914fa04 used ubus/procd signalling to issue a SIGHUP to dnsmasq
when ntp regarded time as valid.  This has highlighted an issue with
procd in a jailed environment in that procd signals the jailing 'helper'
process rather than the 'jailed' process.

The jailer dies, whilst the jailed (dnsmasq) process keeps running.
Procd in response to the jailer dieing respawns dnsmasq processes which
cannot startup properly because ports are already in use. This produces
log spam until procd gives up.  Whilst the original dnsmasq is still
running, it still hasn't received the SIGHUP letting it know time is
valid and hence should check DNSSEC timestamps which is a security
issue.

As a workaround, the ntp hotplug script now uses 'pidof dnsmasq' to
identify and SIGHUP dnsmasq processes.

Fixes FX#464

Tested on Archer C7 v2 in jailed & non jailed environments.

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77test1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


dtaht:
painfully... elegant.

PR #805, opened 2017-02-10
nvram: add nvram init for new created nvram partition

hbl0307106015:
Since the original nvram work based on header with NVRAM_MAGIC,
this command(nvram init) initialize the mtd partition that
new created and named 'nvram'.

Moreover, adding the 'nvram init' command can make the nvram support
on any other platform.

### DIFF
--- a/package/utils/nvram/src/cli.c
+++ b/package/utils/nvram/src/cli.c
@@ -27,13 +27,17 @@
 
 static nvram_handle_t * nvram_open_rdonly(void)
 {
-	const char *file = nvram_find_staging();
+	char *file = nvram_find_staging();
 
 	if( file == NULL )
 		file = nvram_find_mtd();
 
-	if( file != NULL )
-		return nvram_open(file, NVRAM_RO);
+	if( file != NULL ) {
+		nvram_handle_t *h = nvram_open(file, NVRAM_RO);
+		if( strcmp(file, NVRAM_STAGING) )
+			free(file);
+		return h;
+	}
 
 	return NULL;
 }
--- a/package/utils/nvram/src/cli.c
+++ b/package/utils/nvram/src/cli.c
@@ -151,15 +151,16 @@ int main( int argc, const char *argv[] )
 	int done = 0;
 	int i;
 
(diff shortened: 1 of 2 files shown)


nbd168:
If I remember correctly, on many devices the nvram partition also contains some RAM initialization parameters. Thus, calling nvram init after having erased the partition is quite dangerous.
Also, if the boot loader doesn't initialize and use nvram, then we shouldn't probably use it either.
What's your use case for this? In my opinion this might be dangerous to use.

hbl0307106015:
@nbd168 I just want to use the nvram functionality on other platform, such as ramips. And I am sure the nvram init will change nothing if there is a valid nvram magic and parameters in the partition.
As you said, on many devices the nvram partition also contains some RAM initialization parameters. Thus running nvram init on those devices will do nothing, I think it is safe.

On my board, I modified the partition table and add 'nvram' partition, but there is 'ff ff ff ff' on it. So I have to use the 'nvram init' to write the NVRAM MAGIC in the flash so that the other correspond command(get set unset commit) can work fine.

nbd168:
Why do you want to use nvram there? I think we should not support nvram for any cases other than where the boot loader uses it. @jow- what do you think?

hbl0307106015:
@nbd168  nvram is a useful tool. I think it should not only support on just brcm or ar71xx target.
It maybe can used in the case that saving configuration after restore factory setting(firstboot). And in my commit, this function can work on many other devices which boot loader do not support nvram.

nbd168:
We're not going to keep dummy nvram partitions around on devices where they are not used by the default firmware / boot loader. This would waste precious flash space. Without such a partition, nvram wouldn't work either.
If you want to just save some configuration, why don't you do something as simple as storing a .tar file in a partition, it's probably more useful than the nvram tool anyway.

hbl0307106015:
@nbd168  Yes, it is a dummy nvram in my case, one of the advantages of nvram is its key-value feature which seems more convenient than saving a tar file. And the compatibility of tar file approach will be a problem (such as upgrade from 14.07 to 17.01).

PR #809, opened 2017-02-11
ar71xx: fix platform_find_rootfspart()

f00b4r0:
platform_find_rootfspart() fails if the kernel partition comes before the
rootfs partition. The proposed patch fixes this while preserving what I
understand was the original idea: stop at first match.


f00b4r0:
For reference, this is what happens with the current code:
```shell
+ platform_find_rootfspart kernel:rootfs kernel
+ local part
+ [ kernel != kernel ]
+ break
+ local rootfspart=
```

nbd168:

PR #811, opened 2017-02-11
ramips: added image size into dwr-512 DT

fid0did0:


blogic:
merged, Thanks !

PR #812, opened 2017-02-12
ar71xx: archer C7: change reset button to reset function

heinzek:
Changed button description.
This router have only one button named Reset/WPS.
Reset/restart function is more important as WPS.

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c7.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c7.c
@@ -107,9 +107,9 @@ static struct gpio_led archer_c7_leds_gpio[] __initdata = {
 
 static struct gpio_keys_button archer_c7_gpio_keys[] __initdata = {
 	{
-		.desc		= "Reset button",
+		.desc		= "Reset/WPS button",
 		.type		= EV_KEY,
-		.code		= KEY_WPS_BUTTON,
+		.code		= KEY_RESTART,
 		.debounce_interval = ARCHER_C7_KEYS_DEBOUNCE_INTERVAL,
 		.gpio		= ARCHER_C7_GPIO_BTN_RESET,
 		.active_low	= 1,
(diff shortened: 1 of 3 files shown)


pepe2k:
@heinzek, there are also other devices with same "problem", e.g. TL-WDR3600/43x0, what with them? IMHO, we should be consistent in this subject.

Cheers,
Piotr

heinzek:
Great idea. I change this in TL-WDR3600/43x0 and set proper button description.

diizzyy:
I'm going to say no here because it doesn't make any sense.
If you want to reset/restart your router just pull the plug or (if available) use the hardware power switch. You can't replace the WDS button with pulling the plug.

pepe2k:
> I'm going to say no here because it doesn't make any sense.
> If you want to reset/restart your router just pull the plug or (if available) use the hardware power switch. You can't replace the WDS button with pulling the plug.

We use reset button [not only for rebooting the device](https://github.com/lede-project/source/blob/master/package/base-files/files/etc/rc.button/reset#L17). For me it's more like a question how we should deal with buttons which have more than one function in vendor's firmware.

Cheers,
Piotr

blogic:
i agree with @pepe2k 
i'll close this in a day or two if there are no other opinions being voiced

diizzyy:
Wait what? Why can't we have a timer based function? Quick press... WPS, long press reset?

pepe2k:
> Wait what? Why can't we have a timer based function? Quick press... WPS, long press reset?

I don't think there is a way to define more than one input event code for a particular GPIO driven button in kernel (correct me if I'm wrong). We could start using a reset button also for wps function, but that might be misleading for users, especially if their devices include separated buttons for these functions (reset and wps).

Cheers,
Piotr

heinzek:
> If you want to reset/restart your router just pull the plug or (if available) use the hardware power switch

You have right but restart function is not the same as turn power on/off. Some device have harddisk or mmc/sd device. When You power off this device you may lose unsaved data.
reboot function use sync command before reboot device.

> Wait what? Why can't we have a timer based function? Quick press... WPS, long press reset?

Key restart have timer based function:
0-4 sec = reboot
5-99 sec = reset

add wps option to this button is possible, but i dont known how this change will work on another devices that have separate reset and wps button.

I think this is possble to add:
0-1sec - wps
2-4sec - reboot
5-99sec - reset

pepe2k:
If there are no other comments, I would prefer to close this PR. There is [ongoing discussion on the mailing list](http://lists.infradead.org/pipermail/lede-dev/2017-February/006134.html) about possible solution/s for multi-function buttons.

Cheers,
Piotr

heinzek:
OKAY. I'll be waiting.

PR #813, opened 2017-02-12
tools/m4: update 1.4.18

bladeoner:
Patch 100-fix-gets-removal.patch is removed because it's included in the new version.

Compile tested: ar71xx


diizzyy:
As much as I/we appreciate your effort you need to mention if this have been compile and run-tested. You need to **at least** compile test and it's most likely not going to be comitted without at least a run-test.

bladeoner:
I already had been testing the compile on ar71xx.

I only didn't do a run-test yet.

nbd168:

PR #814, opened 2017-02-12
ramips: WN3000RPv3: do not setup switch

f00b4r0:
The WN3000RPv3 is a repeater with a single ethernet port. Setting up the
switch, even to disable it, is unnecessary and possibly confusing.

Configure LAN as eth0 instead.


blogic:
merged, Thanks !

PR #816, opened 2017-02-12
dnsmasq: bump to dnsmasq v2.77test3

kdarbyshirebryant:
New test release (since test1) includes 2 LEDE patches that are
upstream and may be dropped, along with many spelling fixes.

Add forthcoming 2017 root zone trust anchor to trust-anchors.conf.

Backport 2 patches that just missed test3:

Reduce logspam of those domains handled locally 'local addresses only'
Implement RFC-6842 (Client-ids in DHCP replies)


hnyman:
Not quite sure if it would be easier to just switch to git download, scrap the patches and directly download from commit X. 

Carrying upstream patches here is rather unnecessary if we are going to keep gradually upgrading dnsmasq also between the major releases.

kdarbyshirebryant:
The trap that you fall into with that approach is that VERSION is reported as 'unknown', whereas the tarballs at least have version set.

hnyman:
> The trap that you fall into with that approach is that VERSION is reported as 'unknown'

Looks like version detection magic from ./bld/get-version script is used to set the version in dnsmasq's Makefile. 
http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=Makefile;hb=HEAD#l66
http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=bld/get-version;hb=HEAD

A simple way would be to patch the VERSION file that is used as a source by the version magic script. That seems to work. The tarball release version has it in cleartext, while the git download will show only a placeholder "$Format:%d$" (that would need to be patched).
```
$ cat VERSION 
 (HEAD, v2.77test1, origin/master, origin/HEAD, master)
$ ./bld/get-version .
2.77test1

PATCHED VERSION:

$ cat VERSION 
 (HEAD, v2.77testHannu, origin/master, origin/HEAD, master)
$ ./bld/get-version .
2.77testHannu
```

But naturally it is more straightforward to fetch tarballs and then selectively add upstream patches if necessary.

zorun:
Ping @jow- , do you think this can be merged and backported to lede-17.01?

I think that (at least) the SERVFAIL fix should be backported to 17.01 before the release, see https://bugs.lede-project.org/index.php?do=details&task_id=391

diizzyy:
@hnyman 
This is just a temporary workaround until there's a proper release.

dedeckeh:
@zorun Upstepping dnsmasq to a test version in a stable branch like lede-17.01 does not make sense.
However backporting the SERVFAIL dnsmasq patch to the lede-17.01 branch is an option; feel free to launch a PR.

dedeckeh:
Applied into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=6fb66e49bfe25a8a4d408f93dedb4058d2952bb8); thanks

PR #824, opened 2017-02-13
Enable diagnostic LED on Geos, add PC Engines APU2

dwmw2:


dwmw2:
Note that the diag LED doesn't seem to work on the first boot — the defaults are applied too late. Options for fixing this without going back to hard-coded board options in /etc/diag.sh would be welcomed...

target/linux/x86/base-files/etc/board.d/02_network
```diff
@@ -11,6 +11,10 @@ board_config_update
 board="$(cat /tmp/sysinfo/board_name)" 2>/dev/null
 
 case "$board" in
+pc-engines-apu2)
+	ucidef_set_interfaces_lan_wan "eth1" "eth0"
+	ucidef_set_interface_raw "opt" "eth2"
```
obsy: Why not just a bridge / lan for eth1 and eth2?

riptidewave93: @obsy This is done mainly to match other firmware's that run on the device, ex PfSense. If there is a good argument to merge them instead then that path could be taken as well.

obsy: IMO should be wan (eth0) and land (eth1/eth2).  pfSense is a different firmware, we do not need to stick to their rules.

dwmw2: Should I change the above-cited part to just one line
```
        ucidef_set_interfaces_lan_wan "eth1 eth2" "eth0"
```
?

Or drop that patch entirely and submit only the bits I can test, leaving @riptidewave93 to reach consensus, test, and submit the APU2 part?

riptidewave93: @dwmw2 I am fine with this change if it puts the board in line with other LEDE devices. If you don't mind, please give me a short bit to do the testing (30 minutes) and we can decide where to go from there.

dwmw2: I'll push what I have (including the corresponding LED change). It's best if you test precisely what's in the PR.

riptidewave93:
@dwmw2 LGTM and confirmed working on an APU2 unit. :+1:

dwmw2:
Fixed, thanks.

dwmw2:
I took a quick look at making the generic `diag.sh` work earlier in the first boot, by making it use `/etc/board.json` if there's no configuration. And then by making it create `/tmp/board.json` since we want it to work before the root file system is mounted...
And then I realised that `/tmp/sysinfo/model` doesn't even exist, and we weren't getting the diag LED early in the first boot *anyway*. So I might push what I have, but on the whole I'm not too worried about the current approach losing the "first boot" diagnostic LED as it wasn't there anyway.

dwmw2:
OK, all sorted. It'll work as early as `/tmp/sysinfo/model` is available, using the default LED until/unless the configuration is populated.

blogic:
merged, Thanks !

riptidewave93:
@dwmw2 Been doing more testing, seems this fails on squashfs targets as board_detect calls the empty /etc/config/system* files, causing DIAG led to completely fail. Thoughts?

dwmw2:
@riptidewave93 This is not on first boot but later? And `/tmp/sysinfo/model` exists (so it worked before)? But the overlay hasn't yet been mounted so `/etc/config/system` still doesn't?

riptidewave93:
@dwmw2 you got it. This only happens on the squashfs image so its overlay related. This problem does not occur on the ext4 image.

dwmw2:
Would my (abandoned) suggestion of using `/tmp/board.json` have worked here? Or is there a simpler  option with just reordering the diag LED setup vs. the mounting of the overlay?

mkresin:
Could be as easy that target/lib/preinit/20_sysinfo runs to late. The first time the diag led is set via base-files/lib/preinit/10_indicate_failsafe.

For reference the following commit from my staging/boardname branch: https://git.lede-project.org/6ca3681a98d503aa14a583cf46fcd6d73da62706

PR #827, opened 2017-02-14
ar71xx: WNDR4300: enlarge flash storage space

yuanl:
This device has total of 128MB flash. However, 96MB of them was
reserved for doing nothing.
[this openwrt ticket](https://dev.openwrt.org/ticket/17316) was propose
the same thing but never been merge.
[lots of people](https://www.google.com/search?q=wndr4300+121856k)
including myself are using this extra 96MB space without problem.

Signed-off-by: Yuanliang Li yuanl.lee@gmail.com


pepe2k:
Please, always include SoB line and fix line length in the commit message (more information about submitting patches: [here](https://lede-project.org/docs/guide-developer/the-source-code#submitting_patches) and [here](https://github.com/lede-project/source/pull/592)).

nbd168:
Also, it seems that you're overwriting the caldata_backup partition. As far as I can tell, that's the reason why the patch wasn't merged in the first place. Any solution to this issue should find a way around that, maybe by using mtdconcat.

PR #829, opened 2017-02-14
ramips: Clean duplicated status property for Omega2 WMAC in dtsi

HZFrodo:
At the tail of dtsi, wmac is enabled twice, clean the first one


blogic:
merged, Thanks !

PR #832, opened 2017-02-14
tools: patch-image: fix file descriptor leak.

HZFrodo:
Maybe this is committed by mistake, fix it.


blogic:
merged, Thanks !

PR #775, opened 2017-02-03
17.01: ar71xx: Add support for A40/A60 and OM2Pv4/OM2P-HSv4

ecsv:
I am not aware of some policy regarding HW support patches during the 17.01 development. It is therefore quite likely that these patches should not be integrated in 17.01.

PR #774, opened 2017-02-03
ar71xx: Add support for A40/A60 and OM2Pv4/OM2P-HSv4

ecsv:

PR #834, opened 2017-02-14
17.01: ar71xx: Add support for A40/A60 and OM2Pv4/OM2P-HSv4

ecsv:
This is the follow up for #775. I was told by @blogic in #774#issuecomment-279680361 to open an PR to get it backported to LEDE 17.01

@nbd168 Does this approach fulfill all requirements from #775#issuecomment-278907224  or would prefer to run `git cherry-pick 9c2422709075196022fcbc2d0f5f83e01d02b951..3d16c16eb669caf7244461a9e83d5f871591681c` directly?

My personal opinion is, that this PR is not following the rules from https://www.mail-archive.com/lede-dev@lists.infradead.org/msg05725.html - but maybe this only applies to patches send via the mailing list.


blogic:
i only asked for the new PR as the previous did not mention 17.01, if the patches are indeed identical we can cherry-pick them, which is much easier. are they identical ?

ecsv:
Yes, they are identical

blogic:
cherry-picked, Thanks !

PR #836, opened 2017-02-14
Omap uboot updates

mrnuke-adaptrum:


blogic:
the description of 2/2 does not explain why the 2 patches are removed.

mrnuke-adaptrum:
Does a "the patches are removed because they are not needed anyomore" count?

blogic:
@mrnuke-adaptrum if that is the real reason then its more than valid :-)

mrnuke-adaptrum:
@blogic well, in that case, I am honored to be properly documenting the changes.

blogic:
merged, Thanks !

mkresin:
@mrnuke-adaptrum Starting with your uboot-omap patches, the [omap build on our buildbots fails](http://phase1.builds.lede-project.org/builders/omap%2Fgeneric/builds/24). Would you please have a look at the issue.

mrnuke-adaptrum:
@mkresin It looks like your buildroot is missing dtc.

> ./scripts/dtc-version.sh: line 18: dtc: command not found

mrnuke-adaptrum:
That being said, is there a particular thing one has to do to make sure that dtc is pulled in, or is it a buildroot issue?

mkresin:
We do not install any extra packages beside the essential build packages on any of the build bots.

If a package needs some extra tools, these have to be packaged and the package relying on the extra package need to depend on this package.

mrnuke-adaptrum:
Okay, I see now. omap-sunxi and omap-uboot both use a patch to modify the uboot makefile, then specify the path to dtc to be $(LINUX_DIR)/scripts.

I could make a patch to fix this on uboot-omap, but I don't know a way to test if it actually fixes the problem.

mkresin:
Yes please send a patch to fix it the same way.

To test your patch, select your omap board and the bootloader in menuconfig. Do a full build to compile the kernel including the dtc binary. Afterwards run ```make package/uboot-omap/{clean,compile} V=s``` and check the log output for the invocation of dtc. It hopefull shows the full path to the dtc binary.

mrnuke-adaptrum:
It seems that sunxi and mvebu targets use the wrong approach. If you look in include/u-boot.mk:

`DTC=$(wildcard $(LINUX_DIR)/scripts/dtc/dtc)`
`[snip]`
`		$(if $(DTC),DTC="$(DTC)")`

The DTC should already be set correctly. I've verified that the u-boot makefile does indeed respect that. The problem is here:

`DTC=$(wildcard $(LINUX_DIR)/scripts/dtc/dtc)`

This line does not get evaluated correctly, and $(DTC) ends up being empty. I would prefer if someone more knowledgeable than me in make syntax could take a look at this

nbd168:
The uboot-omap Makefile needs to include $(INCLUDE_DIR)/kernel.mk right after rules.mk

mrnuke-adaptrum:
@nbd168 Thank you!!!!

mrnuke-adaptrum:
Resolved in #868

PR #833, opened 2017-02-14
mtdsplit: Add support for Mikrotik NOR firmware

f00b4r0:
This series of patches add mtdsplit support for MikroTik SPI NOR RouterBOARDS.
It also converts RB SPI NOR images to metadata style.

- [X] Generic bits added to the PR
- [X] Arch-dependent bits added to the PR

PR #800, opened 2017-02-07
ar71xx: Rework MikroTik RB SPI-NOR support and add several devices 

f00b4r0:
This series of commits rewrite from scratch the way MikroTik RouterBOARDs with SPI NOR are supported in LEDE.

This is a resubmit of https://patchwork.ozlabs.org/patch/724226/ and following, and an alternative to #778 

A single image (initramfs and sysupgrade) is generated that supports the already supported hAP lite (tested on actual hardware), as well as the mAP lite (tested on actual hardware), hAP (tested on actual hardware by @adron-s), hEX PoE lite (tested on actual hardware by @joaochainho) and hEX lite. The code is documented and kept as compact as possible.

A penultimate commit offers basic kernel-only support for 3 more untested devices (cAP, mAP and wAP): network interfaces should work correctly, as well as LEDs, but GPIO buttons (if any) will not be supported.
The rationale is to give the opportunity for beta testers with this hardware to report a successful boot of their devices and help make them fully supported. There is no risk of bricking the devices since the image cannot be flashed from RouterOS, but instead from a previously netbooted LEDE initramfs.

The last commit in this series offers a switch to firmware mtdsplit system, enabling more efficient use of the flash space. This commit has been tested on actual hardware.

- [X] mAP lite fully working
- [X] hAP lite fully working
- [x] hEX (PoE) lite / hAP fully working
- [X] Serial support enabled
- [x] Code for cAP / mAP / wAP added
- [x] Add mtdsplit support

PR #839, opened 2017-02-15
17.01: ar71xx: Remove images for rb-941-2nd

f00b4r0:
The code that generates this image in 17.01-RC is broken in many ways.

The new code currently in master (from #800 and #833) generates images that will not be compatible with the ones in 17.01. To avoid a migration nightmare, this patch removes image generation for this device in 17.01


jow-:
Comitted with cdeb2322eab605639f1545a2eee0625424cedf03 - thanks!

PR #840, opened 2017-02-15
x86: Add Geos profile for Geode subtarget

dwmw2:
This makes the Geode images actually useful again. The Geos profile
should include the relevant hardware for that board, and the Default
profile adds the via-rhine adapter which seems to have been present in
the net5501 and alix targets killed in commit 9e0759ea265 ("x86: merge
all geode based subtargets into one").

Signed-off-by: David Woodhouse <dwmw2@infradead.org>


blogic:
merged, Thanks ! (also for 17.01)

PR #844, opened 2017-02-16
ramips: fix the number of uarts for MT7688

HZFrodo:
Same as MT7621 and MT7628, MT7688 allows up to 3 uarts


blogic:
merged, Thanks !

PR #847, opened 2017-02-17
libevent2: update to 2.1.8-stable

liudf0716:
Signed-off-by: Dengfeng Liu <liudengfeng@kunteng.org>

PR #848, opened 2017-02-17
libevent2:  version update to 2.1.8-stable

liudf0716:
Compile tested: ar71xx & ramips & ipq806x  OpenWRT Chaos Calmer & LEDE 17.01 RC1
Compile run:  ar71xx & ramips & ipq806x  OpenWRT Chaos Calmer & LEDE 17.01 RC1

update libevent2 from version 2.0.22 to 2.1.8
add patch to fix bug of needing to add depends on libopenssl in all libevent library, according to communicate with  @azat who is a member of libevent2, this bug will be fixed in version 2.1.9. here is issue
https://github.com/libevent/libevent/issues/462

Signed-off-by: Dengfeng Liu <liudengfeng@kunteng.org>

### DIFF
--- a/package/libs/libevent2/Makefile
+++ b/package/libs/libevent2/Makefile
@@ -8,13 +8,16 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libevent2
-PKG_VERSION:=2.0.22
-PKG_RELEASE:=1
+PKG_VERSION:=2.1.8
+PKG_RELEASE:=2
 
-PKG_BUILD_DIR:=$(BUILD_DIR)/libevent-$(PKG_VERSION)-stable
+PKG_SOURCE_PROTO:=git
+PKG_BUILD_DIR:=$(BUILD_DIR)/libevent-$(PKG_VERSION)
 PKG_SOURCE:=libevent-$(PKG_VERSION)-stable.tar.gz
-PKG_SOURCE_URL:=@SF/levent
-PKG_HASH:=71c2c49f0adadacfdbe6332a372c38cf9c8b7895bb73dabeaa53cdcc1d4e1fa3
+PKG_SOURCE_SUBDIR:=libevent-$(PKG_VERSION)
+PKG_SOURCE_URL:=https://github.com/libevent/libevent.git
+PKG_SOURCE_VERSION:=release-$(PKG_VERSION)-stable
+PKG_MIRROR_HASH:=7227b9ae556f22f9d9815c6ba2630b614510f3cb0ac44c3775fe55c5cae720bb
 PKG_MAINTAINER:=Jo-Philipp Wich <jo@mein.io>
 PKG_LICENSE:=BSD-3-Clause
 
@@ -45,7 +48,7 @@ endef
 
 define Package/libevent2
   $(call Package/libevent2/Default)
-  TITLE+= library (version 2.0)
+  TITLE+= library (version 2.1)
 endef
 
 define Package/libevent2/description
@@ -57,7 +60,7 @@ endef
 
 define Package/libevent2-core
   $(call Package/libevent2/Default)
-  TITLE+= core library (version 2.0)
+  TITLE+= core library (version 2.1)
 endef
 
 define Package/libevent2-core/description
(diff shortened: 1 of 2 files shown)


diizzyy:
Again, you're breaking build variants (openssl vs non openssl). There's no point in switching to git especially when you're pulling a release and it's not the preferred way.  See #847 for more information.

liudf0716:
sorry for these annoying pr, I am not good at submiting pr. next time I will attention it

wongsyrone:
OK. I did some research on this one.

If you add `--disable-openssl` , it will not build libevent_openssl.so* anymore. I checked the whatsnew-2.1 file, it shows it will find openssl via pkg-config and link to it.

There are two options:

1. link to openssl unconditionally, this is what @liudf0716 has done
2. patch the build system, including Makefile.am, configure.ac, etc. Other source files may get involved. The downside is hard to keep up with upstream. The good point is we can benefit from the less dependencies.

wongsyrone:
For downloading tarball, I checked the official site, it will redirect to GitHub release (Amazon S3).
Example:
https://github.com/libevent/libevent/releases/download/release-2.1.8-stable/libevent-2.1.8-stable.tar.gz
https://github.com/libevent/libevent/releases/download/release-2.0.22-stable/libevent-2.0.22-stable.tar.gz
These two links are taken from http://libevent.org/
I'm okay with the git way.

liudf0716:
@wongsyrone  did u try cmake of libevent? It can't be compiled with cmake way.

wongsyrone:
Nope. I checked the repo git history, the author is tuning cmakelists. In my opinion, we should stick to the autotools until the author switches to CMake completely.

liudf0716:
I have submitted the issue to libevent project. when they finished it, I will update my PR

wongsyrone:
Do you mean https://github.com/libevent/libevent/issues/462?

liudf0716:
yes

liudf0716:
@wongsyrone I have give you invitation to join our apfree_wifidog project. hope u can accept it, thanks.
the following  is our project address:
https://github.com/liudf0716/apfree_wifidog

liudf0716:
I need to change any of my pull request? how long a pr will be accepted if no problem?

wongsyrone:
The issue has been fixed via https://github.com/libevent/libevent/commit/5d18d87916728df67640bd225967c6d6df4a54f8, we may wait for next release or backport it to get rid of linking to openssl blindly.

liudf0716:
that bug caused by a sample which caused all library linking openssl. Just commented it should be OK.

wongsyrone:
Why not pass `--disable-samples` directly?

wongsyrone:
You should squash all commits into a single one at least.

package/libs/libevent2/Makefile
```diff
@@ -8,13 +8,16 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libevent2
-PKG_VERSION:=2.0.22
-PKG_RELEASE:=1
+PKG_VERSION:=2.1.8
+PKG_RELEASE:=2
```
wongsyrone: Any particular reason to bump PKG_RELEASE?

liudf0716: Can not just use --disable-sample.ok, I will restore PKG_RELEASE

liudf0716: I will cooperate with @azat to test cmake of libevent, after finish that task, I will update this PULL REQUEST.

liudf0716:
@wongsyrone thanks, I have squashed it.

wongsyrone:
Tested for a long time: https://github.com/wongsyrone/lede-1/commit/95ee9ca6de20e7f2a1cfde23490efc5d625c47fd

diizzyy:
https://github.com/diizzyy/source/blob/9f271f09469e49f79b9e4ff8eb49968d52f25f93/package/libs/libevent2/Makefile

This is what I have at the moment...
* All additional packages depends on libevent2
* OpenSSL and pthreads are linked to main lib in additional to the extra libs

The only thing I can't fix is automatic rebuild if you change settings between openssl and pthreads using menuconfig. Settings seems to stick even if you deselect :-/

diizzyy:
@liudf0716 
Do you have any objects with #1080?

diizzyy:
This version has bugs, just wait until a fixed release is available.

diizzyy:
@liudf0716 
Please close this one, we'll just wait for a release by upstream instead as there are no pressing issues to switch right now.

PR #849, opened 2017-02-17
nvram: lots of improvements

hbl0307106015:
e944b30: add usage() function for merging the help message in a single function.
48da92b:  since the incorrect magic code can result in error return, add the help message for 'nvram magic not found'.
7911f8f: improve argc check, return usage message when argc less than 2.
a4e5d1c:  fix memory leak for nvram_open() and nvram_open_rdonly() function.


hbl0307106015:
@jow- @nbd168 Excuses me, any problem with my PR ?

PR #854, opened 2017-02-19
ar71xx: Add more boards to the routerboard serial detection patch

f00b4r0:
This patch enables serial console on the 435 series and the 953gs.

953gs requested here: http://lists.infradead.org/pipermail/lede-dev/2017-January/005197.html
435 requested by @djbuldog


blogic:
merged, Thanks !

PR #855, opened 2017-02-20
mikrotik: Remove console hack and use CONFIG_CMDLINE instead

f00b4r0:
This proposed patch removes a suboptimal hack implemented in
701-MIPS-ath79-add-routerboard-detection.patch : the list was getting
long, would become difficult to maintain and might eventually trigger
false positives.

Instead, we use CONFIG_CMDLINE in the mikrotik subtarget to achieve the
same effect on _all_ mikrotik devices, while allowing local override.


pepe2k:
Hi @f00b4r0,

Two notes:

1. For AR933x based devices (as far as I can see, for now we don't have any AR933x based MikroTik boards supported, correct me if I'm wrong here) this will not work (that kind of boards would need to have `ttyATH0` instead of `ttyS0` in kernel command line).

2. AFAIK, AR934x and QCA955x contain two UARTs:

- UART0 (called low-speed or "LS-UART", 16550 compatible)
- UART1 (called hi-speed or "HS-UART", compatible with the one available in AR933x)

And the question here is: are you sure that all MikroTik boards based on AR934x/QCA955x are using low-speed UART as the main serial console (I know that this is true for most of the QC/A based boards, but I know few which don't follow this rule and use HS-UART as default console)?

Cheers,
Piotr

f00b4r0:
Hi @pepe2k 

> Two notes:
> 
> For AR933x based devices (as far as I can see, for now we don't have any AR933x based MikroTik boards supported, correct me if I'm wrong here) this will not work (that kind of boards would need to have ttyATH0 instead of ttyS0 in kernel command line).

As far as I can tell, there are no (supported) mikrotik boards with AR933x, so that point is fine I think.
 
> AFAIK, AR934x and QCA955x contain two UARTs:
> 
> UART0 (called low-speed or "LS-UART", 16550 compatible)
> UART1 (called hi-speed or "HS-UART", compatible with the one available in AR933x)
> And the question here is: are you sure that all MikroTik boards based on AR934x/QCA955x are using low-speed UART as the main serial console (I know that this is true for most of the QC/A based boards, but I know few which don't follow this rule and use HS-UART as default console)?

I think this is very theoretical: _all_ of the "serial-enabled" boards listed in the patch I'm trying to remove are using ttyS0,115200. It's possible that mikrotik chose to use a different console for some of their boards, but 1) that seems highly unlikely (it would add unnecessary design and maintenance burden) and 2) from what I see in the mach-rb* files, there are currently no such board supported by LEDE. So again, I'm not sure that point applies.

Finally, let's look at what this is about:
- right now, we have a fairly ugly patch that has absolutely no chance of ever going upstream, that will possibly become a nightmare to maintain as more and more boards get added, and that can have undesirable side effects (false positive).
- on the other hand, CONFIG_CMDLINE is provided exactly for this purpose, and the parameters passed by the bootloader should normally override CONFIG_CMDLINE, which is not the case of the current patch.
- Finally, most users will never really care what the serial setting is because they will never interact with serial anyway. Those who will can certainly change their boot parameters.

That being said, now that I've been reading target/linux/ar71xx/image/Makefile, I even wonder why we're not using the CONSOLE parameter in the Device definition. It seems like an even better way of dealing with this 😉 

Maybe there was a specific reason of not doing it that way to being with, and that's what I'd like to hear about. Git blame points at @nbd168 for the original patch, maybe he could provide valuable input here...

Cheers,
Thibaut

pepe2k:
> That being said, now that I've been reading target/linux/ar71xx/image/Makefile, I even wonder why we're not using the CONSOLE parameter in the Device definition. It seems like an even better way of dealing with this 😉

This is exactly what I was going to suggest as a more universal way (than keeping it hard coded it subtarget's kernel config). Here is the related line for cmdline patching: https://github.com/lede-project/source/blob/master/include/image-commands.mk#L146 and the tool source code: https://github.com/lede-project/source/blob/master/tools/patch-image/src/patch-cmdline.c

Cheers,
Piotr

f00b4r0:
I'm going to propose a different PR using patch-cmdline

f00b4r0:
Turns out I don't like what I see in patch-cmdline, and it yields non-bootable kernels. So I'll stick to this proposal.

mkresin:
@f00b4r0 would you please give us an update on this? Is it still RFC?

f00b4r0:
It is. I didn't hear back after my last comment...
Cheers

pepe2k:
IMHO, it doesn't make sense now as of/dts support in ar71xx/ath79 should be ready soon.

Cheers,
Piotr

mkresin:
> It is. I didn't hear back after my last comment...

Not sure what you expect. Pepe suggested to use ```patch-cmdline``` which you don't like/doesn't work. I guess it is as simple as there are no more comments. And there will be no final "do it this way" from any of the devs.

As far as I can see, it's a decision between plague and cholera. Either we force ```ttyS0,115200``` for all mikrotik boards or we have to use the growing if condition.

Forcing the same serial console settings for all boards _might_ break some unknown boards or some boards in future. I really see no real improvement here and agree with pepe that we should do it the proper way as soon as device tree support is available.

f00b4r0:
> Le 3 mars 2017 à 22:42, Mathias Kresin <notifications@github.com> a écrit :
> 
> It is. I didn't hear back after my last comment...
> 
> Not sure what you expect. Pepe suggested to use patch-cmdline which you don't like/doesn't work.
> 
Correct.
> I guess it is as simple as there are no more comments. And there will be no final "do it this way" from any of the devs.
> 
> As far as I can see, it's a decision between plague and cholera.
> 
It is indeed.
> Either we force ttyS0,115200 for all mikrotik boards or we have to use the growing if condition.
> 
Correct.
> Forcing the same serial console settings for all boards might break some unknown boards or some boards in future.
> 
Highly hypothetical. Personally I don't consider "it might break something we don't know about in a distant future" a valid objection, but that's not my call to make.
> I really see no real improvement here and agree with pepe that we should do it the proper way as soon as device tree support is available.
> 
Fair enough: I'm not the one who will be maintaining the growing if() hack so I don't have a strong opinion about this.

Cheers

pepe2k:
> Fair enough: I'm not the one who will be maintaining the growing if() hack so I don't have a strong opinion about this.

It will go away when we switch to device tree ([ramips example](https://github.com/lede-project/source/blob/master/target/linux/ramips/dts/OMEGA2.dtsi#L10)).

>> Forcing the same serial console settings for all boards might break some unknown boards or some boards in future.

> Highly hypothetical. Personally I don't consider "it might break something we don't know about in a distant future" a valid objection, but that's not my call to make.

As I already wrote here before, _there are_ MikroTik boards based on AR933x (ttyATH0!), just not (yet) supported in LEDE.

Cheers,
Piotr

f00b4r0:
Hi @pepe2k 

> As I already wrote here before, there are MikroTik boards based on AR933x (ttyATH0!), just not (yet) supported in LEDE.

And as I already wrote here, this change _will not break them_, if we ever support them. CONFIG_CMDLINE sets defaults, it doesn't override the bootloader's settings (unless CONFIG_CMDLINE_OVERRIDE is set). So it remains perfectly possible to use a different setting _anyway_.

This is why I think this solution is superior to both the existing hack (which unconditionally appends the console setting) *and* patch-cmdline (which entirely and _silently_ trashes whatever is set by CONFIG_CMDLINE): it's straightforward, obvious, and offers a single, clear modification point that's easy to tune at compile time _and_ at boot time.

Finally, we're talking about serial console here. Most users don't care one bit what the serial console setting is, and those who do are fully capable of changing CONFIG_CMDLINE if they wish (and again, this is a much more obvious tunable than editing some obscure patch).

My 2c here 😃

pepe2k:
Hi @f00b4r0,

>> As I already wrote here before, there are MikroTik boards based on AR933x (ttyATH0!), just not (yet) supported in LEDE.
 
> And as I already wrote here before, this change will not break them, if we ever support them. CONFIG_CMDLINE sets defaults, it doesn't override the bootloader's settings (unless CONFIG_CMDLINE_OVERRIDE is set). So it remains perfectly possible to use a different setting anyway.

Ha, good point!

> Finally, we're talking about serial console here. Most users don't care one bit what the serial console setting is, and those who do are fully capable of changing CONFIG_CMDLINE if they wish (and again, this is a much more obvious tunable than editing some obscure patch).

I'm not sure if this is true for _most users_, maybe for _most of MikroTik hardware users_ ;)

Cheers,
Piotr

pprindeville:
Can this be rebased?  Are there any remaining issues with this PR?  It sounded to me like there's convergence...

pepe2k:
Hi @pprindeville,

> Can this be rebased? Are there any remaining issues with this PR? It sounded to me like there's konwergencja...

As in [my first comment](https://github.com/lede-project/source/pull/855#issuecomment-281056100), this will not work for AR933x devices as they use different console. I have one MikroTik with AR9331 in queue for support.

Cheers,
Piotr

f00b4r0:
How about proposing a clean alternative that works for your currently-unsupported-but-in-queue devices?

PR #858, opened 2017-02-20
swconfig: Bugfix switch_port uci option parsing

scrpi:
When not defining 'device' or 'vlan' in relevant switch_port uci
sections, behaviour is inconsistent due to *devn, *port and *vlan
pointers not being zero initialized.

Signed-off-by: Ben Kelly <ben@benjii.net>


nbd168:
Merged, thanks.

PR #859, opened 2017-02-20
dnsmasq: do not forward rfc6761 excluded domains

kdarbyshirebryant:
RFC 6761 defines a number of top level domains should not be forwarded
to the Internet's domain servers since they are not responsible for
those domains.

This change adds a list of domains that will be blocked when 'boguspriv'
is used and augments that which is already blocked by dnsmasq's notion
of 'local service' using '--bogus-priv' i.e. RFC 1918 private addresses
and IPv6 prefixes as defined in RFC 6303.

To make this configurable rather than hard coded in dnsmasq's init
script, a new file /usr/share/dnsmasq/rfc6761.conf is conditionally
included.

The default file matches the RFC 6761 recommendation along with a few
other top level domains that should not be forwarded to the Internet.


dedeckeh:
Merged into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=b2c77e576f9ce6a9664a1d1ecb35c35bafd51316); thanks

PR #860, opened 2017-02-20
x86_64: Add support for EFI framebuffer (FS#515)

alive4ever:
CONFIG_FB_EFI and CONFIG_FB_SIMPLE are needed to display console text on
EFI framebuffer.

CONFIG_FB_EFI is needed when the kernel is directly launched via EFI
shell or EFI startup.nsh script.

CONFIG_FB_SIMPLE is needed when launching the kernel via grub2 efi. In
this case, grub2 has prepared a gfxterm framebuffer and the kernel just
need to use the already prepared grub's gfxterm framebuffer to display
console text.

Signed-off-by: Alif M. Ahmad <alive4ever@live.com>


nbd168:

PR #861, opened 2017-02-21
R6220-release

jessons:


jessons:
release

obsy:
What is R6220-release, why did you deleted enablemodem and bootcount, why wifi is enabled by default?

samm-git:
@obsy as far as i could see it is  Netgear R6220, and +1 that wifi should not be enabled by default.

updateing:
And... why is it R6220_SPI.dts? The router comes with NAND. Did you do the same hardware mod as I did?

Even if you modded your router, there would still be two problems, apart from what others mentioned above:

1. LEDE may not accept patches for non-stock hardware.
2. If you get the files from [my repo](https://github.com/updateing/lede-r6220-spi/commits/r6220-spi), why don't you use my commits directly, or at least keep original author info?

mkresin:
> LEDE may not accept patches for non-stock hardware.

That is completely right. I will close this one. The PR creator hasn't responded to the question so far and I'm not sure if this PR was an accident.

PR #867, opened 2017-02-22
x86: automate building UEFI bootable images

alive4ever:
This is a WIP.
Working: UEFI bootable ext4 image.
Not working: UEFI bootable squashfs image.
Not tested: ISO and ramdisk image.

The resulting image is located inside ./bin/target/x86/ for TARGET_x86_64_efi and ./bin/target/x86/efi for TARGET_x86_efi.


nbd168:
Please don't add more subtargets for EFI. The normal ones should be able to handle adding EFI support just fine. If they need extra kernel config options, just enable them.

alive4ever:
Thanks for the review. I'll create another branch that makes use of profile instead of subtarget.

nbd168:
Please also don't use profiles for it. I think EFI and non-EFI builds should be done in one build run

alive4ever:
New PR has been submitted, with no new profiles nor subtargets.

PR #868, opened 2017-02-23
uboot-omap: Makefile: Include kernel.mk

mrnuke-adaptrum:
This is needed for the definition of $(LINUX_DIR). This is used in
u-boot.mk to change the device-tree compiler to the dtc used by
linux.

Without this, the build will use the system dtc. This results in a
build failure when a dtc does not exist outside the buildroot.

Signed-off-by: Alexandru Gagniuc <alex.g@adaptrum.com>


nbd168:
Merged, thanks.

PR #870, opened 2017-02-23
ar71xx: mikrotik: erase firmware partition on SPI NOR devices in pre_upgrade()

f00b4r0:
On Mikrotik SPI NOR devices, the firmware partition must be erased when flashing
from stock firmware, otherwise leftover bits (in particular a kernel signature)
can trigger a boot loop.

When booted from initramfs (the only way to flash LEDE on these devices),
this patch unconditionally erases the firmware partition in the pre_upgrade()
stage for all supported SPI NOR devices.


nbd168:

PR #876, opened 2017-02-24
dnsmasq: tweaks

kdarbyshirebryant:
Make 'non-wildcard' interfaces the default.
Remove check for existing dhcp server on interface.

### DIFF
--- a/package/network/services/dnsmasq/files/dhcp.conf
+++ b/package/network/services/dnsmasq/files/dhcp.conf
@@ -15,7 +15,7 @@ config dnsmasq
 	option leasefile	'/tmp/dhcp.leases'
 	option resolvfile	'/tmp/resolv.conf.auto'
 	#list server		'/mycompany.local/1.2.3.4'
-	#option nonwildcard	1
+	option binddynamic	1 # bind to & keep track of interfaces
 	#list interface		br-lan
 	#list notinterface	lo
 	#list bogusnxdomain     '64.94.110.11'
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -58,30 +58,6 @@ dhcp_calc() {
 	echo "$res"
 }
 
-dhcp_check() {
-	local ifname="$1"
-	local stamp="${BASEDHCPSTAMPFILE_CFG}.${ifname}.dhcp"
-	local rv=0
-
-	[ -s "$stamp" ] && return $(cat "$stamp")
-
-	# If there's no carrier yet, skip this interface.
-	# The init script will be called again once the link is up
-	case "$(devstatus "$ifname" | jsonfilter -e @.carrier)" in
-		false) return 1;;
-	esac
-
-	udhcpc -n -q -s /bin/true -t 1 -i "$ifname" >&- && rv=1 || rv=0
-
-	[ $rv -eq 1 ] && \
-		logger -t dnsmasq \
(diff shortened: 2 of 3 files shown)


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -58,30 +58,6 @@ dhcp_calc() {
...
-
-	[ $rv -eq 1 ] && \
-		logger -t dnsmasq \
-			"found already running DHCP-server on interface '$ifname'" \
-			"refusing to start, use 'option force 1' to override"
-
-	echo $rv > "$stamp"
-	return $rv
-}
-
```
EricLuehrsen: The makes `force` more than the default, but the only option right? I thought this check is to prevent conflicts that can cause real grief in connecting to our device, debugging, and fixing what are often each our own stupid mistakes. (1) something really messed up in configuration. (2) plugging your WAN cord into a LAN port (or other routed chain).

package/network/services/dnsmasq/files/dhcp.conf
```diff
@@ -15,7 +15,7 @@ config dnsmasq
 	option leasefile	'/tmp/dhcp.leases'
 	option resolvfile	'/tmp/resolv.conf.auto'
 	#list server		'/mycompany.local/1.2.3.4'
-	#option nonwildcard	1
+	option nonwildcard	1 # bind to & keep track of interfaces
```
EricLuehrsen: Most people either backup or put their favorite UCI in the build `files/etc/config` so this may not be adopted. I agree with the forced shift however, because its 99% transparent to most users, but useful for us to build on for future features. Might I suggest further eliminating this option in dnsmasq.init, and using new option `bind_dynamic`. Also, make dnsmasq.init assume a 1 if not included.

package/network/services/dnsmasq/files/dnsmasq.init
kdarbyshirebryant: Hi Eric, yes I agree the original code is largely there as an 'idiot proof' factor.  @jow expressed the opinion on IRC something like 'dnsmasq is the only thing we protect in this way, which makes it an anomaly and perhaps the check should go' - so this change is the result of that opinion.

bobafetthotmail: All times I did one of such mistakes (leaving DHCP on in LAN when connecting it to a lan where there was already a DHCP server of a device with stock firmware or dd-wrt) the openwrt/lede device's DHCP was still active, and I experienced the expected breakage for that setup. 
I never thought that there was a check about that, are you sure the current check works at all?
Or what I experienced is unrelated?

package/network/services/dnsmasq/files/dhcp.conf
kdarbyshirebryant: Ok, so I've done that which of course leads to tweaking luci now as well https://github.com/kdarbyshirebryant/luci/commit/17882e86321749c043b3581bd7540a566c2e1e0b

Not done the translations - no idea how.  The other 'fun' aspect - the 'ignoring/including' interface option was dependent on 'non wildcard' - which is really bogus 'cos ignoring or including interfaces is independent of dnsmasq's binding mechanism.  Ah LEDE, changing one thing leads to 6 other changes....across repos. joy.

kdarbyshirebryant:
Chaps - I've decided to split these 2 commits into separate PRs... I think it'll be easier that way..and I'll deal with the 'make file revision bump conflict; if required depending on what/when gets accepted.

PR #877, opened 2017-02-24
Amazon-SE SoC subtarget for Lantiq

mcmilk:
This patchset will re-add the Amazon-SE SoC to LEDE.

If there are questions, just mail me.


target/linux/generic/patches-4.4/308-mips32r2_tune.patch
```diff
@@ -1,17 +0,0 @@
-Add -mtune=34kc to MIPS CFLAGS when building for mips32r2
```
langerth: Instead of removing this patch, just disable CONFIG_CPU_MIPS32_R2 in the ase kernel config

mcmilk: Ah, okay, I will do that.

nbd168: It seems that MIPS 4KEc is supposed to implement MIPS32 R2. If this patch breaks it, please show me an example crash along with disassembly so that I can see which instruction it breaks on. Maybe there's a less intrusive tweak that we can do here.

mcmilk: -mips32r2 is okay, but your patch also does an -mtune=34kc, this will not work on my 4kc ...

I add this flag to CONFIG_TARGET_OPTIMIZATION: -march=4kec -EB ... this works.

nbd168: As I mentioned above, I need more details than "will not work"

mcmilk: I will add them, the init is crashing with -mtune=34kc then. What commands should I run on the "good" and the "wrong" binary?

mcmilk: I checked the issue again, and it seems that compiling with -march=mips32r2 -mtune=34kc works now as expected on my system. I made an complete rebuild (also the toolchain) ... maybe some bug in gcc in december 2016 was the reason for that issue. But removing the patch solved it 2016 for me... but now, it builds fine, even with the patch ;-)

mcmilk:
Compiling with -march=mips32r2 and -mtune=34kc works now as expected on my system.... the 308-mips32r2_tune.patch is okay.

mkresin:
Hey Tino,

would you please cleanup you commits. The commits do not have the correct prefix and there is a lot of room to improve the commit messages (the commit message should describe how and why something is done). Have a look the [patch submission draft](https://github.com/bobafetthotmail/source/blob/68913cbafbd7b16f9819c5b29db3cda307a79b30/.github/pull_request_template) to prevent the common mistakes.

Please get rid of the "remove Remove the patch 308-mips32r2_tune.patc" and the "Revert "Remove the patch 308-mips32r2_tune.patch" commits.

I would suggest to split your code into the following commits:

- "lantiq: fix amazonse.dsti"
- "lantiq: add Amazon-SE subtarget"
- "lantiq: enable Netgear DGN1000B image build"
- "lantiq: add Allnet ALL0333CJ support"

Please remove your comments from the code. If they contain important information, add them to the commit message.

Please, include also some short description about the device you are adding support for and how to flash LEDE a image to the commit message of the support commit. Have a look at https://github.com/lede-project/source/commit/3f31029b190417e12f4d19d9999cd52ec76d0f85 for an good example. Please also mention in the commit message what is not (yet) working.

Be prepared for more change requests. But we need to get the commits first into a shape that allows me an easy review.

If you have questions about "how to do this with git" feel free to contact me.

mcmilk:
Hm, the new binaries do not work... rebuilding again without removing the 308-mips32r2_tune.patch gives me an image where the kernel is okay, but all other binaries contain some bits'n bytes which cause "Illegal instruction" ... here is a snippet:

The box is running with an image, which was build without the 308'er patch. I have uploaded an new opkg binary via scp under /tmp ... the size of both banries are the same, but /bin/opkg will run, /tmp/opkg not:

```
root@LEDE:/tmp# scp mcmilk@192.168.1.2:opkg ./opkg
mcmilk@192.168.1.2's password: 
opkg                                          100%  110KB 109.6KB/s   00:01    
root@LEDE:/tmp# chmod +x opkg 
root@LEDE:/tmp# ls -l /bin/opkg 
-rwxr-xr-x    1 root     root        112261 Feb 24 08:49 /bin/opkg
root@LEDE:/tmp# ls -l /tmp/opkg 
-rwxr-xr-x    1 root     root        112261 Feb 24 08:53 /tmp/opkg
root@LEDE:/tmp# cmp /bin/opkg /tmp/opkg
/bin/opkg /tmp/opkg differ: char 28, line 1
root@LEDE:/tmp# /tmp/opkg
Illegal instruction
root@LEDE:/tmp# 
```

mcmilk:
What commands should I use for generating more info about the wrong instructions used?

Why is the cflags handling done with patches instead of configuration files. It would make much more sense, when the compiler options for some target arch can be specified within the "target.mk" files... some option like DEFAULT_CFLAGS="..." - just like DEFAULT_PACKAGES

nbd168:
On the device, run 'ulimit -c unlimited', then run the opkg binary that doesn't work.
When it segfaults, it should create a core file in /tmp. Please send that one to me.

mcmilk:
I have uploaded this core file here: https://mcmilk.de/tmp/opkg.1487926362.991.4.core

Thank you for helping with that issue :+1:

nbd168:
Please also send me staging_dir/target-*/root-*/bin/opkg
Thanks

mcmilk:
I have uploaded this file here: https://mcmilk.de/tmp/opkg

nbd168:
Please try setting CPU_TYPE to mips32, and change the kernel config to use MIPS32 Release 1 instead of Release 2

mcmilk:
Oh, I have found a solution.
When leaving CPU_TYPE:=4kec, and adding that "new" cpu type to include/target.mk like this: CPU_CFLAGS_4kec = -march=4kec, then the patch doesn't apply and the binaries are fully optimized to 4kec, as it should be. They run fine, just tested before 3 minutes... but the DSL MEI stuff is not working properly, these are the only ones, which get -mtune=34kc now, I don't know why... see updated logfiles:

I made a make clean + make menuconfig + make
- stderr: https://mcmilk.de/tmp/stderr.txt (200KB)
- stdout: https://mcmilk.de/tmp/stdout.txt (3,7MB)

nbd168:
I just did a comparison, there seems to be no noticeable differences in generated code between -mtune=mips32 and -mtune=4kec.
Because of that, please use mips32 for CPU_TYPE. If you also change the kernel config to use MIPS32 Release 1, builds should work properly and it should not use -mtune=34kc anymore.

mcmilk:
I made again an make clean + make menuconfig + make with CPU_TYPE:=mips32 ... it fails on busybox, I have no idea why ?

* logfile: https://mcmilk.de/tmp/mips32-log.txt (1MB)
* CPU_TYPE:=4kec works, but on the "ltq-adsl-mei-ase" package it uses "-march=mips32r2 -mtune=34kc -Wa,-mips32r2" ...
* logfile with cpu type 4kec: https://mcmilk.de/tmp/log.txt.full.gz (1MB) .. it fails in the end, cause I missed the allnet dts file ... the build was fine in the end, the image worked (but not the dsl modem mei stuff)

nbd168:
Did you make the kernel config change regarding the MIPS CPU type?

mcmilk:
No, I have just used "make clean; make menuconfig; make" as said...

nbd168:
Please make the change, it should work then

mcmilk:
What do you mean?

nbd168:
Run make kernel_menuconfig CONFIG_TARGET=subtarget
Change the CPU type under "CPU selecton" to "MIPS32 Release 1"

mcmilk:
Ah, okay, I make this ... but "MIPS32 Release 2" should be fine I think... but will try and comment...

PR #878, opened 2017-02-24
ramips: include only kmod-mt76-core in EX2700 image

jclehner:
This patch frees up flash space on the EX2700, by
removing unused mt76 drivers and firmware.


blogic:
merged, Thanks !

PR #879, opened 2017-02-25
dnsmasq: remove check for existing dhcp server

kdarbyshirebryant:
Remove check for existing dhcp server on interface.  This check could
already be overridden by using 'force' parameter so this change
forces 'force'.

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77test4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq/test-releases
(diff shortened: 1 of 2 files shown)


bobafetthotmail:
reposting here what I posted in the discussion that lead to split this:

All times I left DHCP on in LAN interface when connecting it to a lan where there was already a DHCP server of a device with stock firmware or dd-wrt, the openwrt/lede device's DHCP remained active, and I experienced the expected breakage for that setup.
I never thought that there was a check about that, are you sure the current check works reliably or at all?
Or what I experienced is unrelated?

kdarbyshirebryant:
Who knows!

blogic:
why ? this check allows booting a router inside a lan where another dhcp server is present. please explain why you want to remove the check

DazzyWalkman:
This feature is quite handy when I test x86 LEDE builds in Virtualbox, and LAN is a host-only adapter where Virtualbox manages VMs IP assignment. I personally prefer modifying config options to avoid possible confusion in the production environment.

blogic:
@kdarbyshirebryant why do you want to remove this check ?

kdarbyshirebryant:
It was borne out of an IRC discussion with jow where he expressed a desire to remove it as it was 'inconsistent with other daemon startup scripts' in that it's the only one that does a service 'sanity' check before starting.  I suspect there's also some motivation to simplify the startup script as it is getting close to being self aware :-) I did the change and offered it up.  However I can see there is a use case for this and it'll probably cause more harm than good.

nbd168:
I'd like to keep this check. It's a pretty common scenario to configure a router by plugging it into your LAN. Having it mess up DHCP leases for other clients by *default* seems like a bad idea.

PR #880, opened 2017-02-25
dnsmasq: make bind-dynamic 'non-wildcard' interfaces default

kdarbyshirebryant:
'non-wildcard' interfaces enables dnsmasq's '--bind-dynamic' mode.  This
binds to interfaces rather than wildcard addresses *and* keeps track of
interface comings/goings via a unique Linux api.

Quoting dnsmasq's author "bind-dynamic (bind individual addresses, keep
up with changes in interface config) ... On linux, there's actually no
sane reason not to use --bind-dynamic, and it's only not the default for
historical reasons."

Rename the 'nonwildcard' option to 'binddynamic' as this is more
descriptive of what dnsmasq is doing underneath.

Let's change history, well on LEDE at least, and change the default!

Companion luci PR https://github.com/openwrt/luci/pull/1040

The advantages of doing this is that multiple dnsmasq instances are easier
to implement if required and interface labels are carried through for the various
tagging options that may be used.

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77test4
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq/test-releases
--- a/package/network/services/dnsmasq/files/dhcp.conf
+++ b/package/network/services/dnsmasq/files/dhcp.conf
@@ -15,7 +15,7 @@ config dnsmasq
 	option leasefile	'/tmp/dhcp.leases'
 	option resolvfile	'/tmp/resolv.conf.auto'
 	#list server		'/mycompany.local/1.2.3.4'
-	#option nonwildcard	1
+	option binddynamic	1 # bind to & keep track of interfaces
 	#list interface		br-lan
 	#list notinterface	lo
 	#list bogusnxdomain     '64.94.110.11'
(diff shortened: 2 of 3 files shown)


EricLuehrsen:
Agree. Its the right thing to do for future growth. It should be transparent function to nearly all users.

dedeckeh:
No objection against enabling bind-dynamic by default but I wonder if it's really necessary to replace the uci parameter wildcard by binddynamic. From experience I know it causes quite some hassle to explain people an uci parameter has been renamed and documentation has to be updated as well.
I prefer to keep the existing name unless there's a very good reason to change the name.

kdarbyshirebryant:
Hi Hans,

Thanks for the review, appreciated.  If I may put a little more paint on this bikeshed :-)  A couple of things. 1) 'nonwildcard' doesn't obviously map to the underlying option used with dnsmasq (not that it's particularly obvious even from the manpage as to the advantages of binddynamic) .  2) People doing upgrades will not inherit the new default as 'nonwildcard' is set false by default in /etc/config/dhcp whereas 'binddynamic' can be set true by default/absence - it's an attempt to make the new default, really the new default.

There's a reason for not doing it...it requires an update to luci...which then also affects compatibility with OpenWrt.

I'll happily do whatever you'd like - even update doc if you can point me at where it should be.

dedeckeh:
Hi Kevin,

I agree the nonwildcard naming was poorly chosen but related to upgrades I'm missing the point.
The dhcp.conf file has nonwildcard commented out meaning indeed bind-dynamic is by default disabled at the moment however if the implementation would be changed to enable it by default (as currently done for binddynamic in the code) people doing an upgrade will have it enabled  I assume ?

Related to the Luci change is this change of behavior acceptable for OpenWRT or does it not matter ?

For the documentation I was referring to https://lede-project.org/docs/user-guide/dhcp_configuration which should be updated.

As this uci option is not widely used by people the hassle by the name change will be minimal; therefore I won't reject it but I'm always cautious when doing such changes.

kdarbyshirebryant:
I'm going to close this and re-visit when/if OpenWrt/LEDE merge.  I think it too complicated to change variables and/or defaults cross platform/luci boundaries.

PR #881, opened 2017-02-25
17.01: backport trunk: busybox: update to 1.26.2

bladeoner:
busybox: update to 1.26.2

Refresh patches, delete patches that have been applied upstream.

Signed-off-by: Magnus Kroken <mkroken@gmail.com>
Tested-by: Koen Vandeputte <koen.vandeputte@ncentric.com>
Signed-off-by: Felix Fietkau <nbd@nbd.name> [fix defaults]

Compiled and performed run test on ar71xx (netgear wndr4300) with LEDE 17.01 and busybox 1.26.2.


nbd168:
Why do you want to backport this? Just for fun, or do you have any specific issue in mind?

bladeoner:
@nbd168 I want to backport this because I saw there are a lot of fixes in 1.26.2.
Most of them are backported to 1.25 but it also makes it easier to go to a new version if they release that if we use the latest version. Maybe we can also reduce the amount of patches applied now.

I agree with you that 1.25.1 also is stable.

It's also that I was busy trying to get 1.26.1 and 1.26.2 working because I experienced some DHCP issues with the trunk and I wanted to know if it could be busybox, so I was glad someone got it working.

mkresin:
Nothing in the written above convinces me that we should this version bump. 17.01 is our stable branch and only bugfix and security related updates should be made to the branch.

Busybox will be updated with the next stable version 17.06.

I'm for rejecting this PR.

PR #883, opened 2017-02-26
x86_64: Add profile for the PC Engines APU2

riptidewave93:
The following patch would add support for the PC Engines APU2 embedded board.

### DIFF
--- a/target/linux/x86/64/profiles/001-PCEngines.mk
+++ b/target/linux/x86/64/profiles/001-PCEngines.mk
@@ -1,0 +1,21 @@
+#
+# Copyright (C) 2017 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+define Profile/APU2
+  NAME:=PC Engines APU2
+  PACKAGES:=beep flashrom libsensors lm-sensors usbutils wpad-mini \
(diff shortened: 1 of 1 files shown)


nbd168:
NACK. The package selection seems rather arbitrary to me. I'd prefer having the default profile work well on the APU2 instead of adding more custom profile hackery.

riptidewave93:
@nbd168 Totally understand, with that said this specific file has been through a few RFCs on the Mailing list [0][1] without objection.

As for making the default profile work, the biggest thing in this change is the apu2 and nct5104d drivers. Would adding these to default be better, or should this be left up to the user to install?

Or is there intent to keep x86_64 as default only? I just ask as other targets have no issues with profiles for "full" board support, but I know x86_64 could easily get out of hand if that is the path that is taken.

[0]: http://lists.infradead.org/pipermail/lede-dev/2016-October/003326.html
[1]: http://lists.infradead.org/pipermail/lede-dev/2016-October/003536.html

nbd168:
Please add the APU2 drivers to the default profile.

Here's my reason for this:
Buildbot can't build multiple normal profiles in one build run, which means it's only going to ever build the default profile anyway. If the default profile lacks these drivers, that's bad for people who want to use the APU2. Adding another build run just for extra images for the APU2 is a waste of precious server resources.

I don't think it's a big deal to add more drivers to the default configuration of x86_64, there's plenty of space for it compared to other targets.

PR #891, opened 2017-02-27
Fix 433: insert_modules: Ignore modprobe's return code

yanosz:
Fix for https://bugs.lede-project.org/index.php?do=details&task_id=433
Work-around for https://bugs.busybox.net/show_bug.cgi?id=9676

### DIFF
--- a/package/base-files/files/lib/functions.sh
+++ b/package/base-files/files/lib/functions.sh
@@ -158,7 +158,10 @@ insert_modules() {
 		if [ -f /etc/modules.d/$m ]; then
 			sed 's/^[^#]/insmod &/' /etc/modules.d/$m | ash 2>&- || :
 		else
-			modprobe $m
+			# Since busybox modprobe returns 0 on loading non-existing
+			# and 255 on loading already loaded modules
+			# its return-code is not useful        
+			modprobe $m || true 
 		fi
 	done
 }


mkresin:
Your commit has formal issues. Please have a look at [draft version of the github submission guidelines](https://github.com/bobafetthotmail/source/blob/68913cbafbd7b16f9819c5b29db3cda307a79b30/.github/pull_request_template) and the [generic submission guidelines](https://lede-project.org/docs/guide-developer/the-source-code#submitting_patches).

- no signed of by line
- no commit message
- wrong/missing prefix in commit title

Please have a look at the LEDE commit history to have some examples on how to do it the right way.

yanosz:
hmm.. hope hope its ok, now

package/base-files/files/lib/functions.sh
```diff
@@ -158,7 +158,10 @@ insert_modules() {
 		if [ -f /etc/modules.d/$m ]; then
 			sed 's/^[^#]/insmod &/' /etc/modules.d/$m | ash 2>&- || :
 		else
-			modprobe $m
+			# Since busybox modprobe returns 0 on loading non-existing
+			# and 255 on loading already loaded modules
+			# its return-code is not useful        
+			modprobe $m || true
```
yousong: better make it `modprobe; [ "$?" = 0 -o "$?" = 255 ]` and adapt the comment accordingly

yanosz: hmm... I don't agree.
This code doesn't ignore modprobe's return code. I think, we should do so, since its bogus: 255 can be an error (if the module doesn't exists) or success if it's already loaded. I

Thus I don't like to make "255" explicit and go for ignore its return code at all.

yousong: ok, got it.  thanks for the efforts ;)

yousong:
Well, I just checked the issue again.  It seems like the modprobe in question here is from the ubox package, not busybox.  We should fix ubox kmodloader instead.

@yanosz , to verify, please check the output of `ls -l $(which modprobe)`

mkresin:
This patch shouldn't be required any longer. The issue was fixed with https://git.lede-project.org/21a4bd04062e16164f4f3e1c67005a062f470b0c, which went into LEDE with https://git.lede-project.org/c8f7031ba76824642035ed0525fd640e00f58926.

Beside of that, no response from the PR author.

PR #892, opened 2017-02-27
Afoundry EW-1200 support

Fanfwe:
Signed-off-by: Francois Goudal <francois@goudal.net>

This adds support for the Afoundry EW-1200 router.
I made a PR to OpenWRT few weeks ago but didn't get any response, so I just decided to switch to LEDE and ported my changed to the LEDE tree.
I have tested it on my own device.
Both 2.4 and 5G radios work, USB port works, Reset button works, RUN LED work.
The USB LED worked on my port to OpenWRT but doesn't work on LEDE, most likely due to the transition from usbdev to usbport trigger. I am pushing this anyway as it doesn't prevent router operation, and can be fixed later.
The device also has "2.4G" and "5G" LEDs but those are not GPIOs of the MT7621 chip. They are controlled by the radio chips and the mt76 driver doesn't currently expose them as GPIOs.

The router runs stable and works fine, except for those two LEDs which are not really important.

The stock firmware is in fact based on some openwrt barrier breaker, with a mediatek SDK kernel, and an afoundry custom made web interface (not LuCI based).
Firmware update page on the stock web interface can not accept sysupgrade images.
At this point, the only working solution I found was to connect to the serial console port (available on J4 header) and to use opkg to install dropbear. Then use dropbear to scp the sysupgrade file and run sysupgrade from console without preserving configuration files.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -19,6 +19,7 @@ board_config_update
 case $board in
 11acnas|\
 all0239-3g|\
+ew1200|\
 hw550-3g|\
 mofi3500-3gn|\
 sap-g3200u3|\
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -88,6 +88,7 @@ get_status_led() {
 		status_led="$board:blue:wlan"
 		;;
 	atp-52b|\
+	ew1200|\
 	ip2202)
 		status_led="$board:green:run"
 		;;
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -63,6 +63,14 @@ define Device/dir-860l-b1
 endef
 TARGET_DEVICES += dir-860l-b1
 
+define Device/ew1200
+  DTS := EW1200
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
+  DEVICE_TITLE := AFOUNDRY EW1200
+  DEVICE_PACKAGES := kmod-usb3 kmod-usb-ledtrig-usbport kmod-ata-core kmod-ata-ahci
+endef
+TARGET_DEVICES += ew1200
+
 define Device/firewrt
   DTS := FIREWRT
   IMAGE_SIZE := $(ralink_default_fw_size_16M)
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -78,6 +78,7 @@ ramips_setup_interfaces()
 	dir-320-b1|\
 	dir-610-a1|\
 	dir-615-h1|\
+	ew1200|\
 	firewrt|\
 	hc5661a|\
 	hlk-rm04|\
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -184,6 +184,9 @@ ramips_board_detect() {
 	*"ESR-9753")
 		name="esr-9753"
 		;;
+	*"EW1200")
+		name="ew1200"
+		;;
 	*"EX2700")
 		name="ex2700";
 		;;
--- a/target/linux/ramips/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/base-files/lib/upgrade/platform.sh
@@ -53,6 +53,7 @@ platform_check_image() {
 	dwr-512-b|\
 	e1700|\
 	esr-9753|\
+	ew1200|\
 	ex2700|\
 	f7c027|\
 	firewrt|\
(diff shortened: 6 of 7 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -18,6 +18,7 @@ board_config_update
 
 case $board in
 11acnas|\
+afoundry-ew1200|\
```
mkresin: PLease remove the ```afoundry-``` prefix from all names (here, the model name in the dts, the led names and so on).

target/linux/ramips/dts/AFOUNDRY-EW1200.dts
```diff
@@ -0,0 +1,126 @@
+/dts-v1/;
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/input/input.h>
```
mkresin: Please include  <[dt-bindings/gpio/gpio.h](http://lxr.free-electrons.com/source/include/dt-bindings/gpio/gpio.h?v=4.4)> here as well and keep alphabetical order..

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -87,6 +87,7 @@ get_status_led() {
 	ht-tm02)
 		status_led="$board:blue:wlan"
 		;;
+	afoundry-ew1200|\
```
pepe2k: Please, follow the general rule and keep all boards in alphabetical order here (in this example, you should move whole block above the one starting with `ai-br100`) and in other scripts your patch changes, don't forget about that after you change the board name (we don't use vendor name as prefix in board name).

target/linux/ramips/image/mt7621.mk
```diff
@@ -63,6 +63,14 @@ define Device/dir-860l-b1
 endef
 TARGET_DEVICES += dir-860l-b1
 
+define Device/afoundry-ew1200
+  DTS := AFOUNDRY-EW1200
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
+  DEVICE_TITLE := AFOUNDRY EW1200
+  DEVICE_PACKAGES := kmod-usb3 kmod-ledtrig-usbport kmod-ata-core kmod-ata-ahci
```
pepe2k: `kmod-ledtrig-usbdev` is obsolete, please use `kmod-usb-ledtrig-usbport` instead and check again if the USB LED works on your board with this package.

target/linux/ramips/base-files/etc/board.d/01_leds
Fanfwe: I based my porting on the ZBT WG2626 as it is the closest device to the one I have in terms of specs. And for that device, board name is prefixed with zbt- everywhere, which is why I prefixed my board with afoundry-
Regardless, I got the point and will amend my commit to remove the prefixes everywhere.

target/linux/ramips/image/mt7621.mk
Fanfwe: My patch already includes kmod-ledtrig-usbport, not -usbdev. I did change that between my OpenWRT PR and this PR, already.
I used the same LED GPIO definition in the LEDE port than the one I used in the OpenWRT port in the DTS, so I know for sure I'm using the right GPIO pins.

target/linux/ramips/dts/AFOUNDRY-EW1200.dts
Fanfwe: Sure, will do.

target/linux/ramips/image/mt7621.mk
pepe2k: Ops, I see my copy&paste mistake, sorry for that!
But anyway, you should us **kmod-usb-ledtrig-usbport**, not kmod-ledtrig-{usbdev,usbport}.

target/linux/ramips/base-files/etc/board.d/01_leds
pepe2k: Actually, the full name of "ZBT" company is (based on their website) "Shenzhen Zhibotong Electronics Co., Ltd" but more often used name is probably "ZBTlink" or just "ZBT". What is important here, all ZBT devices are named "ZBT-something" by the manufacturer, we didn't add "ZBT" as a company prefix in the code.

target/linux/ramips/image/mt7621.mk
Fanfwe: Good catch, I missed the -usb portion. Will update this too, hopefully this will restore LED operation as it was on my previous OpenWRT port.

Fanfwe:
Commit has been amended and pushed, taking into account your comments. Hopefully good to go now.

pepe2k:
> Commit has been amended and pushed, taking into account your comments. Hopefully good to go now.

Please, include `ramips` prefix in commit title (e.g. `ramips: add support for Afoundry EW-1200 support`) and fix line lengths in commit message (max. 75 characters per line). You can read more about this kind of rules here: https://github.com/lede-project/source/pull/592/commits/68913cbafbd7b16f9819c5b29db3cda307a79b30

Thanks!

Fanfwe:
Commit message now complies with guidelines.

pepe2k:
@Fanfwe thanks for all the changes and fixes!
@mkresin LGTM now!

Cheers,
Piotr

mkresin:
Merge, thanks!

PR #893, opened 2017-02-28
mvebu: WRT3200ACM add support for SDIO wlan card

heinzek:
This commit enabled SDIO interface and add wireless driver for internal wireless card.


Noltari:
Merged, thanks!

PR #895, opened 2017-02-28
netifd: remove _6 suffix from 6rd interface names

alinnastac:
This suffix is redundant, 6rd interfaces already use "6rd-" prefix.
Given the maximum length of 15 characters Linux ifnames supports, it
would be best not to waste this limited space with redundancies.

Signed-off-by: Alin Nastac <alin.nastac@gmail.com>


jow-:
Hmm, this will cause the virtual 6rd interface and the underlaying dhcp interface share the same logical name, I do not think that this will properly work as netifd requires interfaces to have distinct names.

To be able to shorten the link name (`6rd-wan_6`) we first need to amend `package/network/ipv6/6rd/files/6rd.sh` to allow overriding the resulting netdev name from a parameter.

alinnastac:
I will propose a different solution.

PR #748, opened 2017-01-28
Iptables: update to 1.6.1

Ansuel:
Refreshed some patches and updated in support of newer kernel (4.9)

From the changelog page:
"This release includes accumulated fixes and enhancements for the
following matches:

* ah
* connlabel
* cgroup
* devgroup
* dst
* icmp6
* ipcomp
* ipv6header
* quota
* set
* socket
* string

and targets:

* CT
* REJECT
* SET
* SNAT
* SNPT,DNPT
* SYNPROXY
* TEE

We also got rid of the very very old MIRROR and SAME targets and the
unclean match, that were removed from the kernel tree long time ago.
We also got patches to update different aspects of our manpages."

With the help of Ralph Sennhauser <ralph.sennhauser@gmail.com> for some modification

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>

PR #896, opened 2017-02-28
Iptables: bump to 1.6.1

Ansuel:
Less modification than old commit (#748) , switch to git repo removed musl patch and refreshed existing patch
(is just some bits bigger because with new version we need 2 package for xtables to work)

Compiled and tested on WRT1900ACS


Ansuel:
@diizzyy can you test this?

diizzyy:
Within a week or so

Please change git:// to http:// and use xz instead of gz.

Ansuel:
@diizzyy fixed

Ansuel:
@diizzyy updated to last commit

nbd168:
I don't see why libnetfilter-conntrack would be required, and libnfnetlink seems non-essential too, since it's used only by the nftables part.
Please try to get rid of this stuff and either switch to an iptables release, or explain why you want to use the latest git version.

Ansuel:
iptables 1.6.1 26KB
iptables 1.4.21 23,1KB
@nbd168 (now i'm trying to compiling with no extra dep)

Ansuel:
@nbd168 connlabel seems to be the problem 
I made some changes and added config.in if someone need to enabled it (don't actually know if we need it, i'm testing) 
I have some problem with the if condition with libxtables (it should include that package if the condition is yes but it doesn't select them.. why? )

Ansuel:
Ok i'm testing it and is working just fine

Ansuel:
@nbd168 this should be good right?

Ansuel:
@diizzyy can you test this?

diizzyy:
Sure, but why aren't we one 1.6.1 or whatever the release is at?

Ansuel:
@diizzyy just to keep things well updated 
(release is at 1.6.1 anyway)

Ansuel:
I changed to relase version (i keep git so we can xz compression if i'm not wrong)

diizzyy:
Nice, compiling... :-)

Ansuel:
@diizzyy did you tested that ?

diizzyy:
My testbox has gone missing... :/ 
I'll try find something else that isn't in use ASAP.

Ansuel:
@nbd168 ok, this is the final version
Compiling and working fine, it's 2 month that i have this and all is working good
I solved problem with extra depends and add 2 configuration so if someone need that feature, he can enabled it. What do you think ?

nbd168:
Merged to my staging tree with a minor cleanup, thanks.

kdarbyshirebryant:
The bcp38 package uses '-m state --state NEW' but unfortunately this iptables doesn't understand the '-m state' switch.

PR #898, opened 2017-02-28
Netfilter package: general update

Ansuel:
Various update package and sync with git repo
Removed included patch

I switch to git (and not follow the release versions) because some package are very old and i think they will never be updated...

Compiled and tested on WRT1900ACS


Ansuel:
@jow- @sbyx @blogic

Ansuel:
@diizzyy can you test this?

Ansuel:
Update to today commit

nbd168:
Are there any particular features/bugfixes that you need from these packages? I have some doubts about simply updating all of them to git versions

Ansuel:
@nbd168 Not so much... some are quite old and they contain bugfix 
What are your doubts about upgrading them? 
It's just to keep all up to date, same size so why not?

jow-:
I'd be fine with the nftables update, for the other components I would prefer released versions as long as there is no particular reason to reference some git version.

Ansuel:
@jow- libnftnl what about this bump ? 
and the removing of the patch with the bump of libnfnetlink

jow-:
nftables + libnft - Git version okay imho, but please switch to http://, not git://
for libnfnetlink - is there an upstream commit resolving the header conflict? If yes I'd prefer to replace the patch with a variant exported from git with format-patch.

Ansuel:
For libnfnetlink can't find the patch but when i upgraded it, the patch failed...
I checked the file that the patch modify and i discover that the 2 include where there... so patch useless

diizzyy:
@Ansuel 
Not within a week or so, lack of time :/

Ansuel:
@diizzyy done

Ansuel:
Package updated

Ansuel:
@diizzyy updated 

tested and compiled well

Ansuel:
Think this way we can include only specific commit and leave what is "not approved" here ahahaha
@diizzyy @jow- @nbd168

LipkeGu:
Will this also fix

```
Thu Apr 27 17:08:32 2017 kern.info kernel: [    1.500012] bridge: automatic filtering via arp/ip/ip6tables has been deprecated. Update your scripts to load br_netfilter if you need this.
```

diizzyy:
Please close this and use releases when possible.
Also, this should probably separate requests.

PR #899, opened 2017-02-28
lantiq: vr9 fxs support: bugfix for usage of VPE with nosmp

kochstefan:
This patch is needed to use FXS if CONFIG_MIPS_MT_SMP
is enabled within kernel configuration and the 'nosmp'
command line argument is given to disable SMP at runtime.

Without this patch CONFIG_MIPS_MT_SMP must be disabled before using FXS.
With this patch setting the 'nosmp' parameter is enough.

In general, concurrent usage of FXS and SMP
is incompatible and will cause kernel panics.

Signed-off-by: Stefan Koch <stefan.koch10@gmail.com>


blogic:
the patch looks correct however the description explains why it is not needed and not what it does. please update the description to explain the content

kochstefan:
I have updated the description now.

blogic:
merged, Thanks !

PR #903, opened 2017-03-01
build: Add make nconfig toplevel target

alive4ever:
Add ``make nconfig`` toplevel target to make use of nconf UI as found on kernel's build system.
lede-nconfig

### DIFF
--- a/include/toplevel.mk
+++ b/include/toplevel.mk
@@ -107,6 +107,9 @@ $(eval $(call rdep,scripts/config,scripts/config/mconf))
 scripts/config/qconf:
 	@$(_SINGLE)$(SUBMAKE) -s -C scripts/config qconf CC="$(HOSTCC_WRAPPER)"
 
+scripts/config/nconf:
+	@$(_SINGLE)$(SUBMAKE) -s -C scripts/config nconf CC="$(HOSTCC_WRAPPER)"
+
 scripts/config/conf:
 	@$(_SINGLE)$(SUBMAKE) -s -C scripts/config conf CC="$(HOSTCC_WRAPPER)"
 
@@ -140,6 +143,12 @@ menuconfig: scripts/config/mconf prepare-tmpinfo FORCE
 		$< Config.in
 
 xconfig: scripts/config/qconf prepare-tmpinfo FORCE
+	if [ \! -e .config -a -e $(HOME)/.openwrt/defconfig ]; then \
+		cp $(HOME)/.openwrt/defconfig .config; \
+	fi
+	$< Config.in
+
+nconfig: scripts/config/nconf prepare-tmpinfo FORCE
 	if [ \! -e .config -a -e $(HOME)/.openwrt/defconfig ]; then \
 		cp $(HOME)/.openwrt/defconfig .config; \
 	fi
--- a/scripts/config/nconf.c
+++ b/scripts/config/nconf.c
@@ -1,0 +1,1561 @@
+/*
+ * Copyright (C) 2008 Nir Tzachar <nir.tzachar@gmail.com?
+ * Released under the terms of the GNU GPL v2.0.
+ *
+ * Derived from menuconfig.
+ *
+ */
+#define _GNU_SOURCE
+#include <string.h>
+#include <stdlib.h>
+
+#include "lkc.h"
+#include "nconf.h"
+#include <ctype.h>
+
+static const char nconf_global_help[] = N_(
+"Help windows\n"
+"------------\n"
+"o  Global help:  Unless in a data entry window, pressing <F1> will give \n"
+"   you the global help window, which you are just reading.\n"
+"\n"
+"o  A short version of the global help is available by pressing <F3>.\n"
+"\n"
+"o  Local help:  To get help related to the current menu entry, use any\n"
+"   of <?> <h>, or if in a data entry window then press <F1>.\n"
+"\n"
+"\n"
+"Menu entries\n"
+"------------\n"
+"This interface lets you select features and parameters for the kernel\n"
+"build.  Kernel features can either be built-in, modularized, or removed.\n"
+"Parameters must be entered as text or decimal or hexadecimal numbers.\n"
+"\n"
+"Menu entries beginning with following braces represent features that\n"
+"  [ ]  can be built in or removed\n"
+"  < >  can be built in, modularized or removed\n"
+"  { }  can be built in or modularized, are selected by another feature\n"
+"  - -  are selected by another feature\n"
+"  XXX  cannot be selected.  Symbol Info <F2> tells you why.\n"
+"*, M or whitespace inside braces means to build in, build as a module\n"
+"or to exclude the feature respectively.\n"
+"\n"
+"To change any of these features, highlight it with the movement keys\n"
+"listed below and press <y> to build it in, <m> to make it a module or\n"
+"<n> to remove it.  You may press the <Space> key to cycle through the\n"
+"available options.\n"
+"\n"
+"A trailing \"--->\" designates a submenu, a trailing \"----\" an\n"
+"empty submenu.\n"
+"\n"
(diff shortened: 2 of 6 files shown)


alive4ever:
Rebased to latest HEAD

alive4ever:
Rebased to latest HEAD.

I really want this to get merged, so `nconfig` ui can be used in addition to the good old `menuconfig`.

Reasons to use `make nconfig` over `make menuconfig`

* Better navigation
Arrow keys are better utilized for navigation. Instead of pressing `Enter` to follow the submenu, `right arrow` can be used. Similarly, instead of choosing `Exit` at every submenu to navigate back to upper menu, `left key` can be used.
`pg-up` and `pg-down` keys can be used to scroll the display on all menus. On `menuconfig`, the `pg-up` and `pg-down` keys cannot be used to scroll the display on `Target System` submenu.

* Better search
Incremental search using `/` key to quickly select the desired menu. The incremental search feature works on all menu display. In comparation, `menuconfig` search doesn't do incremental search selection. It only does symbol search.

alive4ever:
Ping @pepe2k , @diizzyy , @nbd168 , and other reviewers.

diizzyy:
Looks pretty nice! I'm a bit worried though however how well maintained this is upstream. 

As for the search features, does that mean that you can search for packages (and options) directly anywhere?

alive4ever:
> As for the search features, does that mean that you can search for packages (and options) directly anywhere?

Yeah. Try it out
```
git remote add alive4ever https://github.com/alive4ever/source
git fetch alive4ever
git checkout make-nconfig
make nconfig
```

zorun:
This is nice, I used nconf not long ago in another project and it's indeed an improvement over menuconfig.

However your branch does not build on Debian stretch x86_64, see: https://paste.aliens-lyon.fr/X1y
It seems that you are missing some linking flags against ncurses, and there are some strange errors on top.

alive4ever:
On those Debian based distributions, you need to install distributions's `pkg-config` to use `make nconfig`.

This is required to use find and use the host-provided tools and development libraries, since `ncurses` is not part of `hostpkg` on LEDE build system.

zorun:
Ok thanks!  I'm wondering why pkg-config is needed for this, because ncurses was already required before (not sure exactly why), but not pkg-config.

By the way, are you sure the new check is working?  Because even without pkg-config, the check seems to succeed:

```
Checking 'working-make'... ok.
Checking 'case-sensitive-fs'... ok.
...
Checking 'git'... ok.
Checking 'file'... ok.
Checking 'pkg-config'... ok.
Checking 'ldconfig-stub'... ok.
```

Also, the nconfig target fails, but without the error message you added:
```
$ make nconfig
make: scripts/config/nconf: Command not found
/home/build/lede/include/toplevel.mk:145: recipe for target 'nconfig' failed
make: *** [nconfig] Error 127
```

It is only visible with `V=s`, while I think it should be shown without:
```
$ make nconfig V=s
make[1]: Entering directory '/home/build/lede/scripts/config'
You don't have pkg-config installed. Please install pkg-config
For now, you can use 'make menuconfig' instead of 'make nconfig'
make[1]: Leaving directory '/home/build/lede/scripts/config'
if [ \! -e .config -a -e /home/build/.openwrt/defconfig ]; then \
	cp /home/build/.openwrt/defconfig .config; \
fi
scripts/config/nconf Config.in
make: scripts/config/nconf: Command not found
/home/build/lede/include/toplevel.mk:145: recipe for target 'nconfig' failed
make: *** [nconfig] Error 127
```

zorun:
Also, you should add the generate files to `scripts/config/.gitignore`, because git sees them:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	scripts/config/nconf
	scripts/config/nconf_check
```

Other than that, after installing pkg-config and manually cleaning up `scripts/config/*.o`, it works fine, thanks :)

zorun:
`
`
In fact, I have another issue: when switching back to master and trying to use the regular `menuconfig`, it fails:

```
$ make menuconfig V=s
make[1]: Entering directory '/home/build/lede/scripts/config'
/usr/bin/ld: lxdialog/checklist.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lxdialog/util.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lxdialog/inputbox.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lxdialog/textbox.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lxdialog/yesno.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: lxdialog/menubox.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
Makefile:32: recipe for target 'mconf' failed
```

Maybe you don't use the same build options?

alive4ever:
> Other than that, after installing pkg-config and manually cleaning up scripts/config/*.o, it works fine, thanks :)

You should run `make -C ./scripts/config clean` to properly clean up the nconfig leftover objects and binary.

> In fact, I have another issue: when switching back to master and trying to use the regular menuconfig, it fails:

Do you have unusual system configuration, i.e. multiarch or similar?
This works fine for me with pure 64 bit Debian 9.1 container.

Consider cleaning up `./scripts/config` directory using `make -C ./scripts/config clean`.

alive4ever:
For your information, the check of `pkg-config` requirement will place a symbolic link to `/usr/bin/pkg-config` inside `./staging_dir/host/bin` directory. You may remove the symlink if you are going back to the master branch after checking out my branch.

alive4ever:
I shouldn't add pkg-config check during prereq, since the symbolic link of distro's pkg-config will conflict with `pkg-config` built during the build process.

I'll undo the relevant commit to fix the issue.

diizzyy:
Hmm... there are a few interesting quirks

*  Keyboard mapping is a bit weird (using putty)
* F1-F4 can do anything from showing you generic help to exit nconfig (I once got F2 to display package info)
* F5+ seems to work as expected, F8 should probably be mapped to '/'
* Search '/' sometimes work? 
Color scheme needs to be adjusted ;-)

nbd168:
Please rebase on top of the xconfig changes that I merged into my staging tree.

include/toplevel.mk
```diff
@@ -136,6 +141,12 @@ menuconfig: scripts/config/mconf prepare-tmpinfo FORCE
 	[ -L .config ] && export KCONFIG_OVERWRITECONFIG=1; \
 		$< Config.in
 
+nconfig: scripts/config/nconf prepare-tmpinfo FORCE
+	if [ \! -e .config -a -e $(HOME)/.openwrt/defconfig ]; then \
```
pprindeville: Why do you need the `\!` instead of just `!`?

scripts/config/nconf.c
```diff
@@ -0,0 +1,1557 @@
+/*
```
pprindeville: Can we pull `nconf.[hc]` from a URL instead of embedding them directly?

blogic:

alive4ever:
Rebased with the latest codebase. Instead of using buildroot's nconf, use nconf sources from kernel 4.9.75 package.

scripts/config/nconf.c
alive4ever: LEDE/OpenWrt build phases are configuration, source downloads, tools compilation, toolchain compilation, target compilation, and firmware generation.

Adding download during configuration phase doesn't make sense.

PR #904, opened 2017-03-01
build: Add make xconfig toplevel target

alive4ever:
Add ``make xconfig`` toplevel target to make use of Qt-based qconf UI as found on the kernel's build system.
The qconf related files are pulled from linux 4.9.13 archive, which includes support for Qt 5.
lede-xconfig


alive4ever:
I can get ``qconf`` to compile manually, i.e. running ``make qconf`` inside ``scripts/config``.

As long as I call ``make xconfig`` from clean state, ``pkg-config`` on the LEDE build system just can't find the required library.

Any hint to use distro-provided pkg-config instead of pkg-config shell script inside the build system?

alive4ever:
Ok, I can just parse the output of ``whereis -b -S /usr -f pkg-config`` to find distro installed pkg-config.

alive4ever:
Rebased to latest HEAD.

alive4ever:
Rebased to latest HEAD.

Let me know if anybody is interested in xconfig toplevel target, since it's only make sense on desktop systems, which has graphical display.

The xconfig toplevel target doesn't make sense on server without graphical environment set up, although it is possible to use X forwarding feature for this to work - provided that all qt related dependencies are installed properly on the server.

diizzyy:
@alive4ever 
Given the amount of time that has passed I think you can more or less can assume that this will not go into the tree.

nbd168:

PR #905, opened 2017-03-01
ramips: remove kmod-mt76* from EX2700 images

jclehner:
These modules are not needed by the EX2700, since it does not have an external wifi chip (MT7620A is covered by rt2x00).


mkresin:
merged, thanks!

PR #906, opened 2017-03-02
gpio-button-hotplug: leave platform_device.dev.platform_data untouched

HZFrodo:
get platform_data from gpio_keys_button_dev.pdata, and fix a illegal pointer
dereference like this:

[   51.143776] gpio-keys-polled gpio-keys-polled: missing poll_interval value
[   51.150852] gpio-keys-polled: probe of gpio-keys-polled failed with error -22
[  828.159993] gpio-keys-polled gpio-keys-polled: no memory for button data
[  828.166821] gpio-keys-polled: probe of gpio-keys-polled failed with error -12

### DIFF
--- a/package/kernel/gpio-button-hotplug/src/gpio-button-hotplug.c
+++ b/package/kernel/gpio-button-hotplug/src/gpio-button-hotplug.c
@@ -474,7 +474,6 @@ static int gpio_keys_button_probe(struct platform_device *pdev,
 			dev_err(dev, "missing platform data\n");
 			return -EINVAL;
 		}
-		pdev->dev.platform_data = pdata;
 	}
 
 	if (polled && !pdata->poll_interval) {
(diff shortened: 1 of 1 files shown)


lynxis:
Hi @HZFrodo,

thanks for your PR. Could you explain, when the "illegal pointer dereference" happens?

HZFrodo:
Hi @lynxis ,
When I am doing some learning about this kernel module, I modified souce code and compile a new one, so I rmmod gpio-button-hotplug.ko, then try to insmod my modified version, illegal pointer dereference crashs the kernel.
And, maybe gpio-button-hotplug.ko will never be unloaded once be loaded, but we know kernel modules should be reloadable.
Finally, in gpio_keys_polled_probe(), at LINE#603, we get pdata from struct gpio_keys_button_dev.pdata, instead of pdev->dev.platform_data. So, this PR also make the gpio_keys_driver&gpio_keys_polled_driver more consistent.

Thanks.

PR #911, opened 2017-03-03
ramips: kmod-sdhci-mt7620: switch dependence to kmod-mmc

HZFrodo:
mtk-mmc/mtk_sd.ko only depends on mmc_core and mmc_block.
And, we remove kmod-sdhci dependence assignment from all related target devices.

Tested on Onion Omega2+


blogic:
merged, Thanks !

PR #914, opened 2017-03-03
mvebu: enable disk led trigger

Ansuel:
As we have a disk led we enable the disk trigger that work with the easata port


mkresin:
Would you please fix the commit title. It has to be ```mvebu: enable disk led trigger```.

wongsyrone:
@Ansuel 
He means **the commit title** instead of the PR title.

Ansuel:
@mkresin Think this should be commited anyway... we need it anyway on mvebu and we don't have space problem so why not?

PR #917, opened 2017-03-03
lantiq: Add support for Alpha ASL56026

eamaclean:
The ASL56026 is a VDSL2 router with dual 100mbit ethernet,
also known as the ECI B-FOCuS V-2FUb/I.

CPU: Lantiq XRX268 v1.1 at 333MHz
Modem: Lantiq VRX208
RAM: 32MiB DDR2 at 167MHz
Flash: 8MiB NOR, Spansion S29GL064N90TF04

UART is at JP1:
Pin 1 TX
Pin 2 GND
Pin 3 +3.3V
Pin 4 NC
Pin 5 RX

Boot selection pins are exposed via several resistor jumpers:
boot_sel0 is at J15, on the rear of the board. Default is high.
boot_sel1 is at J3, next to the flash - it is also the flash CE# pin. Default is low.
boot_sel2 is at J12, directly below the SoC. Default is low.
boot_sel3 is at J16, on the rear of the board. Default is low.

The boot_sel pins should never be shorted, the jumper must be moved or
a lower value resistor used to change the pull (existing resistors are 4k7, 1k should work)

To install with the stock bootloader you must break the built in image selection process
which uses at least the following vars: f_upgrade_addr, f_upgrade2_addr, loadaddr, kernel_addr, activeregion, committedregion
This is done by setting loadaddr and both f_upgrade_addr vars to the same address:

	VR9 # setenv loadaddr 0xB0040000
	VR9 # setenv f_upgrade_addr 0xB0040000
	VR9 # setenv f_upgrade2_addr 0xB0040000
	VR9 # saveenv

Then flash the firmware image:

	VR9 # tftpboot 0x81000000 lede-lantiq-xrx200-ASL56026-squashfs-sysupgrade.bin
	VR9 # erase B0040000 +${filesize}
	VR9 # cp.b 0x81000000 0xB0040000 ${filesize}


eamaclean:
I'm presently chasing down an ethernet bug and need to make a change to prevent the power LED flashing, will comment again when these issues are fixed.

eamaclean:
Both these issues are now fixed, some documentation has been added to the main commit, and the adjustment of the VG3503J's name has been split out to a second commit.

mkresin:
Thanks a lot for your contribution.

I've merged you commits into [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Would you please review and ACK the changes I've done to the gpio_export node in the DTS!

eamaclean:
Ack, cheers.

PR #918, opened 2017-03-04
base-files: add submission service port

lucize:
prevent postfix start failure: fatal: 0.0.0.0:submission: Unrecognized service


lucize:
I don't know why there are 5 commits to this PR !

regards

bobafetthotmail:
That's because it is using your Master branch. You should do PRs by using a dedicated branch of your forked repository, with only the commits you want to merge.

Considering that it is a single-line change, you can use the Github webinterface too.
You go here https://github.com/lede-project/source/blob/master/package/base-files/files/etc/services and click on the pencil icon on the right (near the | Raw | Blame | History | buttons), then do your change through the webinterface.

mkresin:
04.03.2017 15:12, Alberto Bursi:
> Considering that it is a single-line change, you can use the Github
> webinterface too.
> You go here
> https://github.com/lede-project/source/blob/master/package/base-files/files/etc/services
> and click on the pencil icon on the right (near the | Raw | Blame |
> History | buttons), then do your change through the webinterface.

Don't do it this way! Changes submitted this way are considered as 
untested and will be rejected by me.

bobafetthotmail:
> Don't do it this way! Changes submitted this way are considered as untested and will be rejected by me.

How about asking for logs instead?
Submitting through git on linux does not guarantee that the change was tested either.

lucize:
googled about it and made another PR

regards

PR #920, opened 2017-03-04
base-files: add submission service port

lucize:
prevent postfix start failure fatal: 0.0.0.0:submission: Unrecognized service

### DIFF
--- a/package/base-files/files/etc/services
+++ b/package/base-files/files/etc/services
@@ -76,6 +76,8 @@ afpovertcp	548/tcp
 afpovertcp	548/udp
 nntps		563/tcp		snntp
 nntps		563/udp		snntp
+submission	587/tcp
+submission	587/udp
 ldaps		636/tcp
 ldaps		636/udp
 tinc		655/tcp


stintel:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/stintel/staging.git;a=commit;h=0c0c752f7d7cbf90777ce39e85b68a9949e49a69

stintel:
Sorry, I was too soon. Please correct subject prefix (base-files), and I think we should follow IANA policy to add both TCP and UDP entries, even if the protocol doesn't use UDP.
See https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=587

lucize:
wouldn't it be better to make a diiff with upstream services file and see what's missing, or the size of it does matter and we should add only relevant ports ?

lucize:
renamed and added udp

stintel:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/stintel/staging.git;a=commit;h=8e75efc0fb52d02a0cdc13a5ad819b380de6f3da

PR #922, opened 2017-03-06
kernel: add kmod-leds-pwm

Ham22:
The pistachio target was upstreamed to the linux kernel with a led-pwm compatible string hence can't use the standard kmod-leds-gpio.

Tested on MIPS pistachio target (Ci40)


nbd168:
I think it would be a better idea to simply enable this in the ci40 kernel config

Ham22:
Fair, I shall close this and add it to the config instead.

PR #923, opened 2017-03-06
kernel: tpm: add Infineon i2c support

Ham22:
Based on openwrt commit "kernel: add tpm support" but reworked to match the style of how it's been introduced in LEDE.

Tested on MIPS pistachio target (Ci40)


blogic:
merged, Thanks !

PR #924, opened 2017-03-06
Convert OpenMesh devices to new build commands

ecsv:
@blogic These are the minimal changes to get the OpenMesh device images building with the new image commands. We were briefly talking about that via skype.

Might be easier to integrate them first before switching to ath79

### DIFF
--- a/target/linux/ar71xx/image/generic.mk
+++ b/target/linux/ar71xx/image/generic.mk
@@ -1,3 +1,19 @@
+define Device/OpenMesh
+  FILESYSTEMS := squashfs
+  DEVICE_PACKAGES := om-watchdog
+  KERNEL := kernel-bin | lzma | uImage lzma
+  IMAGES := sysupgrade.bin
+  IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-rootfs | combined-ext-image
+endef
+
+define Device/a60
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh A40/A60
+  DEVICE_PACKAGES += kmod-ath10k ath10k-firmware-qca988x mod-usb-core kmod-usb2
+  BOARD_NAME = A60
+endef
+TARGET_DEVICES += a60
+
 define Device/ap90q
   DEVICE_TITLE := YunCore AP90Q
   BOARDNAME = AP90Q
@@ -196,6 +212,28 @@ define Device/mr16
 endef
 TARGET_DEVICES += mr16
 
+define Device/mr1750
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh MR1750
+  DEVICE_PACKAGES += kmod-ath10k ath10k-firmware-qca988x
+  BOARD_NAME = MR1750
+endef
+TARGET_DEVICES += mr1750
+
+define Device/mr600
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh MR600
+  BOARD_NAME = MR600
+endef
+TARGET_DEVICES += mr600
+
+define Device/mr900
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh MR900
+  BOARD_NAME = MR900
+endef
+TARGET_DEVICES += mr900
+
 define Device/dr531
   DEVICE_TITLE := Wallys DR531
   DEVICE_PACKAGES := kmod-usb-core kmod-usb2
@@ -204,6 +242,28 @@ define Device/dr531
   MTDPARTS = spi0.0:192k(u-boot)ro,64k(u-boot-env),64k(partition-table)ro,7808k(firmware),64k(art)ro
 endef
 TARGET_DEVICES += dr531
+
+define Device/om2p
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh OM2P/OM2P-HS/OM2P-LC
+  BOARD_NAME = OM2P
+endef
+TARGET_DEVICES += om2p
+
+define Device/om5p
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh OM5P/OM5P-AN
+  BOARD_NAME = OM5P
+endef
+TARGET_DEVICES += om5p
+
+define Device/om5pac
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh OM5P-AC
+  DEVICE_PACKAGES += kmod-ath10k ath10k-firmware-qca988x
+  BOARD_NAME = OM5PAC
+endef
+TARGET_DEVICES += om5pac
 
 define Device/wndr3700
   DEVICE_TITLE := NETGEAR WNDR3700
--- a/target/linux/ar71xx/image/legacy-devices.mk
+++ b/target/linux/ar71xx/image/legacy-devices.mk
@@ -1,10 +1,3 @@
-define LegacyDevice/A60
-   DEVICE_TITLE := OpenMesh A40/A60
-   DEVICE_PACKAGES := om-watchdog kmod-ath10k ath10k-firmware-qca988x \
-	mod-usb-core kmod-usb2
-endef
-LEGACY_DEVICES += A60
-
 define LegacyDevice/ALFANX
   DEVICE_TITLE := ALFA Network N2/N5 board
 endef
@@ -479,42 +472,6 @@ define LegacyDevice/WPN824N
   DEVICE_TITLE := NETGEAR WPN824N
 endef
 LEGACY_DEVICES += WPN824N
-
-define LegacyDevice/OM2P
-  DEVICE_TITLE := OpenMesh OM2P/OM2P-HS/OM2P-LC
-  DEVICE_PACKAGES := om-watchdog
-endef
-LEGACY_DEVICES += OM2P
-
-define LegacyDevice/OM5P
-  DEVICE_TITLE := OpenMesh OM5P/OM5P-AN
-  DEVICE_PACKAGES := om-watchdog
-endef
-LEGACY_DEVICES += OM5P
-
-define LegacyDevice/OM5PAC
-  DEVICE_TITLE := OpenMesh OM5P-AC
-  DEVICE_PACKAGES := kmod-ath10k ath10k-firmware-qca988x om-watchdog
-endef
-LEGACY_DEVICES += OM5PAC
-
-define LegacyDevice/MR600
-  DEVICE_TITLE := OpenMesh MR600
-  DEVICE_PACKAGES := om-watchdog
-endef
-LEGACY_DEVICES += MR600
-
-define LegacyDevice/MR900
-  DEVICE_TITLE := OpenMesh MR900
-  DEVICE_PACKAGES := om-watchdog
-endef
-LEGACY_DEVICES += MR900
-
-define LegacyDevice/MR1750
-  DEVICE_TITLE := OpenMesh MR1750
-  DEVICE_PACKAGES := om-watchdog kmod-ath10k ath10k-firmware-qca988x
-endef
-LEGACY_DEVICES += MR1750
 
 define LegacyDevice/ALL0305
   DEVICE_TITLE := Allnet ALL0305
--- a/target/linux/ar71xx/image/legacy.mk
+++ b/target/linux/ar71xx/image/legacy.mk
@@ -813,26 +813,6 @@ define Image/Build/ZyXELNAND
 endef
 
 
-Image/Build/OpenMesh/buildkernel=$(call MkuImageLzma,$(2))
-Image/Build/OpenMesh/initramfs=$(call MkuImageLzma/initramfs,$(2),)
-
-define Image/Build/OpenMesh
-	-sh $(TOPDIR)/scripts/om-fwupgradecfg-gen.sh \
-		"$(4)" \
-		"$(BUILD_DIR)/fwupgrade.cfg-$(4)" \
-		"$(KDIR_TMP)/vmlinux-$(2).uImage" \
-		"$(KDIR)/root.$(1)"
-	-sh $(TOPDIR)/scripts/combined-ext-image.sh \
-		"$(4)" "$(call factoryname,$(1),$(2))" \
-		"$(BUILD_DIR)/fwupgrade.cfg-$(4)" "fwupgrade.cfg" \
-		"$(KDIR_TMP)/vmlinux-$(2).uImage" "kernel" \
-		"$(KDIR)/root.$(1)" "rootfs"
-	if [ -e "$(call factoryname,$(1),$(2))" ]; then \
-		cp "$(call factoryname,$(1),$(2))" "$(call sysupname,$(1),$(2))"; \
-	fi
-endef
-
-
 Image/Build/Zcomax/buildkernel=$(call MkuImageLzma,$(2),$(3) $(4))
 Image/Build/Zcomax/initramfs=$(call MkuImageLzma/initramfs,$(2),$(3) $(4))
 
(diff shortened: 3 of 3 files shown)


target/linux/ar71xx/image/generic.mk
```diff
@@ -196,6 +209,45 @@ define Device/mr16
 endef
 TARGET_DEVICES += mr16
 
+define Device/mr1750
+  DEVICE_TITLE := OpenMesh MR1750
+  DEVICE_PACKAGES := om-watchdog kmod-ath10k ath10k-firmware-qca988x
+  BOARDNAME = MR1750
+  FILESYSTEMS := squashfs
```
mkresin: I might be wrong, but this shouldn't be required

target/linux/ar71xx/image/generic.mk
```diff
@@ -196,6 +209,45 @@ define Device/mr16
...
+  BLOCKSIZE := 4 8 16 64 128 256
```
mkresin: One blocksize and with ```k``` as unit. The blocksize should match the erase block size of the flash chip

target/linux/ar71xx/image/generic.mk
```diff
@@ -196,6 +209,45 @@ define Device/mr16
...
+  KERNEL := kernel-bin | lzma | uImage lzma
```
mkresin: What about moving all the stuff that is identical for all OpenMesh into a ```define Device/OpenMesh``` and include it for each OpenMesh board. PLease check the lantiq image build code (Device/NAND for example)

target/linux/ar71xx/image/generic.mk
```diff
@@ -196,6 +209,45 @@ define Device/mr16
...
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-rootfs | combined-ext-image board=MR1750 rootfs=$$$$@
```
mkresin: ```IMAGE/sysupgrade.bin := ``` should be enough. Same for the factory.bin.

I'm not yet sure what combined-ext-image really does but I'm not that happy that the rootfs is passed by default. Similar is possible with sysupgrade-tar but it is only used for some special cases. sysupgrade-tar without extra param fits in most cases.

It should be possible to use $(BOARDNAME) within the combined-ext-image recipe instead of passing it for each OpenMesh image. Check the existing image build code in include/image-commands.mk for some examples. You can make it even conditional.

target/linux/ar71xx/image/generic.mk
ecsv: in this case it would be possible. but not with OM2P (different flash chips with different erase sizes)

ecsv:
I will drop work on this for now

target/linux/ar71xx/image/generic.mk
ecsv: How would squashfs-only be converted from legacy to new build commands when not with /squashfs + "FILESYSTEMS := squashfs"?

target/linux/ar71xx/image/generic.mk
mkresin: in that case drop this line. i'm quite sure pad-rootfs will to these steps in case nothing else is specified.

target/linux/ar71xx/image/generic.mk
mkresin: Isn't squashfs the default? I mean, it seams to work fine for all the other boards in that file.

But as I tried to indicate by using "should be", I haven't check it. It was more or less an educated guess based on the existing build code.

ecsv: Looks to me like it is fs-types-y by default. And this can be squashfs, ext4, iso, ...

Enabling ext4 without overwriting "FILESYSTEMS := squashfs" will also create a non-working ext4 image for this board or let the build fail. 

My current test setup will simple fail to build when removing the FILESYSTEMS line:

    CONFIG_TARGET_ar71xx=y
    CONFIG_TARGET_ar71xx_generic=y
    CONFIG_TARGET_MULTI_PROFILE=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_a60=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_mr1750=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_mr600=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_mr900=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_om2p=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_om5pac=y
    CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_om5p=y
    CONFIG_KMOD_BATMAN_ADV_BATMAN_V=y
    CONFIG_KMOD_BATMAN_ADV_BLA=y
    CONFIG_KMOD_BATMAN_ADV_DAT=y
    CONFIG_KMOD_BATMAN_ADV_DEBUGFS=y
    CONFIG_KMOD_BATMAN_ADV_MCAST=y
    CONFIG_PACKAGE_kmod-batman-adv=y
    CONFIG_PACKAGE_kmod-crypto-crc32c=y
    CONFIG_PACKAGE_kmod-crypto-hash=y
    CONFIG_PACKAGE_kmod-lib-crc16=y
    CONFIG_PACKAGE_kmod-lib-crc32c=y
    CONFIG_TARGET_EXT4_BLOCKSIZE=4096
    CONFIG_TARGET_EXT4_BLOCKSIZE_4K=y
    CONFIG_TARGET_EXT4_RESERVED_PCT=0
    CONFIG_TARGET_IMAGES_GZIP=y
    CONFIG_TARGET_ROOTFS_EXT4FS=y
    CONFIG_TARGET_ROOTFS_PARTSIZE=256

Here the interesting part of the build:

    /home/sven/tmp/lede/staging_dir/host/bin/padjffs2 /home/sven/tmp/lede/build_dir/target-mips_24kc_musl/linux-ar71xx_generic/tmp/lede-ar71xx-generic-a60-ext4-sysupgrade.bin  64
    padding image to 10150000
    WARNING: Image file /home/sven/tmp/lede/build_dir/target-mips_24kc_musl/linux-ar71xx_generic/tmp/lede-ar71xx-generic-a60-ext4-sysupgrade.bin is too big
    gzip -c -9n /home/sven/tmp/lede/build_dir/target-mips_24kc_musl/linux-ar71xx_generic/tmp/lede-ar71xx-generic-a60-ext4-sysupgrade.bin > /home/sven/tmp/lede/bin/targets/ar71xx/generic/lede-ar71xx-generic-a60-ext4-sysupgrade.bin.gz
    gzip: /home/sven/tmp/lede/build_dir/target-mips_24kc_musl/linux-ar71xx_generic/tmp/lede-ar71xx-generic-a60-ext4-sysupgrade.bin: No such file or directory
    Makefile:113: recipe for target '/home/sven/tmp/lede/bin/targets/ar71xx/generic/lede-ar71xx-generic-a60-ext4-sysupgrade.bin.gz' failed


And when also removing the /squashfs suffix, the resulting image is not usable on the devices.

So in theory, the /squashfs suffix could be removed when the "FILESYSTEM" is kept. But this would also mean that there is a line which can only create valid squashfs image which is not marked that way.

ecsv:
Hm, I can actually keep this open just in case @blogic wants to pick some changes for ath79. But I will not  make the FILESYSTEM related changes

target/linux/ar71xx/image/generic.mk
ecsv: done

target/linux/ar71xx/image/generic.mk
ecsv: done

target/linux/ar71xx/image/generic.mk
ecsv: only the combined-ext-image changes were done

ecsv:
Just to make it clear: The pushed changes are now basically untested

target/linux/ar71xx/image/generic.mk
mkresin: > CONFIG_TARGET_EXT4_BLOCKSIZE=4096
CONFIG_TARGET_EXT4_BLOCKSIZE_4K=y
CONFIG_TARGET_EXT4_RESERVED_PCT=0
CONFIG_TARGET_IMAGES_GZIP=y
CONFIG_TARGET_ROOTFS_EXT4FS=y
CONFIG_TARGET_ROOTFS_PARTSIZE=256

Correct me if I'm wrong, but this is neither is the default config nor the config used for the prebuild LEDE images.

If you have to change the rootfs filesystem type, either something is wrong with the openmesh support in LEDE or you are having some really special kind of use case.

Most likely I miss the point here.

ecsv: No, this was simply a test. I wanted to have more than the default filesystems selected to test what happens when i remove "FILESYSTEMS" and/or the "/squashfs".

The default settings of LEDE work perfectly fine. But non-default settings of LEDE should not cause a build error due to the OpenMesh device image code or create invalid images.

You can use it therefore as an example to see why I prefer to have the FILESYSTEMS line and the /squashfs suffix.

mkresin:
I compile tested with the suggested changes applied on top of your code and at least using the default config as it is in LEDE everything works as expected.

[suggested_changes.patch.txt](https://github.com/lede-project/source/files/830132/suggested_changes.patch.txt)

> Just to make it clear: The pushed changes are now basically untested

Are you going to test the changes on real hardware or are you done with the PR? I've some more remarks regarding factory.bin == sysupgrade.bin and using meta-data for image validation. But no need to waste our time in case you don't plan to work any longer on this.

ecsv:
There was a smaller problem in the changes from yesterday (which made them incompatible to ap51-flash). I have now tested the changes briefly with an OM2P and OM5P-ACv2

mkresin:
Okay, you lost me. You have opened a PR which should not be merged? And you haven't mentioned it anywhere?

Well, I'm really not sure how to proceed. Is it worse the time to add my final comments to get it into a shape that I'm happy to merge it or not?

ecsv:
If this would be fine to merge then it could have been merged. If we have a longer discussion about the FILESYSTEMS line and a longer ping-pong-session then I don't have time for this at the moment. Therefore my comment "I will stop the work on this for now"

target/linux/ar71xx/image/generic.mk
```diff
@@ -205,6 +244,28 @@ define Device/dr531
...
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh OM5P/OM5P-AN
+  BOARD_NAME = OM5P
+endef
+TARGET_DEVICES += om5p
+
+define Device/om5pac
+  $(Device/OpenMesh)
+  DEVICE_TITLE := OpenMesh OM5P-AC
+  DEVICE_PACKAGES := kmod-ath10k ath10k-firmware-qca988x om-watchdog
```
mkresin: Would please fix my typo here. It should be ```DEVICE_PACKAGES +=``` of course.

target/linux/ar71xx/image/generic.mk
```diff
@@ -1,3 +1,20 @@
+define Device/OpenMesh
+  FILESYSTEMS := squashfs
+  DEVICE_PACKAGES := om-watchdog
+  KERNEL := kernel-bin | lzma | uImage lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-rootfs | combined-ext-image
+  IMAGE/factory.bin/squashfs := append-rootfs | pad-rootfs | combined-ext-image
```
mkresin: Drop the factory.bin here. It is identical to the sysupgrade.bin, which means it is waste of build resources on our buildbots and storage resources on our download server.

If someone insists that each board needs a factory.bin in case the sysupgrade.bin can be used for the first install as well, it needs to be fixed in a different way.

target/linux/ar71xx/image/generic.mk
```diff
@@ -1,3 +1,20 @@
+define Device/OpenMesh
+  FILESYSTEMS := squashfs
+  DEVICE_PACKAGES := om-watchdog
+  KERNEL := kernel-bin | lzma | uImage lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-rootfs | combined-ext-image
```
mkresin: As written, in my build test using the default config neither the ```FILESYSTEMS := squashfs``` nor ```/squashfs``` was required. But since ```FILESYSTEMS := squashfs``` isn't set in ```target/linux/ar71xx/Makefile.mk``` as I would expected it, I'm fine to keep this line.

mkresin: Please use as last build step [append-metadata](https://git.lede-project.org/?p=source.git&a=search&h=HEAD&st=grep&s=append-metadata). You can drop the ```platform_check_image_target_openmesh()``` and most likely the  "Combined Extended Image v1" check as well in ```target/linux/ar71xx/base-files/lib/upgrade/openmesh.sh``` afterwards. The metadata (boardname appended to image) is validated against the boardname by ```package/base-files/files/lib/upgrade/fwtool.sh``` during sysupgrade.

If the boardname doesn't match for some reason, you can override the boardname that is used within the metadata using SUPPORTED_DEVICES. Have a look at d2b6bf141662429d542b23e9e8f44245a77710a5 for reference.

It might be worse to add an check for the the [maximum possible image size](https://git.lede-project.org/?p=source.git&a=search&h=HEAD&st=grep&s=check-size).

mkresin:
> If we have a longer discussion about the FILESYSTEMS line and a longer ping-pong-session then I don't have time for this at the moment. 

I guess it is reasonable that I'm not necessarily enthusiastic about PRs which gets opened if the submitter already knows that (s)he doesn't have the time to discuss/fix outlined issues. In such cases I prefer to focus on something different than doing a time consuming review.

target/linux/ar71xx/image/generic.mk
ecsv: Not a typo. The complete line has to be changed to remove the om-watchdog and then change it to +=

target/linux/ar71xx/image/generic.mk
ecsv: @NeoRaider you wanted to have both, any opinions?

ecsv:
Sorry, I thought that blogic will take of it because he was referenced in the initial message and I was talking already about it.

blogic:
@ecsv thanks, i'll merge it into my ath79 tree

target/linux/ar71xx/image/generic.mk
ecsv: append-metadata will not do anything when SUPPORTED_DEVICES is not defined. And defining them will "break" the image (at least it not anymore in the CE01 format). It is maybe the same reason why combined-image are also not created with append-metadata. in the last step. But that is just a wild guess.

I was also thinking about the image size check but it was already done in the ce image script. And this script had access to the actual firmware and kernel sizes at the same time (which is the important factor for the openmesh devices). But maybe the check could be added before the combined-ext-image is called - haven't tested that

mkresin:
I obviously missed that you talked about/before that PR already with blogic. I'll leave it up to him what should be changed and what is fine to merge.

ecsv:
@mkresin yes, happens. I also sometimes skip the cover letter when receiving patches.

diizzyy:
@blogic 
Status on this pull request?

target/linux/ar71xx/image/generic.mk
NeoRaider: Oh, I completely missed this mention. Gluon can well deal without a factory image, please remove the factory.bin.

blogic:

PR #549, opened 2016-11-21
ipq806x: clean up dts files

heinzek:
- Set the pinmux to the values found in the GPL tarballs of the boards.
Remove pinmux which are is not used (like nand pinmux for spi
flash boards).
This allows to use the wan orange led of the C2600 which had a wrong
pinmux before. Might fix buttons or leds of other boards as well.

- Fix the LED color and the ledswitch key code of the C2600. Rename the
ledgnr to ledswitch.

- Add support for indication the boot state using LEDs to the D7800,
R7500 and R7500v2.

- Change GPIO active to readable values in D7800, EA8500, R7500, 
R7500v2 and R7800.

- Add gpio export usb ports on D7800 and R7500v2.

- Add proper "drive strenght" to i2c4_pins and use it for RPM on
C2600, D7800, EA8500, R7500 and R7500v2.

- Remove pcie pinmux from D7800.

- Move pinctrl to correct place in R7800.

Big thanks to @mkresin @dissent1 
Signed-off-by: Henryk Heisig <hyniu@o2.pl>

PR #927, opened 2017-03-07
libpcap: bump to last commit

Ansuel:
Drop of 2 patch pushed upstream, refresh patch and bump to 1.9.0 version.

> Summary for 1.9.0 libpcap release
>   Man page improvements
>   Fix Linux cooked mode userspace filtering (GitHub pull request #429)
>   Fix compilation if IPv6 support not enabled
>   Fix some Linux memory-mapped capture buffer size issues
>   Don't fail if kernel filter can't be set on Linux (GitHub issue
>     #549)
>   Improve sorting of interfaces for pcap_findalldevs()
>   Don't list Linux usbmon devices if usbmon module isn't loaded
>   Report PCAP_ERROR_PERM_DENIED if no permission to open Linux usbmon
>     devices
>   Fix DLT_ type for Solaris IPNET devices
>   Always return an error message for errors finding DAG or Myricom
>     devices
>   If possible, don't require that a device be openable when
>     enumerating them for pcap_findalldevs()
>   Don't put incompletely-initialized addresses in the address list for
>   When finding Myricom devices, update description for regular
>     interfaces that are Myricom devices and handle SNF_FLAGS=0x2(port
>     aggregation enabled)
>   Fix compilation error in DAG support
>   Fix issues with CMake configuration


commodo:
there is a general preference for release tar-balls instead of git repo hashes

especially if the project releases regularly
[ there are some cases where there hasn't been a release tarball in 2+ years, but the git repo has recent stuff ] 

for libpcap, it looks like they release fairly regularly [ every N months or so ]
so, i think it would be preferred to wait for a release tarball ; than to move to a git repo source

Ansuel:
I made this pr as we can drop 2 patch and from the git commit version 1.9.0 looks in the RC stage so...

Ansuel:
Compiled and tested WRT1900ACS

blogic:
is this a release version or just a snapshot file from git ?

Ansuel:
@blogic From the commit looks like a rc to 1.9 version

diizzyy:
Unless you have use-case where you need to upgrade libpcap please wait until 1.9.0 gets a proper release.

blogic:
@Ansuel i have to agree with @diizzyy. please dont bump to a RC unless there is some critical reason to do so.

blogic:
closing, please reopen when rebased ontop of a new stable release.

PR #929, opened 2017-03-08
almond: add support for nutsboard Almond SoM.

wigcheng:
Almond is a System-on-Module for the (IIoT) industrial internet of things
applications, based on TI AM3352, it's already Combined a Ready-To-Go
Walnut Carrier Board.

Specification:

- 600/400/ MHz (CPU/DDR)
- 256 MB of RAM (DDR3)
- 256 MB of NAND FLASH
- 1T1R 2.4 GHz
- Bluetooth v4.2
- Power management support
- 4x 10/100 Mbps Ethernet LAN ports
- 1x 10/100/1000 Mbps Ethernet WAN port
- 2x USB 2.0 Host
- PCIe
- 2x UART for RS232/422/485
- 1x UART for serial console
- 1x CAN bus
- 8x GPIO

SD boot instruction
1. Copy your MLO, u-boot.img, zImage, uEnv.txt and dtb file to the
partition 1 of SD card.
2. Extracting your LEDE tarball file that you compile to the
partition 2 of SD card.

Signed-off-by: Wig C <onlywig@gmail.com>


pepe2k:
Hello @wigcheng,

This is... **huge** and in practice, impossible for review.

So, for now, only general question: why are you adding separated target for this device? As far as I can see, there are already other targets with same CPU, why not then include it there?

BTW. Please, always use full name in Signed-off-by line, no shortcuts, no nicknames, etc.

Cheers,
Piotr

wigcheng:
Hello @pepe2k 

Sorry to scare you.
1. Because NutsBoard not only 1 device in the future, I planing to use different platform for different device.
2. The functions and applications is almost different of the current target "omap", if you want to know the detail, I can list you, I want to avoid conflict with it (ex: the structure of the base-files is very different)

But I still can move to there if you think it's better.
If you think no any concern and I will corrected the Signed-off-by line and send pull request again.

Thank you for your support.

pepe2k:
Hello @wigcheng,

> Sorry to scare you, I will try to reduce content of this commit.

I'm afraid we can't accept your changes in the current shape.

> Because NutsBoard not only 1 device in the future, I planing to use different platform for different device.

You might be wrong about what target and device terms mean in LEDE. We definitely don't want one device per target and if your device can be included in the omap target, it should go there, not to a separate and new target.

> The functions and applications is almost different of the current target "omap", if you want to know the detail, I can list you, I want to avoid conflict with it (ex: the structure of the base-files is very different)

From what I can see after brief review of your changes, you are not using LEDE userspace scripts/tools in the proper way and also including lot of code, scripts and also binary blobs which is also wrong.

As the GitHub is not very often visited by other LEDE developers, I would suggest you to ask about new target and the best way to include your custom device in LEDE on the [official development mailing list](http://lists.infradead.org/mailman/listinfo/lede-dev).

For now, it's NAK from me and if there won't be any comments from others, I will close this PR in few days.

Cheers,
Piotr

wigcheng:
Thank you, this is my first pull request, sorry for much issues, I will improve it, anyway, thanks for your comments.

PR #925, opened 2017-03-07
ar71xx: add support for RB LHG 5nD

f00b4r0:
This patch adds support for the MikroTik RouterBOARD LHG 5
https://routerboard.com/RBLHG-5nD

Specifications:
- SoC: Atheros AR9344 (600MHz)
- RAM: 64MB
- Storage: 16MB NOR SPI flash
- Wireless: builtin AR9344 5GHz, 2x2:2
- Ethernet: 1x100M

Leveraged the existing SPI NOR code. Tested on actual hardware. Sysupgrade works.

PR #931, opened 2017-03-08
ar71xx: add support for MikroTik hAP ac lite

rmounce:
This patch adds support for the MikroTik RouterBOARD hAP ac lite (RB952Ui-5ac2nD). The hAP ac lite is nearly identical to the hAP, with an added QCA9887 5GHz radio.

The 2.4GHz radio ID is also changed in the hAP ac lite. Handling for this case has been added in #925 upon which this patch is based, so this PR depends upon #925.

There has also been some discussion in #925 regarding different target profiles for SPI NOR based RouterBOARDs. As the hAP ac lite also depends upon ath10k and firmware for the QCA9887 it probably deserves its own profile.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -106,6 +106,9 @@ case "$FIRMWARE" in
 	unifiac-pro)
 		ath10kcal_extract "EEPROM" 20480 2116
 		;;
+	rb-952ui-5ac2nd)
+		ath10kcal_from_file "/sys/firmware/routerboot/ext_wlan_data" 20480 2116
+		;;
 	esac
 	;;
 "ath10k/cal-pci-0000:01:00.0.bin")
--- a/target/linux/ar71xx/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ar71xx/base-files/lib/upgrade/platform.sh
@@ -638,6 +638,7 @@ platform_check_image() {
 	rb-750up-r2|\
 	rb-941-2nd|\
 	rb-951ui-2nd|\
+	rb-952ui-5ac2nd|\
 	rb-lhg-5nd|\
 	rb-mapl-2nd)
 		return 0
@@ -690,6 +691,7 @@ platform_pre_upgrade() {
 	rb-750up-r2|\
 	rb-941-2nd|\
 	rb-951ui-2nd|\
+	rb-952ui-5ac2nd|\
 	rb-lhg-5nd|\
 	rb-mapl-2nd)
 		# erase firmware if booted from initramfs
--- a/target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
+++ b/target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
@@ -942,6 +942,7 @@ config ATH79_MACH_RBSPI
 	  MikroTik RouterBOARD mAP lite
 	  MikroTik RouterBOARD hAP lite
 	  MikroTik RouterBOARD hAP
+	  MikroTik RouterBOARD hAP ac lite
 	  MikroTik RouterBOARD hEX PoE lite
 	  MikroTik RouterBOARD hEX lite
 	  MikroTik RouterBOARD LHG 5
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
@@ -4,6 +4,7 @@
  *  - MikroTik RouterBOARD mAP L-2nD
  *  - MikroTik RouterBOARD 941L-2nD
  *  - MikroTik RouterBOARD 951Ui-2nD
+ *  - MikroTik RouterBOARD 952Ui-5ac2nD
  *  - MikroTik RouterBOARD 750UP r2
  *  - MikroTik RouterBOARD 750 r2
  *  - MikroTik RouterBOARD LHG 5nD
@@ -23,6 +24,7 @@
  *  by the Free Software Foundation.
  */
 
+#include <linux/pci.h>
 #include <linux/platform_device.h>
 #include <linux/phy.h>
 #include <linux/routerboot.h>
@@ -47,6 +49,7 @@
 #include "dev-usb.h"
 #include "dev-wmac.h"
 #include "machtypes.h"
+#include "pci.h"
 #include "routerboot.h"
 
 #define RBSPI_KEYS_POLL_INTERVAL 20 /* msecs */
@@ -59,6 +62,7 @@
 #define RBSPI_HAS_SSR		BIT(4)	/* has an SSR on SPI bus 0 */
 #define RBSPI_HAS_POE		BIT(5)
 #define RBSPI_HAS_MDIO1		BIT(6)
+#define RBSPI_HAS_PCI		BIT(7)
 
 #define RB_ROUTERBOOT_OFFSET    0x0000
 #define RB_BIOS_SIZE            0x1000
@@ -490,6 +494,9 @@ static void __init rbspi_peripherals_setup(u32 flags)
 
 	if (flags & RBSPI_HAS_USB)
 		ath79_register_usb();
+
+	if (flags & RBSPI_HAS_PCI)
+		ath79_register_pci();
 }
 
 /*
@@ -597,7 +604,7 @@ static void __init rbhapl_setup(void)
 }
 
 /*
- * The hAP, hEX lite and hEX PoE lite share the same platform
+ * The hAP, hAP ac lite, hEX lite and hEX PoE lite share the same platform
  */
 static void __init rbspi_952_750r2_setup(u32 flags)
 {
@@ -606,8 +613,11 @@ static void __init rbspi_952_750r2_setup(u32 flags)
 
 	rbspi_peripherals_setup(flags);
 
-	/* GMAC1 is HW MAC + 1, WLAN0 MAC IS HW MAC + 5 */
-	rbspi_network_setup(flags, 1, 5, 0);
+	/*
+	 * GMAC1 is HW MAC + 1, WLAN0 MAC IS HW MAC + 5 (hAP),
+	 * WLAN1 MAC IS HW MAC + 6 (hAP ac lite)
+	 */
+	rbspi_network_setup(flags, 1, 5, 6);
 
 	if (flags & RBSPI_HAS_USB)
 		gpio_request_one(RB952_GPIO_USB_POWER,
@@ -628,21 +638,29 @@ static void __init rbspi_952_750r2_setup(u32 flags)
 }
 
 /*
- * Init the hAP hardware (QCA953x).
+ * Init the hAP (ac lite) hardware (QCA953x).
  * The 951Ui-2nD (hAP) has 5 ethernet ports, with ports 2-5 being assigned
  * to LAN on the casing, and port 1 being assigned to "internet" (WAN).
  * Port 1 is connected to PHY4 (the ports are labelled in reverse physical
  * number), so the SoC can be set to connect GMAC0 to PHY4 and GMAC1 to the
  * internal switch for the LAN ports.
  * The device also has USB, PoE output and an SSR used for LED multiplexing.
+ * The 952Ui-5ac2nD (hAP ac lite) is nearly identical to the hAP, it adds a
+ * QCA9887 5GHz radio via PCI and moves 2.4GHz from WLAN0 to WLAN1.
  */
 static void __init rb952_setup(void)
 {
-	u32 flags = RBSPI_HAS_WLAN0 | RBSPI_HAS_WAN4 | RBSPI_HAS_USB |
+	u32 flags = RBSPI_HAS_WAN4 | RBSPI_HAS_USB |
 			RBSPI_HAS_SSR | RBSPI_HAS_POE;
 
 	if (rbspi_platform_setup())
 		return;
+
+	/* differentiate the hAP from the hAP ac lite */
+	if (strstr(mips_get_machine_name(), "952Ui-5ac2nD"))
+		flags |= RBSPI_HAS_WLAN1 | RBSPI_HAS_PCI;
+	else
+		flags |= RBSPI_HAS_WLAN0;
 
 	rbspi_952_750r2_setup(flags);
 }
--- a/target/linux/ar71xx/image/mikrotik.mk
+++ b/target/linux/ar71xx/image/mikrotik.mk
@@ -31,9 +31,15 @@ define Device/rb-nor-flash-16M
   LOADER_TYPE := elf
   KERNEL_INSTALL := 1
   KERNEL := kernel-bin | lzma | loader-kernel
-  SUPPORTED_DEVICES := rb-750-r2 rb-750up-r2 rb-941-2nd rb-951ui-2nd rb-lhg-5nd rb-mapl-2nd
+  SUPPORTED_DEVICES := rb-750-r2 rb-750up-r2 rb-941-2nd rb-951ui-2nd rb-952ui-5ac2nd rb-lhg-5nd rb-mapl-2nd
   IMAGE/sysupgrade.bin = append-kernel | kernel2minor -s 1024 -e | pad-to $$$$(BLOCKSIZE) | \
 	append-rootfs | pad-rootfs | append-metadata | check-size $$$$(IMAGE_SIZE)
 endef
 
-TARGET_DEVICES += rb-nor-flash-16M
+define Device/rb-nor-flash-16M-ac
+  $(Device/rb-nor-flash-16M)
+  DEVICE_TITLE := MikroTik RouterBoard with 16 MB SPI NOR flash and ac
+  DEVICE_PACKAGES += kmod-ath10k ath10k-firmware-qca988x ath10k-firmware-qca9887
+endef
+
+TARGET_DEVICES += rb-nor-flash-16M rb-nor-flash-16M-ac
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -239,7 +239,8 @@ rb-750up-r2)
 rb-941-2nd)
 	ucidef_set_led_timer "user" "USR/ACT" "rb:green:user" "1000" "1000"
 	;;
-rb-951ui-2nd)
+rb-951ui-2nd|\
+rb-952ui-5ac2nd)
 	ucidef_set_led_timer "user" "USER" "rb:green:user" "1000" "1000"
 	ucidef_set_led_netdev "port1" "port1" "rb:green:port1" "eth0"
 	ucidef_set_led_switch "port2" "port2" "rb:green:port2" "switch0" "0x10"
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -142,6 +142,7 @@ ar71xx_setup_interfaces()
 	rb-750up-r2|\
 	rb-951ui-2hnd|\
 	rb-951ui-2nd|\
+	rb-952ui-5ac2nd|\
 	routerstation|\
 	tl-wr710n|\
 	tl-wr720n-v3|\
--- a/target/linux/ar71xx/base-files/etc/board.d/03_gpio_switches
+++ b/target/linux/ar71xx/base-files/etc/board.d/03_gpio_switches
@@ -28,7 +28,8 @@ rb-912uag-5hpnd)
 	ucidef_add_gpio_switch "usb_power_switch" "USB Power Switch" "52" "1"
 	;;
 rb-750up-r2|\
-rb-951ui-2nd)
+rb-951ui-2nd|\
+rb-952ui-5ac2nd)
 	ucidef_add_gpio_switch "usb_power_switch" "USB Power Switch" "45" "1"
 	ucidef_add_gpio_switch "poe_passthrough" "PoE Passthrough" "14"
 	;;
--- a/target/linux/ar71xx/base-files/etc/diag.sh
+++ b/target/linux/ar71xx/base-files/etc/diag.sh
@@ -306,6 +306,7 @@ get_status_led() {
 	rb-912uag-5hpnd|\
 	rb-941-2nd|\
 	rb-951ui-2nd|\
+	rb-952ui-5ac2nd|\
 	rb-lhg-5nd|\
 	rb-mapl-2nd)
 		status_led="rb:green:user"
--- a/target/linux/ar71xx/base-files/lib/ar71xx.sh
+++ b/target/linux/ar71xx/base-files/lib/ar71xx.sh
@@ -905,6 +905,9 @@ ar71xx_board_detect() {
 	*"RouterBOARD 951Ui-2nD")
 		name="rb-951ui-2nd"
 		;;
+	*"RouterBOARD 952Ui-5ac2nD")
+		name="rb-952ui-5ac2nd"
+		;;
 	*"RouterBOARD LHG 5nD")
 		name="rb-lhg-5nd"
 		;;
--- a/target/linux/ar71xx/files/arch/mips/ath79/machtypes.h
+++ b/target/linux/ar71xx/files/arch/mips/ath79/machtypes.h
@@ -176,7 +176,7 @@ enum ath79_mach_type {
 	ATH79_MACH_RB_941,			/* MikroTik RouterBOARD 941-2nD */
 	ATH79_MACH_RB_951G,			/* Mikrotik RouterBOARD 951G */
 	ATH79_MACH_RB_951U,			/* Mikrotik RouterBOARD 951Ui-2HnD */
-	ATH79_MACH_RB_952,			/* MikroTik RouterBOARD 951Ui-2nD */
+	ATH79_MACH_RB_952,			/* MikroTik RouterBOARD 951Ui-2nD / 952Ui-5ac2nD */
 	ATH79_MACH_RB_CAP,			/* Mikrotik RouterBOARD cAP2nD */
 	ATH79_MACH_RB_LHG5,			/* Mikrotik RouterBOARD LHG5 */
 	ATH79_MACH_RB_MAP,			/* Mikrotik RouterBOARD mAP2nD */


target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -106,6 +106,9 @@ case "$FIRMWARE" in
 	unifiac-pro)
 		ath10kcal_extract "EEPROM" 20480 2116
 		;;
+	rb-952ui-5ac2nd)
+		ath10kcal_from_file "/sys/firmware/routerboot/ext_wlan_data" 20480 2116
+		;;
```
f00b4r0: What about the MAC address? Is it properly set?

f00b4r0:
I can't comment on the ath10k bits since I'm not familiar with this part, but the code looks otherwise good to me, with 3 caveats:

- What about the 5GHz mac address, is it correctly set?
- Please add the new device reference at the top of the mach-rbspi.c file
- Likewise please update Kconfig.openwrt help text

I don't think a separate profile is necessary. Plenty of flash space available and other devices can be supported with this base, while keeping the maintenance burden low :)

f00b4r0:
Also the commit you included from #925 isn't the last one, please update :)

target/linux/ar71xx/image/mikrotik.mk
```diff
@@ -26,12 +26,12 @@ TARGET_DEVICES += nand-64m nand-large
 
 define Device/rb-nor-flash-16M
   DEVICE_TITLE := MikroTik RouterBoard with 16 MB SPI NOR flash
-  DEVICE_PACKAGES := rbcfg rssileds
+  DEVICE_PACKAGES := rbcfg rssileds ath10k-firmware-qca9887 kmod-ath10k
```
pepe2k: So, again back to roots, I have here same question as in https://github.com/lede-project/source/pull/925#discussion_r104659643: do all below boards need both the ath10k firmware and the driver? Well, we all know the answer.

Argument for keeping `rssileds` package for all these boards was, more or less, like _it's just a few KB and all these boards have 16 MB_ - fine, I agree with this after considering all pros and cons. But this time it's not _just a few KB_, I simply don't like this approach.

@blogic, @mkresin, @nbd168, what do you think about this?

@rmounce, please rebase on current master and don't include in your PR commits which are not yet in the repository.

f00b4r0:
> Le 8 mars 2017 à 19:53, Piotr Dymacz <notifications@github.com> a écrit :
> 
> @pepe2k commented on this pull request.
> 
> In target/linux/ar71xx/image/mikrotik.mk:
> 
> > @@ -26,12 +26,12 @@ TARGET_DEVICES += nand-64m nand-large
>  
>  define Device/rb-nor-flash-16M
>    DEVICE_TITLE := MikroTik RouterBoard with 16 MB SPI NOR flash
> -  DEVICE_PACKAGES := rbcfg rssileds
> +  DEVICE_PACKAGES := rbcfg rssileds ath10k-firmware-qca9887 kmod-ath10k
> So, again back to roots, I have here same question as in #925 (comment): do all below boards need both the ath10k firmware and the driver? Well, we all know the answer.
> 
> Argument for keeping rssileds package for all these boards was, more or less, like it's just a few KB and all these boards have 16 MB - fine, I agree with this after considering all pros and cons. But this time it's not just a few KB, I simply don't like this approach.
> 
I am in favor of this approach.
Please explain what is(are?) the problem(s?) with it.
> @blogic, @mkresin, @nbd168, what do you think about this?
> 
> @rmounce, please rebase on current master and don't include in your PR commits which are not yet in the repository.
> 

He can't do that: he needs code that's in #925 (which underlines how all these boards are one single cohesive platform, btw).

mkresin:
> Please explain what is(are?) the problem(s?) with it. Please be more specific than "I don't like it".

1. Carrying unnecessary packages => waste of flash space
2. As soon as this code needs to be splitted for some reason, it will be pain to get which package is required for which board. Which most likely cause regressions, which will cause bugtickets, which one has to take care of
3. it is done different as in all other targets

This has **not** to be read as NAK. But that is why I don't like this approach. But I do acknowledge that doing it this way, saves resources in terms of build time and storage for images on the download server.

pepe2k:
> I am in favor of this approach. Please explain what is(are?) the problem(s?) with it.

Fully agree with @mkresin here.

Cheers,
Piotr

f00b4r0:
As I'm suspecting you're still worrying about space usage, here's a little more data: two builds of these images.

First build without the extra `ath10k-firmware-qca9887 kmod-ath10k` packages:
```
3551056 Mar  8 20:41 lede-ar71xx-mikrotik-rb-nor-flash-16M-initramfs-kernel.bin
1493836 Mar  8 20:41 lede-ar71xx-mikrotik-rb-nor-flash-16M-kernel.bin
3735790 Mar  8 20:41 lede-ar71xx-mikrotik-rb-nor-flash-16M-squashfs-sysupgrade.bin
```

Second build, exact same code/config, only difference being these two packages added to device definition:
```
3552732 Mar  8 20:44 lede-ar71xx-mikrotik-rb-nor-flash-16M-initramfs-kernel.bin
1493836 Mar  8 20:41 lede-ar71xx-mikrotik-rb-nor-flash-16M-kernel.bin
3735790 Mar  8 20:44 lede-ar71xx-mikrotik-rb-nor-flash-16M-squashfs-sysupgrade.bin
```

That's a 1,676 bytes difference to the initramfs image.

_Notice how, because of the space lost in padding, it makes **NO difference** to the final squashfs image size._

Can we get over this argument now?
Cheers.

f00b4r0:
> Carrying unnecessary packages => waste of flash space

That is not the case, as proven above.

> As soon as this code needs to be splitted for some reason, it will be pain to get which package is required for which board. Which most likely cause regressions, which will cause bugtickets, which one has to take care of

It will _NOT_ need to be split, that's the whole point here. The boards are exactly the same. The hardware is identical, some extra chips are or aren't populated.

> it is done different as in all other targets

It's done differently in other targets for specific reasons. It is done exactly the same for the NAND variant of the mikrotik subtarget though.

Cheers.

pepe2k:
> It will NOT need to be split, that's the whole point here. The boards are exactly the same. The hardware is identical, some extra chips are or aren't populated.

Is MikroTik roadmap publicly available somewhere? How can you say for sure what other, new NOR-based boards will have? I can imagine they will use different ath10k chips which simply means more ath10k firmware packages.

f00b4r0:
> I can imagine they will use different ath10k chips which simply means more ath10k firmware packages.

And we still have plenty of space to waste before we fill a full 64KB erase-block, if every firmware blob "wastes" 1.5KB. Please take a look at the numbers.

mkresin:
I'm really not interested in a flame ware (again). But 640Kb are enough for everyone. My two cents on making predictions on the future.

But as long as someone else takes care of cleaning it up in case it doesn't work any longer this way, I'm fine if it gets merged.

f00b4r0:
> But as long as someone else takes care of cleaning it up in case it doesn't work any longer this way,

I am ready to step up if that happens.

target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
```diff
@@ -455,28 +522,34 @@ static void __init rbspi_network_setup(u32 flags, int gmac1_offset,
...
-	if (flags & RBSPI_HAS_WLAN)
-		rbspi_wlan_init(wmac_offset);
+	if (flags & RBSPI_HAS_WLAN0)
+		rbspi_wlan_init(0, wmac0_offset);
+
+	if (flags & RBSPI_HAS_WLAN1)
+		rbspi_wlan_init(1, wmac1_offset);
+
+	if (flags & RBSPI_HAS_PCI)
+		ath79_register_pci();
```
f00b4r0: the register_pci() should go in rbspi_peripheral_setup() I think.

pepe2k:
> It's done differently in other targets for specific reasons. It is done exactly the same for the NAND variant of the mikrotik subtarget though.

This is actually not true, we have support for `911g-5hpacd` which is NAND one with ath10k, and as far as I can see, dedicated image doesn't have any of the ath10k stuff included.

Cheers,
Piotr

f00b4r0:
@pepe2k 

> This is actually not true, we have support for 911g-5hpacd which is NAND one with ath10k, and as far as I can see, dedicated image doesn't have any of the ath10k stuff included.

I wonder what kind of "support" we actually have for that particular board: I see mention of it in ar71xx.sh, but the corresponding machine code in mach-rb91x.c does _not_ seem to support that variant.

And if it did, the lack of ath10k binaries would be a bug IMO.

Cheers.

target/linux/ar71xx/base-files/lib/upgrade/platform.sh
```diff
@@ -688,6 +690,7 @@ platform_pre_upgrade() {
 	rb-750up-r2|\
 	rb-941-2nd|\
 	rb-951ui-2nd|\
+	rb-lhg-5nd|\
 	rb-mapl-2nd)
```
f00b4r0: you need to add your board here too

target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
```diff
@@ -931,6 +932,7 @@ config ATH79_MACH_RBSPI
 	  MikroTik RouterBOARD hAP
 	  MikroTik RouterBOARD hEX PoE lite
 	  MikroTik RouterBOARD hEX lite
+	  MikroTik RouterBOARD LHG5
 	  MikroTik RouterBOARD cAP (EXPERIMENTAL)
```
f00b4r0: Add your board here too.

target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
```diff
@@ -6,6 +6,7 @@
  *  - MikroTik RouterBOARD 951Ui-2nD
  *  - MikroTik RouterBOARD 750UP r2
  *  - MikroTik RouterBOARD 750 r2
+ *  - MikroTik RouterBOARD LHG 5nD
  *
```
f00b4r0: Add your board here too.

pepe2k:
@f00b4r0

>> This is actually not true, we have support for 911g-5hpacd which is NAND one with ath10k, and as far as I can see, dedicated image doesn't have any of the ath10k stuff included.

> I wonder what kind of "support" we actually have for that particular board: I see mention of it in ar71xx.sh, but the corresponding machine code in mach-rb91x.c does not seem to support that variant.

I'm not the author of support and I have never used this device, but: https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-rb922.c#L230

> And if it did, the lack of ath10k binaries in the image would be a bug IMO.

So, missing ath10k binaries (still can be installed manually later) in capable device is bug for you, but having ath10k binaries in not capable device (without any way to remove them) is... feature? :)

Cheers,
Piotr

f00b4r0:
> I'm not the author of support and I have never used this device, but: https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-rb922.c#L230

I see. That's why having everything in a single file is clearer. The naming of this file is confusing, even though I see why it's named that way.

> So, missing ath10k binaries (still can be installed manually later) in capable device is bug for you, but having ath10k binaries in not capable device (without any way to remove them) is... feature? :)

Yes. It's practical and pragmatic.

pepe2k:
> Yes. It's practical and pragmatic.
> This is also why we have ppp in all targets, as I understand it. And no way to remove it either.

It was already discussed on IRC, ppp is kind of essential packet for a _router_ device. How can user with some PPPoE Internet access download ppp packages from Internet after a flash... without actually having already ppp packages in image?

Cheers,
Piotr

f00b4r0:
> My suggestions:
> 
> don't include ath10k firmware+driver by default for all boards, or
> separate profile into device with and without 802.11ac hardware

Again, I disagree. It makes no sense. I have given plenty of reasons why the benefits totally outweigh the presumed downsides (which are honestly hard to establish given the above: no space is wasted and I volunteer to do the work if this needs to be split down the road).

Furthermore you seem to assume that people flashing their devices won't be put off if part of the hardware isn't working right away (I can already picture the bug reports such as "wifi not working after flashing LEDE"); and that they will also know what extra packages they need to install.

I think these are very bold assumptions: making life _harder_ for the end user is typically why open source projects struggle to gain foothold. You're going to say that this can be documented: sure, that's _more work_ and it's _less effective_.

My rationale is based on making the _full_ hardware work **out of the box**, which I believe to be a reasonable expectation from the end user's perspective, _and_ reduce maintenance burden, resource usage and fragmentation. In plain words: I just want to make everybody's life easier.

I am not interested in arguing further on this: I believe I have made my point. I will stick to this rationale and rest my case now, somebody must decide.

target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
rmounce: The MAC address is set correctly for both radios. I assume that in the ath10k case this is contained in the EEPROM along with caldata.

rmounce:
I have rebased this patch on the updated branch for #925 and made some fixes suggested in review. Basically #925 needs to be merged and my branch needs another rebase before this can be merged.

Without wanting to get too involved in the single vs. multiple profile debate, I have updated my patch to generate a separate image containing ath10k and qca9887 firmware to reflect what appears to be majority opinion, and also to seek feedback on the style of this split target profile.

I also note that kmod-ath10k pulls in ath10k-firmware-qca988x by default, I'm not sure how best to handle this.

Some comparison between image and most importantly IMO jffs2 partition size
Without 802.11ac support:
3368128 lede-ar71xx-mikrotik-rb-nor-flash-16M-initramfs-kernel.bin
3539203 lede-ar71xx-mikrotik-rb-nor-flash-16M-squashfs-sysupgrade.bin
/dev/mtdblock8           12800       468     12332   4% /overlay

With 802.11ac support (including CONFIG_DRIVER_11AC_SUPPORT, kmod-ath10k and ath10k-firmware-qca9887, explicitly excluding):
3667736 lede-ar71xx-mikrotik-rb-nor-flash-16M-qca9887-initramfs-kernel.bin
3866883 lede-ar71xx-mikrotik-rb-nor-flash-16M-qca9887-squashfs-sysupgrade.bin
/dev/mtdblock8           12480       472     12008   4% /overlay

So by all metrics 802.11ac support costs ~300KB in this case.

rmounce:
Boot log as requested in IRC:
http://pastebin.com/nsxZ4y26

Also, the WLAN LED that was already defined for the "RB952 series" does not appear to function. Perhaps this is populated on hAP/hEX lite?

f00b4r0:
> Without wanting to get too involved in the single vs. multiple profile debate, I have updated my patch to generate a separate image containing ath10k and qca9887 firmware to reflect what appears to be majority opinion,
> 

I don't think it is.

f00b4r0:
> So by all metrics 802.11ac support costs ~300KB in this case.

The kernel has a different size in your comparison. You are comparing two different configs. 300KB is completely over estimated, see my previous comparison.

f00b4r0:
> Also, the WLAN LED that was already defined for the "RB952 series" does not appear to function. Perhaps this is populated on hAP/hEX lite?

does it exist on your board?

Also, I see a several firmware loading errors in your bootlog. Are these expected?

rmounce:
> I don't think it is.

3/4 in this discussion is an apparent majority to me, but again as I said I am only updating the patch as a RFC.

> 300KB seems completely over estimated, see my previous comparison.

I did see your comparison, it appeared to be under estimated so I performed my own. Adding the bare minimum to support the 5GHz radio resulted in a clear change of just over 300KB.

As a sanity check I added ath10k_core.ko, ath10k_pci.ko and the QCA9887 firmware to a .tar.gz which compressed to 348KB.

> Also why do you add CONFIG_DRIVER_11AC_SUPPORT to the second image when you don't add it to your commit. Is it needed?

It is set with kmod-ath10k, I only noted it in the comment for the sake of a full "diff" between the images I compiled.

> does it exist on your board?

No, I do not believe there is a WLAN LED on any of the hAP (ac lite)/hEX (PoE) lite series hence my asking.

> Also, I see several firmware loading errors in your bootlog. Are these expected?

Yes. I see exactly the same on my Archer C7 v2 (probably the most popular ath10k board).

f00b4r0:
> On 09 Mar 2017, at 13:16, Ryan Mounce <notifications@github.com> wrote:
> 
> I don't think it is.
> 
> 3/4 in this discussion is an apparent majority to me, but again as I said I am only updating the patch as a RFC.
> 
> 
I wasn’t aware this was about quantity.
> 300KB seems completely over estimated, see my previous comparison.
> 
> I did see your comparison, it appeared to be under estimated so I performed my own. Adding the bare minimum to support the 5GHz radio resulted in a clear change of just over 300KB.
> 
> As a sanity check I added ath10k_core.ko, ath10k_pci.ko and the QCA9887 firmware to a .tar.gz which compressed to 348KB.
> 
> 
That suggests that your commit does _not_ list all the necessary packages for it to work.
> Also why do you add CONFIG_DRIVER_11AC_SUPPORT to the second image when you don't add it to your commit. Is it needed?
> 
> It is set with kmod-ath10k, I only noted it in the comment for the sake of a full "diff" between the images I compiled.
> 
> does it exist on your board?
> 
> No, I do not believe there is a WLAN LED on any of the hAP (ac lite)/hEX (PoE) lite series hence my asking.
> 
> 
It does on the hAP. That’s why it’s in the code.

Best

rmounce:
> I wasn’t aware this was about quantity.

It's not, but in the absence of anything else... Put another way, you are the only participant this discussion who feels (strongly or otherwise) that ~300KB of firmware should be included in all images for SPI NOR RouterBOARDs.

> That suggests that your commit does _not_ list all the necessary packages for it to work.

How do you figure?

> It does on the hAP. That’s why it’s in the code.

Ok.  As it is the same PCB as the hAP ac lite with different parts populated, where is the WLAN LED located on the hAP?

f00b4r0:
> I wasn’t aware this was about quantity.
> 
> It's not, but in the absence of anything else... Put another way, you are the only participant this discussion who feels (strongly or otherwise) that ~300KB of firmware should be included in all images for SPI NOR RouterBOARDs.
> 

300KB doesn't make sense. The figure that is. It's 20% of the kernel size. I doubt that driver + firmware can be that big. I also don't understand how you get a figure 200x what I got.
> That suggests that your commit does not list all the necessary packages for it to work.
> 
> How do you figure?
> 

Because when I use your make file changes my image size increases by only 1.5KB.
> It does on the hAP. That’s why it’s in the code.
> 
> Ok. As it is the same PCB as the hAP ac lite with different parts populated, where is the WLAN LED located on the hAP?
> 

I don't have that particular hw. The led is listed in Mikrotik's patch and iirc one of the testers reported it working.

rmounce:
> I doubt that driver + firmware can be that big.

Don't be so optimistic :) At a minimum the following (uncompressed) files are added.

```
334144  ath10k_core.ko
 42812  ath10k_pci.ko
234560  firmware-5.bin
```

> Because when I use your make file changes my image size increases by only 1.5KB.

You need a fresh config file when changing image profiles if defconfig has already been run. If you run scripts/diffconfig.sh after a naive change to the -qca9887 image profile in menuconfig it will show as if the new packages have been explicitly deselected. I don't understand all the quirks of the build system, suffice to say that this behaviour is not specific to my commit and diffconfig.sh is your friend :)

> I don't have that particular hw. The led is listed in Mikrotik's patch and iirc one of the testers reported it working.

I think they may have been referring to the user/"USR" LED that can be similarly designated as a WLAN LED in RouterOS or LEDE. Short of confirmation that this is the case I think the WLAN LED configuration should be removed to avoid confusion.

target/linux/ar71xx/image/mikrotik.mk
```diff
@@ -31,9 +31,15 @@ define Device/rb-nor-flash-16M
...
+  SUPPORTED_DEVICES := rb-750-r2 rb-750up-r2 rb-941-2nd rb-951ui-2nd rb-952ui-5ac2nd rb-lhg-5nd rb-mapl-2nd
   IMAGE/sysupgrade.bin = append-kernel | kernel2minor -s 1024 -e | pad-to $$$$(BLOCKSIZE) | \
 	append-rootfs | pad-rootfs | append-metadata | check-size $$$$(IMAGE_SIZE)
 endef
 
-TARGET_DEVICES += rb-nor-flash-16M
+define Device/rb-nor-flash-16M-qca9887
+  $(Device/rb-nor-flash-16M)
+  DEVICE_TITLE := MikroTik RouterBoard with 16 MB SPI NOR flash and QCA9887 radio
+  DEVICE_PACKAGES += ath10k-firmware-qca9887 kmod-ath10k
```
rmounce: ath10k-firmware-qca988x is included by default with kmod-ath10k. IMO it is best for each target to explicitly include its firmware and change this default so that unnecessary firmware is not included in images for non-988x targets.

f00b4r0:
-- 
Envoyé depuis ma Remington portative
> Le 9 mars 2017 à 14:25, Ryan Mounce <notifications@github.com> a écrit :
> 
> I doubt that driver + firmware can be that big.
> 
> Don't be so optimistic :) At a minimum the following (uncompressed) files are added.
> 
> 334144  ath10k_core.ko
>  42812  ath10k_pci.ko
> 234560  firmware-5.bin
What is firmware-5.bin?
> Because when I use your make file changes my image size increases by only 1.5KB.
> 
> You need a fresh config file when changing image profiles if defconfig has already been run. If you run scripts/diffconfig.sh after a naive change to the -qca9887 image profile in menuconfig it will show as if the new packages have been explicitly deselected. I don't understand all the quirks of the build system, suffice to say that this behaviour is not specific to my commit and diffconfig.sh is your friend :)
> 

Ok, thanks for explaining. Can't test now, about to board a plane ;)

Regardless, 300KB is a small amount considering the available flash space, I think.
> I don't have that particular hw. The led is listed in Mikrotik's patch and iirc one of the testers reported it working.
> 
> I think they may have been referring to the user/"USR" LED that can be similarly designated as a WLAN LED in RouterOS or LEDE. Short of confirmation that this is the case I think the WLAN LED configuration should be removed to avoid confusion.
> 

No. They're separate GPIOs. Please look at the code. Please also look at this picture:

https://wiki.openwrt.org/_detail/media/mikrotik/hap1.png?id=toh%3Amikrotik%3Arb941_2nd

rmounce:
> What is firmware-5.bin?

The QCA9887 firmware located within /lib/firmware/ath10k/QCA9887/hw1.0/

> No. They're separate GPIOs. Please look at the code. Please also look at this picture:

I had seen the code, simply skeptical as to how closely MikroTik's patch aligned with reality :)

'LD603' is not populated on the hAP ac lite. Given that this is the more expensive variant it seems unlikely that a single LED has been removed for cost reasons, probably the GPIO itself has been repurposed for some QCA9887-related function? In this case it should definitely not be defined as the WLAN LED on the ac lite, not that it should be anyway.

f00b4r0:
> Le 9 mars 2017 à 15:24, Ryan Mounce <notifications@github.com> a écrit :
> 
> What is firmware-5.bin?
> 
> The QCA9887 firmware located within /lib/firmware/ath10k/QCA9887/hw1.0/
> 
> No. They're separate GPIOs. Please look at the code. Please also look at this picture:
> 
> I had seen the code, simply skeptical as to how closely MikroTik's patch aligned with reality :)
> 

It works on their hardware. I'd say that's a pretty good hint.
> 'LD603' is not populated on the hAP ac lite.
> 

It is populated on the hap, see picture.
> Given that this is the more expensive variant it seems unlikely that a single LED has been removed for cost reasons, probably the GPIO itself has been repurposed for some QCA9887-related function?
> 
Highly unlikely. The QCA9887 is on the pci bus: it doesn't need gpios for setup/control. A more likely explanation is that led output from the qca9887 required extra complication in pcb routing and mikrotik deemed unnecessary.

Mikrotik's strategy regarding cost cutting is rather odd. I wouldn't be surprised if, if you soldered a LED, it worked (as 2.4GHz radio phy LED).

I'll take a look at this when I get a chance but iirc the hap ac uses the same device definition as the hap in mikrotik's code.

If that led isn't populated and the gpio is unused (or used to control the absent LED, as I suspect), I see no reason to remove it from the code: it's the same situation on the hEX. You're welcome to remove it from the 01_leds file for your device however.
> In this case it should definitely not be defined as the WLAN LED on the ac lite,
> 
That's correct. You can update 01_led, although you'll notice you'll end up duplicating a lot of text...

rmounce:
> Mikrotik's strategy regarding cost cutting is rather odd. I wouldn't be surprised if, if you soldered a LED, it worked (as 2.4GHz radio phy LED).

I'll give this a shot tomorrow :)

> A more likely explanation is that led output from the qca9887 required extra complication in pcb routing and mikrotik deemed unnecessary.

I had considered this. "We can't be bothered breaking out a 5GHz LED, so let's remove 2.4GHz so that it's even."

I would have thought that MikroTik's particular client base is generally appreciative of all the LEDs they can get, but oh well. These are the little things that make my favourite brand of quirky Latvian networking equipment what they are.

> I'll take a look at this when I get a chance but iirc the hap ac uses the same device definition as the hap in mikrotik's code.


It does, I guess MikroTik hide extra LEDs in userspace for ac lite and hEX lite.

f00b4r0:
> I had considered this. "We can't be bothered breaking out a 5GHz LED, so let's remove 2.4GHz so that it's even."
> 

Exactly
> I would have thought that MikroTik's particular client base is generally appreciative of all the LEDs they can get, but oh well. These are the little things that make my favourite brand of quirky Latvian networking equipment what they are.
> 
;)
> I'll take a look at this when I get a chance but iirc the hap ac uses the same device definition as the hap in mikrotik's code.
> 
> It does, I guess MikroTik hide extra LEDs in userspace for ac lite and hEX lite.
> 

Very possibly. It's easy to do, especially in their totally locked-down system.

Re splitting images: if that ends up being the decision, I might have figured a nice/effective (user _and_ maintainer friendly) way of doing this. It'll probably go against some untold orthodoxy again but at least I'll try being constructive ;)

Will detail more when I get back in touch with proper connectivity.

rmounce:
> I wouldn't be surprised if, if you soldered a LED, it worked

It does, now there's a small party inside my router :)
https://i.imgur.com/LAEAqka.jpg
https://gfycat.com/DapperGoodnaturedGemsbuck

This being the case, I would be happy to leave the WLAN LED defined as a bonus for those with a soldering iron.

f00b4r0:
> I wouldn't be surprised if, if you soldered a LED, it worked
> 
> It does, now there's a small party inside my router :)
> 

Glad to be proved right ;^)

f00b4r0:
> Re splitting images: if that ends up being the decision, I might have figured a nice/effective (user _and_ maintainer friendly) way of doing this. It'll probably go against some untold orthodoxy again but at least I'll try being constructive ;)
> 
> Will detail more when I get back in touch with proper connectivity.

OK so, I still think this is wrong, for the reasons I highlighted previously. I'll also add that if some specific user needs to squeeze out every single byte of those 16MB of flash, they'll probably know exactly how to do that (either by using the image builder or by building from scratch).

If that opinion isn't deemed reasonable, I suggest we do 2 different images:
 - a "-minimal" one that would essentially be the current rb-spi-nor-16M
 - a "-full" one with all the necessary drivers to get complete hardware operation of _ALL_ supported mikrotik NOR devices.

The rationale being that it still keeps the maintenance burden relatively low, still reduces potential user confusion (by not relying on specific hardware details the end user may not be familiar with, such as chip names) and provide a good idea to the storage-constrained user of how much space they could reclaim by building their own custom image.

TL;DR: the "carefree" user could go for the full image and be sure that everything will work out of the box, while the "conscious" user can start from the base image and add what they need.

My 2c on this matter :)

nbd168:
I think -minimal and -full is not a good idea. It doesn't give users any clear indication of what to flash for which device.

f00b4r0:
OK well. I give up then.

jow-:
Is the discussion around ath10k / rssleds included the only thing holding up this PR?
Personally I'd favor having them included by default but if this is blocking this PR then please by all means pick the support code without the package selection changes at least :)

pepe2k:
@jow- 

> Is the discussion around ath10k / rssileds included the only thing holding up this PR?

I don't think that rssileds is a problem here, but we weren't able to agree (even the [PR author had doubts](https://github.com/lede-project/source/pull/931#issuecomment-285336285)) about ath10k related stuff.

> Personally I'd favor having them included by default but if this is blocking this PR then please by all means pick the support code without the package selection changes at least :)

If we agree about including everything by default, should we then include ath10k related stuff also in images for NAND based MikroTik devices? It's same issue - we have there one supported device with built-in ath10k WLAN, as [I already wrote before](https://github.com/lede-project/source/pull/931#issuecomment-285175406).

Personally, I'd prefer to avoid feeding images with useless (1 of 7 supported devices need it) binary blobs and kernel modules, but if others don't have problem with that and I'm in minority, I'm fine with it as far as we make it as some kind of rule and apply same approach for NAND images.

f00b4r0:
> (1 of 7 supported devices need it)

need it _right now_. More devices that use ath10k can be added (hAP AC, wAP AC...) with little extra work thanks to the code design.

pepe2k:
@f00b4r0

>> (1 of 7 supported devices need it)

> need it right now. More devices that use ath10k can be added (hAP AC, wAP AC...) with little extra work thanks to the code design.

And for me it's good reason to have separated image for devices with and without AC WLAN embedded. But we already had this discussion ;)

rmounce:
@jow- 
> Is the discussion around ath10k / rssileds included the only thing holding up this PR?

This PR is also held up as it depends upon code in #925.

@f00b4r0 
I think we had almost arrived at an agreeable compromise with -minimal and -full, however I agree with nbd that this naming isn't very descriptive.

I have renamed the additional image profile -ac. This includes ath10k, firmware for hAP ac lite (QCA9887) and the potential future devices wAP ac and hAP ac (QCA9880/x, this firmware is unavoidably included by default with ath10k but that is a separate matter).

Both images will boot on all supported "RBSPI" devices, the only difference being 802.11ac support and ~400KB (as this includes multiple firmwares by default).

target/linux/ar71xx/image/mikrotik.mk
```diff
@@ -31,9 +31,15 @@ define Device/rb-nor-flash-16M
...
-  SUPPORTED_DEVICES := rb-750-r2 rb-750up-r2 rb-941-2nd rb-951ui-2nd rb-lhg-5nd rb-mapl-2nd
+  SUPPORTED_DEVICES := rb-750-r2 rb-750up-r2 rb-941-2nd rb-951ui-2nd rb-952ui-5ac2nd rb-lhg-5nd rb-mapl-2nd
   IMAGE/sysupgrade.bin = append-kernel | kernel2minor -s 1024 -e | pad-to $$$$(BLOCKSIZE) | \
 	append-rootfs | pad-rootfs | append-metadata | check-size $$$$(IMAGE_SIZE)
 endef
 
-TARGET_DEVICES += rb-nor-flash-16M
+define Device/rb-nor-flash-16M-ac
+  $(Device/rb-nor-flash-16M)
+  DEVICE_TITLE := MikroTik RouterBoard with 16 MB SPI NOR flash and ac
```
f00b4r0: maybe ..."with 16MB SPI NOR flash and 802.11ac wireless" ?
"ac" might be a bit too stern here.

rmounce:
Agreed, but anything I think a couple of characters past "ac" will be
truncated in menuconfig. "and 802" is even less descriptive :)

Perhaps the "MikroTik RouterBoard" prefix for these image profiles should
be changed to simply "RouterBOARD"

pepe2k:
@rmounce, please rebase on current master.

rmounce:
> please rebase on current master.

@pepe2k consider it rebased

pepe2k:
@rmounce, thank you for your contribution! I have merged your change/s into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

PR #932, opened 2017-03-08
ar71xx: add support to TP-Link TL-WR942N v1

xssa:
TP-Link TL-WR942N v1 is a 2,4 GHz single-band N450 router, based on Qualcomm/Ath
eros
QCA9561.

Specifications:

- 775/650/258 MHz (CPU/DDR/AHB)
- 128 MB of RAM (DDR2)
- 16 MB of FLASH (SPI NOR)
- 3T3R 2.4 GHz
- 5x 10/100 Mbps Ethernet
- 2 USB 2.0 ports
- 11x LED (most are controled by 74HC595), 2x buttons
- UART header on PCB, but serial console is disabled in OEM non beta firmwares and
 corresponding GPIO pins 14 and 15 are assigned to control USB1 and USB2 LEDs by
 production (non-beta) u-boot and firmware.

Currently not working:
1. U-Boot Firmware recovery for both OEM and LEDE images (seems to be a TP-Link fault).
2. USB1 and USB2 LEDs are disabled due they are hardwired to UART RX and TX pins.
3. Serial console in case you install LEDE on top of OEM production image. You
 need to obtain and install factory betas containing u-boot with working UART and then
 install LEDE on top of it to have working UART console in both u-boot and LEDE.

Flash instruction:
1. Download lede-ar71xx-generic-tl-wr942n-v1-squashfs-factory.bin
2. Go to WEB interface and make usual firmware upgrade.

Flash instruction under U-Boot, using UART (only can be done with preinstalled UART
 enabled u-boot!!!):

1. Use "tpl" to stop autobooting and obtain u-boot CLI
2. Setup ip addresses for u-boot and your tftp server
3. tftp 0x81000000 lede-ar71xx-generic-tl-wr942n-v1-sysupgrade.bin
4. erase 0x9f020000 +$FILESIZE
5. cp.b 0x81000000 0x9f020000 $FILESIZE
6. reset


target/linux/ar71xx/base-files/lib/ar71xx.sh
```diff
@@ -963,6 +963,9 @@ ar71xx_board_detect() {
 	*TEW-823DRU)
 		name="tew-823dru"
 		;;
+	*"TL-WR942N v1")
```
pepe2k: Please, follow the general rule and keep all boards in alphabetical order in all scripts.

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr942n-v1.c
```diff
@@ -0,0 +1,252 @@
...
+	},
+	{
+		.desc			= "RFKILL button",
+		.type			= EV_KEY,
+		.code			= KEY_RFKILL,
+		.debounce_interval	= TL_WR942N_V1_KEYS_DEBOUNCE_INTERVAL,
+		.gpio			= TL_WR942N_V1_GPIO_BTN_RFKILL,
+		.active_low		= 1,
+	},
+/*	{
```
pepe2k: Please, drop code which is commented out anyway or include information why it should be here.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -880,6 +880,21 @@ define Device/tl-wr1043nd-v4
...
+  DEVICE_PACKAGES := kmod-usb-core kmod-usb2 kmod-usb-ledtrig-usbport
+  BOARDNAME := TL-WR942N-V1
+  TPLINK_BOARD_NAME := TL-WR942N
+  DEVICE_PROFILE := TLWR942NV1
+  IMAGE_SIZE := 14464k
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade
+  IMAGE/factory.bin := append-rootfs | tplink-safeloader factory
+  MTDPARTS := spi0.0:128k(u-boot)ro,1344k(kernel),13120k(rootfs),64k(product-info)ro,64k(partition-table)ro,64k(soft-version)ro,64k(support-list)ro,640k(config)ro,576k(certificate)ro,64k(usb-config)ro,128k(log)ro,64k(radio-bk)ro,64k(ART)ro,14464k@0x20000(firmware)
```
pepe2k: Please, combine all consecutive not used mtd partitions in one, read-only partition, like it's done e.g. for Archer C59.

target/linux/ar71xx/base-files/etc/board.d/02_network
```diff
@@ -424,7 +424,8 @@ ar71xx_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0@eth0" "1:lan:1" "2:lan:2" "3:lan:3" "4:lan:4"
 		;;
-	tl-wr842n-v2)
+	tl-wr842n-v2|\
+	tl-wr942n-v1)
 		ucidef_set_interfaces_lan_wan "eth1.1" "eth0"
 		ucidef_add_switch "switch0" \
 			"0@eth1" "1:lan:4" "2:lan:1" "3:lan:2" "4:lan:3"
```
xssa: > Just checking, have you tested this, e.g. check if the switch port are correctly mapped to the RJ45 connectors (using swconfig tool or what LuCI shows on switch page)?

Switch ports are checked and mapped correctly accordind to port definition writings on backside of the router, LuCI switch page and LEDs.

target/linux/ar71xx/base-files/lib/ar71xx.sh
xssa: Lines from 965 to 1136 are not in alphabetical order. Can I sort all boards in this range?

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr942n-v1.c
xssa: ok

target/linux/ar71xx/image/tp-link.mk
xssa: will look

target/linux/ar71xx/base-files/etc/board.d/02_network
xssa: There is a piece of uboot Makefile which redefines GPIOs for uart to unpopulated 4 and 5.

`ifeq ($(CFG_BOARD_TYPE), ap151)
ifneq ($(CONFIG_NO_SERIAL),)
        @echo '#define UART_RX5_TX4           1'      >>include/config.h
        @echo '#define CONFIG_942RU_NO_SERIAL           1'      >>include/config.h
else
ifneq ($(filter $(PRODUCT_NAME), WR942NV1 WR942HPV1 ARCHERC59V1 ARCHERC58V1),)
        @echo '#define UART_RX15_TX14           1'      >>include/config.h
else`

There is another piece of code in board956x.c which defines GPIOs 14 and 15 as output and use it for USB1 and USB2 LEDs

`#ifdef CONFIG_942RU_NO_SERIAL
// set GPIO 14~22 as output
    ath_reg_rmw_clear(GPIO_OE_ADDRESS, (1 << 14) | (1 << 15) | (1 << 16) | (1 << 17) | \
                                       (1 << 18) | (1 << 19) | (1 << 20) | (1 << 21) | \
                                       (1 << 22));
#else
// set GPIO 16~22 as output
    ath_reg_rmw_clear(GPIO_OE_ADDRESS, (1 << 16) | (1 << 17) | (1 << 18) | (1 << 19) | \
                                       (1 << 20) | (1 << 21) | (1 << 22));
#endif
`

USB LEDs and UART Tx and Rx pins are hardwired on the board and AFAIK can not be used simultaneously. Is this information useful?

target/linux/ar71xx/image/tp-link.mk
xssa: I would mostly like to be as close as possible to new tp-link "pharos" so called partitioning schema. It is also used during their new multypart image firmware update process and is fully defined in this file for other tplink devices like tl-wr1043 v4, eap120, re450, cpe510/520. Why whould you recomend to use simplified one which is used on C59 only? 
IMHO This simplification will make reverting to factory images much harder if we adopt full support for multypart fw image as some this partitions needs to be owerwritten. As for now reverting by LEDE sysupgrade to oem firmware is possible only to the same fw which was before LEDE if previously dumped oem partition "firmware" is used as sisupgrade image. Attempt to revert with different oem version dump leeds to non fully functional oem WEB interface. So this problem has to be learned and discused.

target/linux/ar71xx/base-files/etc/board.d/02_network
pepe2k: > Switch ports are checked and mapped correctly accordind to port definition writings on backside of the router, LuCI switch page and LEDs.

OK, thanks for confirmation.

> USB LEDs and UART Tx and Rx pins are hardwired on the board and AFAIK can not be used simultaneously. Is this information useful?

You can override what U-Boot configured, later, in kernel (re-assign GPIO14/15 to UART RX/TX signals), this is not a problem at all.

Question here is what is more important for the user: working USB LEDs or serial console and what we should support by default if there is no way to have both things working at the same time. As the USB LEDs are available on front panel I would prefer to support them, as it's done in vendor _not beta version_ firmware.

Are you sure there are no other GPIO signals available on the PCB which we could assign console to? Is some high resolution photo of the PCB of this device available somewhere on the Internet?

target/linux/ar71xx/base-files/lib/ar71xx.sh
pepe2k: > Lines from 965 to 1136 are not in alphabetical order. Can I sort all boards in this range?

I know about broken order, I plan to fix it later, for now please just put your board somewhere under TL-WR941...

target/linux/ar71xx/image/tp-link.mk
pepe2k: > I would mostly like to be as close as possible to new tp-link "pharos" so called partitioning schema. It is also used during their new multypart image firmware update process and is fully defined in this file for other tplink devices like tl-wr1043 v4, eap120, re450, cpe510/520. Why whould you recomend to use simplified one which is used on C59 only?

First of all, all of that additional partitions are marked as **read-only** which means that even if you want, you can't change their content during the upgrade. If it's possible (take this as a general rule), we never touch/change/overwrite vendor additional data, that's the reason why we keep it in read-only marked partitions.

Secondly, if you look into the [tools/firmware-utils/src/tplink-safeloader.c](https://github.com/lede-project/source/blob/master/tools/firmware-utils/src/tplink-safeloader.c#L134), for CPE510 we don't define all partitions in kernel command line, they are combined in `product-info` and `config` partitions.

> IMHO This simplification will make reverting to factory images much harder if we adopt full support for multypart fw image as some this partitions needs to be owerwritten.

Please refer to tplink-safeloader.c, we don't touch other partitions than kernel and rootfs (os-image and file-system) during the upgrade.

> As for now reverting by LEDE sysupgrade to oem firmware is possible only to the same fw which was before LEDE if previously dumped oem partition "firmware" is used as sisupgrade image. Attempt to revert with different oem version dump leeds to non fully functional oem WEB interface. So this problem has to be learned and discused.

Are there any differences in partition layout between different vendor firmware images?

target/linux/ar71xx/base-files/etc/board.d/02_network
bobafetthotmail: > Question here is what is more important for the user: working USB LEDs or serial console and what we should support by default if there is no way to have both things working at the same time.

Can't this be set that if no serial is detected on boot it will switch to controlling USB leds?

pepe2k: > Can't this be set that if no serial is detected on boot it will switch to controlling USB leds?

I don't think so, at least I don't have any idea how we could detect connected serial console adapter.

xssa: Here http://fex.net/#227411653165 is my own photos and untouched flash dump right from the box.
For now there is two in power of two different scenarios:
1) We want support USB Leds with both factory and beta u-boots.
2) We want to have working uart.
Both variants needs that GPIO assignments has to be checked and reassigned during boot if nessesary.
But the best is enemy of good. So I think this has to go to TODO list. And, may be some voting and discussing on forum? For now there is nothing to discuss as there is no LEDE support on public.

pepe2k: > Here http://fex.net/#227411653165 is my own photos and untouched flash dump right from the box.

Any chance for picture of the bottom of PCB?

> For now there is two in power of two different scenarios:
> 
> 1. We want support USB Leds with both factory and beta u-boots.
> 2. We want to have working uart.
> Both variants needs that GPIO assignments has to be checked and reassigned during boot if nessesary.
> But the best is enemy of good. So I think this has to go to TODO list. And, may be some voting and discussing on forum? For now there is nothing to discuss as there is no LEDE support on public.

From what I can see, the UART console connector (JP1) is missing some jumper resistors on TX/RX lines, which means that it's not usable without some additional, soldering work, am I right here? If yes, then for me it should support USB LEDs by default (as it is done in the vendor firmware) and then, eventually, we can think how to solve the console problem.

bobafetthotmail: > I don't think so, at least I don't have any idea how we could detect connected serial console adapter.

How about writing in the serial "press a key to keep this serial connection open" and if no key is received after say 5 seconds it assumes there is no serial attached and switches it off to operate the USB leds?

Something like this is already done for failsafe and for debug level, on serial boot in my LEDE devices I get 
```
Press the [f] key and hit [enter] to enter failsafe mode
Press the [1], [2], [3] or [4] key and hit [enter] to select the debug level
```

pepe2k: @bobafetthotmail if you know how to make this, I would be happy to review your code.

bobafetthotmail: @pepe2k I cannot test this easily with my hardware, but there are ways to free GPIO pins used by UART in older Atheros chips https://wiki.openwrt.org/doc/recipes/turnoff.uart.to.gpio 
Also it seems it is possible to free JTAG pins https://wiki.openwrt.org/doc/recipes/turnoff.jtag.to.gpio
I don't know if that method works with the newer one here, but it's worth trying.

If that works, it's not hard to clone **package/base-files/files/lib/preinit/30_failsafe_wait** and make a script that does what I said above.

pepe2k: @bobafetthotmail my previous comment is still valid.

bobafetthotmail: @pepe2k I know how to read, thanks. I posted in case xssa was interested in that.

xssa: > Any chance for picture of the bottom of PCB?

I did. Look at http://fex.net/#227411653165

target/linux/ar71xx/image/tp-link.mk
xssa: > Are there any differences in partition layout between different vendor firmware images?

Definitely they can as new image parts catalog and new layot is defined in the first 512 bytes of the oem update images. Will try to investigate what they are already change in known updates. :)

xssa: > fwup-ptn os-image base 0x10b35 size 0x10776e
fwup-ptn file-system base 0x1182a3 size 0x978679
fwup-ptn product-info base 0xa9091c size 0x0009f
fwup-ptn partition-table base 0x00800 size 0x00800
fwup-ptn soft-version base 0xa909bb size 0x00015
fwup-ptn support-list base 0xa909d0 size 0x00094
fwup-ptn profile base 0xa90a64 size 0x02658
fwup-ptn default-config base 0xa930bc size 0x0254d

I think we should have at least five last partitions rw as they are always updated during oem updates to have posibility to revert back to any oem firmware, leave partition _product-info_ ro (it contains MAC and PIN) and combine all space between _default-config_ and _ART_ into one another ro partition.

target/linux/ar71xx/base-files/etc/board.d/02_network
xssa: > How about writing in the serial "press a key to keep this serial connection open" and if no key is received after say 5 seconds it assumes there is no serial attached and switches it off to operate the USB leds?
This is meaningless in case of production u-boot installed as RX GPIO is already assigned as OUTPUT by it. This will work with beta u-boots only. I think we should check GPIO 14 and 15 asignments on board initialisation time on kernel side and reassign it to LEDS if it is not.

target/linux/ar71xx/image/tp-link.mk
pepe2k: > I think we should have at least five last partitions rw as they are always updated during oem updates to have posibility to revert back to any oem firmware, leave partition product-info ro (it contains MAC and PIN) and combine all space between default-config and ART into one another ro partition.

NAK, please keep all vendor specific data partitions marked as _read-only_, as it's done for _all_ other similar (using tplink-safeloader) boards.

If, for any reasons, reverting back from LEDE to vendor firmware requires using combination of kernel+rootfs partitions from the same version of the vendor firmware as it was on device before upgrade to LEDE, make it clear in commit message and/or in the Wiki.

target/linux/ar71xx/base-files/etc/board.d/02_network
pepe2k: > This is meaningless in case of production u-boot installed as RX GPIO is already assigned as OUTPUT by it. This will work with beta u-boots only. I think we should check GPIO 14 and 15 asignments on board initialisation time on kernel side and reassign it to LEDS if it is not.

Or make it static (always set GPIO14/15 as regular GPIO), without checking what U-Boot configured earlier.

In future, when we switch ar71xx target to dts, there will/should be easy way to produce two images with different dtb appended to the kernel - one with regular/default GPIO definition (for USB LEDs) and one with UART support on these GPIOs. And that wouldn't require to have any _hacks_ inside board specific code.

xssa: Here is oem bootlog http://pastebin.com/ZBr5YLw6 and current LEDE http://pastebin.com/gvdxQ3V1

target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
```diff
@@ -1622,6 +1622,18 @@ config ATH79_MACH_TL_WR940N_V4
 	select ATH79_DEV_M25P80
 	select ATH79_DEV_WMAC
 
+config ATH79_MACH_TL_WR942N_V1
+	bool "TP-LINK TL-WR942N v1 support"
+	select SOC_QCA956X
+	select ATH79_DEV_AP9X_PCI if PCI
```
pepe2k: This is not needed here, the board you are adding support for doesn't have/use pci/-e. Please, drop it from here and also clean-up mach-*.c file from related header files.

xssa: done

pepe2k:
@xssa, sorry for the delay.

There is still one unsolved issue here: USB LEDs.

You have included `kmod-usb-ledtrig-usbport` package but [my previous comments](https://github.com/lede-project/source/pull/932/commits/4f8b43d2ebf7e0543fce444878688701f2a13e70#r105561898) about making USB LEDs working by default (as the serial console is disabled in _hardware_ anyway and your PCB pictures clearly show that) hasn't been addressed.

Your comment "USB1 and USB2 LEDs are disabled due they are hardwired to UART RX and TX pins." from commit message is misleading. In fact, the UART console header is not hardwired anywhere (missing jumper resistors make it unusable by default) and USB1/2 LEDs are hardwired by default to the SOC GPIOs.

If you want to make it _somehow universal_ (based on what U-Boot configures), you can:

- add both USB1 and UBS2 LEDs definitions into `tl_wr942n_v1_leds_gpio[]`
- inside `mach-tl-wr942n-v1.c`, before `ath79_register_leds_gpio` call, check value of the corresponding QCA956X_GPIO_REG_OUT_FUNCx value and based on that change gpio value in `tl_wr942n_v1_leds_gpio[]` (if USB LEDs GPIOs, 14 and 15 based on your information, are used as regular GPIO, keep them assigned to USB LEDs, if not: change corresponding GPIO numbers for USB LEDs to unpopulated GPIO4/5)

With this approach, users with production U-Boot version would be able to make use of USB LEDs and users with _modified hardware_ and beta version of the U-Boot would be able to use UART console instead.

Let me know if you are able to make this change yourself or I will do it and just ask you for tests on real hardware.

Cheers,
Piotr

pepe2k:
@xssa are you still interested in finishing this one?

Cheers,
Piotr

xssa:
@pepe2k Sorry for delay. I just went back from business trip.

xssa:
> check value of the corresponding QCA956X_GPIO_REG_OUT_FUNCx value and based on that change gpio value
@pepe2k I can't find this definition in kernel sources. Can you show me your version of this check?

pepe2k:
@xssa
>> check value of the corresponding QCA956X_GPIO_REG_OUT_FUNCx value and based on that change gpio value

> @pepe2k I can't find this definition in kernel sources. Can you show me your version of this check?

These registers are added in [621-MIPS-ath79-add-support-for-QCA956x-SoC.patch](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/patches-4.4/621-MIPS-ath79-add-support-for-QCA956x-SoC.patch), you should have them in your build, inside `build_dir/target-mips_24kc_musl/linux-ar71xx_generic/linux-4.4.61/arch/mips/include/asm/mach-ath79/ar71xx_regs.h` and `build_dir/target-mips_24kc_musl/linux-ar71xx_generic/linux-4.4.61/arch/mips/ath79/gpio.c` files.

My idea here is to use them to check what U-Boot wrote there and based on that, setup GPIOs for USB LEDs.

Cheers,
Piotr

xssa:
@pepe2k thats it. USB LEDs checked on hardware and confirmed working well as expected on factory and beta uboots. 
One more thing. Digging on the uboot sources and PCB I found that there is gpio bit 0 on 74hc595 extender wired to RESET pin on GL850G USB HUB. They call it HUB_RESET in uboot sources. It also tracks to USB power controlles but there is missing resistors there. So only USB HUB can be reseted by this GPIO. I write down some code like on other boards. How can we use this? Can we reset eg, hanged 3g/4g usb modem by it? Is there some code examples?
Cheers, Serg

pepe2k:
@xssa,

> @pepe2k thats it. USB LEDs checked on hardware and confirmed working well as expected on factory and beta uboots.

Perfect, thanks for fixing this!

> One more thing. Digging on the uboot sources and PCB I found that there is gpio bit 0 on 74hc595 extender wired to RESET pin on GL850G USB HUB. They call it HUB_RESET in uboot sources. It also tracks to USB power controlles but there is missing resistors there. So only USB HUB can be reseted by this GPIO. I write down some code like on other boards. How can we use this? Can we reset eg, hanged 3g/4g usb modem by it? Is there some code examples?

It looks fine in current version of your patch. To be honest, I'm not sure what you can do with HUB reset, but I suppose it issues all connected devices to re-initialize on the USB bus. Hard to say if that's enough for a hanged USB modem, I suppose it depends on the model and how bad it is.

Anyway, thank you for your first-time contribution! I have finally (sorry for the delay) merged your PR into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary), with some minor fixes.

Cheers,
Piotr

PR #933, opened 2017-03-09
brcm47xx: remove HW_RANDOM config option

neheb:
Was surprised to find /dev/hwrng on my E1000v2 (cross-flashed to E1200v1) after upgrade to LEDE final.

@rmilecki thoughts?


neheb:
closing this since it seems the b43 driver uses it. My unit does not have the driver but no sense in crippling other units.

PR #934, opened 2017-03-09
ar71xx: fix the GigE switch on the RB2011-UiAS not working

nightexcessive:
On the RB2011-UiAS (identified by model name 2011r5), the GigE switch does not work without this patch.

See https://wiki.openwrt.org/toh/mikrotik/rb2011uias for more information.

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-rb2011.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-rb2011.c
@@ -53,6 +53,7 @@
 #define RB2011_FLAG_SFP		BIT(0)
 #define RB2011_FLAG_USB		BIT(1)
 #define RB2011_FLAG_WLAN	BIT(2)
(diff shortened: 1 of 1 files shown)


f00b4r0:
The mikrotik kernel patch suggests that the 2011R5 uses the same ethernet subsystems as found on other 2011-series routers.

The [original patch submitter](https://lists.openwrt.org/pipermail/openwrt-devel/2014-December/029949.html) suggested testing the change would work on other 2011 devices to confirm, doing so would probably be a good idea if possible: they might benefit from this change.

nightexcessive:
I've tested this patch on hardware (an RB2011-UiAS-2HnD-IN) with the OpenWRT kernel. I haven't had a chance to compile or test this exact code yet and I won't until the weekend.

I don't have any other Mikrotik/RB boxen to test it on, unfortunately.

nightexcessive:
I'm actually going to withdraw this; I won't end up being able to test it because I managed to break my RB2011. Sorry to waste your time!

PR #935, opened 2017-03-09
Introduce Pistachio target

Ham22:
The Pistachio target is a MIPS interaptiv based SoC developed by
Imagination Technologies. It includes blocks for i2c, spi, audio,
usb and WiFi.

This also adds the base support for the 'Creator Ci40 (marduk)'
device which uses the Pistachio SoC to create an IoT hub by
including Bluetooth, WiFi and 6lowpan on one board. Additionally 2x
Mikrobus ports are available to expand with further RF technologies
or add sensors. You can find out more here http://creatordev.io.

Note, this commit is just the initial board support hence the
following are not expected to work yet:
 * WiFi
 * Bluetooth
 * 6lowpan
 * Audio
 * Mikrobus uarts, user leds (clock dependency of 6lowpan chip)

The aim of this commit is to essentially have the same level of
support that currently exists in the mainline kernel.

---
This depends on https://github.com/lede-project/source/pull/923

I fully expect there to be comments before this is ready to be merged hence it's currently two logical commits and I have left a comment in the upgrade script because ci40 uses "dual nand" i.e. two full copies of the image side by side which are upgraded alternately to give some robustness.


target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_CRYPTO_RNG=y
+CONFIG_CRYPTO_RNG2=y
+CONFIG_CRYPTO_SHA1=y
+CONFIG_CRYPTO_SHA256=y
+CONFIG_CRYPTO_WORKQUEUE=y
+CONFIG_CSRC_R4K=y
+CONFIG_DEFAULT_HOSTNAME="localhost"
+CONFIG_DEFAULT_MMAP_MIN_ADDR=32768
+CONFIG_DEVTMPFS=y
+CONFIG_DEVTMPFS_MOUNT=y
```
blogic: the 4 config symbols above look wrong

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+# CONFIG_DM_SNAPSHOT is not set
+CONFIG_DM_VERITY=y
+# CONFIG_DM_VERITY_FEC is not set
+CONFIG_DTC=y
+CONFIG_DWMAC_GENERIC=y
+CONFIG_EARLY_PRINTK=y
+CONFIG_EARLY_PRINTK_8250=y
+CONFIG_EXT4_FS=y
+CONFIG_EXT4_FS_POSIX_ACL=y
+CONFIG_EXT4_FS_SECURITY=y
```
blogic: is ext4 really required ?

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_HAVE_MOD_ARCH_SPECIFIC=y
+CONFIG_HAVE_NET_DSA=y
+CONFIG_HAVE_OPROFILE=y
+CONFIG_HAVE_PERF_EVENTS=y
+CONFIG_HAVE_REGS_AND_STACK_ACCESS_API=y
+CONFIG_HAVE_SYSCALL_TRACEPOINTS=y
+CONFIG_HAVE_VIRT_CPU_ACCOUNTING_GEN=y
+CONFIG_HOTPLUG_CPU=y
+CONFIG_I2C=y
+CONFIG_I2C_BOARDINFO=y
```
blogic: is I2C really required ?

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_I2C_IMG=y
+CONFIG_IIO=y
+CONFIG_IIO_BUFFER=y
+CONFIG_IIO_KFIFO_BUF=y
+CONFIG_IIO_TRIGGER=y
+CONFIG_IIO_TRIGGERED_BUFFER=y
+CONFIG_IMGPDC_WDT=y
+CONFIG_IMG_MDC_DMA=y
+CONFIG_INITRAMFS_SOURCE=""
+CONFIG_INPUT=y
```
blogic: the 2 above should be removed

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_IRQCHIP=y
+CONFIG_IRQ_DOMAIN=y
+CONFIG_IRQ_DOMAIN_HIERARCHY=y
+CONFIG_IRQ_FORCED_THREADING=y
+CONFIG_IRQ_MIPS_CPU=y
+CONFIG_IRQ_WORK=y
+CONFIG_JBD2=y
+CONFIG_KEYS=y
+CONFIG_LEDS_PWM=y
```
blogic: leds and keys should also be removed

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_MIPS_MT=y
+CONFIG_MIPS_MT_FPAFF=y
+CONFIG_MIPS_MT_SMP=y
+CONFIG_MIPS_NO_APPENDED_DTB=y
+CONFIG_MIPS_PERF_SHARED_TC_COUNTERS=y
+# CONFIG_MIPS_RAW_APPENDED_DTB is not set
+CONFIG_MIPS_SPRAM=y
+# CONFIG_MIPS_VPE_LOADER is not set
+CONFIG_MMC=y
+CONFIG_MMC_BLOCK=y
```
blogic: mmc support aswell

target/linux/pistachio/config-4.9
```diff
@@ -0,0 +1,376 @@
...
+CONFIG_RFS_ACCEL=y
+CONFIG_RPS=y
+CONFIG_RTC_CLASS=y
+CONFIG_RTC_I2C_AND_SPI=y
+CONFIG_RTC_MC146818_LIB=y
+CONFIG_SCHEDSTATS=y
+CONFIG_SCHED_HRTICK=y
+CONFIG_SCHED_INFO=y
+CONFIG_SCSI=y
+CONFIG_SCSI_SPI_ATTRS=y
```
blogic: i think there are way to many thinks linked in directly. please review the config and remove anything that can be installed as a kmod

target/linux/pistachio/base-files/etc/board.d/02_network
```diff
@@ -0,0 +1,19 @@
...
+#
+# Copyright (C) 2017 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+. /lib/functions/uci-defaults.sh
+. /lib/pistachio.sh
+
```
blogic: no need to include the pistachio.sh

target/linux/pistachio/config-4.9
Ham22: I'll look into if we had any reasons for CONFIG_DEFAULT_MMAP_MIN_ADDR as it was also upstreamed to the kernel with that value. My guess is legacy debugging while bringing up the SoC.

target/linux/pistachio/config-4.9
Ham22: See boot explanation below.

target/linux/pistachio/config-4.9
Ham22: Agree, missed this one when moving tpm to kmod. What would be your suggestion for the `CONFIG_I2C_IMG` as `make kernel_menuconfig` would strip it out without I2C enabled and a grep through other targets seem to do the same as above in this situation?

target/linux/pistachio/config-4.9
Ham22: CONFIG_INITRAMFS_SOURCE seems to get put there by `make kernel_menuconfig` hence most configs have it.

CONFIG_INPUT I'll switch to kmod, is there a preferred way of using buttons in LEDE?

target/linux/pistachio/config-4.9
Ham22: There was some concern about using a kmod for pwm leds so put it back into the config, see https://github.com/lede-project/source/pull/922.

target/linux/pistachio/config-4.9
Ham22: See boot explanation below.

target/linux/pistachio/config-4.9
Ham22: @blogic the Ci40 is a development board designed to be as flexible as possible. We therefore support booting from several locations. I understand that kmods can not be used for anything required to boot as they are stored in the filesystem meaning we require all of these for the different booting scenarios:
* NAND boot: `CONFIG_SPI_*`, `CONFIG_MTD_NAND_*`
* USB boot: `CONFIG_USB_*`, `CONFIG_SCSI_*`, `CONFIG_EXT4_FS_*`
* MMC boot: `CONFIG_MMC_*`, `CONFIG_EXT4_FS_*`

Any suggestions on a better way of doing this?

target/linux/pistachio/base-files/etc/diag.sh
```diff
@@ -0,0 +1,33 @@
...
+		status_led_blink_failsafe
+		;;
+	preinit_regular)
+		status_led_blink_preinit_regular
+		;;
+	upgrade)
+		status_led_blink_slow
+		;;
+	done)
+		status_led_set_heartbeat
```
pepe2k: Can you use here `status_led_set_on` as in (almost) all other targets?

In fact, [I was going to drop both](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=log;h=refs/heads/20170309_failsafe-firstboot) `status_led_set_heartbeat` and `status_led_blink_slow` functions as the first one is used only in ar7 target and second one seems to be there since introduction in 2013 and... has never been used (based on what I see in history).

Ham22: I am happy to drop the use of status_led_blink_slow if it helps. However I'd quite like to keep the use of heartbeat:
* For info of lockup/panic, debatable but it can give you information of system status
* For consistency, the dts has been upstreamed to the kernel naming that led heartbeat and defining a default trigger of heartbeat
* There would be a lot of public documentation that would be inconsistent

If there are major concerns I can change this but believe there are valid reasons for keeping it. Also just because LEDE isn't using that function doesn't mean there aren't any forks that do make full use of it.

pepe2k: My idea (more like a long-term plan) was (still is, probably) to move out state-related status LED setup code from all target specific "diag.sh" files to a common place/function and then eventually allow override/extend defaults in target (we do other things per state, [like in ar71xx](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/base-files/etc/diag.sh#L492)).

At now we already fill "diag.sh" file per (new) target by copy&paste same file from other place/s and in almost all (alive/maintained) targets "set_state()" function has the exact same or very similar content (`grep -rn "set_state()" target/linux/*/base-files/ -A999`). This approach doesn't seem correct for me. If we include some new, project-wide state (in fact, I have here something in mind: firstboot triggered inside failsafe), we would need to include same change in all target specific "diag.sh" files.

Anyway, I really don't know the reason/s, why by default status LED is set to "on" (and not to a "heartbeat" for example) in "done" state (this seems to be a general rule, across all targets, excluding old and almost forgotten ar7 which I want to "fix"), but I'm pretty sure it's done this way not without a reason. Maybe because of some historical reasons or just because we use "blinking" patterns during all "preceding" states (failsafe, preinit, etc.) and "heartbeat" might be confusing (I wasn't able to find any better argument here :confused:).

IMHO, this topic should be discussed in larger group and lede-dev mailing list looks like a correct place for it.

> I am happy to drop the use of status_led_blink_slow if it helps.

At least in "ramips" and "ipq806x" targets we use for "upgrade" state the same pattern as for "preinit_regular" ([ramips/base-files/etc/diag.sh#L271](https://github.com/lede-project/source/blob/master/target/linux/ramips/base-files/etc/diag.sh#L271)).

> If there are major concerns I can change this but believe there are valid reasons for keeping it. Also just because LEDE isn't using that function doesn't mean there aren't any forks that do make full use of it.

Personally, it doesn't make sense for me to keep functions which are not used at all, across whole code. If some fork/s make use of it, they can add them in fork as well.

Ham22: Sounds like a good plan to help code duplication and refactoring effort. I shall switch the functions to your suggestions, any further discussion on the topic I will leave to the mailing list.

target/linux/pistachio/image/Makefile
```diff
@@ -0,0 +1,38 @@
...
+	IMAGE/factory.ubi := append-ubi
+	IMAGE/sysupgrade.tar := sysupgrade-tar
+endef
+
+define Device/marduk
+	DEVICE_DTS := img/pistachio_marduk
+	BLOCKSIZE := 256KiB
+	PAGESIZE := 4KiB
+	DEVICE_TITLE := Creator Ci40
+	DEVICE_PACKAGES := kmod-input-gpio-keys kmod-input-evdev \
```
nbd168: Why are you using kmod-input-gpio-keys? For Buttons, kmod-gpio-button-hotplug should be a better fit and it won't pull in the input subsystem.

target/linux/pistachio/image/Makefile
```diff
@@ -0,0 +1,38 @@
...
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/image.mk
+
+KERNEL_LOADADDR := 0x80400000
+
+define Device/Default
+	PROFILES := Default $$(DEVICE_NAME)
```
nbd168: Please remove $$(DEVICE_NAME), it is unnecessary here.

target/linux/pistachio/base-files/lib/pistachio.sh
```diff
@@ -0,0 +1,40 @@
...
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+PISTACHIO_BOARD_NAME=
+PISTACHIO_MODEL=
+
+pistachio_board_detect() {
+	local name=$(grep "^machine" /proc/cpuinfo | sed "s/machine.*: IMG \(.*\) (\(.*\))/\1/g" | awk '{print tolower($1)}')
+	local model=$(grep "^machine" /proc/cpuinfo | sed "s/machine.*: IMG \(.*\) (\(.*\))/\2/g")
```
nbd168: I think it would be a good idea to rely on package/base-files/files/lib/preinit/10_sysinfo instead of adding a custom script to parse /proc/cpuinfo.

target/linux/pistachio/Makefile
```diff
@@ -0,0 +1,27 @@
...
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+include $(TOPDIR)/rules.mk
+
+ARCH:=mipsel
+BOARD:=pistachio
+BOARDNAME:=MIPS pistachio
+FEATURES:=fpu usb usbgadget squashfs targz nand ubifs
```
nbd168: Do you really need an ubifs image? squashfs images are strongly preferred.

target/linux/pistachio/base-files/lib/pistachio.sh
Ham22: The problem with that is that you would get these values:
```
/tmp/sysinfo/board_name = img,pistachio-marduk
/tmp/sysinfo/model = IMG Marduk (Creator Ci40)
```
hence we currently parse the data. The two solutions I can think of are either:
* move the parsing to a custom 10_sysinfo for marduk and switch to use proc
* patch the dts file and drivers to use expected strings

My preference is the former, I can find one similar thing done [here](https://github.com/lede-project/source/blob/master/target/linux/rb532/base-files/lib/preinit/10_sysinfo).

nbd168: I think img,pistachio-marduk as a string to identify the board in base-files is fine.
As for IMG Marduk (Creator Ci40), I think that should be changed in the device tree file instead.

target/linux/pistachio/base-files/etc/diag.sh
Ham22: @pepe2k I believe `status_led_set_on` should just be `status_led_on`, looks like a bug in your staging tree?

pepe2k: @Ham22 Yes, thanks, I will update that branch soon.

target/linux/pistachio/base-files/lib/pistachio.sh
Ham22: Managed to delete my comment, sorry. Anyhow after reviewing what every other target does I changed it to work the same way. Please take a look at the latest version, do you feel it would still be best to make the changes you have suggested?

target/linux/pistachio/image/Makefile
Ham22: Legacy, switched to LEDE way as we have no dependencies on it

nbd168:
Merged to my staging tree with some follow-up fixes and cleanups. Please a look at those.
Thanks!

Ham22:
The follow up changes look sensible, I didn't realise there was a base implementation for board detect functions. I have kicked some CI build/tests so our system can make sure it's all still working, can be found [here](http://jenkins.creatordev.io/job/CreatorDev/job/lede-source/job/nbd-staging/) if your interested.

nbd168:
Please re-run the CI test. There were some issues in my commits and I've fixed those and I've also run some tests on my own Ci40 to verify them.

PR #936, opened 2017-03-09
ipvsadm: add kernel configuration tool and utilities for Virtual Server Load Balancer

ezplanet:
ipvsadm is used to set up, maintain or inspect the virtual server table in the Linux kernel. The Linux Virtual Server can be used to build scalable network services based on a cluster of two or more nodes. The active node of the cluster redirects service requests to a collection of server hosts that will actually perform the services. Supported features include two protocols (TCP and UDP), three packet-forwarding methods (NAT, tunneling, and direct routing), and eight load balancing algorithms (round robin, weighted round robin, least-connection, weighted least-connection, locality-based least-connection, locality-based least-connection with replication, destination-hashing, and source-hashing).

This change adds the following:
- ipvsadm: kernel netfilter Virtual Server Load Balancer administration tool package
- Virtual Server Load Balancer LEDE configuration file
- Virtual Server Load Balancer LEDE init script
- Virtual Server Load Balancer automated monitoring and failover tool

this change depends on kernel ip_vs modules

### DIFF
--- a/package/kernel/linux/modules/netfilter.mk
+++ b/package/kernel/linux/modules/netfilter.mk
@@ -77,6 +77,17 @@ endef
 $(eval $(call KernelPackage,nf-conntrack))
 
 
+define KernelPackage/nf-conntrack-ftp
+  SUBMENU:=$(NF_MENU)
+  TITLE:=FTP connection tracking
+  KCONFIG:=CONFIG_NF_CONNTRACK_FTP
+  DEPENDS:=kmod-nf-conntrack
+  FILES:=$(LINUX_DIR)/net/netfilter/nf_conntrack_ftp.ko
+endef
+
+$(eval $(call KernelPackage,nf-conntrack-ftp))
+
+
 define KernelPackage/nf-conntrack6
   SUBMENU:=$(NF_MENU)
   TITLE:=Netfilter IPv6 connection tracking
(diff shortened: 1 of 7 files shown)


ezplanet:
Linux Virtual Server documentation here: http://www.linuxvirtualserver.org/
ipvsadm kernel tool here: https://www.kernel.org/pub/linux/utils/kernel/ipvsadm/

ezplanet:
Added a Luci pull request for luci configuration and monitoring pages:

https://github.com/openwrt/luci/pull/1061

commodo:
it looks to me like this PR also adds the kernel module

in any case: i was reading into IPVS a while back, but this looks like it would be more fitted to as part of the package feeds here: https://github.com/openwrt/packages

i usually feel that core is a bit more oriented towards embedded than servers & clustering
it would be an idea to add in the feeds a kernel package like in core (i.e. this one https://github.com/lede-project/source/tree/master/package/kernel/linux )

jow-:
I merged this PR into my staging tree and took care of dropping the unrelated kmod commit.
Had to adjust a number of details in your Makefile.

See https://git.lede-project.org/d3f8cea.

ezplanet:
@commodo ipvs is a kernel module. ipvsadm is distributed as an accessory to the linux kernel, therefore it belongs with the linux kernel configuration which is in master, not packages.
If you are not familiar with the architecture of ipvs, just think of iptables.
It is as if you were asking to move iptables to packages.

PR #865, opened 2017-02-22
ramips: Add support for HiWiFi HC5962

gigibox:
This patch adds supports for the HiWiFi HC5962(gee4) http://www.hiwifi.com

Short specification:
- MT7621AT + MT7612EN + 7603EN
- 256MB DDR3 RAM
- 128MB NAND flash
- 1+3 x 1000M Ethernet
- 1x USB 2.0 port. 1x USB 3.0 port
- UART pad on PCB (JP3: TX, RX, GND, 3.3V)

Flash instruction:

1. Download lede-ramips-mt7621-hc5962-squashfs-factory.bin
2. Login as root via SSH on 192.168.199.1 and then copy factory.bin(using wget or nc or...) to /tmp/
3. use the following commands:
   $ mtd write /tmp/lede-ramips-mt7621-hc5962-squashfs-factory.bin firmware
   $ mtd erase firmware_backup && reboot
After reboot you should be able to login as root via SSH on 192.168.1.1

Signed-off-by: ZengFei Zhang <zhangzengfei@kunteng.org>

PR #937, opened 2017-03-10
ramips: add support for Netgear EX3700

jclehner:
Specifications:
* SoC: MT7620A
* RAM: 64 MB DDR
* Flash: 8MB NOR SPI flash
* WiFi: MT7612E (5Ghz) and builtin MT7620A (2.4GHz)
* LAN: 1x100M

This patch would not have been possible without the help and input of
Paul Oranje (@poranje).

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -165,6 +165,10 @@ wn3000rpv3)
 	ucidef_set_led_default "power_r" "POWER (red)" "$board:red:power" "0"
 	set_wifi_led "$board:green:router"
 	;;
+ex3700)
+	ucidef_set_led_netdev "wlan5g" "ROUTER (green)" "$board:green:router" "wlan0"
+	ucidef_set_led_netdev "wlan2g" "DEVICE (green)" "$board:green:device" "wlan1"
+	;;
 f5d8235-v1)
 	ucidef_set_led_default "lan" "lan" "$board:blue:wired" "0"
 	set_usb_led "$board:blue:storage"
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -23,6 +23,7 @@ get_status_led() {
 	dch-m225|\
 	dir-860l-b1|\
 	e1700|\
+	ex3700|\
 	fonera20n|\
 	kn_rc|\
 	kn_rf|\
--- a/target/linux/ramips/dts/EX3700.dts
+++ b/target/linux/ramips/dts/EX3700.dts
@@ -1,0 +1,162 @@
+/* This file is released into the public domain */
+
+/dts-v1/;
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+#include "mt7620a.dtsi"
+
+/ {
+	compatible = "ralink,mt7620a-soc";
+	model = "Netgear EX3700";
+
+	chosen {
+		bootargs = "console=ttyS0,57600";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power_g {
+			label = "ex3700:green:power";
+			gpios = <&gpio2 23 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
+
+		power_a {
+			label = "ex3700:amber:power";
+			gpios = <&gpio2 28 GPIO_ACTIVE_LOW>;
+			default-state = "off";
+		};
+
+		router_g {
+			label = "ex3700:green:router";
+			gpios = <&gpio2 25 GPIO_ACTIVE_LOW>;
+		};
+
+		router_r {
+			label = "ex3700:red:router";
+			gpios = <&gpio2 24 GPIO_ACTIVE_LOW>;
+		};
+
+		device_g {
+			label = "ex3700:green:device";
+			gpios = <&gpio2 20 GPIO_ACTIVE_LOW>;
+		};
+
+		device_r {
+			label = "ex3700:red:device";
+			gpios = <&gpio2 21 GPIO_ACTIVE_LOW>;
+		};
+
+		wps {
+			label = "ex3700:green:wps";
+			gpios = <&gpio2 27 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio2 26 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps {
+			label = "wps";
+			gpios = <&gpio0 2 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+};
+
+&gpio0 {
+	status = "okay";
+};
+
+&gpio2 {
+	status = "okay";
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+
+		partition@0 {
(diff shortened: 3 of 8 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -165,6 +165,11 @@ wn3000rpv3)
 	ucidef_set_led_default "power_r" "POWER (red)" "$board:red:power" "0"
 	set_wifi_led "$board:green:router"
 	;;
+ex3700)
+	ucidef_set_led_default "power_a" "POWER (amber)" "$board:amber:power" "0"
```
mkresin: Drop this LED here, it is already handled by diag.sh.

I would suggests to use the green:power LED in diag.sh. Everything reddish indicates an error to me. You can drop the ```default-state = "on";``` from the green power LED in dts afterwards.

In case the amber LED is is switched on by the bootloader, add a ```default-state = "off";``` to the amber power led in the DTS (http://lxr.free-electrons.com/source/Documentation/devicetree/bindings/leds/leds-gpio.txt).

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -231,6 +231,7 @@ get_status_led() {
 	wsr-600)
 		status_led="$board:green:power"
 		;;
+	ex3700|\
```
mkresin: Please move the whole block a few lines up so that the file is in alphabetical order again.

target/linux/ramips/dts/EX3700.dts
```diff
@@ -0,0 +1,162 @@
+/* This file is released into the public domain */
+
+/dts-v1/;
+
+#include "mt7620a.dtsi"
+
+#include <dt-bindings/input/input.h>
```
mkresin: Would you please swap the include lines so that they are in alphabetical order.

target/linux/ramips/dts/EX3700.dts
```diff
@@ -0,0 +1,162 @@
...
+
+&pinctrl {
+	state_default: pinctrl0 {
+		default {
+			ralink,group = "i2c", "rgmii2", "spi refclk";
+			ralink,function = "gpio";
+		};
+	};
+};
+
```
mkresin: Drop this blank line.

target/linux/ramips/base-files/etc/board.d/01_leds
jclehner: The blinking amber led on bootup is used by the stock firmware on this device, so this behaviour was retained for LEDE.

target/linux/ramips/base-files/etc/diag.sh
jclehner: Done.

target/linux/ramips/dts/EX3700.dts
jclehner: Done.

target/linux/ramips/dts/EX3700.dts
jclehner: Done.

target/linux/ramips/base-files/etc/board.d/01_leds
pepe2k: > The blinking amber led on bootup is used by the stock firmware on this device, so we wanted to replicate this in LEDE. I did a similar thing in the EX2700 port.

We had similar discussion about LEDs in one of recent PR: https://github.com/lede-project/source/pull/865#discussion_r102938528.

If this device contains two LEDs which can be used as status LED, I also prefer to select the green one over the red/amber/orange.

target/linux/ramips/dts/EX3700.dts
```diff
@@ -0,0 +1,162 @@
...
+&ethernet {
+	mtd-mac-address = <&factory 0x28>;
+};
+
+&wmac {
+	ralink,mtd-eeprom = <&factory 0x0>;
+};
```
f00b4r0: what about mac-address? Is it correctly set? On the WN3000RPv3 the wireless MAC was stored at a different offset in the factory partition...

poranje: Several tests have shown that the MAC address is correct (same as when using the stock image, on two different devices).

target/linux/ramips/image/mt7620.mk
```diff
@@ -79,6 +79,18 @@ define Device/wn3000rpv3
 endef
 TARGET_DEVICES += wn3000rpv3
 
+define Device/ex3700
```
mkresin: nope. this file is a mess anyway. I will sort it after the patch is merged.

poranje: It is, but currently the order is: ex2700, wn3000rpv3, ex3700. And that seems incorrect.

target/linux/ramips/dts/EX3700.dts
f00b4r0: Ok, thanks. I'm curious: if you change either of the MACs in u-boot, is that change correctly reflected in LEDE?
I notice this device's MTD appears to be structured quite differently from the wn3000, which surprises me.

mkresin: @f00b4r0 not sure what you are trying to find here. They have tested it on two devices and it was fine on both. It should be enough confirmation that the way it is done is the correct way.

There is no need to let them do more work than necessary. Especially since @jclehner has a good track record in submitting patches.

f00b4r0: @mkresin I noticed on the wn3000 that if I didn't specify the correct MAC address location, the wlan would seemingly get the right MAC address, but that address wasn't updated when changed in u-boot (contrary to stock firmware). Hence my concern. Forgive me if I look like I'm splitting hair here, I'm just trying to ensure that we get most accurate support.

mkresin: @f00b4r0 The ralink eeprom on flash has normally a device specific mac address set. According to the ethernet node, the mac address is at offset 0x28 of the ralink eeprom/caldata and is used as base mac for the ethernet interface as well (done this way by most boards). I'm not even sure if the mt76 driver allows to override the mac address set in the eeprom. At least for the "old" rt2x00 driver I've added the missing bits [a few weeks ago](http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=9766cb709089fc460fda66c06a3da756af4a2930).

It might be that there is another location for mac addresses which is written from the bootloader to. But I've never seen an u-boot that writes settings to something different than to its u-boot env. Looks like your board is some kind of alien.

As long as the eeprom has a unique mac address, matching the one used by the stock firmware, I'm fine with it. It don't see the need to change the device specific mac address at all, as long as you don't have a mac prefix assigned to you self.

f00b4r0: @mkresin ok. It's quite possible the ex2700/wn3000 are more different from the ex3700 than I expect, which would explain the discrepancy in DTS. As for the ability of changing the device MAC address, I'm told the primary use is to resolve collisions, though I never experienced one myself :)

jclehner: @f00b4r0 The EX2700 and WN3000* are both manufactured by Delta Networks, while the EX3700's ODM is Foxconn, so this explains the difference in MTD layouts. The stock firmware *does* store MAC addresses in the `config` and `nvram` partitions, but so far, these matched the EEPROM MACs on all tested devices.

@mkresin I've updated the patch, so the diagnostic LED thingy is the only thing holding back this PR at this point, correct?

f00b4r0: > @f00b4r0 The EX2700 and WN3000* are both manufactured by Delta Networks, while the EX3700's ODM is Foxconn, so this explains the difference in MTD layouts. The stock firmware does store MAC addresses in the config and nvram partitions, but so far, these matched the EEPROM MACs on all tested devices.

Thanks for the clarification! Re MAC: I suppose this means that the WLAN MAC cannot be changed with the current DTS config then?

mkresin: > @mkresin I've updated the patch, so the diagnostic LED thingy is the only thing holding back this PR at this point, correct?

Correct

jclehner: Latest patch fixes that too.

mkresin:
merged, thanks.

poranje:
tested snapshot images of monday 2017-03-13:
sysupgrade works
flash of factory image works
failsafe (with green blinking power LED) works.

PR #938, opened 2017-03-10
kirkwood: add default leds in NSA310 and NSA325

bobafetthotmail:
creating a generic uci helper function because it would make no sense to have 3-4 uci helpers that differ only in name and trigger name.


mkresin:
Is there any specific reason why you can not use ```ucidef_set_led_usbport``` with all your usb ports as arguments, similar to how it is done in all other targets?

I'm not that thrilled of adding a new function to ```package/base-files/files/lib/functions/uci-defaults.sh```, which is only used by one target. IMHO you should use a target specific function here.

Regarding CONFIG_ATA_LEDS, I would like to see that the port specific blinking gets integrated into the upstream [LED Disk Trigger](http://cateee.net/lkddb/web-lkddb/LEDS_TRIGGER_DISK.html). This way we can drop one more LEDE specific patch.

bobafetthotmail:
> Is there any specific reason why you can not use ucidef_set_led_usbport with all your usb ports as arguments, similar to how it is done in all other targets?

As said in the commit: "to keep consistent behaviour with the Sata leds that show sata activity, this led uses usb-host trigger to show usb activity."

usbport trigger lights up the led if there are devices connected, and turns it off if there is nothing connected. It's not an usb activity led currently.

Also, I personally prefer using leds to show activity for storage devices. I can check their "connected" status pretty easily by looking at the physical ports, while I can't check activity with the same ease.

> I'm not that thrilled of adding a new function ... which is only used by one target. 

I placed it there because it is very generic, it allows to set any simple led trigger like the ones I'm using (**usb-host**, **ata1**, **ata2**), but it can easily consolidate the current **ucidef_set_led_ide** and **ucidef_set_led_wlan** (which actually is _exactly the same_ as mine, but with a device-specific name), and probably also some target-only triggers or stuff that is written in the dts file like the **disk-activity** trigger in **mvebu** target. I volunteer to do this change treewide if you agree.

Or, if placing these triggers in dts is the way to go, I'll do that. I don't like that very much though.

> Regarding CONFIG_ATA_LEDS, 

@dangowrt  (the patch author) said he tried to upstream it a long time ago and it was dropped with no explanation, see here http://lists.infradead.org/pipermail/lede-dev/2017-February/006358.html

If you want that upstreamed I'm afraid you will have to discuss this with him, as I don't understand the patch enough, I can only send it as-is and hope they don't request changes.

bobafetthotmail:
Is there anything that needs to be done here? 
I'm not sure if what @mkresin said above is a request for changes or not.

mkresin:
> I'm not sure if what @mkresin said above is a request for changes or not.

No, it was meant as question/starter for a discussion. It is just that I don't really like the new function and  therefore I will not merge the PR.

But I'm not going to reject this PR. Might be that other devs see the need of the new function and I'm the one who is wrong here.

bobafetthotmail:
>  It is just that I don't really like the new function and therefore I will not merge the PR.

@mkresin If I make that a device-specific function, will you merge this? 
Can I move my uci function inside **target/linux/kirkwood/base-files/lib/kirkwood.sh**  or is that a wrong way to make it device-specific? 

I suspect that a serious discussion about led helper functions is too low in priorities of everyone (and I'm not even blaming them, lol), so I doubt we will see anyone else here.

blogic:
@bobafetthotmail  looks ok, but please make ucidef_set_led_usbport() and all other helpers setting triggers use this new function as to avoid code duplication

bobafetthotmail:
ok, updated.
Since after the cleanup any simple led trigger function is a one-liner, I've added the ones I used to uci-defaults.
If it's not OK I can move them to local as said above, but these functions aren't specific to kirkwood devices.

bobafetthotmail:
Umm, is there something I can do to help the review of this PR? Splitting more the commits, making a different PR for the led functions consolidation, whatever.
I understand that led things are not high-priority, but I would like to have these two kirkwoods set up correctly.

@blogic

jow-:
I'll take it :)

jow-:
Merged into my staging tree - thanks!

PR #939, opened 2017-03-12
kernel: bump 4.9 to 4.9.14

Ansuel:
Drop 070 generic patch (pushed upstream), refresh patch


Ansuel:
Compiling for testing right now...

nbd168:
Only parts of patch 070 were applied, not all of it. Please only drop the get_devce/put_device related chunks, not all of it.

Ansuel:
@nbd168 ok sorry 
should be good now (don't know if it applyies with offset... think not)

nbd168:
It looks like you've reverted a change to 702-phy_add_aneg_done_function.patch, please fix

nbd168:
If you want to make sure that there are no patch offsets, just run make target/linux/refresh. Please do this with every platform that uses 4.9

Ansuel:
@nbd168 i know but i'm recompiling all the toolchain and stuff at the moment and i can't do it
 
fixed 702 patch

PR #946, opened 2017-03-13
hostapd: Default disable b rates, add legacy_rates

nicklowe:
The require_mode option can currently be configured to a value of g or n for 2.4 GHz BSSes or n or ac for 5 GHz BSSes to set the minimum mode that connecting clients need to support to be able to connect.

This patch introduces a new require_mode of b and defaults to a require_mode of g where this is unspecified.

The rationale for this, which is stronger now than in 2014, can be found in:

https://mentor.ieee.org/802.11/dcn/14/11-14-0099-00-000m-renewing-2-4ghz-band.pptx

The balance of equities between compatibility with b clients and the detriment to the 2.4 GHz ecosystem as a whole in 2017 strongly favors disabling b support by default.

Issues previously in the hostapd.sh shell script for 2.4 GHz BSSes for the require_mode were:

1) Where a require_mode of g was configured, the supported rates still included b rates yet the basic rates would not include b rates.

2) Where a require_mode of g was configured, the basic rates would be set to 60 120 240 that would override the configuration of the basic_rate option rather than filtering to exclude b rates.

3) Where a require_mode of n was configured, the b rates were still used as basic rates which is not airtime efficient. The g require_mode achieved better airtime efficiency for broadcast/multicast traffic. Only where the require_mode was set to g would the basic rates be set to not include b rates.

Now:

1) Where a require_mode is not configured to b and the supported_rates option has not been configured, this will be changed to OFDM rates only:

60 90 120 180 240 360 480 540

2) Where a require_mode is not configured to b and the supported_rates option has been configured, this will be filtered to include OFDM rates only.

3) Where a require_mode is not configured to b and the basic_rate option has not been configured, this will be changed to OFDM rates only:

60 120 240

4) Where a require_mode is not configured to b and the basic_rate option has been configured, this will now be filtered to include OFDM rates only.

Signed-off-by: Nick Lowe <nick.lowe@gmail.com>

### DIFF
--- a/package/network/services/hostapd/files/hostapd.sh
+++ b/package/network/services/hostapd/files/hostapd.sh
@@ -64,6 +64,7 @@ hostapd_common_add_device_config() {
 	config_add_string country
 	config_add_boolean country_ie doth
 	config_add_string require_mode
+	config_add_boolean legacy_rates
 
 	hostapd_add_log_config
 }
@@ -75,12 +76,13 @@ hostapd_prepare_device_config() {
 	local base="${config%%.conf}"
 	local base_cfg=
 
-	json_get_vars country country_ie beacon_int doth require_mode
+	json_get_vars country country_ie beacon_int doth require_mode legacy_rates
 
 	hostapd_set_log_options base_cfg
 
 	set_default country_ie 1
 	set_default doth 1
+	set_default legacy_rates 0
 
 	[ -n "$country" ] && {
 		append base_cfg "country_code=$country" "$N"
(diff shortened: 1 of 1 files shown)


nbd168:
First of all, I like the idea of disabling non-OFDM rates by default. However, I think that using require_mode is not the way to do it, since it refers to rate capabilities at assoc time, not actual rate usage.

I think a better approach would be to have a separate bool option to enable CCK support (and have it set to 0 by default).

Some nitpicks: Please use a proper prefix for the commit (e.g. `hostapd: `) and reduce the line length of the commit message to something like 75.

nicklowe:
In practice, this does stop most stations from using the non-OFDM rates and should also retain compatibility with g or n stations that were to make still to use of barker coding/CCK modulation for some frames. I think this is a preferable approach therefore for default behaviour while also being sufficient to achieve the airtime efficiency goals.

nbd168:
Okay, let me clarify things a bit. Here's what I see in your proposal that I agree with:
- Dropping CCK rates from supported_rates by default
- Dropping CCK rates from basic_rates when require_mode is 'n'

Here's what I don't agree with:
- Having require_mode affect supported_rates. There should be a separate option for that.

I meant for 'require_mode' to specify the minimum mode that a client needs to support to be able to connect. It should not have other side effects on the actual rates being used.

nicklowe:
Thanks for the review. require_mode no longer affects supported_rates and I have introduced a legacy_rates option.

nbd168:
Merged a cleaned up version of this to my staging tree, thanks.

PR #952, opened 2017-03-14
dnsmasq: don't point --resolv-file to default location unconditionally

pprindeville:
If noresolv is set, we should not generate a --resolv-file parameter.


pprindeville:
@jow-

dedeckeh:
One unconditionally read of the resolvfile uci parameter has been removed; however the resovlfile uci parameter is still read unconditionally (see https://github.com/lede-project/source/blob/master/package/network/services/dnsmasq/files/dnsmasq.init#L791). Thus even if the noresolv is set to 1 the --resolv-file parameter is still be generated atm

pprindeville:
> Thus this line needs also to be removed if the --resolv-file parameter does not to be generated if noresolv is set to 1.

Fixed with latest version.

dedeckeh:
Pulled into my staging tree https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=95ae5ae18e87eab07dec0457d9a10465005093f4

PR #957, opened 2017-03-15
ramips: clean kernel config symbol conflicts for MT7688 PWM

HZFrodo:
pwm-mediatek should be selected in Other Modules submenu, not default to yes.

Just make mt7688 consistent with mt7620 mt7621 mt7628. 
And we should assign kmod-pwm-mediatek to DEVICE_PACKAGES explictly only if some target devices require pwm exactly.
As for the definition of pwm-mediatek,  it is already at the head of target/linux/ramips/modules.mk


mkresin:
> pwm-mediatek should be selected in Other Modules submenu, not default to yes.

Why? What kind of issue do you have? I miss an explanation in the commit messages why this change is required at all.

Beside the wrtnode2r all MT7688 based boards have pwm nodes in their device tree source file. Makes totally sense to me to assume that pwm is used for all MT7688 boards and therefore should be enabled in the kernel config by default.

blogic:
the patch is incomplete. please also add the definition inside target/linux/ramips/modules.mk

HZFrodo:
Just make mt7688 consistent with mt7620 mt7621 mt7628. And we should assign kmod-pwm to DEVICE_PACKAGES explictly only if some target devices require pwm exactly.
The definition is already inside target/linux/ramips/modules.mk

HZFrodo:
@mkresin @blogic

blogic:
merged, Thanks !

PR #958, opened 2017-03-15
utils/f2fs-tools: Update to 1.8.0

diizzyy:
Update f2fs-tools to 1.8.0
Refresh patches
Remove selinux patch, use configure argument instead.
Switch to xz tarball
Adjust url to avoid redirects


bobafetthotmail:
It seems this version fixes a pretty bad issue of fsck.f2fs that hoses the f2fs partition instead of fixing it on **ar71xx** devices, as tested here by myself https://forum.lede-project.org/t/tp-link-tl-1043nd-v2-freezes/2379/9
(same issue is also reported here https://github.com/openwrt/packages/issues/1848#issuecomment-262851090 )

I think this update should go also in LEDE 17.01 when possible.

nbd168:

PR #959, opened 2017-03-15
curl: Adjust URLs

diizzyy:
Update mirror list, add main site as last resort
Source: https://github.com/curl/curl-www/blob/master/latest.pl


hauke:
thank you, this was merged

PR #960, opened 2017-03-16
ar71xx: add support for TP-LINK TL-WR840N v2 and v3

svobodavac:
This patch adds support for the TP-LINK TL-WR840N v2 and TP-LINK TL-WR840N v3.
- SoC: Qualcomm QCA9533-BL3A (650MHz)
- RAM: 32 MiB (Zentel A3S56D40GTP)
- Flash: 4 MiB
   (v2: Macronix MX 25L323F)
   (v3: Winbond 25Q32FVSIG)
- LAN: 4x 100M
- WAN: 1x 100M

Installation:
It is possible to install OpenWRT/LEDE from the web GUI on older stock firmwares.
Such upgrades are locked on newer stock firmwares. The only way is to use
TFTP recovery mode of the bootloader.


target/linux/ar71xx/base-files/etc/board.d/01_leds
```diff
@@ -679,6 +679,12 @@ tl-wr941nd-v6)
 	ucidef_set_led_switch "lan4" "LAN4" "tp-link:blue:lan4" "switch0" "0x02"
 	ucidef_set_led_wlan "wlan" "WLAN" "tp-link:blue:wlan" "phy0tpt"
 	;;
+tl-wr840n-v2|\
+tl-wr840n-v3)
+	ucidef_set_led_netdev "wan" "WAN" "tp-link:green:wan" "eth1"
+	ucidef_set_led_netdev "lan" "LAN" "tp-link:green:lan" "eth0"
```
heinzek: Use eth0 as lan led trigger is bad idea, because eth0 interface is on when br-lan is up. Better to use:
ucidef_set_led_switch "lan" "LAN" "tp-link:green:lan" "switch0" "0x1E"

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr840n-v2.c
```diff
@@ -0,0 +1,127 @@
+/*
+ *  TP-LINK TL-WR841N/ND v9/v11 / TL-WR842N/ND v3
```
pepe2k: Please, drop both mach-*.c files and include changes needed by these boards inside mach-tl-wr841n-v9.c, as the code is almost exact copy and it really doesn't make sense to multiply it.

[You can take a look how it was done with TL-WR842N/ND v3 support](https://github.com/lede-project/source/commit/0b45bec22c59cc6e6b2fa71dc9a88386c5587132).

target/linux/ar71xx/base-files/etc/board.d/01_leds
pepe2k: @heinzek is right here.

svobodavac: No problem, I'm going to fix it. 
But it looks like it works someway - the ethernet switch replicates LAN ports link states to the port link state.
When all external LAN ports are diconnected, the eth0 port has link down state:

root@LEDE:~# dmesg
...
[  236.689696] eth0: link down
[  236.694266] br-lan: port 1(eth0.1) entered disabled state
[  373.189175] eth0: link up (1000Mbps/Full duplex)

Even if br-lan is UP.

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr840n-v2.c
svobodavac: OK, I'm preparing changes, but are you really sure?
It would add some code (yes, only a few bytes) to the firmware, which isn't necessary...

target/linux/ar71xx/base-files/etc/board.d/01_leds
svobodavac: Done.

svobodavac:
The EU factory image works on EU model and is accepted by new stock firmware during upgrade process.

heinzek:
Please add:
# CONFIG_ATH79_MACH_TL_WR840N_V2 is not set
# CONFIG_ATH79_MACH_TL_WR840N_V3 is not set

to files:
target/linux/ar71xx/mikrotik/config-default
target/linux/ar71xx/nand/config-default

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr840n-v2.c
pepe2k: With your current approach, you are adding two new mach-*.c files which will result in much more more bytes added to the kernel. I'm asking you to drop these files as the same code already exists in other place and there is no reason to duplicate it.

svobodavac: It's already done;)
No problem, just for clarification:
Every mach-*.c file results to one one mach-*.o file.
Only one mach-*.o file is linked to the kernel according to device type.
So, I woud say that many small mach-*.c files are better from the kernel size point of view.

svobodavac: You're right, I'm sorry.
The linker takes only used variables/functions from the object file to the kernel.

svobodavac:
Done:
# CONFIG_ATH79_MACH_TL_WR840N_V2 is not set
# CONFIG_ATH79_MACH_TL_WR840N_V3 is not set

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr841n-v9.c
```diff
@@ -1,9 +1,10 @@
 /*
- *  TP-LINK TL-WR841N/ND v9/v11 / TL-WR842N/ND v3
+ *  TP-LINK TL-WR840N v3/v4 / TL-WR841N/ND v9/v11 / TL-WR842N/ND v3
```
bkil: There's a typo here: TL-WR840N v2/v3

svobodavac: Fixed, thanks.

pepe2k:
@svobodavac, I have included your changes [in my staging repository, in separate branch](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=shortlog;h=refs/heads/ar71xx_pr960_20170321), with few minor changes (list below), please take a look/test and let me know if you accept them.

- change device name in image name to tl-wr840n (for now it doesn't look there is TL-WR840ND available)
- dropped redundant factory-us image (this device isn't listed on TP-Link US website)
- simplify image code in tp-link.mk
- dropped CONFIG_ATH79...WR840N_V3 (there is no separated config in Kconfig.openwrt for it)

Cheers,
Piotr

pepe2k:
@svobodavac, thank you for your contribution! I have merged your change/s into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

PR #961, opened 2017-03-16
openvpn: add extra respawn parameters

sch-m:
This change protects the openvpn instances to be marked as "in a crash loop"
and thereby the connection retries will run infinitely.

### DIFF
--- a/package/network/services/openvpn/files/openvpn.init
+++ b/package/network/services/openvpn/files/openvpn.init
@@ -68,6 +68,9 @@ openvpn_add_instance() {
 		--config "$conf"
 	procd_set_param file "$dir/$conf"
 	procd_set_param respawn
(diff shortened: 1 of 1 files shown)


blogic:
by default it should not tru to respawn more then 5 times. please explain the problem that you are seeing.

sch-m:
When the remote site of an openvpn connection goes down for some time (network failure etc.) the openvpn instance in an openwrt/lede device should not stop retrying to establish the connection.

With the current limit of 5 retries, there is a user interaction required, which isn't really what you want when the device should simply do everything to keep the vpn connection up.

blogic:
please add that explanation to the patch description

sch-m:
ok, updated the patch description

blogic:
merged, Thanks !

PR #962, opened 2017-03-16
ipq806x: support independent core clocks in cpufreq driver with kernel 4.9+

hnyman:
There is a regression with the ipq806x jump to kernel 4.9 regarding the CPU frequency scaling. Earlier with 4.4 the CPU cores scaled frequency independently, but with 4.9 both cores scale always simultaneously.

The reason for that are upstream Linux changes that have removed the platform data structure used by the qcom driver that is created by the patch 0045. To enable the driver to work with kernel 4.9 the patch 0046 was created to remove the independent core functionality from the driver.

My patch removes the simplification patch 0046 and replaces it with one that adds the independent_clocks definition to the new data structure in the new cpufreq-dt.h and modifies the driver to use that.

----

Add back support for the independent_clocks definition that has been removed between kernel 4.4 and 4.9 by upstream commits:
* https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/drivers/cpufreq/cpufreq-dt.c?h=linux-4.9.y&id=eb96924acddc709db58221c210ca05cd9effb1df
* https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/drivers/cpufreq/cpufreq-dt.c?h=linux-4.9.y&id=e86eee6bc2aaa6b3637f6497b26beee09a91bde9

Summary of changes:
* extend the new cpufreq_dt_platform_data definition in cpufreq-dt.h
* revert the removal of its usage in cpufreq-dt.c
* use new cpufreq-dt.h in qcom-cpufreq.c

Run-tested with R7800 (with ipq8065) as of LEDE master r3761-6bada71e51

@blogic @nbd168 
cc: @dissent1 

image


dissent1:
Just for clarification, this platform data has been removed from qcom cpufreq driver in favor of generic cpufreq-dt-platdev driver that now stores common platform data for all cpufreq board specific drivers instead of having duplicate code in each one.

If someone could help incorporate it in a generic way utilizing platdev driver, I suppose it would have been a cleaner solution. 
@mkresin

hnyman:
> this platform data has been removed from qcom cpufreq driver in favor of generic cpufreq-dt-platdev driver

I think that a more realistic interpretation is:  platform data has been removed from upstream because there have been no upstream users of the data. And then the usage of that data structure has been removed from the qcom cpufreq driver to enable it to work with kernel 4.7+ but with reduced functionality. My patch just restores the functionality to the previous level.  (There is no 8064/8065 implementation of the cpufreq-dt-platdev usage, so "in favor of" is rather bold.)

Too bad that the qcom cpufreq driver for 8064/8065 has not been properly upstreamed :-(
Would be great if somebody figures out a better way to implement this.

dissent1:
Actually platform data has been present in most cpufreq drivers, check compatibility list of platdev and go through its log along with board specific drivers log. Mind that some of them are not in cpufreq dir, for example omap-cpufreq platform data has been in mach-omap dir.
Example https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/?id=7694ca6e1d6f01122f05039b81f70f64b1ec4063

As for independent freq scaling it seems yes that it hasn't been used anywhere, but instead of correctly adjusting qcom cpufreq driver, guys from qcom have just removed platform data from the driver excluding mentioned functionality.
Maybe because of such approach the krait driver hasn't made its way upstream.

blogic:
merged, Thanks !

i will look into rewriting this using the generic driver, for now this hack at least makes the target work so that we can bump it to v4.9

PR #963, opened 2017-03-17
firewall[RFC]: add ipsec_nat_pass option

feckert:
Outgoing IPsec traffic should `not` be masquraded on the wan interface.

Signed-off-by: Florian Eckert <fe@dev.tdt.de>

### DIFF
--- a/package/network/config/firewall/patches/0001-add-ipsec-nat-pass.patch
+++ b/package/network/config/firewall/patches/0001-add-ipsec-nat-pass.patch
@@ -1,0 +1,58 @@
+--- a/zones.c
++++ b/zones.c
+@@ -75,6 +75,7 @@ const struct fw3_option fw3_zone_opts[]
+ 
+ 	FW3_OPT("mtu_fix",             bool,     zone,     mtu_fix),
+ 	FW3_OPT("custom_chains",       bool,     zone,     custom_chains),
++	FW3_OPT("ipsec_nat_pass",      bool,     zone,     ipsec_nat_pass),
+ 
+ 	FW3_OPT("log",                 bool,     zone,     log),
+ 	FW3_OPT("log_limit",           limit,    zone,     log_limit),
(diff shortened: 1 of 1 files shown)


jow-:
Hm, this seems rather arbitrary to me. I'd prefer to add a per-zone `option nat_extra`  which can be used to add additional arguments to the generated MASQUERADING rules.

feckert:
On https://wiki.openwrt.org/doc/uci/firewall#redirect i already have seen the extra option with `-m policy --dir in` for IPsec. But i have only the target `DNAT/SNAT`. For IPsec I need `ACCEPT` so the `xfrem lookup` will get a unmodified package (http://inai.de/images/nf-packet-flow.png)

feckert:
@jow- any annotations? Or are there any other opinion or solution to solve my problem with ipsec?

feckert:
@jow- i have thought about your hint. Do you mean with `option nat_extra` a new configuration param with iptables syntax `-m policy --dir out --pol ipsec -j ACCEPT` or do you mean that i can add my `ipsec_pass_nat` param to this new config option and FW3 will do the job?

feckert:
No comments anymore so i will close this PR

jow-:
@feckert - I thought some more about it and it seems we actually implemented support for nat ACCEPT rules years ago.

Can you try if the following configuration solves your use case?

    config nat
        option src 'wan'
        option extra '-m policy --dir out --pol ipsec'
        option target 'ACCEPT'

If you specify src wan the rule is put into `zone_wan_postrouting`, else into `POSTROUTING`.

feckert:
@jow- thanks for comments but i think you are right it is arbitrary. My colleague mybe found an other solution. He will fix the issue in the ipsec directly upstream.
https://github.com/strongswan/strongswan/pull/71

If this will not work as well then i will try your solution with.
```
config nat
    option src 'wan'
    option extra '-m policy --dir out --pol ipsec'
    option target 'ACCEPT'
```

PR #967, opened 2017-03-18
ramips: fix labels for LAN interfaces on Zbtlink ZBT-WE2026

svobodavac:
This patch fixes incorrect reverse order LAN ports labels on ZBT-WE2026.


pepe2k:
Hello @svobodavac, I suppose that "fix switch port mapping" would be a better description of what your change does.

Cheers,
Piotr

blogic:
merged, Thanks !

PR #968, opened 2017-03-19
odhcp6c: check for empty variable before reading

updateing:
`IFACE_464XLAT` is set by dhcpv6.sh only when
`iface_464xlat` exists. Thus, if 464XLAT support is
installed but interface is not configured to use it,
`$IFACE_464XLAT` will be empty and the `!= 0` check
will be true, causing the script to go along the path
as if 464XLAT is enabled. Adding a check for variable
length before reading solves the issue.

Signed-off-by: Tian Hao <haotia@gmail.com>

### DIFF
--- a/package/network/ipv6/odhcp6c/files/dhcpv6.script
+++ b/package/network/ipv6/odhcp6c/files/dhcpv6.script
@@ -167,7 +167,7 @@ setup_interface () {
 		[ -n "$IFACE_DSLITE_DELEGATE" ] && json_add_boolean delegate "$IFACE_DSLITE_DELEGATE"
 		json_close_object
 		ubus call network add_dynamic "$(json_dump)"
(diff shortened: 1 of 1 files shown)


updateing:
Actually all three if-elif branches have the same problem, but the other 2 branches have checks for expressions which can not be true if corresponding protocol is not chosen. Only the branch for 464XLAT exposed this issue. Shall I add `-n` checks for other branches?

dedeckeh:
Not checking for an empty IFACE_464xlat is on purpose as in this case the 464xlat interface name will be created automatically as ${INTERFACE}_4. Only if iface_464xlat is set explicitly as 0 464xlat interface creation will be disabled in all other cases it will either use the configured 464xlat interface name or will create automatically an interface name if iface_464xlat is either empty or set to 1.

PR #970, opened 2017-03-19
ipq806x: scm firmware fixes

dissent1:
**1. ipq806x: remove scm firmware clocks**
At the moment as a workaround definition for scm firmware in DT is used as if it is apq8064 board. This leads to incomplete scm firmware initialization and as a result cpuidle driver fails to configure.

By design unlike other qcom boards ipq do not use clocks to connect to scm.

Considering this we're removing clocks for ipq boards from DT and scm driver.

As a result cpuidle does not produce errors about failed configuration anymore.

**2. ipq806x: route gpio interrupts to APPS processor through scm firmware**
For IPQ806x targets, TZ protects the registers that are used to configure the routing of interrupts to a target processor.
To resolve this, this patch uses scm call to route GPIO interrupts to application processor. Also the scm call interface is changed.

@hnyman

### DIFF
--- a/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064.dtsi
+++ b/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064.dtsi
@@ -206,10 +206,7 @@
 
 	firmware {
 		scm {
-			compatible = "qcom,scm-apq8064";
-
-			clocks = <&rpmcc RPM_DAYTONA_FABRIC_CLK>;
-			clock-names = "core";
+			compatible = "qcom,scm-ipq806x";
 		};
 	};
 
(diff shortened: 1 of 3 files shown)


hnyman:
Seems to work for R7800 with ipq8065 (at r3778-312b9dcd65).
Router boots ok and there is one error less in boot log.

```
[    2.525196] cpuidle: enable-method property 'qcom,kpss-acc-v1' found operations
[    2.525242] CPUidle arm: CPU 0 failed to init idle CPU ops
-->
[    2.525340] cpuidle: enable-method property 'qcom,kpss-acc-v1' found operations
[    2.525450] cpuidle: enable-method property 'qcom,kpss-acc-v1' found operations
```
But I am not quite sure how the actual effects are otherwise visible in a running router.

blogic:
merged, Thanks !

PR #972, opened 2017-03-20
mvebu: Enable eSATA on Linksys WRT-devices

diizzyy:
Enable eSATA port on WRT1200AC, WRT1900AC, WRT1900ACv2, WRT3200ACM
and WRT1900ACS by default.


heinzek:
Hi @diizzyy. I use WRT3200ACM with 3,5' HDD connected to esata port but I not use any kmod-ata* package.
Disk is detected wthout any additional packages on current stable LEDE build (except kmod-fs-ext4).

diizzyy:
Yep, you're right... I shouldn't do this half asleep. Thanks!

PR #973, opened 2017-03-20
mvebu: Enable USB-support by default

diizzyy:
Enable basic USB-support for mvebu


anomeome:
Does adding kmod-usb2 kmod-usb3 bring more to the party? I thought all USB support was now baked into the kernel and including those was no longer required; at least I do not add them and have USB support, but I only use extx FS. Some LED action kmod-usb-ledtrig-usbport?

diizzyy:
You're correct, it never occurred to me to test without any USB-modules.

PR #974, opened 2017-03-21
ramips: add initial support for TP-LINK TL-WR840N v4

svobodavac:
This patch adds initial support for wireless router TP-LINK TL-WR840N v4.

Specification:
- SoC: MediaTek MT7628NN (580MHz)
- RAM: 64 MiB (Winbond W9751G6JB-25)
- Flash: 8 MiB (GigaDevice GD25Q64CSIG)
- LAN: 4x 100M
- WAN: 1x 100M
- Antennas: 2x non-detachable
- LED: 5x green
- Buttons: 1x reset

Instalation:
TP-Link firmware uses recent firmware v3 header, please see
  https://github.com/xdarklight/mktplinkfw3.
Therefore direct upgrade from TP-Link web UI is not supported yet.

Upgrade from the TP-Link firmware is possible using common TFTP recovery mode.
Rename file lede-ramips-mt7628-tl-wr840n-v4-squashfs-factory.bin
to tp_recovery.bin and put it on a TFTP server with IP 192.168.0.66.
Hold reset buton when connecting power until TFTP transfer begins.

For upgrade from LEDE use generic sysupgrade method.

LEDE uses uImage firmware header instead of the TP-Link v3 firmware
header. In order to  recover the original TP-Link firmware please
use TFTP recovery mode. The header has to be removed from
the original firmware:
  dd if=TL-WR840Nv4_0.9.1_3.16_up_boot\(161011\).bin \
    bs=512 skip=1 > tp_recovery.bin

Currently not working:
The only LAN LED indicates activity only on LAN port 2. The LAN LED
seems to be hardwired to the internal switch, port 2.


target/linux/ramips/dts/TL-WR840NV4.dts
```diff
@@ -0,0 +1,142 @@
...
+			gpios = <&gpio1 4 GPIO_ACTIVE_LOW>;
+		};
+
+		wireless {
+			label = "tl-wr840n-v4:green:wireless";
+			gpios = <&gpio1 12 GPIO_ACTIVE_LOW>;
+		};
+
+		ethernet {
+			label = "tl-wr840n-v4:green:ethernet";
```
heinzek: I think better name for this led is lan.

target/linux/ramips/dts/TL-WR840NV4.dts
```diff
@@ -0,0 +1,142 @@
...
+			gpios = <&gpio1 10 GPIO_ACTIVE_LOW>;
+		};
+
+		internet {
+			label = "tl-wr840n-v4:green:internet";
```
heinzek: and here wan

target/linux/ramips/dts/TL-WR840NV4.dts
```diff
@@ -0,0 +1,142 @@
+/dts-v1/;
+
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
```
heinzek: wrong alphabetical first gpio.h than input.h

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -323,6 +323,12 @@ tiny-ac)
 	set_wifi_led "$board:orange:wifi"
 	set_usb_led "$board:green:usb"
 	;;
+tl-wr840n-v4)
+	set_wifi_led "$board:green:wireless"
+	ucidef_set_led_netdev "lan" "lan" "$board:green:ethernet" eth0.1
+	ucidef_set_led_netdev "wan" "wan" "$board:green:internet" eth0.2
+	ucidef_set_led_default "wps" "wps" "$board:green:wps" "0"
```
heinzek: Drop ucidef_set_led_default here. They shouldn't be required.

target/linux/ramips/dts/TL-WR840NV4.dts
svobodavac: Fixed.

target/linux/ramips/dts/TL-WR840NV4.dts
```diff
@@ -0,0 +1,142 @@
...
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio1 6 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
```
svobodavac: Fixed.

target/linux/ramips/dts/TL-WR840NV4.dts
svobodavac: I do agree. User gude contains description "ethernet", but fixed.

target/linux/ramips/dts/TL-WR840NV4.dts
svobodavac: Fixed.

target/linux/ramips/base-files/etc/board.d/01_leds
svobodavac: Fixed.

svobodavac:
> Hi. Did You try fix it by adding "ephy" or "p0led_an", "p1led_an", "p2led_an", "p3led_an", "p4led_an", to pinctrl0?

I have tried all of them - ephy and pxled_an.
It seems the p2led_an is the right one for MT7628AN, port2:

```
&pinctrl {
         state_default: pinctrl0 {
                gpio {
                        ralink,group = "i2c", "i2s", "uart1", "uart2", "p2led_an" ;
                        ralink,function = "gpio";
                };
        };
 };
```
I'm unable to switch pins 39-43 to gpio and manually control LAN LEDS.
Wifi LED (pin 44, wled_an) is probably set to gpio and I can turn the LED on/off.

Please, any idea?

target/linux/ramips/image/mt7628.mk
```diff
@@ -26,6 +30,17 @@ define Device/miwifi-nano
 endef
 TARGET_DEVICES += miwifi-nano
 
+define Device/tl-wr840n-v4
+  DTS := TL-WR840NV4
+  IMAGE_SIZE := 7808k
+  DEVICE_TITLE := TP-Link TL-WR840N v4
+  DEVICE_PACKAGES :=
+  KERNEL := $(KERNEL_DTB) | prepend-to 448 | uImage lzma
```
mkresin: please use the existing ```pad-extra``` instead of your custom prepend-to.

```
pad-extra 448 | $(KERNEL_DTB) | uImage lzma
```

svobodavac: I'm afraid it's not possible.
`KERNEL_DTB = kernel-bin | patch-dtb | lzma`
Calling _pad-extra_ before making KERNEL_DTB causes, that all leading zeroes crated by pad-extra are subject of _lzma_ a will be comprimed:(
I can't find better/simpler solution.

mkresin:
Totally missed that you managed to fix the LED issue. Better add a new comment next time so that everyone gets noticed.

I'm still struggling to understand why this extra padding is required. What I might to understand so far is that the 128KByte extra padding of the factory image is used to ensure that the tftp recovery doesn't overwrite the bootloader.

The 448 bytes extra padding is required since you don't create a valid tp-link header, right? 

Assuming that the bootloader works the same way as the lantiq tp-link uboots, the padding can be dropped by using a valid tp-link header as it is done in target/linux/lantiq/image/tp-link.mk. It still doesn't create a valid v3 tp-link header that is accepted by the tp-link webinterface, but it creates a tp-link header where the update u-boot flag is unchecked.

Any chance that you give it a try or are you done with it?

svobodavac:
@mkresin 
Extra padding:
1) Uboot TFTP recovery mode takes all the firmware file, but it strips TP-Link header (size of 0x200 B - 512 B) and boot partition (size of 0x20000 B - 128 kB). Then it writes rest of the firmware file to the firmware partition (including second TP-Link partition).
So we need prepend fake TP-Link header and fake boot partition to factory LEDE image.
It seems the uBoot TFTP recovery mode doesn't rewrite the boot partition. (I can confirm this behavior just for current bootloader version;-))

2) When we create the entire LEDE firmware we use uImage header (64B) instead of TP-Link header (512 B) in order to better parse firmware and to find the beginning of the root_fs.
It also help us to provide sysupgrade using uImage a standard way.
We have to put extra padding (512-64B) after the uImage header to ensure the code entry point is on address where the bootloadrer expects it.

3) Yes, I can give it a try to use the TP-Link header. Thank you for the hint with TP-Link MTD splitting.

mkresin:
@svobodavac any update on using the TP-Link header instead of faking it? Would you please rebase your PR to fix the conflicting files.

sanitariu:
So far router seems to work. I found only one major problem. This is wifi txpower. Seems it is working on 25mw or sort of power which is too low. You loose the device in the next room.
Anyone knows a way to make the txpower working normal like 50-100mw ?

root@telnet-router:~# iwinfo wlan0 info
wlan0     ESSID: "LEDE-ROUTER"
          Access Point: 18:D6:C7:DF:12:46
          Mode: Master  Channel: 5 (2.432 GHz)
          Tx-Power: unknown  Link Quality: unknown/70
          Signal: unknown  Noise: unknown
          Bit Rate: unknown
          Encryption: mixed WPA/WPA2 PSK (CCMP)
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: unknown [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

root@telnet-router:~# iwinfo wlan0 txpowerlist
   0 dBm (   1 mW)
   1 dBm (   1 mW)
   2 dBm (   1 mW)
   3 dBm (   1 mW)
   4 dBm (   2 mW)
   5 dBm (   3 mW)
   6 dBm (   3 mW)
   7 dBm (   5 mW)
   8 dBm (   6 mW)
   9 dBm (   7 mW)
  10 dBm (  10 mW)
  11 dBm (  12 mW)
  12 dBm (  15 mW)
  13 dBm (  19 mW)
  14 dBm (  25 mW)
  15 dBm (  31 mW)
  16 dBm (  39 mW)
  17 dBm (  50 mW)
  18 dBm (  63 mW)
  19 dBm (  79 mW)
  20 dBm ( 100 mW)

Power selection is missing.

mkresin:
I will close this Pull Request due to inactivity. Please reopen the PR if you have the time to fix the outlined issues.

fapirini92:
oh boy ... It's a shame I know nothing abt this so I could help.I was looking foward to it because i bought this router 3 weeks ago w/o knowing it hasn't custom firmwares =/ ... Well, I hope some good guy can keep this request going ... I'm cheering fo it =)

dim45:
help error "git apply 974.patch"

target/linux/ramips/dts/TL-WR840NV4.dts
```diff
@@ -0,0 +1,141 @@
...
+	status = "disabled";
+};
+
+&ohci {
+	status = "disabled";
+};
+
+&wmac {
+	status = "okay";
+	mtd-mac-address = <&factory 0xf100>;
```
worm202: I'm trying to investigate poor wireless performance, which was described by @sanitariu 
The thing I noticed here is that the EEPROM address was not defined, so it uses default value from mt7628an.dtsi, which is wrong for this hardware:
`mediatek,mtd-eeprom = <&factory 0x0000>;`
Real EEPROM is located in radio partition, so we must declare this:
```
&spi0 {
...
	m25p80@0 {
...
		radio: partition@07f0000 {
...
&wmac {
...
	mediatek,mtd-eeprom = <&radio 0x0000>;
};
```

But performance is still bad though.

sanitariu: Did you make some patch for the correct eeprom ?

9000000:
Can you give me the link to download the firmware

gleidsonhenriques:
I'm also interested in this firmware.

sanitariu:
No link. Just get the source and compile. Wifi performance is really bad so do not use in production.
I tried to patch as worm202 pointed but no luck. Also tried other stuff from .dtsi but no luck.

gleidsonhenriques:
@sanitariu Overall performance or distance performance ?

f85ac21de1:
@sanitariu im interested in computing performance not Wifi, how can i compile it?

sanitariu:
Its not so hard. Just clone the code. Do make menuconfig choose the arch you like and then compile with make. Output files will be in bin folder.
About 840 ver.4 wifi - problem is with TX power. It's too low i think default like 15-30mw or internal antenna working only (distance performance).

alangregory:
@sanitariu Do you think wifi performance is related to TX power problem ?

alangregory:
Support for 840Nv4 and 841Nv13 was added recently , please check 24043a0d2e01b9843c0dc529205b3b0bc7ecbbf9 . I'm  testing some wireless performance problems. It would be of great help if ppl could test and give feedback.

sanitariu:
Wifi is still bad. The command "iwinfo wlan0 txpowerlist" does not show actual txpower but i think it is set default 30 or 50mw. I compared to tp-link 740 version 4 which is 63mw and i see it stronger in the next room.

target/linux/ramips/dts/TL-WR840NV4.dts
Lipton2kirov: You are the author of this https://raxp.ru/index.php/2017/06/21/lede-for-wr841n-v13wr840n-v4-ru/ ?
Can share the developments?

mkresin:
Please use the [LEDE forum ](https://forum.lede-project.org/) for any further discussions or send a mail to the [LEDE devel mailing list](https://lede-project.org/contact).

This PR was never merged, support for the TL-WR840N was added with https://github.com/lede-project/source/commit/24043a0d2e01b9843c0dc529205b3b0bc7ecbbf9.

PR #979, opened 2017-03-21
target.mk: check that CPU_TYPE has known CPU_CFLAGS mapping

pprindeville:
If someone creates a target and indicates a CPU_TYPE, but there's
no corresponding support for that CPU_TYPE's flags in include/target.mk
then that should probably be indicated rather than silently ignored.


pprindeville:
@hauke @nbd168 @ffainelli @KanjiMonster @jow- @blogic

nbd168:

PR #980, opened 2017-03-21
gpio-ich/itco-wdt/w83627hf-wdt: add modules for Xeon platforms

pprindeville:
Drivers for supporting current Xeon motherboards.

### DIFF
--- a/package/kernel/linux/modules/other.mk
+++ b/package/kernel/linux/modules/other.mk
@@ -1018,3 +1018,18 @@ define KernelPackage/tpm-i2c-infineon/description
 endef
 
 $(eval $(call KernelPackage,tpm-i2c-infineon))
+
+
+define KernelPackage/w83627hf-wdt
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=Winbond 83627HF Watchdog Timer
+  KCONFIG:=CONFIG_W83627HF_WDT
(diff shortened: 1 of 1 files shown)


blogic:
please run "git log --oneline" ont he folders you are patching to see what the patches prefix should be. in this case it would need to be "kernel:"

nbd168:
I think these should probably be enabled in the x86/64 kernel config instead of being packaged.

pprindeville:
> please run "git log --oneline" ont he folders you are patching to see what the patches prefix should be. in this case it would need to be "kernel:"

@blogic: fixed, thanks.

pprindeville:
> I think these should probably be enabled in the x86/64 kernel config instead of being packaged.

@nbd168: that might make sense for the ich and itco drivers; the w83627hf is obviously a 3rd party chipset and is less common.

Occasionally you end up with situations where two different drivers will match the same chipset (for instance, the w83627ehf and the nct6775 will both match the same PCI ID's, but the nct6775 is the better driver for the nct6775/6 (because it's more complete) and, equally importantly, isn't deprecated.

nbd168:
Okay, please change the commit to only package w83627hf

pprindeville:
> Okay, please change the commit to only package w83627hf

Done!

blogic:
merged, Thanks !

PR #983, opened 2017-03-21
x86: add subtargets and profiles for atom64-based platforms

pprindeville:
Add a generic Atom64 target based on the Pineview microarchitecture
as well as a specific instance of an Atom64-based profile such as
the Lanner fw-8756 network appliance.


diizzyy:
Why can't this use the generic 64-bit profile?

pprindeville:
> Why can't this use the generic 64-bit profile?

They all could, but the generic x86_64 profile was intended for K8 generation processors, now 14 years old.

We have 17 variants of CPU_FLAGS_* for ARM.  Why?  Because the variants are sufficiently different that one build version can't apply to all of them.  You could in theory use the ARMv7 core instruction set and generate images that would run on everything from an A5 to an A9.  You'd of course miss out on all of the optimization and specialized IP that the A9 bundles, but it would be possible.  In fact you could even run it on A53 and A57 hardware... but you'd be ignoring the 64-bit capabilities.

This applies as much to x86_64 hardware.

diizzyy:
There's actually very little performance difference between 64-bit cores overall hence why you don't see several different versions of it in distros. Do you have a use-case when there's actually a significant difference?

nbd168:
I think it makes no sense at all to add a new subtarget for atom64. These devices should run just fine with the generic x86/64 subtarget. If you still think that the CFLAGS are important for performance reasons, then please show some hard numbers to back this up.
I'm pretty sure that it either doesn't matter at all, or can be improved by changing the subtarget to optimize for a newer CPU by default.

pprindeville:
> Do you have a use-case when there's actually a significant difference?

@stintel was reporting that using the AES-NI driver, IPsec ESP throughput went from 63MB/s to 140MB/s, a more than doubling of throughput.

That's using the encryption co-processor that the Xeon and iCore processors have.

I'll see if I can compare some numbers.

But the other reason for having a target or profile is so that there's an off-the-shelf recipe which includes the correct bill-of-materials, kernel configuration, etc. that's required.

diizzyy:
The AES-NI driver should work fine if you just add it.

pprindeville:
> The AES-NI driver should work fine if you just add it.

And that would accelerate AES-CBC and -EBC encryption, sure.  But nothing else.  My point was that if a point optimization for block ciphering can achieve that sort of performance increase, then perhaps compiling the kernel and apps to the architecture can get a global improvement as well, even if it's not as marked in all places.

I'll see if I can find time to do some benchmarking.

Perhaps @dwmw2 can point me at where we'd see the biggest gains on a Xeon platform (Broadwell in my case) running Westmere vs. K8 instructions and scheduling.

target/linux/x86/atom64/config-default
```diff
@@ -0,0 +1,264 @@
+CONFIG_64BIT=y
```
alive4ever: Rather than introducing a new target, include necessary kernel configurations on ``target/linux/86/64/config-default``.

target/linux/x86/atom64/profiles/001-fw8756.mk
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (C) 2006-2015 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+define Profile/fw8756
+  NAME:=fw8756
+  PACKAGES:=kmod-w83627-wdt kmod-hwmon-nct6775 lcdproc
```
alive4ever: For this specific package inclusion, ``target/linux/x86/64/target.mk`` can be used.

target/linux/x86/atom64/target.mk
```diff
@@ -0,0 +1,29 @@
+ARCH:=x86_64
+BOARDNAME:=atom64
+FEATURES:=squashfs ext4 pcie usb gpio rtc display -vmdk
+DEFAULT_PACKAGES += kmod-button-hotplug kmod-e1000e kmod-igb kmod-ixgbe \
```
alive4ever: The package selection can be included on ``target/linux/x86/64/target.mk``.

PR #984, opened 2017-03-21
x86_64: add support for Xeon-based servers and appliances

pprindeville:
x86_64: add support for Xeon-based servers and appliances
    
Assumes 64-bit native, and Westmere or more recent processor.
    

diizzyy:
Not sure if we want to randomly add more sub-targets, I'm going to assume it's mainly because you want newer CPU instructions?

pprindeville:
> Not sure if we want to randomly add more sub-targets, I'm going to assume it's mainly because you want newer CPU instructions?

A few reasons.  I want the "xeon" target for generic 1U pizza box servers like Lanners, Supermicros, Dells, etc.

I want the "fw8771" target as a concrete example of a fleshed out box with all of the associated packages (both modules and support packages, like `rngd`, `pciutils`, `lm-sensors`, etc).

And lastly, I want to have an image optimized to the architecture to see how much better that performs than a very generic `x86_64` lowest-common denominator instruction set with very naive scheduling assumptions.  And @stintel has done some useful work in modularization for various encryption and hashing algorithms that should have a target to pair them up with (his AES-NI work doubled the throughput of IPsec ESP, for instance).

I understand that we have limited infrastructure for doing automated builds; but that's an argument for improving the build system, or perhaps increasing the appeal of LEDE so that more build servers get donated.  I don't think it's an argument for limiting the number of targets.

Perhaps we need a ranking system for how often particular targets get built, as a way of prioritizing the use of scarce build resources.

diizzyy:
All these should be run-time detectable by the software so adding it to the stock profile shuoldn't be an issue?

nbd168:
I agree with @diizzyy, support for xeon servers should simply be added to the regular x86/64 subtarget

pprindeville:
> All these should be run-time detectable by the software so adding it to the stock profile shuoldn't be an issue?

But the x86_64 generic image is compiled using the k8 instruction set and pessimistic scheduling.

So I don't understand what you're suggesting.  Are you suggesting fat binaries with several different instruction streams and detecting the appropriate one at run-time?  That's not widely supported.  And it's mostly done for .so's.

target/linux/x86/xeon/config-default
```diff
@@ -0,0 +1,262 @@
+CONFIG_64BIT=y
```
alive4ever: Instead of introducing a new build target, the kernel configuration changes should be included in ``target/linux/x86/64/config-default``.

The configuration can be done with ``make CONFIG_TARGET=subtarget kernel_menuconfig`` after selecting ``x86_64`` as subtarget.

target/linux/x86/xeon/target.mk
```diff
@@ -0,0 +1,11 @@
+ARCH:=x86_64
+BOARDNAME:=xeon
+DEFAULT_PACKAGES += kmod-button-hotplug kmod-e1000e kmod-e1000 kmod-igb \
```
alive4ever: Default packages can be included on ``target/linux/x86/64/target.mk``.

nbd168:
Feel free to propose new CFLAGS that should run well on most x86_64 processors without affecting xeon too much. If you're still convinced that generic CFLAGS hurt xeon performance too much, you need to provide some numbers to back that up.

PR #986, opened 2017-03-22
libs/openssl: Refresh mirror list

diizzyy:
Refresh mirror list, some doesn't offer OpenSSL and add main site as last resort.
Source: https://www.openssl.org/source/mirror.html


blogic:
merged, Thanks !

PR #987, opened 2017-03-22
libs/libpcap: Rework URLs

diizzyy:
Add mirror and use main site as last resort.
Source: http://www.tcpdump.org/mirrors.html


blogic:
merged, Thanks !

PR #990, opened 2017-03-22
utils/util-linux: Disable more

diizzyy:
Disable more as it fails to build on some platforms as a quick-fix.


diizzyy:
I'm testing this right now but build box isn't all that fast so please wait a little while for merging.

diizzyy:
Still breaks...

PR #993, opened 2017-03-22
grub2: update to 2.02~rc2

alive4ever:
Update to version 2.02~rc2.

Full changelog can be found [here](https://git.savannah.gnu.org/cgit/grub.git)


nbd168:

PR #994, opened 2017-03-23
hostapd: fix reload frequency change patch

abhilashtuse:
When sta is configured, hostapd receives 'stop' and 'update' command from
wpa_supplicant. In the update command, hostapd gets sta parameters with
which it configures ap.

Problem is, with the default wireless configuration:
mode:11g freq:2.4GHz channel:1
If sta is connected to 5GHz network, then ap does not work. Ideally with
340-reload_freq_change.patch hostapd should reload the frequency changes
and start ap in 5GHz, but ap becomes invisible in the network.

This issue can be reproduced with following /etc/config/wireless:
config wifi-device  radio0
        option type     mac80211
        option channel  1
        option hwmode   11g
        option path     'virtual/uccp420/uccwlan'
        option htmode   'none'

config wifi-iface 'ap'
        option device 'radio0'
        option encryption 'none'
        option mode 'ap'
        option network 'ap'
        option ssid 'MyTestNet'
        option encryption none

config wifi-iface 'sta'
       option device radio0
       option network sta
       option mode sta
       option ssid TestNet-5G
       option encryption psk2
       option key 12345

This change updates current_mode structure based on configured hw_mode
received from wpa_supplicant. Also prepare rates table after frequency
selection.

Signed-off-by: Abhilash Tuse <Abhilash.Tuse@imgtec.com>

### DIFF
--- a/package/network/services/hostapd/patches/340-reload_freq_change.patch
+++ b/package/network/services/hostapd/patches/340-reload_freq_change.patch
@@ -1,6 +1,56 @@
+From 359b11bacffdaa7ae0b636f0ece67c0a314f8bda Mon Sep 17 00:00:00 2001
+From: Abhilash Tuse <Abhilash.Tuse@imgtec.com>
+Date: Mon, 20 Mar 2017 14:33:23 +0530
+Subject: [PATCH] hostapd: fix reload frequency change patch
+
+When sta is configured, hostapd receives 'stop' and 'update' command from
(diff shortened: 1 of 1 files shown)


wongsyrone:
Does this have been sent upstream?

abhilashtuse:
This patch just updates existing 340-reload_freq_change.patch
And its not sent to hostapd upstream.

nbd168:
Merged to my staging tree with some cleanup

PR #997, opened 2017-03-24
ar71xx: wpj531: fix GPIOs for LED

yogo1212:
The existing values match those in the compex wiki and this document:

http://www.compex.com.sg/wp-content/uploads/2017/01/WPJ531-Hardware-Guide-v1.pdf

That document also contains different values for GPIO pins in the `LED Header` table (`Up pin`).
Those second values produce working LED initialisation.

On 7A03, the LEDs are working with this patch.
Should someone with access to a 7A04 confirm?


blogic:
merged, Thanks !

PR #998, opened 2017-03-24
kernel: Add kmod-crypto-crc32 (crc32_generic.ko)

InBetweenNames:
This is needed for F2FS support (it depends on crc32_generic)

Signed-off-by: Shane Peelar <lookatyouhacker@gmail.com>

### DIFF
--- a/package/kernel/linux/modules/crypto.mk
+++ b/package/kernel/linux/modules/crypto.mk
@@ -307,6 +307,17 @@ endef
 
 $(eval $(call KernelPackage,crypto-crc32c))
 
+define KernelPackage/crypto-crc32
+  TITLE:=CRC32 CRC module
+  DEPENDS:=+kmod-crypto-hash
+  KCONFIG:=CONFIG_CRYPTO_CRC32
+  HIDDEN:=1
+  FILES:=$(LINUX_DIR)/crypto/crc32_generic.ko@ge4.9
+  AUTOLOAD:=$(call AutoLoad,04,crc32_generic@ge4.9,1)
+  $(call AddDepends/crypto)
+endef
+
+$(eval $(call KernelPackage,crypto-crc32))
 
 define KernelPackage/crypto-des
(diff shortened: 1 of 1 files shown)


diizzyy:
@bobafetthotmail 
Can you test this? You didn't need to install this?

InBetweenNames:
It worked on my system -- an RT-AC87U (bcm53xx) with kernel 4.9.14.  Without this module, I was getting an error with F2FS whenever I tried to mount the drive:

F2FS: Cannot load crc32 driver

Including this module fixed that.  Apparently the CRC32C module is not sufficient for F2FS.

diizzyy:
Hmm.. I did a quick Google on this issue and it seems like ~4.6+ requires this module. I didn't go the kernel commit log but it seems correct as far as I can tell.  You can make an addition pull req where it's added as a dep for the f2fs kernel module?

InBetweenNames:
Sure thing!

InBetweenNames:
I created a separate PR to modify the F2FS dependencies: https://github.com/lede-project/source/pull/999

bobafetthotmail:
@diizzyy my devices have kernel 4.4, no issues with f2fs there.

diizzyy:
I see, I think it's related to the kernel bump as I mentioned above.

InBetweenNames:
Indeed, this doesn't seem to be required for 4.4 (at least I had no problems).  It's only for 4.9.x that I had to have this module installed for F2FS to work.

diizzyy:
@nbd168 
Should that module be included in 4.9 by default or is there a way to add dep depending on kernel version?

nbd168:
In the FILES and AUTOLOAD part, add @ge4.9 after the module name, afterwards you can unconditionally select this package from the f2fs module.
Please also mark it as HIDDEN.

InBetweenNames:
Done!  See my other pull request https://github.com/lede-project/source/pull/999 for the F2FS dependency fix.

blogic:
merged, Thanks !

PR #1000, opened 2017-03-24
Cleanup vmmc patches

kochstefan:
- First patch: remove unused code, ... (compile tested with xway, run time tested with xrx200)
- Second patch: align falcon code (not compile tested, in general vmmc builds not for falcon because some headers are not available)


blogic:
merged, Thanks !

PR #953, opened 2017-03-15
package/boot: update uboot-kirkwood to 2017.03

p-wassi:
Maintainer: @lperkov 
Compile tested: kirkwood
Run tested: kirkwood/dockstar

Description:
Bump uboot-kirkwood to upstream release 2017.03
Logs available at https://pwassi.privatedns.org/lede/dockstar/

PR #1001, opened 2017-03-25
package/boot/uboot-kirkwood: fix build errors for nsa310 and nsa325

p-wassi:
Maintainer: @lperkov

As @mkresin has pointed out in [1], my last update introduced build errors for the
nsa310 and nsa325 targets of this package. This is related to some upstream changes.
It's fixed here; every uboot-kirkwood target is compile tested.
Maybe @bobafetthotmail can run test the nsa310/nsa325 targets?

Some parts of the configuration must not be done in the header file any more,
but instead in the <target>_defconfig file. Further, some warnings due to redefinitions
in both of those files have been clean up

[1]:
https://github.com/lede-project/source/pull/953#issuecomment-288877812

### DIFF
--- a/package/boot/uboot-kirkwood/patches/007-nsa310-uboot-generic.patch
+++ b/package/boot/uboot-kirkwood/patches/007-nsa310-uboot-generic.patch
@@ -528,10 +528,11 @@ new file mode 100644
 index 0000000..d26ef35
 --- /dev/null
 +++ b/configs/nsa310_defconfig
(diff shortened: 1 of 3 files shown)


bobafetthotmail:
thanks, will test them today or tomorrow and report back.

bobafetthotmail:
Ok it's working on both nsa310 and nsa325.

nbd168:

PR #1003, opened 2017-03-25
ar8xxx: merge QSDK fixes for switch driver

dissent1:
Merging fixes for ar8xxxx driver found in Qualcomm SDK repository.

This commit:
- changes MTU size configuration logic
- adds certain switch HOL registers configuration during init as
proposed by QCA switch dev team
- adds support for ar8327 switch revision found in ap136/ap137/ap143/ap147/ap152
- some coding clean ups

https://source.codeaurora.org/quic/qsdk/oss/kernel/linux-msm/commit/drivers/net/phy/ar8216.c?h=release/endive_preview&id=f68b461945b3e68bdcc4c39b80045cc9a986652d
https://source.codeaurora.org/quic/qsdk/oss/kernel/linux-msm/commit/drivers/net/phy/ar8216.c?h=release/endive_preview&id=6ff699c6f50263910d188ed993dd3bc365956003

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/ar8216.c
+++ b/target/linux/generic/files/drivers/net/phy/ar8216.c
@@ -2106,6 +2106,7 @@ ar8xxx_phy_config_aneg(struct phy_device *phydev)
 static const u32 ar8xxx_phy_ids[] = {
 	0x004dd033,
 	0x004dd034, /* AR8327 */
+	0x004dd035,
 	0x004dd036, /* AR8337 */
 	0x004dd041,
 	0x004dd042,
--- a/target/linux/generic/files/drivers/net/phy/ar8327.c
+++ b/target/linux/generic/files/drivers/net/phy/ar8327.c
@@ -3,6 +3,7 @@
  *
  * Copyright (C) 2009 Felix Fietkau <nbd@nbd.name>
  * Copyright (C) 2011-2012 Gabor Juhos <juhosg@openwrt.org>
+ * Copyright (c) 2017 The Linux Foundation. All rights reserved.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -667,7 +668,7 @@ ar8327_init_globals(struct ar8xxx_priv *priv)
 	    (AR8327_PORTS_ALL << AR8327_FWD_CTRL1_BC_FLOOD_S);
 	ar8xxx_write(priv, AR8327_REG_FWD_CTRL1, t);
 
-	/* enable jumbo frames */
+	/* setup MTU */
 	ar8xxx_rmw(priv, AR8327_REG_MAX_FRAME_SIZE,
 		   AR8327_MAX_FRAME_SIZE_MTU, 9018 + 8 + 2);
 
@@ -975,9 +976,8 @@ ar8327_sw_set_ports(struct switch_dev *dev, struct switch_val *val)
 		struct switch_port *p = &val->value.ports[i];
 
 		if (p->flags & (1 << SWITCH_PORT_FLAG_TAGGED)) {
-			if (val->port_vlan == priv->pvid[p->id]) {
+			if (val->port_vlan == priv->pvid[p->id])
 				priv->vlan_tagged |= (1 << p->id);
-			}
 		} else {
 			priv->vlan_tagged &= ~(1 << p->id);
 			priv->pvid[p->id] = val->port_vlan;
@@ -1152,7 +1152,7 @@ ar8327_sw_hw_apply(struct switch_dev *dev)
 	if (ret)
 		return ret;
 
-	for (i=0; i < AR8XXX_NUM_PHYS; i++) {
+	for (i = 0; i < AR8XXX_NUM_PHYS; i++) {
 		if (data->eee[i])
 			ar8xxx_reg_clear(priv, AR8327_REG_EEE_CTRL,
 			       AR8327_EEE_CTRL_DISABLE_PHY(i));
@@ -1161,6 +1161,33 @@ ar8327_sw_hw_apply(struct switch_dev *dev)
 			       AR8327_EEE_CTRL_DISABLE_PHY(i));
 	}
 
+	return 0;
+}
+
+static int
+ar8327_sw_set_max_frame_size(struct switch_dev *dev,
+ 			const struct switch_attr *attr,
+ 			struct switch_val *val)
+{
+ 	struct ar8xxx_priv *priv = swdev_to_ar8xxx(dev);
+ 
+ 	ar8xxx_rmw(priv, AR8327_REG_MAX_FRAME_SIZE,
+ 			AR8327_MAX_FRAME_SIZE_MTU, val->value.i + 8 + 2);
+ 	return 0;
+}
+ 
+static int
+ar8327_sw_get_max_frame_size(struct switch_dev *dev,
+ 			const struct switch_attr *attr,
+ 			struct switch_val *val)
+{
+	u32 v;
+	struct ar8xxx_priv *priv = swdev_to_ar8xxx(dev);
+
+	v = ar8xxx_read(priv, AR8327_REG_MAX_FRAME_SIZE);
+	v &= AR8327_MAX_FRAME_SIZE_MTU;
+ 
+	val->value.i = v;
 	return 0;
 }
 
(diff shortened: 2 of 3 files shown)


target/linux/generic/files/drivers/net/phy/ar8216.c
```diff
@@ -296,6 +297,28 @@ ar8xxx_rmw(struct ar8xxx_priv *priv, int reg, u32 mask, u32 val)
...
+	bus->write(bus, 0x18, 0, page);
+	wait_for_page_switch();
+
+	ret = ar8xxx_mii_read32(priv, 0x10 | r2, r1);
+	ret &= mask;
+
+	mutex_unlock(&bus->mdio_lock);
+
+	return ret;
+}
```
nbd168: This function is rather useless, since it does the same thing as ar8xxx_read with the addition of returning ret & mask. Please remove it.

nbd168:
This should be split up into several commits.

target/linux/generic/files/drivers/net/phy/ar8216.c
dissent1: Aw, sorry, my bad, how stupid of me to not notice this. I'll add mask into it.

dissent1:
@nbd168 there's similar patch that sets HOL QM registers http://patchwork.ozlabs.org/patch/743523/
Which one do you want to go for? I'll split and rebase this commit if opposite is going to be accepted.

@VittGam

VittGam:
Hello @dissent1. I'd split this pull request in several commits too:

- Add 0x004dd035 to ar8xxx_phy_ids.
- Change back page sleep wait from 5us to 1-2ms.
- Allow MTU to be configured.

But I've got some questions:

- Disable jumbo frames by default: why?
- HOL registers and RGMII delay settings: this is the same as my patch, but I've properly split the magic values into register defines, and put them in the right places...
- Change global flow control threshold default value: it raises the global base transmit off threshold from 0x188 (chip default) to 0x1f0. But why do it only for AR8327? The AR8337 has the same 0x188 default...

Cheers,
Vittorio

VittGam:
Hello,

By the way, changing back page sleep wait from 5us to 1-2ms will greatly interfere with CPU usage when I try to implement the get_port_stats function (needed to make the switch LEDs blink on tl-wr1043nd-v4 for example).

I'm not sure this is really needed, since it seems to work fine even without the 1-2ms busy wait...

Cheers,
Vittorio

dissent1:
Sorry for late reply, I'm extremely busy at the moment.
Considering your questions:
1. Disable jumbo frames by default. At deeper look it doesn't seem relevant for me anymore. I need to do more tests. I may do it.
2. I think that your patch is better, but why not to leave RGMII_RXCLK_DELAY in INIT_GLOBALS?
3. Flow control threshold for ar8327 is a recommendation by qca, I'm not sure if it's needed for ar8337. I can assume that it may be also applicable, but that's only assumption.
4. Considering your report page sleep should better leave intact. How great is cpu usage increase?

blogic:
@dissent1 @VittGam both patches have good and bad parts. i will merge this into one patch now, test it on AP148 and then put it into my staging tree for testing today

VittGam:
Hi @dissent1, no problem with late reply, I'm late too this time, we're somehow all busy... :)

1. Did jumbo frames cause problems in your setup?
2. Since AR8327_REG_PAD5_MODE is already being set (with board-specific values, but pad5 mode usually is always left to 0) in hw_init, I thought it would be better to have this WAR there too. But in fact, with my patch, boards using device tree instead of platform_data are not actually getting the fix...
3. Yes, you're right, we can only assume... (This and the phy fixups seem to be the only two things done only for the ar8327; while the HOL registers and RGMII port5 delay setting are only done for ar8337 instead. For the rest, the two chips seem to be treated the same way.)
4. It's randomic, from 10% up to 50%... That's because the get_port_stats function has to query every port many times per second, if the LED switch trigger is enabled. Also, with the greater sleep time, more kworker threads are created; and they spend some time every 2-3 seconds in uninterruptible sleep "switching" the CPU load between themselves...

Hi @blogic, thanks for your feedback, I'll wait for your staging tree update so that I can test your final version of the patch on an AP152-based board too (TL-WR1043ND-v4, QCA956x + AR8337)

Cheers,
Vittorio

blogic:
@VittGam i have merged your patch as is with a minor correction to the commented out code.
@dissent1 i will rebase your patch tomorrow. need to do some other stuff today

dissent1:
@blogic I'll rebase myself, thanks :)

Also rgmii delay should better be in init_globals, I'll take care of that as well

VittGam:
@blogic Perfect, thanks!

@dissent1 I'd add it to `ar8327_hw_config_of` too, instead; something like this:
```patch
--- a/target/linux/generic/files/drivers/net/phy/ar8216.c
+++ b/target/linux/generic/files/drivers/net/phy/ar8216.c
@@ -608,6 +608,17 @@
 		}
 	}
 
+	if (chip_is_ar8337(priv)) {
+		/*
+		 * Workaround: RGMII RX delay setting needs to be
+		 * always specified for AR8337 to avoid port 5
+		 * RX hang on high traffic / flood conditions
+		 */
+		ar8xxx_rmw(priv, AR8327_REG_PAD5_MODE,
+			   AR8327_PAD_RGMII_RXCLK_DELAY_EN,
+			   AR8327_PAD_RGMII_RXCLK_DELAY_EN);
+	}
+
 	return 0;
 }
 #else
```

Otherwise, this one. It does not duplicate the code; but I'd prefer the first approach, as `init_globals` is run quite a lot after `hw_init`.
```patch
--- a/target/linux/generic/files/drivers/net/phy/ar8216.c
+++ b/target/linux/generic/files/drivers/net/phy/ar8216.c
@@ -506,14 +506,6 @@
 	ar8xxx_write(priv, AR8327_REG_PAD0_MODE, t);
 
 	t = ar8327_get_pad_cfg(pdata->pad5_cfg);
-	if (chip_is_ar8337(priv)) {
-		/*
-		 * Workaround: RGMII RX delay setting needs to be
-		 * always specified for AR8337 to avoid port 5
-		 * RX hang on high traffic / flood conditions
-		 */
-		t |= AR8327_PAD_RGMII_RXCLK_DELAY_EN;
-	}
 	ar8xxx_write(priv, AR8327_REG_PAD5_MODE, t);
 	t = ar8327_get_pad_cfg(pdata->pad6_cfg);
 	ar8xxx_write(priv, AR8327_REG_PAD6_MODE, t);
@@ -680,6 +672,15 @@
 		data->eee[i] = false;
 
 	if (chip_is_ar8337(priv)) {
+		/*
+		 * Workaround: RGMII RX delay setting needs to be
+		 * always specified for AR8337 to avoid port 5
+		 * RX hang on high traffic / flood conditions
+		 */
+		ar8xxx_rmw(priv, AR8327_REG_PAD5_MODE,
+			   AR8327_PAD_RGMII_RXCLK_DELAY_EN,
+			   AR8327_PAD_RGMII_RXCLK_DELAY_EN);
+
 		/* Update HOL registers with values suggested by QCA switch team */
 		for (i = 0; i < AR8327_NUM_PORTS; i++) {
 			if (i == AR8216_PORT_CPU || i == 5 || i == 6) {
```

Cheers,
Vittorio

dissent1:
@VittGam
I'm not sure there's noticeable negative impact if it's ran a bit later (in init_globals). Is there?

dissent1:
@nbd168 
Function ar8xxx_rmr adds additional argument 'mask' that is absent in ar8xxx_read. Adding mask into ar8xxx_read leads to necessity of adjusting all other functions that call it. 

Isn't adding ar8xxx_rmr along with ar8xxx_rmw (that has the same difference comparing to ar8xxx_write) is a better approach?

nbd168:
@dissent1 This makes no sense at all. Why would you even add a mask argument to the read function?

ar8xxx_rmr(priv, reg, mask) is equivalent to ar8xxx_read(priv, reg) & mask.
The latter does not need any extra functions, extra arguments and is actually more readable.

Unlike your extra function, ar8xxx_rmw actually makes sense because it saves a page switch compared to calling ar8xxx_read and ar8xxx_write.

dissent1:
@nbd168 
Okay, I've rebased, split, and adjusted the mtu frame size function.

dissent1:
If there's still smth wrong with this PR please let me know

target/linux/generic/files/drivers/net/phy/ar8327.c
```diff
@@ -680,7 +672,20 @@ ar8327_init_globals(struct ar8xxx_priv *priv)
...
+
 	if (chip_is_ar8337(priv)) {
+
+		/*
+		 * Workaround: RGMII RX delay setting needs to be
+		 * always specified for AR8337 to avoid port 5
+		 * RX hang on high traffic / flood conditions
+		 */
+		ar8xxx_write(priv, AR8327_REG_PAD5_MODE,
+			AR8327_PAD_RGMII_RXCLK_DELAY_EN);
```
nbd168: Did you really intend to overwrite the entire register here instead of just setting the extra bit?

blogic:

PR #1005, opened 2017-03-26
uhttpd: Enable integrated Lua by default

Ansuel:
If someone want to enable the integrated Lua interpreter, he needs to modify the config file to actually enable it. As the purpose of uhttpd-mod-lua is to actually enable it, we modify it by default.

Compiled and tested.

### DIFF
--- a/package/network/services/uhttpd/Makefile
+++ b/package/network/services/uhttpd/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=uhttpd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(LEDE_GIT)/project/uhttpd.git
@@ -46,8 +46,11 @@ define Package/uhttpd/config
   config PACKAGE_uhttpd_debug
     bool "Build with debug messages"
     default n
+  config uhttpd_lua
+    depends on PACKAGE_uhttpd-mod-lua
+    bool "Enable Integrated Lua interpreter"
+	default y
 endef
-
 
 define Package/uhttpd-mod-lua
   $(Package/uhttpd/default)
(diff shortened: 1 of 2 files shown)


package/network/services/uhttpd/Makefile
```diff
@@ -90,6 +93,9 @@ define Package/uhttpd/install
 	$(INSTALL_BIN) ./files/uhttpd.init $(1)/etc/init.d/uhttpd
 	$(INSTALL_DIR) $(1)/etc/config
 	$(INSTALL_CONF) ./files/uhttpd.config $(1)/etc/config/uhttpd
+	$(if $(CONFIG_uhttpd_lua),
+		sed '/option lua/s/#//g' -i $(1)/etc/config/uhttpd ,
```
wongsyrone: Why do you need the trailing semicolon? `$(if condition,then-part[,else-part])`

blogic:
imho this should be done in a postinst hook or simply ship the default config with lua enabled if that does not make uhttpd blow up when not installed.

shipping an ipkg that has a content dependent on dependency modules is always quirky

package/network/services/uhttpd/Makefile
Ansuel: You mean the extra space? Just to keep all tidy and more understandable.

Ansuel:
@blogic i have a question, in releation to this option i need to make a change to the index.html of luci, but i can't find a way to make the luck makefile check this condition... Any idea? 
(i will move that part to a postinst hook)

Ansuel:
This should be good (tested and working)

Ansuel:
Don't know if this is the right way but it works good.
(i add another "if" that check if the index page is present and modify it)

Ansuel:
@blogic @nbd168 what do you think?

Ansuel:
Can we include this?

hnyman:
That LuCI index.html change does not look that good, as it's success depends on the package installation order. If uhttpd-mod-lua is installed after luci-base, this might work, but if luci-base is installed later, index.html is going be unchanged.  (In general, manipulating files belonging to another package is rather questionable.)

Regarding the config file change, as the lines are commented with `#` in the shipped default config file, they may have been already removed from the config file by uci, if the user has edited any uhttpd options via uci. So, using sed to change the options is rather risky and may easily fail because of missing lines. Better approach would be to use uci commands directly to add the options with default values, if the options have not already been configured by the user.


I am not sure if the config editing & Luci index.html editing part of this PR are necessary. If a rarely used add-on package is installed, it is rather typical that some manual config changes are needed to the system.

Ansuel:
@hnyman what do you think about this? 
I tested and i find that it makes no change having the option with # or no... So we enabled it by default...

Can't find a way to make the index page modification in a better way...
(make a pr in the luci github? )

hnyman:
> @hnyman what do you think about this?

My knowledge about uhttpd internals is not good enough.
We need feedback from @jow- @nbd168 @blogic  or other uhttpd gurus.

jow-:
That change was unfortunately merged already. The current approach of modifying other packages files from postinstall is a no-go as it introduces undefined behaviour.

In general the LuCI packages need to support the uhttpd Lua interpreter, not the other way around.

PR #1006, opened 2017-03-27
ar71xx: add support for Mikrotik RB2011iL

rangeles:
From dmesg:

` [    0.000000] Kernel command line: parts=1 boot_part_size=4194304 gpio=249403 HZ=300000000 mem=64M kmac=4C:5E:0C:37:82:E6 board=2011r5 boot=0 mlc=5 rootfstype=squashfs noinitrd`
`[    0.126556] MIPS: machine is Mikrotik RouterBOARD 2011iL`

In kernel, the `2011r5` is mapped to `Mikrotik RouterBOARD 2011UiAS(-2Hnd) `. While not ideal, it still works for me. This just means that the kernel will attempt to turn on the RB2011UiAS's feature set: SFP, USB, and WLAN. These do not exist on the RB2011iL. 

The RB2011iL matches the RB2011L offerings more closely. So basically, any userland references to RB2011L should also be accompanied by RB2011iL.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -256,6 +256,7 @@ rb-952ui-5ac2nd)
 rb-962uigs-5hact2hnt)
 	ucidef_set_led_timer "user" "USER/SFP" "rb:green:user" "1000" "1000"
 	;;
+rb-2011il|\
 rb-2011l|\
 rb-2011uas|\
 rb-2011uias|\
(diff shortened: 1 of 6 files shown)


pepe2k:
Hello @rangeles, please use correct alphabetical order ("rb-2011il" should be in front of "rb-2011l", i < l) and update your PR.

Cheers,
Piotr

rangeles:
@pepe2k, thanks for the review. I've updated the patch per latest comments.

pepe2k:
@rangeles,

> In kernel, the 2011r5 is mapped to Mikrotik RouterBOARD 2011UiAS(-2Hnd). While not ideal, it still works for me. This just means that the kernel will attempt to turn on the RB2011UiAS's feature set: SFP, USB, and WLAN. These do not exist on the RB2011iL.

If you think that this is a mistake, you can send PR with fix in mach-rb2011.c. The only question is if this is a real bug or the bootloader from other models of this board also provide `board=2011r5` in kernel command line.

Anyway, thank you for your contribution, I have merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) with just a small change in commit subject.

Cheers,
Piotr

PR #1008, opened 2017-03-27
build: Removes absolute path to perl and replaces by /usr/bin/env perl

bkchr:
This patch removes the absolute path to perl. This change is need for distributions like Nixos, that do not support this kind of absolute paths.

Signed-off-by: Bastian Köcher <git@kchr.de>

### DIFF
--- a/scripts/checkpatch.pl
+++ b/scripts/checkpatch.pl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl 
 # (c) 2001, Dave Jones. (the file handling bit)
 # (c) 2005, Joel Schopp <jschopp@austin.ibm.com> (the ugly bit)
 # (c) 2007,2008, Andy Whitcroft <apw@uk.ibm.com> (new conditions, test suite)
@@ -7,6 +7,7 @@
 # Licensed under the terms of the GNU GPL License version 2
 
 use strict;
+use warnings;
 
 my $P = $0;
 $P =~ s@.*/@@g;
--- a/scripts/cleanfile
+++ b/scripts/cleanfile
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl
 #
 # Clean a text file -- or directory of text files -- of stealth whitespace.
 # WARNING: this can be a highly destructive operation.  Use with caution.
@@ -6,6 +6,7 @@
 
 use bytes;
 use File::Basename;
+use warnings;
 
 # Default options
 $max_width = 79;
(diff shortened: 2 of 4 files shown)


alive4ever:
Are only those files that has ``perl`` location hardcoded?

This matters on BSD, which doesn't include perl in the base. On NetBSD, perl installation is under ``/usr/pkg/bin/``. FreeBSD puts the perl binary inside ``/usr/local/bin``.

I think the changes are okay and the perl scripts should run just fine and more portable with it.

NB: should this kind of change be applied for ``#!/bin/bash`` shebang as well, so that the bash scripts can be more portable?

bkchr:
As far as I know these are the only locations where the perl path is hard-coded for files in this repository. 

Yeah I think for bash it's the same :)

alive4ever:
If so, these changes are looking good to me.

bkchr:
Perfect :) So I hope this gets merged.

alive4ever:
Yeah, the ``./scripts/download.pl`` seems to have correct ``#!/usr/bin/env perl`` shebang instead of hardcoded perl location.

This should do no harm to be committed.

I know that maintaining a pull request without clear judgment from the project maintainers is somewhat troublesome. Project maintainers should have the guts to decide whether a PR should be committed or discarded, so the pull request submitter can move on to another task.

For now, let's be patient and wait.

scripts/checkpatch.pl
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl -w
```
alive4ever: ``#!/usr/bin/env perl -w`` is not valid.

Change this to ``#!/usr/bin/env perl`` and add ``use warnings;`` on the first script execution line.

scripts/cleanpatch
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl -w
```
alive4ever: See my previous comment.

scripts/cleanfile
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl -w
```
alive4ever: See my previous comment.

alive4ever:
``env`` will treat the argument afterwards as one literal command. These changes will be evaluated as literal ``perl -w`` command instead of ``perl`` with ``-w`` argument.

``/usr/bin/env: ‘perl -w’: No such file or directory``

Please make changes as I suggested.

bkchr:
Yeah will do that tomorrow! :)

bkchr:
Done :)

nbd168:

PR #1014, opened 2017-03-31
apm821xx: Merge Meraki MX60/MX60W images

riptidewave93:
The following patches do the following:
 * Set core packages used by the nand subtarget as default
 * Merge the MX60/MX60W boards, as they can now share the same image.

This is done to fix the current broken MX60W images, which when compiled in a multi image environment, lack swconfig and ath9k, making the device inaccessible.


nbd168:

PR #1018, opened 2017-03-31
ar71xx: add support for UniFi-AC-Mesh

Brother-Lal:
This adds the build option for the new UniFi AC Mesh.
It is a direct hardware copy from the AC Lite.

- SoC: QCA9563-AL3A (775Mhz)
- RAM: 128MiB
- Flash: 16MiB - dual firmware partitions!
- LAN: 1 1000M - POE
- Wireless:
        2.4G: QCA9563
          5G: UniFi Chip, QCA988X compatible

Thanks to Frank Dietz(@tuennes) for a lot of patience while testing.

See also https://github.com/freifunk-gluon/gluon/issues/1021

### DIFF
--- a/target/linux/ar71xx/image/ubnt.mk
+++ b/target/linux/ar71xx/image/ubnt.mk
@@ -133,6 +133,11 @@ define Device/ubnt-unifiac-lite
   BOARDNAME := UBNT-UF-AC-LITE
 endef
 
+define Device/ubnt-unifiac-mesh
+  $(Device/ubnt-unifiac-lite)
+  DEVICE_TITLE := Ubiquiti UniFi AC-Mesh
+endef
+
 define Device/ubnt-unifiac-pro
   $(Device/ubnt-unifiac)
   DEVICE_TITLE := Ubiquiti UniFi AC-Pro
@@ -147,7 +152,7 @@ define Device/ubnt-unifi-outdoor
   BOARDNAME := UBNT-U20
   DEVICE_PROFILE := UBNT UBNTUNIFIOUTDOOR
 endef
-TARGET_DEVICES += ubnt-unifi ubnt-unifiac-lite ubnt-unifiac-pro ubnt-unifi-outdoor
+TARGET_DEVICES += ubnt-unifi ubnt-unifiac-lite ubnt-unifiac-mesh ubnt-unifiac-pro ubnt-unifi-outdoor
 
 define Device/ubnt-nano-m-xw
   $(Device/ubnt-xw)
--- a/target/linux/ar71xx/base-files/lib/ar71xx.sh
+++ b/target/linux/ar71xx/base-files/lib/ar71xx.sh
@@ -1154,7 +1154,7 @@ ar71xx_board_detect() {
 	*"UniFi AP Pro")
 		name="uap-pro"
 		;;
-	*"UniFi-AC-LITE")
+	*"UniFi-AC-LITE/MESH")
 		name="unifiac-lite"
 		;;
 	*"UniFi-AC-PRO")
--- a/target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
+++ b/target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
@@ -1763,7 +1763,7 @@ config ATH79_MACH_UBNT
 	select ATH79_DEV_USB
 
 config ATH79_MACH_UBNT_UNIFIAC
-	bool "Ubiquiti UniFi AC (LITE/LR/PRO) support"
+	bool "Ubiquiti UniFi AC (LITE/LR/MESH/PRO) support"
 	select SOC_QCA956X
 	select ATH79_DEV_AP9X_PCI if PCI
 	select ATH79_DEV_ETH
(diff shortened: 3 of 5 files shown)


tuennes:
:-) Many thanks to you @Brother-Lal for your patience with a dumb LEDE Newbie...

target/linux/ar71xx/image/ubnt.mk
```diff
@@ -133,6 +133,14 @@ define Device/ubnt-unifiac-lite
   BOARDNAME := UBNT-UF-AC-LITE
 endef
 
+define Device/ubnt-unifiac-mesh
+  $(Device/ubnt-unifiac)
```
pepe2k: You could just put here `$(Device/ubnt-unifiac-lite)` instead and then overwrite just one variable which is different, the `DEVICE_TITLE`. Result: 3 lines less.

Brother-Lal: sounds like a more concise solution, will do

pepe2k:
@Brother-Lal, sorry for previous comment, I should first think about this more...

It looks ok, but I'm now just not convinced about having separate image for this device. There are no differences in code, image building code and the only real difference would be name in image filename.

What about dropping additional `ubnt-unifiac-mesh` image and just change define of `ubnt-unifiac-lite` image to something like (name for image can be discussed, it's just fast idea... also keep rest of the changes which are ok):

```
 define Device/ubnt-unifiac-lite-mesh
   $(Device/ubnt-unifiac)
   DEVICE_TITLE := Ubiquiti UniFi AC-Lite/AC-Mesh
   DEVICE_PACKAGES := kmod-ath10k ath10k-firmware-qca988x
   DEVICE_PROFILE := UBNT UBNTUNIFIACLITE
   BOARDNAME := UBNT-UF-AC-LITE
 endef
```
Cheers,
Piotr

NeoRaider:
We also have identical images for a lot of other devices, for example TP-Link used the same model IDs for its TL-WR740 and TL-WR741 series a while ago. I think adding multiple model names to the image filenames is confusing for users.

pepe2k:
> We also have identical images for a lot of other devices, for example TP-Link used the same model IDs for its TL-WR740 and TL-WR741 series a while ago. I think adding multiple model names to the image filenames is confusing for users.

For a lot? TL-WR741N/ND v1/v2 and TL-WR941ND v2/v3 are the only examples which I know about (and I was thinking about combining them into single image file). IMHO, it's wasting of resources (in terms of disk space and build time) when two identical images are generated with just difference in filename.

@nbd168 wrote about idea to have something like symbolic links instead of duplicated images in such cases but... it's still just an idea.

Cheers,
Piotr

Brother-Lal:
Just to give a heads up, the unifi ac mesh pro will probably be the same case, as it is identical to the unifi ac pro.
Maybe an alias command to link them under different names, because for the user who just looks through the build dir on a webserver different image files is the best option. No knowledge required which image to flash on which hardware, just see the name and it fits.

pepe2k:
@Brother-Lal
> Just to give a heads up, the unifi ac mesh pro will probably be the same case, as it is identical to the unifi ac pro.

OK, just to make it clear: I'm not against including your change (in fact, it's already in my local tree). I just wanted to point out that there is a problem with duplicated images (and I mean here exact 1:1 copy, not images with just difference in kernel cmdline or TP-Link header etc., we are talking here about copies which AFAIK are not generated with simple cp x y during building [someone corrects me if I'm wrong here]). And the ar71xx/ath79 is a great example of the target which, in terms of number of images, grows really fast.

> Maybe an alias command to link them under different names, because for the user who just looks through the build dir on a webserver different image files is the best option. No knowledge required which image to flash on which hardware, just see the name and it fits.

Yes, I agree, but on the other hand I don't think that user will be confused with e.g. `tl-wr741nd-v1-v2` or `ubnt-unifiac-lite-mesh` image filenames, but I might be wrong here. What's more, we have wiki and we can/should make use of it.

I would like to see also others comments about that, maybe someone is already working on some "alias/symlink" solution. Lets give them some time and wait for their comments.

Cheers,
Piotr

Brother-Lal:
I had a look at some more of the other implementations, and i probably could implement something along the lines of ubnt_xm_board_detect() with data from the eeprom partition.
At least the machine name will be correct after flashing, and the image name could be adapted to reflect usage for 2 devices.
Doesn't solve the problem with the duplicate images for the future, but probably would steer around it for now .

Brother-Lal:
I could not find any easily usable binary numbers that would point to different hardware - die location used in ubnt_xm_board_detect is working, but is sadly the same for ac lite & ac mesh and ac pro &ac mesh pro.

Can we just rename the target for both to ac (mesh)lite and ac(mesh) pro?
One image with a mixed name for both targets?

Or 2 copies of the image with just the different machine name?

pepe2k:
@Brother-Lal,

I have merged your changes with minor fix into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

When we will have some way for dealing with the issue/s discussed in this PR, we can make these boards use it.

Cheers,
Piotr

yunhai20082008:
> This adds the build option for the new UniFi AC Mesh. It is a direct hardware copy from the AC Lite.
> 
> * SoC: QCA9563-AL3A (775Mhz)
> * RAM: 128MiB
> * Flash: 16MiB - dual firmware partitions!
> * LAN: 1 1000M - POE
> * Wireless:
>   2.4G: QCA9563
>   5G: UniFi Chip, QCA988X compatible
> 
> Thanks to Frank Dietz(@tuennes) for a lot of patience while testing.
> 
> See also [freifunk-gluon/gluon#1021](https://github.com/freifunk-gluon/gluon/issues/1021)
could you do me a favor about this?https://forum.openwrt.org/t/need-a-copy-content-of-full-flash-about-uap-ac-m/140438

PR #1024, opened 2017-04-07
mvebu add support for ClearFog Base

ratkaj:
This patch set does two things:
1. The conventional ClearFog model is now known as the "Clearfog Pro", so rename it
2. Add support for SolidRun ClearFog Base board

### DIFF
--- a/package/boot/uboot-mvebu/Makefile
+++ b/package/boot/uboot-mvebu/Makefile
@@ -23,7 +23,7 @@ endef
 
 define U-Boot/clearfog
   NAME:=SolidRun ClearFog A1
-  BUILD_DEVICES:=armada-388-clearfog-pro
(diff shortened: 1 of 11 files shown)


nbd168:
Could you please backport the actual upstream changes with their full commit messages, one patch file per upstream commit?

ratkaj:
That was my first approach. However, i found that backporting all of those changes breaks the switch on ClearFog Pro. They add a lot of changes to the entire armada3xx device tree hierarchy and it might be that LEDE versions of mvneta and marvell's phy drivers cant deal with them.

That's why i decided for a minimal change that will not break ClearFog Pro or any other board using armada3xx DT files while making it easy to add new boards from ClearFog family.

nbd168:
Makes sense, thanks. I'll take another look later

PR #1028, opened 2017-04-10
Enable RTC_SUPPORT as a config option

nwf:
Move RTC_SUPPORT from target/Config.in to config/Config-build.in, making it subject to manipulation by .config.  Some platforms that do not have onboard RTCs can nevertheless have them added (using, e.g. a i2c-tiny-usb and a ds3231 I2C RTC module) and being blocked from building the kernel modules is annoying.

This may not be the correct course of action; RTC_SUPPORT is used sparsely enough around the tree that perhaps removing it is the right answer instead.

### DIFF
--- a/config/Config-build.in
+++ b/config/Config-build.in
@@ -39,6 +39,10 @@ menu "Global build settings"
 		bool "Show packages that require graphics support (local or remote)"
 		default n
 
+	config RTC_SUPPORT
+		bool "Enable RTC support"
+		default n
+
 	config BUILD_PATENTED
 		default y
 		bool "Compile with support for patented functionality"
--- a/target/Config.in
+++ b/target/Config.in
@@ -33,9 +33,6 @@ config USB_SUPPORT
 config USB_GADGET_SUPPORT
 	bool
 
-config RTC_SUPPORT
-	bool
-
 config BIG_ENDIAN
 	bool
 


pprindeville:
> [...] RTC_SUPPORT is used sparsely enough around the tree that perhaps removing it is the right answer instead.

When you say "sparsely enough around the tree" are you including where it's invoked by the feature alias `rtc`?

```
target/linux/mxs/Makefile:FEATURES:=ext4 rtc usb gpio
target/linux/sunxi/modules.mk:    $(call AddDepends/rtc)
target/linux/sunxi/modules.mk:    $(call AddDepends/rtc)
target/linux/sunxi/Makefile:FEATURES:=fpu usb ext4 display rtc squashfs
target/linux/x86/xeon/target.mk:FEATURES:=squashfs ext4 pcie usb gpio rtc display -vmdk
target/linux/x86/atom64/target.mk:FEATURES:=squashfs ext4 pcie usb gpio rtc display -vmdk
target/linux/gemini/Makefile:FEATURES:=squashfs pci rtc
target/linux/oxnas/Makefile:FEATURES:=gpio nand pcie usb ramdisk rtc squashfs ubifs
target/linux/zynq/Makefile:FEATURES:=fpu gpio rtc usb usbgadget targz
target/linux/armvirt/Makefile:FEATURES:=fpu pci rtc usb
target/linux/imx6/Makefile:FEATURES:=audio display fpu gpio pcie rtc usb usbgadget squashfs targz nand ubifs
target/linux/orion/Makefile:FEATURES:=broken rtc
target/linux/ar71xx/nand/target.mk:FEATURES += squashfs nand rtc
target/linux/ramips/mt7621/target.mk:FEATURES+=usb rtc nand
```

BTW, when I run:

```
% grep -r 'AddDepends/rtc' .
```

I don't get any hits for a `define`...

nwf:
Mm, my comment was made based solely on the results of `git grep RTC_SUPPORT`, which suggested to me that it's used only to gate off some kernel modules.

pprindeville:
Have a look at `scripts/target-metadata.pl`.

nwf:
I am unsure what you want me to conclude from `scripts/target-metadata.pl`.  I understand that when the `rtc` feature is turned on by the target, then the config system is told to select `RTC_SUPPORT`; am I missing something else?

In any case, to be more precise, the only DEPENDS-encies I can see on `RTC_SUPPORT` are:

```
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
package/kernel/linux/modules/other.mk:  DEPENDS:=@RTC_SUPPORT +kmod-i2c-core
```

Running `git grep rtc package` doesn't make it look like I'm missing anything else, even if there exists some other mechanism of specifying dependencies.  There are no occurrences of `RTC_SUPPORT` in the package repository nor in luci's.

pprindeville:
>  I understand that when the rtc feature is turned on by the target, then the config system is told to select RTC_SUPPORT

Yup, that's it.

nbd168:
The reason that RTC_SUPPORT was added is the fact that compiling some RTC drivers as module in the release build pushes some bloat into the kernel image.
I've pushed a different change in my staging tree that should also address the problem. Please review it and let me know if this covers your use case as well.

https://git.lede-project.org/79daae59ba1fdb670d379b31f3572ead418e1fb2

nwf:
That looks like it will work fine.  Thank you, @nbd168.

PR #1029, opened 2017-04-11
 tools: host/include/getline.h: Fix cygwin compile

neheb:
This was probably true in the past. Today, it throws a redefinition error.

### DIFF
--- a/tools/include/getline.h
+++ b/tools/include/getline.h
@@ -36,7 +36,7 @@
 #include <osreldate.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


nbd168:
It looks like getline.h is completely unnecessary and it should be removed instead (including the reference to it in tools/mtd-utils/Makefile

neheb:
From what I can tell, the file is for compatibility with EOLd versions of OS X and FreeBSD. Yeah I'll remove this.

Will open up a new request.

PR #1030, opened 2017-04-12
build: no target  in the sdk directory like in openwrt

eyal-rounds:
When building an image plus the sdk for that image, in openwrt 15.04 you 
get 3 directories under toolchain one of them is called target.... and contains
all the .so's that were installed on the target image. In Lede project this dir
is missing, apparently the reason is the drop of the line i've added. If it was 
done on purpose i don't understand the reason, if by mistake, well , there 
you go, the fix :-)

### DIFF
--- a/target/sdk/Makefile
+++ b/target/sdk/Makefile
@@ -32,7 +32,7 @@ EXCLUDE_DIRS:=*/ccache/* \
 	*/doc \
 	*/share/locale
 
-SDK_DIRS = \
+SDK_DIRS = 	$(STAGING_SUBDIR_TARGET)\
 		$(STAGING_SUBDIR_HOST) \
 		$(STAGING_SUBDIR_TOOLCHAIN)
 


nbd168:
This was done on purpose.
To be able to link against these libraries properly, you need the source packages anyway to have all the dependency information, often including relevant config symbols etc.
Since full compatibility is not possible with the shipped prebuilt binaries, and using ./scripts/feeds to install packages pulls in dependencies anyway, we prefer to save a lot of space on the download servers by not including these files.

eyal-rounds:
@nbd168 if u ask me it misses the purpose of the sdk, why would i need specific SDK per image if it misses  my .so and .h files ? i understand the need to save space on the server, so maybe i can add a flag to the makefile to keep this dir when need to and not by default so my builds can keep it and i can still be aligned with the lede branch, will that be acceptable ?

PR #1031, opened 2017-04-12
tools: host/include/getline.h: Remove

neheb:
Still doesn't build with Cygwin but a step in the right direction.


nbd168:

PR #1032, opened 2017-04-13
download: allow specifying git clone options

fake666:
- for git, there is an inaccessible OPTS variable already
- allow setting it via PKG_SOURCE_OPTS
- can be used for adding "--depth 1" to the git command for shallow clones

Signed-off-by: Thomas 'fake' Jakobi <thomas@derfake.com>

### DIFF
--- a/include/download.mk
+++ b/include/download.mk
@@ -162,7 +162,7 @@ define DownloadMethod/git
 		cd $(TMP_DIR)/dl && \
 		rm -rf $(SUBDIR) && \
 		[ \! -d $(SUBDIR) ] && \
-		git clone $(OPTS) $(URL) $(SUBDIR) && \
+		git clone $(GIT_OPTS) $(URL) $(SUBDIR) && \
 		(cd $(SUBDIR) && git checkout $(VERSION) && \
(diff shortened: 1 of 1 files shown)


pepe2k:
@nbd168, @jow- ?

jow-:
Hm, can't see any problems with this. LGTM

fake666:
valid point. i'll rename it...

fake666:
@jow ok like this?

fake666:
... knock, knock... :)

nbd168:
I've reviewed this again, and to me this does not make much sense. First of all, if you set --depth 1 to create a shallow clone, you won't be able to select a specific version anymore.
As such, this would only seem to work for strongly-discouraged 'moving target' makefiles that just use whatever version happens to be the most recent one in the git repository.
If you care about cloning a package fast, you could use CONFIG_SOURCE_TREE_OVERRIDE and the git-src symlink. This will ensure that after you've cloned the repo once, the build system can just take it from there. For personal use this should be more convenient as well.

blogic:
@fake666 i share @nbd168 feelings, sorry ...

PR #1034, opened 2017-04-14
omap: Add a profile for the Chiliboard

bartekk942:
The following commit adds support for the GRINN Chiliboard within
the omap target.

The changes are as follows:
 - bump uboot version to 2017.03 to obtain formal support
 - enable usb1 working by default


nbd168:
The profile is completely unnecessary and contains some broken package entries as well. Please leave it out.

bartekk942:
All right, but what if current default profile does not compile functional uboot for my device? Could we just add an additional uboot compilation for this device instead of an additional profile?

mkresin:
> All right, but what if current default profile does not compile working uboot for my device? Could we just add an additional uboot compilation for this device instead of an additional profile?

Well, select the u-boot your are looking for in menuconfig. The [buildbot compiles the omap](https://downloads.lede-project.org/snapshots/targets/omap/generic/) target with all packages selected anyway.

Nevertheless your PR can not be merged as it is. The u-boot updates should be separate commits, if not separate PRs. The u-boot updates need to be tested on already supported boards for possible regressions.

Adding boards for two different targets in one PR doesn't seam to me a good idea either, since the PR can be only merged if the changes to both targets are fine.

While having a brief look at your PR, there are some changes which are either not necessary, will cause issues or should be send upstream first:

- UBOOT_IMAGE:=u-boot.img => UBOOT_IMAGE:=u-boot.imx, not sure if this really is a typo
- changes to 0005-fix-boot-uimage-liteboard.patch which should be send upstream or otherwise explained in the commit message
- a lot of removals in target/linux/imx6/config-4.9 which I have no explanation for
- the imx dts files should be added as patches since they are backports, please mention backports in the commit message
- change of the KERNEL_LOADADDR=0x80800000 in target/linux/imx6/image/Makefile which most likely will cause issues for already supported boards

Finally your commits do not have a signed off by line. Please be a bit more verbose in your commit messages.

mkresin:
Closed due to inactivity. Please reopen the pull request if you have the time to fix the outlined issues.

PR #1036, opened 2017-04-14
ar71xx: add support for EnGenius ENS202EXT

ntzrmtthihu777:
EnGenius ENS202EXT is an outdoor wireless access point with
2-port 10/100 switch, detachable antennas and proprietery PoE.

The device is based on Qualcomm/Atheros AR9341 v1.

Specifications:
- 535/400/200 MHz (CPU/DDR/AHB)
- 64 MB of RAM
- 16 MB of FLASH
- UART (J1) header on PCB (unpopulated)
- 2x 10/100 Mbps Ethernet
- 2.4 GHz, up to 26dBm
- 2x external, detachable antennas
- 7x LED, 1x button

Flash instructions:

You have three options:

- Use the vendor firmware upgrade page on the web interface and give
it the factory.img. This is the easiest way to go about it.
- If you have serial access during u-boot, interrupt the normal boot
(any key before timeout) and run `run failsafe_boot`; this will bring
you to a minimal openwrt luci image on ip 192.168.1.1 useful if you've
bricked the normal firmware.
- Use the vendor's management cli, which can be accessed via telnet
with the same credentials as the web login (default admin:admin), then
issue the following commands:
```
*** Hi admin, welcome to use cli(V-1.6.7) ***
---========= Commands Help =========---
      stat -- Status
       sys -- System
    wless2 -- 2.4G-Wireless
      mgmt -- Management
      tree -- Tree
      help -- Help
    reboot -- Reboot
ens202ext>mgmt
  Management
---========= Commands Help =========---
     admin -- Administration
     mvlan -- Management VLAN settings
      snmp -- SNMP settings
    backup -- Backup/Restore settings
    autorb -- Auto reboot settings
   fwgrade -- Firmware upgrade
      time -- Time settings
   wifisch -- Wifi schedule
       log -- Log
      diag -- Diagnostics
      disc -- Device Discovery
    logout -- Logout
      help -- Help
      exit -- Exit
ens202ext/mgmt>fwgrade
  Management --> Firmware upgrade
---========= Commands Help =========---
      fwup -- Firmware upgrade
      help -- Help
      exit -- Exit
ens202ext/mgmt/fwgrade>fwup http://web.server/lede-ar71xx-generic-ens202ext-squashfs-factory.bin
```


ntzrmtthihu777:
Also, I'm having a slight issue with the firmware I built with these patches, in that my configs (set either with uci commands or direct editing of /etc/config files) do not survive a reboot.

bobafetthotmail:
check if the jffs2 partition is mounted/mountable, as it's the writable partition.

ntzrmtthihu777:
```bash
/dev/root on /rom type squashfs (ro,relatime)
proc on /proc type proc (rw,nosuid,nodev,noexec,noatime)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,noatime)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,noatime)
tmpfs on /tmp/root type tmpfs (rw,noatime,mode=755)
tmpfs on /dev type tmpfs (rw,nosuid,relatime,size=512k,mode=755)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,mode=600)
debugfs on /sys/kernel/debug type debugfs (rw,noatime)
/dev/mtdblock5 on /overlay type jffs2 (rw,noatime)
overlayfs:/overlay on / type overlay (rw,noatime,lowerdir=/,upperdir=/overlay/upper,workdir=/overlay/work)
```
Yep, as you can see from the output of this mount command, it is mounted and rw.

ntzrmtthihu777:
Nevermind the config issue. must have been something odd in my setup. Booting to failsafe and formatting the rootfs_data partition and redoing config is persistent.

pepe2k:
Hello @ntzrmtthihu777,

Thank you for your contribution.

After a short review, I see one problem: we don't want to include more devices in legacy building code. Are you able to move out related code into generic?

Cheers,
Piotr

ntzrmtthihu777:
@pepe2k Sure, sounds alright. Could you point out a good example device to pattern after?

pepe2k:
> @pepe2k Sure, sounds alright. Could you point out a good example device to pattern after?

As far as I can see, there are just two devices using `Image/Build/Senao` recipe, including the one you are adding support for. You can find some commits related with legacy->generic move in history, like d8662ac3c6a2ecd3834d109c5d3abd864fb23a50 or 5d04dcedb4fb773404dcb0625786eb7706284298, but of course they are not related with `Image/Build/Senao` which probably would also need to be moved/rewritten.

Cheers,
Piotr

ntzrmtthihu777:
On another note, this device has two partitions, `custom` and `failsafe`, which seem to be of little use once you flash lede.
`custom` contains a jffs2 partition which contains `sysupgrade.tgz`, which appears to be factory defaults for uci's configs, and `failsafe` contains a uImage, I believe its for some form of web recovery (never managed to get it to work); Could a 'fat' image be made which uses that space (320k and 2048k respectively) for more rootfs space?

pepe2k:
> On another note, this device has two partitions, custom and failsafe, which seem to be of little use once you flash lede.

If it's used by vendor firmware we definitely shouldn't remove it and mark as read-only.

> custom contains a jffs2 partition which contains sysupgrade.tgz, which appears to be factory defaults for uci's configs, and failsafe contains a uImage, I believe its for some form of web recovery (never managed to get it to work);

I suppose it's a kernel with initramfs, if you have access to serial console and U-Boot, try to boot it with `bootm` and see what it's doing.

> Could a 'fat' image be made which uses that space (320k and 2048k respectively) for more rootfs space?

General convention is to keep vendor firmware specific data safe, just in case.

If you are sure (and can prove it _somehow_) that this data isn't needed and used by vendor firmware and return to vendor firmware would still be possible without it, then we can make bigger image by default. If not, then it's better to keep this data untouched (which would for sure save us some bug reports from users in future).

Cheers,
Piotr

ntzrmtthihu777:
Yeah, I've used bootm at that memory address, it boots and whatnot, but once that happens nothing further is output from serial (not even garbage)

ntzrmtthihu777:
Seems I've remembered wrong, on the ens202ext it works, brings up a limited fallback environment with wifi and such, but it lacks the tools to determine the proper IP to access it from. Also, this recovery env is completely undocumented in the vendor documentation/manuals.

pepe2k:
> Seems I've remembered wrong, on the ens202ext it works, brings up a limited fallback environment with wifi and such, but it lacks the tools to determine the proper IP to access it from. Also, this recovery env is completely undocumented in the vendor documentation/manuals.

What about this one: http://www.engenius-europe.com/index.php/support-menu/knowledge-base?kbartid=47? ENS202EXT is listed in "Applies To" list.

Cheers,
Piotr

ntzrmtthihu777:
Yeah, but if you check that link you'll see its the scummy easywrt folks who won't release source and will charge you for use of their recovery ramdisk.

ntzrmtthihu777:
also this recovery ramdisk they speak of is meant to be tftpboot'd and does not refer to the failsafe partition on the flash itself.

pepe2k:
> Yeah, but if you check that link you'll see its the scummy easywrt folks who won't release source and will charge you for use of their recovery ramdisk.

Bastards!

> also this recovery ramdisk they speak of is meant to be tftpboot'd and does not refer to the failsafe partition on the flash itself.

Can you dump and share it somewhere?

Cheers,
Piotr

ntzrmtthihu777:
sure.
base64'd:
https://gist.github.com/ntzrmtthihu777/9b7f9cb94b4bc86f31f165fc44716b1f

ntzrmtthihu777:
Ah, it seems I spoke too soon. The failsafe works, but you can't access it over wifi, even though it broadcasts a signal. The failsafe has an ip of 192.168.1.1, which conflicts with my existing network setup.

ntzrmtthihu777:
It seems this is not ready for merge yet anyways (even discounting the legacy images bit); I was using your u-boot mod, which did not verify the image hashes, so I need to get that fixed up.

ntzrmtthihu777:
@pepe2k how do you feel about the idea of adding a new make file, `engenius.mk` or `senao.mk` to separate the specifics of senao-made devices? Or, tell me this: there is nothing magic about the stock images from the manufacturer, its simply a root.squashfs and uImage-lzma.bin wrapped in a .tar.gz, is there a target that makes that sort of image from the getgo?

pepe2k:
> @pepe2k how do you feel about the idea of adding a new make file, engenius.mk or senao.mk to separate the specifics of senao-made devices?

Seems as a good idea.
We already keep at least TP-Link and Ubiquity specific image code in separated files.

> Or, tell me this: there is nothing magic about the stock images from the manufacturer, its simply a root.squashfs and uImage-lzma.bin wrapped in a .tar.gz, is there a target that makes that sort of image from the getgo?

There is _something_ in `ramips` target:
https://github.com/lede-project/source/blob/master/target/linux/ramips/image/mt7621.mk#L5

Cheers,
Piotr

ntzrmtthihu777:
@pepe2k hows this look? Its kinda ugly, but to be frank the lede/openwrt build system is very... intricate, shall we say, and its rather fun to figure out what variables are available to you where.

target/linux/ar71xx/base-files/etc/board.d/01_leds
```diff
@@ -220,6 +220,14 @@ eap300v2)
 	ucidef_set_led_netdev "lan" "LAN" "engenius:blue:lan" "eth0"
 	ucidef_set_led_wlan "wlan" "WLAN" "engenius:blue:wlan" "phy0tpt"
 	;;
+ens202ext)
+	ucidef_set_led_power "power" "POWER AMBER" "engenius:amber:power" "1"
```
mkresin: Please drop this line! The led is already handled by diag.sh.

ntzrmtthihu777: @mkresin ah! So it is. Also, does the rssileds bit look alright? I don't quite understand the proper application here.

mkresin: good point. the rssileds package is missing. please add DEVICE_PACKAGES containing the rssileds  package to your image build,

target/linux/ar71xx/image/Makefile
```diff
@@ -17,6 +17,7 @@ include ./tp-link.mk
 include ./ubnt.mk
 include ./generic.mk
 include ./legacy-devices.mk
+include ./senao.mk
```
mkresin: please don't add a distinct senao.mk. add your code to generic.mk.

ntzrmtthihu777: @mkresin @pepe2k suggested it to be a good idea to separate it out (currently this is only the ens202ext board, but I've got board support code set up also for the ens200 which I'll be pr'ing once I get this merged, and a host of other EnGenius devices which are identical as far as firmware format/layout is concerned).

ntzrmtthihu777:
also, I'm having a bit of issue causing `/etc/fw_env.config` to exist in the image. What am I missing?

target/linux/ar71xx/image/senao.mk
```diff
@@ -0,0 +1,30 @@
...
+	dd if=$(KDIR)/root.squashfs \
+		of=$(KDIR_TMP)/$(1)/openwrt-senao-$(1)-root.squashfs bs=64k conv=sync
+	( \
+		cd $(KDIR_TMP)/$(1);  \
+		$(TAR) -cz -f $(call factoryname,squashfs,$(1)) * \
+	)
+	rm -rf $(KDIR_TMP)/$(1)
+endef
+
+define Device/ens202ext
```
mkresin: Using sysupgrade.tar here is wrong. sysupgrade.tar is for used for NAND flash chips with ubi. There is no need to have a dedicated kernel and rootfs partition. You can combine them to a firmware partition and use the dynamic firmware splitter in LEDE.

Would you please try the following instead:

ntzrmtthihu777: That wouldn't work as above; the factory.img _has_ to have two separate files with that exact naming scheme, or else the vendor firmware will not accept it for upgrading to even get into LEDE. Now, regarding sysupgrade I don't know, but the factory image (which my understanding says is the way from vendor->lede, correct me if I'm wrong) has to come out the same as my current recipe has (that is, a separate uImage-lzma.bin and root.squashfs with the exact naming scheme).

mkresin: Please have a closer look at the code! The factory image still has two separate files. It is just that openwrt-senao-$(1)-uImage-lzma.bin is the kernel + as much of the rootfs that fits into the 1536k kernel partition that is used by the stock firmware. openwrt-senao-$(1)-root.squashfs contains the remaining bits of the rootfs. This way the oem kernel + rootfs partition (== firmware in LEDE) has the content expected by the LEDE firmware partition splitter.

mkresin: >  Now, regarding sysupgrade I don't know, but the factory image (which my understanding says is the way from vendor->lede, correct me if I'm wrong) has to come out the same as my current recipe has (that is, a separate uImage-lzma.bin and root.squashfs with the exact naming scheme, tar'd and gzipped).

Just to make this more clear. the factory image format is _only_ used to upgrade from OEM firmware to LEDE. LEDE has its own image formats to do an upgrade from your current LEDE version to a later one.

sysupgrade.tar is one of the formats used by LEDE but it is only used if you have a NAND flash chip which might uses UBI. In your case a SPI flash chip is used and you need to use the ["normal" LEDE image format](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/Makefile#L110) which is added to the ar71xx default image build recipe. By using ```IMAGES += factory.img``` you keep the sysupgrade.bin and add the factory.img (```+=``` vs. ```:=```).

ntzrmtthihu777:
Back-burner this for the moment. I somehow messed up my ability to restore stock firmware to the device, so I'm going to have to grab a new one (at least temporarily long enough to do a full flash backup of the stock firmware).

target/linux/ar71xx/image/Makefile
poranje: The ENH200(EXT) pose the exact same porting problems.

ntzrmtthihu777: Oh, do you have some of those? I have code for that ready to go as well.

poranje:
Yep, I stalled more or less at the same point as you: got a ENH200EXT working but integration into the LEDE build environment did not make it more fun.
(http://patchwork.ozlabs.org/patch/745796 <http://patchwork.ozlabs.org/patch/745796>)

Besides that I ran into a rather nasty side effect which caused me for the time being to concentrate on other work: with LEDE runing on de device the radio could not be powered more then 16 dBm (maybe it’s just what user space is reporting).

Now seeing that you have created a port for the ENS202EXT which I have installed at a client of mine, but were the stock software is really causing problems (the location is a rural camping and with more than 10 clients all kind of connection problems seem to occur randomly). Running LEDE with it would enable me to handle the device properly !
You didn’t experience any loss of txpower ?

For the next days I will not be able to pick-up the port again, but I surely want to and knowing that someone else does as well helps.
Your remarks on EasyWrt makes it obvious that your aim is to benefit many by sharing, not keeping. I like that.

BTW, you can execute a recovery of either stock or initramfs image with TFTP. When the reset button is pressed while powering the device it will try to connect to a tftp server running at 192.168.1.101 and try to download an image named “vmlinux-art-ramdisk”. This works on the ENH20* and probably works with the ENS202* as well.

Good luck and we will be in contact, bye,
-- 
Paul Oranje
M	+31 (6) 21278389
T	+31 (20) 4941306
Achterlaan 20, 1027 AK Zunderdorp

GPG key ID 0xB833CA27

ntzrmtthihu777:
@poranje if you could, it would be of great help to my porting efforts to get a firmware dump from the ens202ext you have; I apparently neglected to do so on mine.

target/linux/ar71xx/image/Makefile
poranje: A dump of the MTDs has been prepared for you, please send your e-mail address to por~at~xs4all.nl so that I can can send you the dump directly.

ntzrmtthihu777: @poranje Misread, I thought you had an ens202ext

poranje: Yes I got one as well and the MTDs dump is thereof.

ntzrmtthihu777:
Ok, I have building factory.img and sysupgrade.bin, but sysupgrade.bin does not update (at least correctly) the rootfs hash in u-boot, and as such it will boot to failsafe after sysupgrade.

mkresin:
Please apply the following patch on top of your changes: https://gist.githubusercontent.com/mkresin/e55c309feffc0e2ff42ebaba07fea69a/raw/3704d30f22a0b9520fa6feaff060aa0291959491/ens202ext.patch

It contains everything we already talked about in IRC. Including the hash update in the uboot env.

mkresin:
@pepe2k Image build code is fix, sysupgrade works now and commit message is updated. Will remove the "work in progress" label added by my self now.

target/linux/ar71xx/files/arch/mips/ath79/Kconfig.openwrt
```diff
@@ -1060,7 +1069,7 @@ config ATH79_MACH_WNR2000_V3
 	select ATH79_DEV_LEDS_GPIO
 	select ATH79_DEV_M25P80
 
-	config ATH79_MACH_WNR2200
```
pepe2k: Please, don't include that kind of whitespace fixes withing your patch as it's not related at all.
If you want to fix this, do it in a separate commit.

target/linux/ar71xx/image/senao.mk
```diff
@@ -0,0 +1,28 @@
...
+
+	$(TAR) -czf $@ -C $@.senao .
+	rm -rf $@.senao
+endef
+
+define Device/ens202ext
+  DEVICE_TITLE := EnGenius ENS202EXT
+  BOARDNAME := ENS202EXT
+  DEVICE_PACKAGES += rssileds
+  CONSOLE := ttyS0,115200
```
pepe2k: This is a default value, defined in `target/linux/ar71xx/image/Makefile` (take a look please at `define Device/Default`), you can drop it.

target/linux/ar71xx/image/senao.mk
```diff
@@ -0,0 +1,28 @@
...
+  KERNEL_SIZE := 1536k
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImage lzma | check-size $$(KERNEL_SIZE)
+  FILESYSTEMS := squashfs
```
pepe2k: Can you please explain this? This is a default value for generic subtarget in ar71xx.

ntzrmtthihu777:
@pepe2k agreeable?

pepe2k:
@ntzrmtthihu777, thank you, I have just merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1044, opened 2017-04-20
DTS for AR71xx

809205580:
When did ar71xx start using DTS?


pepe2k:
@809205580, for asking questions, please use [official forum](https://forum.lede-project.org/), [mailing lists](https://lede-project.org/contact#mailing_lists) or [IRC channels](https://lede-project.org/contact#irc_channels).

Cheers,
Piotr

PR #1046, opened 2017-04-21
packages/utils/xfsprogs: Hide until updated

diizzyy:
3.2.X and newer uses a new on-disk format, disable (hide) xfsprogs until
it's updated to avoid unnecessary (bad) surprises.
Source: https://wiki.archlinux.org/index.php/XFS#Integrity


nbd168:
I've pushed an update to my staging tree.

PR #1048, opened 2017-04-21
busybox: Enable sendfile by default

diizzyy:
Enable sendfile system call by default.
http://lists.busybox.net/pipermail/busybox-cvs/2014-November/034831.html


diizzyy:
Tested on mvebu, ramips and ar71xx.

nbd168:
Merged, thanks.

PR #1049, opened 2017-04-21
dgn3500 - blink switch port leds on activity

kdarbyshirebryant:
2 commits here, 1 to enable initvals from DTS for the rtl8366rb switch, and another to take advantage of the new feature and set the LED register to blink switch leds on activity as per the stock firmware.

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/rtl8366rb.c
+++ b/target/linux/generic/files/drivers/net/phy/rtl8366rb.c
@@ -282,6 +282,31 @@ static int rtl8366rb_reset_chip(struct rtl8366_smi *smi)
 static int rtl8366rb_setup(struct rtl8366_smi *smi)
 {
 	int err;
+#ifdef CONFIG_OF
+	unsigned i;
+	struct device_node *np;
+	unsigned num_initvals;
+	const __be32 *paddr;
+
+	np = smi->parent->of_node;
+
+	paddr = of_get_property(np, "realtek,initvals", &num_initvals);
+	if (paddr) {
+		dev_info(smi->parent, "applying initvals from DTS\n");
+
+		if (num_initvals < (2 * sizeof(*paddr)))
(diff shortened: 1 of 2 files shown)


kdarbyshirebryant:
Hi Mathias!  Well this has been a little voyage of discovery hasn't it :-)
FYI: I pulled out all the 'chipdefault2' values from u-boot driver and then stripped out everything about 0x8000 (which is the ASIC PHY_ACCESS_CTRL_REG)  I did that on the basis that the u-boot code enables write access to that register and then sends the initval register/data pairing for register values == & 0xBE00  - ie.

while(chipDefault2[index][0] != 0xFFFF && chipDefault2[index][1] != 0xABCD)
                {       
                        /*PHY registers setting*/       
                        if(0xBE00 == (chipDefault2[index][0] & 0xBE00))
            rtl8368s_setAsicReg(RTL8368S_PHY_ACCESS_CTRL_REG,RTL8368S_PHY_CTRL_WRITE));
                  
 rtl8368s_setAsicReg(chipDefault2[index][0],chipDefault2[index][1]));
                                
                        index ++;       
                }

Stripping the codes out that I suspect are PHY setup related significantly reduces the number of initvals *AND* more importantly, the RRCP frames are still suppressed.

mkresin:
Merged to my staging tree.

> Well this has been a little voyage of discovery hasn't it

Oh yes it was. And I'm getting really annoyed by these realtek switches. Even the datasheets that can be found, do not list all register and they do not list the real register addresses at all.

I've still no idea what the 0xBExx registers are and what the values written to this register really do. Even more confusing are the initvals for 0x8000. They enable/disable the PHY_CTRL_WRITE, which is exactly the same the ```rtl8368s_setAsicReg(RTL8368S_PHY_ACCESS_CTRL_REG,RTL8368S_PHY_CTRL_WRITE));``` line does.

PR #1052, opened 2017-04-21
dnsmasq: don't split dhcp_option entries on spaces

puck:
Previosuly dhcp_option in /etc/config/dnsmasq would be split on any
spaces. This is unexpected, and prevents useful usecases.

This change is based on https://dev.openwrt.org/ticket/7952 but updated to
work with version 15.05 of OpenWRT. This is tested on OpenWRT 15.05,
but not with LEDE (yet).

In my case I'm trying to setup network boot for a Raspberry Pi, when adding
this to /etc/config/dhcp:

config mac 'raspbian'
	option mac 'aa:bb:cc:*:*:*'
       [snip]
	list dhcp_option '43,Raspberry Pi Boot'

Resulted in this:

dhcp-option=raspbian,43,Raspberry
dhcp-option=raspbian,Pi
dhcp-option=raspbian,Boot

Not very helpful. With this patch I now get this result:

dhcp-option=raspbian,43,Raspberry Pi Boot

Which is what I expect, and want.

Signed-off-by: Andrew Ruthven <andrew@etc.gen.nz>

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -546,11 +546,16 @@ dhcp_option_add() {
 
 	[ "$force" = "0" ] && force=
 
-	config_get dhcp_option "$cfg" dhcp_option
-	for o in $dhcp_option; do
-		xappend "--dhcp-option${force:+-force}=${networkid:+$networkid,}$o"
-	done
+	config_list_foreach "$cfg" dhcp_option dhcp_option_add_one "$networkid" "$force"
+}
 
+dhcp_option_add_one() {
+	local networkid="$2"
+	local force="$3"
+
+	[ "$force" = "0" ] && force=
+
+	xappend "--dhcp-option${force:+-force}=${networkid:+$networkid,}$1"
 }
 
 dhcp_domain_add() {


dedeckeh:
The patch cannot be accepted as it will break existing configurations which still use "option dhcp_option ..." which means backwards config compatibility is broken

puck:
Existing configurations which don't have spaces in them will continue to work (I have examples of this in my dhcp config file), and those that do have spaces are generating broken files which won't do what they want...

dedeckeh:
Existing configurations which use "option dhcp_option 'A B C' " are translated as
dhcp-option=A
dhcp-option=B
dhcp-option=C

With the patch such configurations are not parsed anymore as the options are not defined as a list config breaking existing configurations and thus backwards compatibility.
I proposed a similar patch in the past (http://www.mail-archive.com/openwrt-devel@lists.openwrt.org/msg38671.html) which was rejected for this reason.

puck:
Oh, people do that?

What if we use the new behaviour only if the dhcp_option parameter matches the regex ^\d+,[^,]$ ? So that if you have "12,Foo 43,Bar" it still results in:

dhcp-option=12,Foo
dhcp-option=43,Bar

dedeckeh:
@puck After discussion with @jow commit https://git.lede-project.org/?p=source.git;a=commit;h=9412fc294995ae2543fabf84d2ce39a80bfb3bd6 has been pushed which allows to configure dhcp_option as a list while also keeping backwards compatibility.

puck:
Fantastic, thank you both @dedeckeh  and @jow, that's an excellent solution.

This will allow folks to network boot their Raspberry Pis using LEDE running dnsmasq as the DHCP server. And I'm sure there'll be other useful use cases as well!

bobafetthotmail:
Can you (either) update the wiki with this change? https://lede-project.org/docs/user-guide/dhcp_configuration

If none steps up I'll do it myself but I'm less knowledgeable about this.

dedeckeh:
@bobafetthotmail I checked already https://lede-project.org/docs/user-guide/dhcp_configuration and strange enough it does not require any update as it was already describing dhcp_option as a list option while in the actual implementation this was not the case ...

bobafetthotmail:
So I wasn't the only one that was a bit puzzled. But this change isn't going in the release too, right? Do you think the wiki still needs to mention the older way for those that are on the release branch?

dedeckeh:
Will discuss with @jow if we're going to backport the fix to the release and depending on the outcome of that discussion the necessary updates will be done in the wiki

PR #1056, opened 2017-04-22
Iproute2: Bump to 4.10.0

Ansuel:
Drop various patch (pushed upstream), refresh all other patch, updated cake patch to include ingress option. This new version needs xtables to work, we need it anyway to support iptables.

Compiled and tested on mvebu


wongsyrone:
why not bump to 4.10?

Ansuel:
@wongsyrone compilation problem with bfq and something... can't underestand why...

wongsyrone:
You may want to try my fork and see if there is any problem: https://github.com/wongsyrone/lede-1/tree/master/package/network/utils/iproute2

Ansuel:
@wongsyrone Thx for your patch, actually they solve the compilation problem. 
I remove the second one (999-2) cause if you see 110 does the same thing

diizzyy:
Are typefixes needed? Seems odd that it hasn't been merged upstream if its an issue?

wongsyrone:
The previous 010-type_fixes.patch just remove the underscore(_), i.e. convert `u_int8_t` to `uint8_t`. While per 999-05-Add-missing-sys-types.h-include.patch,

> The u_intXX_t types are defined in <sys/types.h>

this is why he add another include directive here.
And for _certain C library_ , the compilation may fail due to the undefined type.

diizzyy:
Care to elaborate?
Both Arch/Alpine Linux use musl and do this just fine without patching it.
https://git.alpinelinux.org/cgit/aports/tree/main/iproute2?h=master
https://git.archlinux.org/svntogit/packages.git/tree/trunk?h=packages/iproute2

wongsyrone:
This patch is taken from buildroot: https://git.buildroot.net/buildroot/tree/package/iproute2/0002-Add-missing-sys-types.h-include.patch

diizzyy:
Sure but is it still needed? Both alpine and arch uses musl

wongsyrone:
Not needed if we only take musl-libc into consideration.

diizzyy:
So it's basically just uclibc that's affected?

kdarbyshirebryant:
The. Iproute2 bump and the effective 'cake' bump should be separate commits.  Change one thing per commit.  This way we can (in theory) revert either should we need to do so.

wongsyrone:
Haven't tested this on uclibc. I suggest keeping it in case any related error.

diizzyy:
IMHO, we have a lot of "stray" patches that fixes "something" without actually specifying what. I think patches in general should be avoided as they only increase overhead. I've noticed that in the past ppl seemed to "blindly" import patches from Debian without any explanation, most of these are now gone but documentation needs to be more specific what imported patches actually fix. Unless we can actually find a use case I think this modification should go.

Ansuel:
Tell me what to drop...
999 are necessary ad they fix the compilation problem

diizzyy:
999-05 for instance, I'm just wondering why we need this while alpine and arch doesn't (both use musl).

kdarbyshirebryant:
Gentlemen,  Russell Senior (maintainer of iproute2) did do a bump to 4.10 a while back if you look into the lede-dev email list - it also went into patchwork but then got lost somehow.  Russell got distracted from the bump and also said 'there were a couple of features he wanted to split out' before doing so.  I've been running the Russell bump for quite a while now, the patch pulled off patchwork I think before it went missing: https://github.com/kdarbyshirebryant/lede-source/commit/713e03af2961f5bfdb6554489ba9e5d465f86888

diizzyy:
@kdarbyshirebryant 
That looks cleaner imo
https://github.com/kdarbyshirebryant/lede-source/tree/iproute410/package/network/utils/iproute2/patches

nbd168:
Superseded by #1123

PR #1057, opened 2017-04-22
Iptables: fix wrong depends for nftable support

Ansuel:
The dep for the nftables support was wrong, if someone actually enable that option gain a compilation error. This fix this problem.

Compiled and tested mvebu


nbd168:
Merged, thanks.

PR #1058, opened 2017-04-22
ramips: fix kn_rc/kn_rf gpios and wan port

Spider84:
In Device Tree configuration files was wrong GPIO registers definitions.

Signed-off-by: Alexey Belyaev <spider@spider.vc>

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -85,6 +85,7 @@ ramips_setup_interfaces()
 	hc5661a|\
 	hc5962|\
 	hlk-rm04|\
+	kn_rc|\
 	mac1200rv2|\
 	miwifi-mini|\
 	miwifi-nano|\
@@ -227,7 +228,6 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "5:lan" "0:wan" "6@eth0"
 		;;
-	kn_rc|\
 	kn_rf)
 		ucidef_add_switch "switch0" \
 			"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6@eth0"
--- a/target/linux/ramips/dts/kn_rc.dts
+++ b/target/linux/ramips/dts/kn_rc.dts
@@ -10,22 +10,18 @@
 
 	gpio-leds {
 		compatible = "gpio-leds";
-
 		wan {
 			label = "kn_rc:green:wan";
-			gpios = <&gpio0 38 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
 		};
-
 		usb {
 			label = "kn_rc:green:usb";
-			gpios = <&gpio0 39 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio1 15 GPIO_ACTIVE_LOW>;
 		};
-
 		wifi {
 			label = "kn_rc:green:wifi";
-			gpios = <&gpio0 72 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
 		};
-
 		power {
 			label = "kn_rc:green:power";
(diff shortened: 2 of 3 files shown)


mkresin:
Hey Alexey,

thanks a lot for your contribution. Unfortunately it can not be merged as it is.

Your commit message mentions only changes to the GPIOs but you have touched the uboot-envtools config and the switch config as well. You either need to add an explanation for the changes to the commit message or drop these changes.

Your commit message has formal issues as well. The commit subject is your signed of by and what I would expect as commit subject is in the actually commit text. Please have a look at https://lede-project.org/submitting-patches for some hints how it should look like. Please take care of adding a commit subject prefix.

target/linux/ramips/dts/kn_rc.dts
```diff
@@ -69,6 +65,18 @@
 	};
 };
 
+&gpio1 {
+	status = "okay";
+};
+
+&gpio2 {
```
mkresin: please drop this node. gpio2 isn't used in your dts.

target/linux/ramips/dts/kn_rf.dts
```diff
@@ -69,6 +69,18 @@
 	};
 };
 
+&gpio1 {
+	status = "okay";
+};
+
+&gpio2 {
```
mkresin: please drop this node. gpio2 isn't used in your dts.

Spider84: Done! 👍

mkresin: Nope, still there. You possibly forgot to add the changed file.

mkresin:
> - in Device Tree configuration file changed portmap from "wllll" to "llllw"
>- fixed wrong switch port mapping lan4<->wlan

"wlan" is a typo, right?

> add support of u-boot-env tool

Why? The commit message should explain why something is done. We usually do not add support for manipulating the u-boot env as long as it isn't required to get LEDE working or some fixups need to be done like fixing the mac address.

Furthermore I prefer full sentences instead of bullet items in the commit message. If I would have written the commit message it would look like this:

```
ramips: fix kn_rc/kn_rf gpios and wan port

Each GPIO bank has only 24 gpios. Use and enable the correct banks for
GPIOs  > 24.

The wan port is connected to switch port 0. Fix the mediatek,portmap as
well as the default switch config.
```

target/linux/ramips/dts/kn_rf.dts
Spider84: Sorry, but how?

Spider84:
Is it right now?

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -227,7 +227,10 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "5:lan" "0:wan" "6@eth0"
 		;;
-	kn_rc|\
+	kn_rc)
```
mkresin: Sorry for not noticing this earlier. Please merge the kn_rc into the 3g-6200n, ac1200pro, ai-br100... block which already uses ```"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "6@eth0"``` instead of duplicating existing code. Please test your changes afterwards, to make sure that it still does work as expected.

Spider84:
screen
Looks like this.

mkresin:
Well that doesn't look right. There should be two vlans. One for all the LAN ports and one for the WAN ports. According to the screenshot your wan port is used as additional lan port.

Please reset the router to the factory defaults to make sure there isn't some kind of old config active. PLease make sure that using the default config your wan port can get an IP-Address from an upstream router.

Spider84:
all work fine. I just deleted second VLAN for test that them realy deleting (in some old version of OpenWRT 
 deleting was not worked).
Before my changes BLUE port of device was detect as LAN4 and LAN4 as WAN. Now all looks good.

mkresin:
Great! I've merged a slightly modified version of your PR to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I splitted you changes into two commits. It is the best to have one commit per change topic. I will push it to master during the next days.

PR #1059, opened 2017-04-23
clearfog: rework boot-script

Artox:
As I originally hinted the boot.script that was added with the initial clearfog support can be done much more generic, and this is what I came up with!

On the way I also updated u-boot to its latest release (because I can), and so I didn't have to rebase my distro-boot patches ona year-old release.


Artox:
Note that it won't boot when early-printk is enabled, see mailinglist about that piece.

Ansuel:
One question @Artox Changes to new uboot?

Artox:
> One question @Artox Changes to new uboot?

@Ansuel I am not sure I fully understand the question.
Reason for u-boot update is that I developed the distro-boot patch on top of u-boot head, and then rebased it on latest u-boot stable.
I don't think there is a good reason to stick with the older 2016.03 release.

Ansuel:
@Artox I'm not against the upgrade... Just want to know if there are some major improvement between the 2 version

Artox:
@Ansuel First thing that comes to my mind is support for running EFI applications (e.g. grubarm.efi).
Other than that I really don't know.

package/boot/uboot-mvebu/patches/0003-clearfog-enable-distro-boot-code.patch
```diff
@@ -0,0 +1,97 @@
...
++ * Keep device tree and initrd in lower memory so the kernel can access them
++ * Set name of fdt file
++ * specify console device
++ */
++#define CONFIG_EXTRA_ENV_SETTINGS	\
++	"kernel_addr_r=0x02000000\0" \
++	"fdt_addr_r=0x05c14dc0\0" \
++	"pxefile_addr_r=0x05d09000\0" \
++	"scriptaddr=0x05d09000\0" \
++	"ramdisk_addr_r=0x05dfd240\0" \
```
nbd168: What's up with these odd addresses for ramdisk_addr_r, fdt_addr_r?

Artox: > What's up with these odd addresses for ramdisk_addr_r, fdt_addr_r?

It is something I came up with previously, to ensure loading files to thelowest possible memory adress while not calculating the addresses dynamically.
The comment explains what maximum sizes I guessed for kernel and fdt, and then ramdisk comes last so its size is not limited.
This thinking target devices where I didn't know in advance if the board would come up with 512MB RAM, or more.

I am pretty fond of these values, even though most people tend to use rounder addresses. I see no issue though. Most likely the question will come up when I try to submit this patch upstream, where the final decision should take place.
For now, these addresses work on both clearfog models.

Artox:
@nbd168 anything else?

package/boot/uboot-mvebu/patches/0003-clearfog-enable-distro-boot-code.patch
nbd168: You still didn't explain why these addresses are odd though. I find it rather confusing and it doesn't make much sense to me. If this will come up again during upstream submission, why not remove the source of confusion now?

Artox: Are they odd? They are as aligned as they need be.
Other than that, I am not interested in coming up with a different set of values and testing them. I will change them when needed, which is either because it breaks, or because the submission process mandates it.
Up to that point I see no problem here. It is functional, and could be replaced at any point in the future without users notice.

nbd168: 0x05dfd240 and 0x05c14dc0 look pretty odd to me, since they're not on boundaries like the 1M specified in your description. I still have no clue where they're coming from or why they have these weird offsets (not even aligned to memory page boundary).

Artox: Oh, nono. My description tells how much space there is after e.g. kerneladdr, before the next location where stuff can be loaded to.
Now kerneladdr is the well-known 0x02000000, where the kernel should be loaded some place after that, and not before.
Anything else is +=, e.g. fdtaddr=kerneladdr+63M

Of course the resulting values could be rounded up to more zero digits, e.g. 0x05e0 but I see no point.

nbd168: Ah, now it's starting to make sense. You're using a rather unusual (for programming at least) definition of 1M, 0xf4240 == 1000000 instead of the regular 0x100000 == 1048576.

nbd168: I guess it's not really important then, but it would be nice to have this fixed anyway. I really don't think I will be the only one hit by WTF moments reading those values ;)

Artox: True that. Is it worth switching to Mebibytes?
But to do that I certainly got to remember why I selected 63M in the first place. I'll try to give it some thought.

Artox: Maybe I should make it clear that M refers to MB != MiB?

nbd168: Yes, it's worth switching to MiB, because it's the standard unit of measurement for RAM related stuff.

Artox: Done. MiB is the right thing to use here.

Artox:
Note that I didn't revert teh split into 2 seperate boot-scripts for -Base and -Pro.
Instead for now I just copied my script twice, and added a setenv fdtfile to the -base script.

Ideally though, users should just setenv fdtfile once on first boot. Really.

Artox:
this last commit eases development, so I can use the *same* sdcard for testing on all variants in a row.
As can be observed, the only difference between clearfog base and pro now is setenv fdtfile armada-388-clearfog-base.dtb.

Artox:
If these last 2 commits are not agreeable, don't let that stop the previous ones please.

nbd168:

PR #1065, opened 2017-04-26
target: mpc85xx: switch to kernel 4.9

commodo:
Did a successful boot-up on a P1020 derivative.

Seems to run fine so far.

```
root@lede:~# uname -a
Linux lede 4.9.20 #0 SMP Wed Apr 26 08:32:27 2017 ppc GNU/Linux

root@lede:~# uptime
 15:01:41 up 29 min,  load average: 0.09, 0.05, 0.00
```


pepe2k:
Hi @commodo,

One note about this one:

> target: mpc85xx: make kernel_menuconfig
> 
> It's the best method I know (so far) to refresh
> the kernel config to a default state.

As far as I can see, you have refreshed only target kernel config. What about subtargets (e.g. `make kernel_menuconfig CONFIG_TARGET=subtarget`)? At least `p1020` subtarget kernel config is not empty and might need refresh.

Cheers,
Piotr

commodo:
ah ; good point ; will re-spin

commodo:
@pepe2k 
seems no diffs were produced for `p1020`

i tried your suggestion: `make kernel_menuconfig CONFIG_TARGET=subtarget`
and also did a `make menuconfig`, change subtarget to `P1020` and do a `make kernel_menuconfig` to make sure

lynxis:
Do we actually need the p1020 target. Sorry for jumping in.

commodo:
hmm ; i wonder why i don't see comments on this PR;
or maybe I just ignore them by accident

@lynxis 
no idea ; i personally don't need the p1020 target, since the subtargets [on which we use mpc85xx] are internal/specific boards
i think the decision may need a broader discussion here
i have no opinion on whether to keep it or not 

AFAICT, the only benefit of the P1020 target is to add LEDE/OpenWrt support to the P1020 RDB.
Which is an old development board anyway.

nbd168:
Merged to my staging tree and tested on WDR4900, thanks.

nbd168:
I also think removing the P1020 subtarget might be a good idea

PR #1066, opened 2017-04-26
ar71xx: Add Aerohive AP-121 Support

riptidewave93:
This adds support for the Aerohive AP-121 access point.

Specs:
SoC: Atheros AR9344-BC2A at 560MHz
WiFi 1: 2.4GHz Atheros AR9340? - SoC
WiFi 2: 5.0GHz Atheros AR9382-AL1A
Memory: 128MB from 2x Nanya NT5TU32M16DG-AC
SPI: 1MB Macronix MX25L8006E
NAND: 128MB Hynix H27U1G8F2BTR-BC
Ethernet: Atheros AR8035-A
USB: 1x 2.0
TPM: Atmel SC3204

Flashing:
1. Hook into UART (9600 baud) and enter U-Boot. You may need to enter a password of administrator or AhNf?d@ta06 if prompted.
2. Once in U-Boot, boot the initramfs image of LEDE:

```
dhcp;
setenv serverip 192.168.1.1xx;
tftpboot 0x81000000 lede-ar71xx-nand-hiveap-121-initramfs-kernel.bin;
bootm;
```

3. Once booted, flash via sysupgrade to install LEDE. The first flash may take a bit as the UBI partition is formatted.

TPM info can be found in the bringup repo at
https://github.com/riptidewave93/LEDE-HiveAP-121


target/linux/ar71xx/config-4.4
```diff
@@ -113,6 +113,7 @@ CONFIG_ATH79_MACH_GL_INET=y
 CONFIG_ATH79_MACH_GL_MIFI=y
 CONFIG_ATH79_MACH_GS_MINIBOX_V1=y
 CONFIG_ATH79_MACH_GS_OOLITE=y
+CONFIG_ATH79_MACH_HIVEAP_121=y
```
pepe2k: As this board belongs to nand subtarget, please enable it only in nand subtarget config-default and disable for other (generic, mikrotik).

riptidewave93: @pepe2k Is this new for the ar71xx target? I ask as this does not seem to be standard atm for the rest of the NAND targets in ar71xx. (Ex, Meraki MR18, Airtight C-60, etc).

pepe2k: @riptidewave93, it's nothing new, it's just... nobody took care about keeping everything clean and correct here.

The whole ar71xx kernel configuration is just _messy_... as you can see, `generic` subtarget kernel config is almost empty (in fact, it will be empty if you refresh it as the target kernel config already contains that single line from `generic` one) and we use target config for it (yes, ar71xx target kernel config == ar71xx generic subtarget kernel config... and to be honest, I understand why). Then, we have `nand` and `mikrotik` subtargets which are different and should include only related differences in configuration (like enabling boards only related with nand/mikrotik subtargets) but... back to begin - it's messy so instead of keeping `nand` only boards selected in `nand` subtarget kernel config, we have them enabled across all subtargets here (enabled in target config, which results in nand boards code included in... all 3 subtarget kernels: generic, nand and mikrotik).

For now, just make it the correct way (disable in target config, enable in nand subtarget config) please. I have already talked with @nbd168 about possible solutions for this config craziness. One of them is to get rid off all boards Kconfig entries and introduce some kind of group configs for subtargets. With this approach we should end up with clean subtargets configs and make everything less confusing for users.

I just haven't got time to cleanup all of this, too busy this month.

Cheers,
Piotr

pepe2k: Take a look also here for example of how it should look like with current approach: https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=commit;h=e74a4385271f684e892ec3c9311859e686956610

riptidewave93:
@pepe2k Thank you for explaining in detail, it's appreciated. The requested changes have been made.

mkresin:
@riptidewave93 would you please add the flashing instructions to the commit message. Content on external pages tend to vanish and having the flashing instructions along the support patch should be the safest.

riptidewave93:
@mkresin Requested change has been added.

pepe2k:
@riptidewave93,

I have pulled your PR and did some minor fixes:
- minor text changes in commit subject and description
- fixed alphabetical order in `target/linux/ar71xx/base-files/etc/diag.sh`
- use only model name in `target/linux/ar71xx/base-files/lib/ar71xx.sh` as majority of the boards listed there follow this approach
- fixed some code style issues in `target/linux/ar71xx/files/arch/mips/ath79/mach-hiveap-121.c` (*)

Would you please have a look at the [commit in my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=commitdiff;h=c27b414f608014a0facf2617c8870b81cb3daba8;hp=afd9aa95bfe76f18a77233878d3c3bd023759e86) and confirm that the code changes I made are still fine.

\* If you want to make sure that your PR doesn't have any obvious style problems (trailing whitespace, too long lines, etc.) please, use `scripts/checkpatch.pl`. For example, you can use it like (assuming you want to check last applied commit) below:

`git format-patch --stdout -1 | ./scripts/checkpatch.pl -`

Also, two more question:
- installation requires use of initramfs image which isn't build by default in ar71xx/nand subtarget
- there are many mtd partitions defined without 'ro' flag, like `u-boot-env` or `hw-info`, are you sure it's safe to keep them writable?

Cheers,
Piotr

riptidewave93:
@pepe2k
Thanks, code changes look good to me. As for questions:
1. Correct, install requires an initramfs image similar to the Meraki MR18. To get around this the user would either need to build an image, or use the one at my bringup repo (under releases). This is also partly why that repo was linked for setup. While this is out of my realm of control, it may be worth a discussion to allow initramfs building for these devices that require it for flashing.
2. The flags for partitions were based on the stock OS. In reality, u-boot-env can stay r/w as it is ignored by the custom u-boot on the board. As for the rest, they were not touched. I have no issues setting most of them to RO if you prefer. Just let me know, and I can revise this PR or you can do it from your staging.

pepe2k:
@riptidewave93,

> Correct, install requires an initramfs image similar to the Meraki MR18. To get around this the user would either need to build an image, or use the one at my bringup repo (under releases). This is also partly why that repo was linked for setup. While this is out of my realm of control, it may be worth a discussion to allow initramfs building for these devices that require it for flashing.

As this wouldn't be the first board in ar71xx/nand subtarget with requirement for an initramfs image, I think it would make sense to enable it by default. But, is there really no other way to flash the image on this device, especially as the serial console and U-Boot access is required anyway? Does the stock U-Boot support writing to NAND?

> The flags for partitions were based on the stock OS. In reality, u-boot-env can stay r/w as it is ignored by the custom u-boot on the board. As for the rest, they were not touched. I have no issues setting most of them to RO if you prefer. Just let me know, and I can revise this PR or you can do it from your staging.

Decision about mtd is up to you, as I don't know this device at all. I asked because general approach is to always mark vendor specific data as read-only, just to keep users safe and prevent against deleting (accidentally) something which might result in breaking their board or make it impossible to use stock firmware (examples: radio calibration data, gui translation data for vendor firmware, MAC addresses, S/N, etc.). If you decide to change something in mtd configuration, just let me know, I will update it locally.

Cheers,
Piotr

mkresin:
Hey Chris,

Piotr asked for some help with reviewing the image build code. Hopefully I do not open a can of worms here. Just a word of warning, I don't have much experience with ar71xx boards. It might be that there are some showstopper I'm not aware of.

As far as I can see, there is some kind of dual image layout. How to you ensure that the kernel is written to the current active kernel partition? Are you able to control the to be used kernel partition? If yes, what about using the following partition layout instead:

```
spi0.0:512k(u-boot)ro,64k(u-boot-env),64k(hw-info),64k(boot-info),64k(boot-sinfo)ro;
ar934x-nfc:4096k(u-boot-1),4096k(u-boot-env-1),5m(kernel),111m(ubi),4096k(wifi-info)
```

This would allow to create an ubinized image:

```
KERNEL_SIZE := 5120k
UBINIZE_OPTS := -E 5
IMAGES += factory.bin
IMAGE/factory.bin := append-kernel | pad-to $$$$(KERNEL_SIZE) | append-ubi
```

Which can be written to the nand via u-boot:

```
# tftp factory.bin
# nand erase 0x800000 $(filesize)
# nand write $(loadaddr) 0x800000 $(filesize)
```

target/linux/ar71xx/image/nand.mk
```diff
@@ -23,6 +23,21 @@ define Device/domywifi-dw33d
...
+define Device/hiveap-121
+  DEVICE_TITLE := Aerohive HiveAP-121
+  DEVICE_PACKAGES := kmod-usb-core kmod-usb2 kmod-i2c-gpio-custom kmod-spi-gpio kmod-ath9k kmod-tpm-i2c-atmel
+  BOARDNAME = HiveAP-121
+  BLOCKSIZE := 64k
+  CONSOLE = ttyS0,9600
+  MTDPARTS = spi0.0:512k(u-boot)ro,64k(u-boot-env),64k(hw-info),64k(boot-info),64k(boot-sinfo)ro;ar934x-nfc:4096k(u-boot-1),4096k(u-boot-env-1),5m(kernel),5m(kernel-1),25m(rootfs-unused),25m(rootfs-1),56m(ubi),4096k(wifi-info)
+  IMAGES := sysupgrade.tar
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImage lzma
+  KERNEL_INITRAMFS := kernel-bin | patch-cmdline | lzma | uImage lzma
```
mkresin: You can drop this line. See https://git.lede-project.org/?p=source.git;a=blob;f=include/image.mk;hb=54ea0f45c8b375ff29478c1f0d659d1548e89fc9#l322

target/linux/ar71xx/image/nand.mk
```diff
@@ -23,6 +23,21 @@ define Device/domywifi-dw33d
...
+  IMAGE/sysupgrade.tar := sysupgrade-tar
+endef
+
```
mkresin: If you have to touch this file, please drop this blank line.

riptidewave93:
@pepe2k

You can write to the NAND from the Stock U-Boot, but the choice for an initramfs vs a system image for flashing via U-Boot was simply simplicity for users. If you think it would be better to make a factory image using the changes recommend by @mkresin, I am fine with that and I can test it on my local unit before anything gets merged.

As for the MTD layouts, here is the purpose for each after more digging:
SPI:
  * 64k(u-boot-env) - Empty space, unused
  * 64k(hw-info) - Used to store info, ex mac address
  * 64k(boot-info) - Unknown purpose, seems to be used u-boot
  * 64k(boot-sinfo)ro - "static" boot info, used by u-boot

NAND:
  * 4096k(u-boot-1) - empty, unused
  * 4096k(u-boot-env-1) - empty, unused
  * 5m(kernel) - Kernel to boot
  * 5m(kernel-1) - Empty and unused
  * 25m(rootfs-unused) - Holds rootfs (initramfs image) that is loaded with the kernel
  * 25m(rootfs-1) - Empty and unused
  * 56m(ubi) - UBI space
  * 4096k(wifi-info) - "empty" caldata partition, as this board has the actual caldata written to the OTP's on the wireless chips. Stock may use this for other things.

Due to the above, the mounts should probably be changed to be similar to:

	spi0.0:512k(u-boot)ro,64k(u-boot-env),64k(hw-info)ro,64k(boot-info)ro,64k(boot-sinfo)ro;ar934x-nfc:4096k(u-boot-1),4096k(u-boot-env-1),5m(kernel),5m(kernel-1),25m(rootfs-unused),25m(rootfs-1),56m(ubi),4096k(wifi-info)ro

@mkresin

As for the dual image, please see the above. While the NAND is formatted for dual image, from my understanding this feature is not yet working on this specific AP, and both partitions are completely empty on my unit specifically. We also can't use that extra space as we can't control the "load" address for the boot kernel, as this address is statically set in the custom u-boot environment, and any changes to the boot arg in u-boot is ignored and not saved.

Once we come to an agreement on the above things, I can update this PR unless @pepe2k doesn't mind doing it from his staging.

Regards,
Chris B

pepe2k:
@riptidewave93 

> You can write to the NAND from the Stock U-Boot, but the choice for an initramfs vs a system image for flashing via U-Boot was simply simplicity for users. If you think it would be better to make a factory image using the changes recommend by @mkresin, I am fine with that and I can test it on my local unit before anything gets merged.

To be honest, I don't see now any benefits from using initramfs here. In fact, it adds one step more for the user during the update process (first you need to boot the initramfs and then do the update, on the other hand, with regular image you just do the upgrade).

As it's possible to write to NAND from U-Boot, I vote for using ubinized image (factory) for this device, instead of the initramfs.

I also like the idea of using whole available space for ubi partition (which will cover: kernel-1, rootfs-unused, rootfs-1 and ubi), as @mkresin suggested.

Summary:
- factory image: ubinized image, use whole available space
- mark mtd partitions with "ro" flag as you suggested
- change block size to 128k (according to the [NAND datasheet](http://natisbad.org/NAS/refs/Hynix_NAND_128Mo_H27U1G8F2BT.pdf))

> Once we come to an agreement on the above things, I can update this PR unless @pepe2k doesn't mind doing it from his staging.

I have update my [custom branch](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=shortlog;h=refs/heads/ar71xx_pr-1066_20170509). To make it easier, I have also run test build, you can grab [ready images here](https://www.dropbox.com/s/9j1xx64bd1oc766/hiveap-121.tar.gz?dl=0).

Please, do the test and also provide updated/tested flashing instruction - I will then update commit message and finally merge it.

@mkresin 

Thanks for help with this one!

Cheers,
Piotr

bobafetthotmail:
> To be honest, I don't see now any benefits from using initramfs here.

Using sysupgrade from initramfs image works fine if there are bad blocks on flash, flashing from uboot may or may not work.

I've had cases where flashing from uboot did not work due to bad blocks (the system could not boot afterwards), and lately it happened again with a user of nsa310 because he had bad blocks, he was able to install LEDE only by using initramfs-then-sysupgrade method.

IMHO it's better to install from initramfs wherever the bootloader allows it, if the flash is NAND.

mkresin:
Well, I rather would expect that the bootloader has bad block (table) support build in. Otherwise bad blocks on the kernel partition would brick the board.

bobafetthotmail:
> Well, I rather would expect that the bootloader has bad block (table) support build in. Otherwise bad blocks on the kernel partition would brick the board.

On a device that will receive less than 10 updates tops in its life? I'm skeptical.

In my experience I said that flashing from uboot isn't always working, while from initramfs installs I never had issues.

mkresin:
> On a device that will receive less than 10 updates tops in its life? I'm skeptical.

You might be shocked, but NAND is that unreliable that chips might have bad blocks at the time they leave the factory.

> In my experience I said that flashing from uboot isn't always working, while from initramfs installs I never had issues.

And what I tried to point you to is that it could be as simple as your (self compiled) u-boot might not have proper bad block support. It seams to me it is more or less a lucky coincidence that your initramfs workaround does the job. Probably because the kernel partition doesn't have these bad blocks.

Anyway, this PR isn't the right place for this discussion.

bobafetthotmail:
> And what I tried to point you to is that it could be as simple as your (self compiled) u-boot might not have proper bad block support.

Or stock's, as the flashing was done with stock uboot. I also had issues with flashing from stock uboot, not on kirkwood nor on OpenWRT anyway.

riptidewave93:
@pepe2k I am sadly out of town and away from the unit until Friday, but I will test your image then and give you the u-boot flashing commands for the commit message.

Regards,
Chris B

pepe2k:
@riptidewave93 

> @pepe2k I am sadly out of town and away from the unit until Friday, but I will test your image then and give you the u-boot flashing commands for the commit message.

OK, take your time and thanks for letting us know!

Cheers,
Piotr

riptidewave93:
@pepe2k Factory image looks good on my end. Flashing commands are:
```
dhcp;
setenv serverip tftp-server-ip; 
tftpboot 0x81000000 lede-ar71xx-nand-hiveap-121-squashfs-factory.bin;
nand erase 0x800000 0x800000;
nand write 0x81000000 0x800000 0x800000;
reset;
```
:+1: from me! :)

pepe2k:
@riptidewave93, thanks for tests and confirmation! I have updated flash instruction and merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

@mkresin, again, thanks for your help!

Cheers,
Piotr

PR #677, opened 2017-01-06
ar71xx: add support to TP-Link Archer C59v1 and C60v1

heinzek:
TP-Link Archer C59v1 is a dual-band AC1350 router, based on Qualcomm/Atheros
QCA9561+QCA9886.

Specification:
- 775/650/258 MHz (CPU/DDR/AHB)
- 128 MB of RAM (DDR2)
- 16 MB of FLASH (SPI NOR)
- 3T3R 2.4 GHz
- 2T2R 5 GHz
- 5x 10/100 Mbps Ethernet
- USB 2.0 port
- 8x LED (controled by 74HC595), 3x button
- UART header on PCB

TP-Link Archer C60v1 is a dual-band AC1350 router, based on Qualcomm/Atheros
QCA9561+QCA9886.

Specification:
- 775/650/258 MHz (CPU/DDR/AHB)
- 64 MB of RAM (DDR2)
- 8 MB of FLASH (SPI NOR)
- 3T3R 2.4 GHz
- 2T2R 5 GHz
- 5x 10/100 Mbps Ethernet
- 7x LED, 2x button
- UART header on PCB

Currently not working:
- Port LAN1 on C59, LAN4 on C60
- WiFi 5GHz (missing ath10k firmware for QCA9886 chip)
- Update from oficial web interface ( tplink-saveloader not support "product-info")

Flash instruction:
1. Set PC to fixed ip address 192.168.0.66
2. Download lede-ar71xx-generic-archer-cXX-v1-squashfs-factory.bin
and rename it to tp_recovery.bin
3. Start a tftp server with the file tp_recovery.bin in its root directory
4. Turn off the router
5. Press and hold Reset button
6. Turn on router with the reset button pressed and wait ~15 seconds
7. Release the reset button and after a short time
the firmware should be transferred from the tftp server
8. Wait ~30 second to complete recovery.

Flash instruction under U-Boot, using UART:
1. tftp 0x81000000 lede-ar71xx-...-sysupgrade.bin
2. erase 0x9f020000 +$filesize
3. cp.b $fileaddr 0x9f020000 $filesize
4. reset

Add and fix some missing regs to QCA956X.

PR #1069, opened 2017-04-27
ar71xx: add support for TP-Link Archer C25 v1

Brother-Lal:
The TP-Link Archer C25 is a low-cost Dualband Router:

 CPU:	Atheros QCA9561 775 Mhz
 RAM:	64 MB
 Flash:	8MB
 Wifi: 	3x3 2.4Ghz integrated
		1x1 5 Ghz QCA9887
 NET:	5x 10/100 Mbps Ethernet

Some LEDS are controlled by an additional 74HC595 chip.

Currently not working: ~~Factory~~  ~~Image~~

If anybody has some ideas to fix the outstanding issues, i would be grateful to try any suggestions.
As starting point bootlog, iw listing and the error message produced by the factory firmware update helper:
https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c
 
Edit: all is working now, so please do a review!


NeoRaider:
ath10k can't load its calibration data, so I suspect the offset in /etc/hotplug.d/firmware/11-ath10k-caldata is incorrect.

target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c25-v1.c
```diff
@@ -0,0 +1,236 @@
...
+	u8 *mac = (u8 *) KSEG1ADDR(0x1f7e0008);
+	u8 *art = (u8 *) KSEG1ADDR(0x1f7f0000);
+
+	ath79_register_m25p80(NULL);
+
+	spi_register_board_info(archer_c25_v1_spi_info,
+			   ARRAY_SIZE(archer_c25_v1_spi_info));
+	platform_device_register(&archer_c25_v1_spi_device);
+	
+		gpio_request_one(ARCHER_C25_GPIO_SHIFT_OE,
```
heinzek: redundant tab...

Brother-Lal:
Thanks for the tipp!
As soon as the partitions are loaded correctly (booting with the correct kernel cmdline), the radio starts up.
This means 5Ghz should be fixed when partiotions are correctly loaded(proabably after flashing the device).

target/linux/ar71xx/files/arch/mips/ath79/mach-archer-c25-v1.c
```diff
@@ -0,0 +1,236 @@
...
+	/* LAN ports */
+	ath79_eth1_data.phy_if_mode = PHY_INTERFACE_MODE_GMII;
+	ath79_eth1_data.speed = SPEED_1000;
+	ath79_eth1_data.duplex = DUPLEX_FULL;
+	ath79_switch_data.phy_poll_mask |= BIT(4);
+	ath79_switch_data.phy4_mii_en = 1;
+	ath79_register_eth(1);
+
+	ath79_register_wmac(art + ARCHER_C25_V1_WMAC_CALDATA_OFFSET, mac);
+	ap91_pci_init(/*art + ARCHER_C25_V1_PCI_CALDATA_OFFSET*/ NULL, NULL);
```
heinzek: Why You comment this code?

Brother-Lal: Its an open TODO, i just removed it because it is the wrong location.
At least the pci driver complained about invalid caldata, and it works without it.
Left it as a reminder. Ill add a TODO for now

Brother-Lal:
@heinzek Thanks for the review, added some comments for the pcicaldata for now, and added the other improvements.
SoB should be in the commit, but removed it for the pull request for now, as i really want to get the factory image to work :)

heinzek:
Did You try recovery mode?

Can You send factory bootlog and recovery log?

Brother-Lal:
You mean a flash via tftp?
Haven't tried it yet, i assume it has more or less the same firmware checks like the webinterface updater.
Added the factory bootlog here: https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c#file-factorybootlog

I have a serial console attached, so its no hassle and ill try the tftp flash process.
Might be a viable workaround if it works.

NeoRaider:
The TFTP upgrader is often more lenient than the web interface; on many devices, it allows to bypass region checks, and I remember that I could even flash images without a soft-version partition on the CPE510...

I'll have a look at your factory image generation tomorrow, the format should not be too different from already supported devices with safeloader images.

heinzek:
Did You make backup all of mtd partitions?

In factory bootlog:
Uncompressing Multi-File Image ... OK

Change build image code to create kernel with: uImageArcher lzma

Brother-Lal:
I have a complete flash backup.

Tried the uImageArcher lz7ma setting too, but the updater still won't accept the image.
I suspect they added some additional infos somewhere in the image file and those are missing or wrong...

Brother-Lal:
I can now use a generate imagefile as a factory image (see the modifications to tplink-safeloader.c).

The flashed image works and boots correctly directly after flashing.
At the second boot, the Uboot tries to access the partition table, and fails to verify it.
See https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c#file-ubootpartcheck .

Any suggestions for a possible workaround?

Brother-Lal:
Ok, all is working properly now.
I added a config partition to shield the config infos from getting erased at squasfs buildtime.

Thanks for all the help and suggestions!

Please review (especially the safeloader and partitions parts) :)

heinzek:
Great job!!!

Fix commit message if factory image is OK.
Wifi 5Ghz is OK?

Brother-Lal:
Yes, everything works.
Message is fixed.

NeoRaider:
The rootfs partition still seems to big. 0x620000 are 6272k, so you should pass the same in the kernel cmdline. It's not good to overwrite unrelated partitions.

Brother-Lal:
Changed to 128k  config partition. Thanks!

tools/firmware-utils/src/tplink-safeloader.c
```diff
@@ -729,6 +771,15 @@ static struct image_partition_entry read_file(const char *part_name, const char
 	return entry;
 }
 
+/** Creates a new image partition from arbritary data */
```
H4ndl3: Typo, should be `arbitrary`.

Brother-Lal:
Thanks, fixed the typo!

Brother-Lal:
@jannic tested the pr and noticed that sysupgrade is not working.

```
root@LEDE:/tmp# sysupgrade sysuparch.bin 
Image metadata not found
Sysupgrade is not yet supported on archer-c25-v1.
Image check 'platform_check_image' failed.
```
Anybody has an idea how to fix this?

heinzek:
Add board name to target/linux/ar71xx/base-files/lib/upgrade/platform.sh
sample:
https://github.com/lede-project/source/pull/677/commits/47f3690d3b8b2830d8646b87aa09be61615129ae#diff-99173fa14409dec55cab43950a64f5ef

Brother-Lal:
Thank you!
All those little files where just one line needs to be added :(

Sysupgrade is now working.

pepe2k:
Hi @Brother-Lal,

Is this one ready (discussion got long here)? Can you write short summary, what's working, what not, if you need help with anything else?

Cheers,
Piotr

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -104,6 +104,22 @@ define Device/tplink-16mlzma
...
+  BOARDNAME := ARCHER-C25-V1
+  TPLINK_BOARD_NAME := ARCHER-C25-V1
+  DEVICE_PROFILE := ARCHERC25V1
+  IMAGE_SIZE := 7936k
+  LOADER_TYPE := elf
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade
+  IMAGE/factory.bin := append-rootfs | tplink-safeloader factory
+  MTDPARTS := spi0.0:128k(factory-uboot)ro,64k(u-boot)ro,1536k(kernel),6272k(rootfs),128k(config),64k(art)ro
```
jannic: I had to add a firmware partition to MTDPARTS to make sysupgrade working:
`MTDPARTS := spi0.0:128k(factory-uboot)ro,64k(u-boot)ro,1536k(kernel),6272k(rootfs),128k(config),64k(art)ro,7808k@0x30000(firmware)`
Without that, I got:
`Could not open mtd device: firmware` during sysupgrade.
(Only tried on 17.01 based gluon build, _maybe_ this works on lede master, but I don't see how)

Brother-Lal:
@jannic Interesting, but i had no such problem.
Did you add a firmware partition somewhere in your backport? Did you do a full reflash with the factory image(with the sysupgrade changes) before testing the sysupgrade image?

Thanks @pepe2k , everything you expect should now be working, wifi,lan/wan,leds,sys+factory, buttons.
I would say it is commit ready if no one has any more issues. But we might wait if jannic made an error in backporting or if it is indeed something that works for me, but not in another environment ;)

jannic:
It's not really a separate partition, it's just an alias for (kernel + rootfs), so the image can be written in a single write command. AFAIKT the sysupgrade script writes the firmware via `default_do_upgrade()` from `package/base-files/files/lib/upgrade/common.sh`.

The command boils down to `cat firmware_image | mtd write - firmware`, which needs a partition named `firmware`.

This is consistent to my observations. But the call to `default_do_upgrade` comes after several indirections, perhaps I'm missing something and on LEDE `platform_do_upgrade_combined` or some other variation gets called?

I'll do another build using your current PR and just try it out :-)

target/linux/ar71xx/image/tp-link.mk
heinzek: Add "ro" atribute to config part.
@jannic have right. add firmware part.

pepe2k:
@jannic, indeed, without the `firmware` mtd partition sysupgrade will not work on this board. @Brother-Lal, please, take a look how support for other boards with `tplink-safeloader` image type was done - all of them have `firmware` partition which covers `kernel` and `rootfs`.

@Brother-Lal, please, explain also changes in `tplink-safeloader.c`, it looks for me that you are trying to fix some data which has been accidentally overwritten/erased because of the wrong partition sizes/offsets.

Cheers,
Piotr

jannic:
Just checked with a clean compile of 9590c69 (the PR by @Brother-Lal ):
```
root@LEDE:/tmp# sysupgrade lede-ar71xx-generic-archer-c25-v1-squashfs-sysupgrade.bin 
Image metadata not found
Saving config files...
Sending TERM to remaining processes ... odhcpd dnsmasq ntpd ubusd logd netifd 
Sending KILL to remaining processes ... 
Switching to ramdisk...
Performing system upgrade...
Could not open mtd device: firmware
Can't open device for writing!
Upgrade completed
Rebooting system...
```

→ the sysupgrade has not been performed.

RubenKelevra:
Just a sidenote:

This are the numbers of a pretty similar router tested by the FCC. If the numbers hold up, the 5 GHz coverage should be outstanding, in opposite to older low cost devices like the WDR3500.

img_1255

So the range should be very good, for 2.4 as well as 5 GHz wifi, can you confirm this works as expected @Brother-Lal ?

Brother-Lal:
@jannic Thanks for testing, finding my bugs and then even providing a solution!
Seems i should be more carfeul, sysupgrade runs != sysupgrade is flashed... :(

@RubenKelevra define range? Seems like the 5G wifi has 200mW/23dBm and the 2.4 has 316mW/25dBm if no country option is set, output by iwinfo txpowerlist. Haven't done any rangetests, though.

@pepe2k 
The changes to the safeloader partitions are to add the extra-para partition, which i suspect contains some metadata about the image/bootinformation, because if it is not there, the factory flasher complains about too many bootpartitions (see here: https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c#file-firmwareorigerror )
The data in the partition is taken 1:1 from the original firmware file, so i doubt that the flash process writes this partition on the device. Haven't investigated further, to be honest, my intenetion was to get the factory image working...

Brother-Lal:
I added the firmware partition definition and the ro for the config partition.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -104,6 +104,22 @@ define Device/tplink-16mlzma
   IMAGE_SIZE := 15872k
 endef
 
+define Device/archer-c25-v1
+  DEVICE_TITLE := TP-LINK Archer C25 v1
+  DEVICE_PACKAGES := kmod-ath10k ath10k-firmware-qca9887
+  BOARDNAME := ARCHER-C25-V1
+  TPLINK_BOARD_NAME := ARCHER-C25-V1
+  DEVICE_PROFILE := ARCHERC25V1
+  IMAGE_SIZE := 7936k
```
heinzek: image size can not exceed the size of the firmware partition. Change size to 7808k

Brother-Lal: thank you, fixed

heinzek:
Check if MAC address on LAN/WAN/WLAN/WLAN5G are the same as in the factory firmware.

Brother-Lal:
Mac is the same as printed on the device itself, for wireless and +1 for lan/wan.
Sadly i noticed another issue, i cannot flash back onto the original firmware anymore in tftp mode, complains about software version mismatch (flashing an openwrt factory image works though...)

I will try to see if my tests with different images might have overwritten anything by writing my backup to flash and the installing lede with all the fixes fresh....

pepe2k:
@Brother-Lal,

> The changes to the safeloader partitions are to add the extra-para partition, which i suspect contains some metadata about the image/bootinformation, because if it is not there, the factory flasher complains about too many bootpartitions (see here: https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c#file-firmwareorigerror )
> The data in the partition is taken 1:1 from the original firmware file, so i doubt that the flash process writes this partition on the device. Haven't investigated further, to be honest, my intention was to get the factory image working...

Can you maybe provide me a full flash dump from this device?

Cheers,
Piotr

Brother-Lal:
Github has problems with the file, so here: https://upthis.de/dlf/6yh2zz/mtd0archerc25.zip

I'll extract all changes to the safeloader.c into a separate commit, with a description.

pepe2k:
> Github has problems with the file, so here: https://upthis.de/dlf/6yh2zz/mtd0archerc25.zip

It doesn't work :(

Brother-Lal:
$%&%$/(/(%$&!§$&§//( try this: https://upthis.de/dl/6yh2zz/mtd0archerc25.zip

pepe2k:
@Brother-Lal,

> $%&%$/(/(%$&!§$&§//( try this: https://upthis.de/dl/6yh2zz/mtd0archerc25.zip

Thanks, I will look at it later.

Also, please test your changes with `scripts/checkpatch.pl`. For example, you can use it like below, assuming you want to check last applied commit in active branch:

`git format-patch --stdout -1 | ./scripts/checkpatch.pl -`

Cheers,
Piotr

Brother-Lal:
I split the commit, hopefully without any errors :/

Tried the tftp-flash with original after factory flash, still get the same error:
https://gist.github.com/Brother-Lal/d3114b76bdc0740749dce0f9da8e673c#file-tftporigfactoryafterlede

Seems some data is indeed overwritten or missing from the factory file to allow going back to original firmware.

heinzek:
Can You compare partition "soft-version" after flashing LEDE?
I think tplink-safeloader.c need change and add custom "soft-version" partition with data looks like:
`...Q....soft_ver:1.0.0 Build 20160725 Rel. 77078.fw_id:985B2ECC075F299461B49FB2F69DABE7...`

Maybe only soft_ver:1.0.0 is needed.

RubenKelevra:
@Brother-Lal 200 mW sounds like artificial restrictions but it's definitely better than an old cheap WDR3500. Can you do some easy range tests, if the value holds up, just to be sure that the radio works fine. :)

jannic:
As @heinzek suspected, writing "soft_ver:1.0.0\n" is sufficient.
Writing a firmware built with the patch at https://github.com/jannic/source/commit/f6bf42e8e565ff1f5e244eecdf9af71d79d9c661, using the original firmware's GUI, allows for a later tftp recovery back to the original firmware without the error message `the firmware software version dismatched`.

Please note that a normal `sysupgrade` doesn't write the config partition, so once you 'broke' tftp recovery, writing a fixed firmware using `sysupgrade` doesn't fix tftp.

I successfully tried two methods to recover from this situation:

- manually wrote a fixed config partion using `mtd write fixed_config_image config`
- manually wrote the original firmware using `mtd write kernel_and_rootfs_from_orignal_firmware firmware`, then booted original firmware and flashed a fixed firmware (or the original one, again) via the GUI

heinzek:
@jannic Great Job. 
I test this on C59/C60, because factory image upgrade not working on original firmware GUI.

jannic:
Wouldn't it be great to check the boot loader source code, so we could be sure what we write to soft_ver partition matches what the boot loader is looking for?
Well, U-Boot is GPL, so it should be possible. So, I got the GPL source archive from http://static.tp-link.com/resources/gpl/Archer_C25v1_GPL.tar.bz2 and looked for the error message.

First step was fine: The error message, `[NM_Error](nm_checkUpdateContent) 01164: the firmware software version dismatched`, obviously comes from U-Boot, generated by the following macro in `openwrt/qca/src/qca-legacy-uboot/nvram_mngr/nm_lib.h`:
```
#define NM_ERROR(fmt, args...)  \
        printf("[NM_Error](%s) %05d: "fmt"\r\n", __FUNCTION__, __LINE__, ##args)
```

However, searching for a function called `nm_checkUpdateContent` only gave a single hit, in `openwrt/qca/src/qca-legacy-uboot/nvram_mngr/nm_fwup.c`. But the function is between line numbers 596 and 751, the whole file has only 893 lines. How can the error be generated in line 1164? And other parts of the error message, especially the slightly unusual word `dismatched`, can't be found anywhere.

So I'm quite sure that TP-Link doesn't deliver the source code of the boot loader actually used. At least not in the file they offer for the Archer C25. :-(

Brother-Lal:
@heinzek very nice that the c59/c60 is working!
i could try to implement this around the weekend, but i'm a poor programmer...
A tool to dissect tplink-firmware would be nice, too :D

@jannic good idea with the uboot, totally forgot they have to publish the sources!
They have 2 uboot versions installed in the archer (thus 2 uboot partitions), so there might be some confusion on their side. It probably is a license violation otherwise... maybe asking nicely will help?

RubenKelevra:
> A tool to dissect tplink-firmware would be nice, too :D
> 

You may want to try this tool, if you got different versions:
https://security.googleblog.com/2016/03/bindiff-now-available-for-free.html?m=1

Brother-Lal:
@RubenKelevra  Ahh, nice. bookmarked!
I was more thinking about a tool more suited for TP-Link factory file formats, as they seem to have a more or less constant format. Run it over a new file and see if its a format already known and give info about the partitions, checksums, etc. A binwalk signature/plugin would probably be the most useful way.

jannic:
I already wrote a friendly email to GPL@tp-link.com, asking for the source code, on sunday.
Today I got an answer pointing out that there are u-boot sources in the tar file, but completely missing the point that these are not the sources of the version installed on the router.
Well, sent another friendly email, asking for the correct version. :-)

RubenKelevra:
Thanks for your work Jan! :)

jannic:
Just got an email from tp-link, telling me that they updated the GPL source code of u-boot. And really, it looks like the code checking the soft_ver version is included now.
Currently I don't have time to have a closer look. The tarball is available here: http://static.tp-link.com/resources/gpl/Archer_C25v1_GPL.tar.bz2 (linked from http://www.tp-link.com/en/support/gpl/) and the relevant code is in openwrt/qca/src/qca-legacy-uboot/nvram_mngr/nm_fwup.c

Brother-Lal:
Nice that TP-Link folks do their job :D

Now that the firmware-utils are merged, i'll do a rebase and then hopefully the fully functional support can be merged :)

pepe2k:
> Now that the firmware-utils are merged, i'll do a rebase and then hopefully the fully functional support can be merged :)

If you can, please rebase on my [staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) and I will pull it.

Cheers,
Piotr

RubenKelevra:
Great work guys, we really appreciate that!

Brother-Lal:
I have it rebased, and pushed here.

Currently building, if it works i'll update :)

Brother-Lal:
Ok, build flashes, wifi can be enabled, and a reflash onto the factory firmware is possible.
Seems all the blocking kinks are out ;)

JamesT42:
I was succesful in installing it on my device.
I found one bug so far: the green and amber wan leds are switched. In default config after install the green led is setup in luci, but it lights amber.

Brother-Lal:
Can reproduce, will fix! Thanks!!

rotanid:
@pepe2k anything missing before you can/will pull it?

pepe2k:
@Brother-Lal, thank you for your contribution and all others involved here in discussion! I have finally :innocent: merged your changes (with minor fixes) into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

For now, I will keep your changes in `tplink-safeloader.c` as it but I expect that we would need `extra-para` partition also for other devices in near future.

As it turned out, it's needed for devices with dual U-Boot images (C25 seems to be just a first example of dual U-Boot and safeloader image type combination). From GPL sources `openwrt/qca/src/qca-legacy-uboot/nvram_mngr/nm_fwup.c`:

```C
/*******************************************************************
 * Name		: nm_checkUpgradeMode
 * Abstract	: check the boot mode
 * Input	: 
 * Output	: 
 * Return	: OK/ERROR
 */
int nm_checkUpgradeMode(NM_PTN_STRUCT *ptnStruct)
{
    NM_PTN_ENTRY *ptnEntry = NULL;
    int upDbootFound = FALSE;
    int flashDbootFound = FALSE;
	NM_PTN_EXTRA_PARA *ptnSysParaData = NULL;
	
	/* not allowed to upgrade between double boot version 
	 * and single boot version, "extra-para" all exist or 
	 * not and double boot flag must equal
	 */

	/* check partition-table partition base address is equal or not*/
#if defined(CONFIG_DOUBLE_UBOOT_FACTORY) || defined(CONFIG_DOUBLE_UBOOT_SECOND)
	if ((ptnEntry = nm_lib_ptnNameToEntry(ptnStruct, NM_PTN_NAME_PTN_TABLE)) == NULL)
	{
		NM_ERROR("Up file partition-table not found!\n");
		return ERROR;
		
	}

	if (ptnEntry->base != NM_PTN_TABLE_BASE)
	{
		NM_ERROR("Up file partition-table base address changed up(0x%08x) flash(0x%08x)!\n",
			ptnEntry->base, NM_PTN_TABLE_BASE);
		return ERROR;		
	}
#endif

	/* do not allow double boot upgrade to single boot or single to double */
	if ((ptnEntry = nm_lib_ptnNameToEntry(ptnStruct, NM_PTN_NAME_EXTRA_PARA)) != NULL)
	{
		/* double boot up file must have extra-para partition for boot mode check */
		if (ptnEntry->upgradeInfo.dataType == NM_FWUP_UPGRADE_DATA_FROM_FWUP_FILE)
		{
			ptnSysParaData = (NM_PTN_EXTRA_PARA *)((char*)ptnEntry->upgradeInfo.dataStart + 8);

			if (ptnSysParaData->dbootFlag == NM_FWUP_IS_DOUBLE_BOOT)
			{
				upDbootFound = TRUE;
			}
		}
	}

#if defined(CONFIG_DOUBLE_UBOOT_FACTORY) || defined(CONFIG_DOUBLE_UBOOT_SECOND)
	flashDbootFound = TRUE;
#endif

	if (upDbootFound != flashDbootFound)
	{
		NM_ERROR("Double boot flag up %d flash %d not ok!",
			upDbootFound, flashDbootFound);
		return ERROR;
	}
	
	return OK;
}
```

Cheers,
Piotr

Brother-Lal:
Thanks you for the supervision and merging!

I think a small rework for the safeloader.c would be in order anyway - best would probably be to make arbitrary partitions possible ... and an inspection capability like mktplinkfw2 would be be nice too.

rubo77:
Heißt der jetzt eigentlich ac900 oder c25?

belzebub40k:
@rubo77 Die offizielle Modellbezeichnung ist Archer C25.

PR #1070, opened 2017-04-27
ramips: add new device ZyXEL Keenetic as kn

Spider84:
Specification:

- SoC: Ralink RT3052 (MIPS24Kc) @384MHz
- RAM: 32 MiB
- Flash: 8 MiB
- WLAN: WiSoC 2T2R/300Mbps (2.4GHz)
- LAN: 4x100M
- WAN: 1x100M
- USB: 1x2.0

This device exactly same as NBG-419N but with USB port and USB Led.

- add USB Led definition and OTG Port in Device Tree
- add kn definition in all necessary scripts like switch configuration and led configuration...

**Installation through U-Boot via TFTP.** _(copy-paste from OpenWRT Wiki)_
_NOTE: Serial connection and configured TFTP server needed!_

- After successfull compiling firmware, move bin/ramips/lede-snapshot-r4048-4b4f739-ramips-rt305x-kn-squashfs-sysupgrade.bin file into your TFTP server's directory.
- Then rename firmware to something shorter, for example "sysupgrade.bin" (the file name must be shorter than 32 symbols).
- Connect serial port to device (speed is 57600 8N1).
- When you power on device, and see U-Boot log, immediatly push "2" once.

You will see this message:
```
2: System Load Linux Kernel then write to Flash via TFTP. 
 Warning!! Erase Linux in Flash then burn new one. Are you sure?(Y/N)
```
Push "y", and enter: device IP, then TFTP server's IP, and then openwrt's firmware file name:
 ```
Please Input new ones /or Ctrl-C to discard
        Input device IP (192.168.1.1) ==:10.10.1.1
        Input server IP (192.168.1.2) ==:10.10.1.72
        Input Linux Kernel filename () ==:sysupgrade.bin
```
If TFTP server installed correctly, the firmware would be downloaded at ~30 seconds. Then it will be flashed into device (It will take about 2 minutes).

Signed-off-by: Alexey Belyaev <spider@spider.vc>

### CODE REVIEW

File: target/linux/ramips/dts/kn.dts [hunk 1]
```diff
@@ -0,0 +1,112 @@
+/dts-v1/;
+
+#include "rt3050.dtsi"
+
+#include <dt-bindings/input/input.h>
```
mkresin: Please include <[dt-bindings/gpio/gpio.h](http://lxr.free-electrons.com/source/include/dt-bindings/gpio/gpio.h?v=4.4)> here as well and keep alphabetical order.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -265,6 +265,7 @@ mzk-ex750np)
 na930)
 	set_usb_led "$board:blue:status"
 	;;
+kn|\
```
mkresin: please move the whole block up to keep alphabetical order

target/linux/ramips/base-files/lib/upgrade/platform.sh
```diff
@@ -105,6 +105,7 @@ platform_check_image() {
 	mzk-ex750np|\
 	mzk-w300nh2|\
 	mzk-wdpr|\
+	kn|\
```
mkresin: please keep alphabetical order

target/linux/ramips/dts/kn.dts
```diff
@@ -0,0 +1,112 @@
...
+			label = "firmware";
+			reg = <0x50000 0x3b0000>;
+		};
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power {
+			label = "nbg-419n:green:power";
```
mkresin: ```kn:green:power``` please. Some applies to all other leds as well. I'm quite sure neither your usb nor the power led in diag does work for you.

mkresin:
Even if your board is similar to an existing one, would you please add a brief description of the hardware and outline how LEDE can be installed on the board to the commit message. Have a look at the recent ramips board additions for some good examples.

mkresin:
Please get rid of the merge commit.

> Installation through U-Boot via TFTP.

Well, that isn't really a description how the installation could be done. Again, please check the recent ramips board additions for some _good_ examples.

mkresin:
@Spider84 I've squashed your commits and merged your PR into [my staging](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I've slightly adjusted your commit message.

Would you please have a look at the commit in my staging tree and confirm that the installation instructions and the code changes are still fine.

mkresin:
merged, thanks!

PR #1071, opened 2017-04-27
ar71xx: fix tl-wr841n-v8 switch port mapping

oldium:
The switch port is eth1. The order of switch ports (shown in `swconfig dev eth1 show`) is CPU,
LAN 4, LAN 1, LAN 2, LAN 3.

Signed-off-by: Oldřich Jedlička <oldium.pro@gmail.com>

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -300,7 +300,6 @@ ar71xx_setup_interfaces()
 	smart-300|\
 	tl-mr3420-v2|\
 	tl-wdr6500-v2|\
-	tl-wr841n-v8|\
 	tl-wr940n-v4|\
 	tl-wr941nd-v5|\
 	tl-wr941nd-v6|\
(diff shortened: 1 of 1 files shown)


Bracket-:
Tested and confirmed to match [port indexes on OpenWRT wiki](https://wiki.openwrt.org/toh/tp-link/tl-wr841nd#hardware_v8). All ports work with this PR while port 3 (going by case marking) doesn't work on master.

Disclaimer: I have no idea how any of this works, just stating what I see and is reported by swconfig

For what it's worth, port indexes don't match the case's as far as I can tell: `1=4, 2=1, 3=2, 4=3`
I believe the wiki says the case doesn't match hardware, though?

pepe2k:
@Bracket-

> All ports work with this PR while port 3 (going by case marking) doesn't work on master.

Can you explain this one, please? Port3 is not working at all or is not mapped properly?

I should have somewhere v8... will do the test ASAP.

Cheers,
Piotr

pepe2k:
@oldium, thank you for your contribution! I have merged your change into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

I have also included TL-MR3420 v2 and TL-WR941N/ND v5 in the fix as they all share same reference design.

Cheers,
Piotr

PR #1072, opened 2017-04-28
ramips: support jumbo frame on mt7621 up to 2k

igvtee:
This patch is base on Gaetano Catalli work.

i change mtk_eth_soc.c driver part. because mt7621 don't need to
set FE_GDMA1_FWD_CFG register to enable jumbo frame. and keep
MAX_RX_LENGTH because other chips need it.

Signed-off-by: Michael Lee <igvtee@gmail.com>


airend:
Sorry for spamming here, but does `MT7620A` support jumbo frames? I can't tell for sure from the code, but changing MTU to 1508 doesn't work.

PR #1074, opened 2017-04-28
procd: skip reload_config if it's already running

angt:
Signed-off-by: Adrien Gallouët <adrien@gallouet.fr>


nbd168:
Did you actually run into this issue?
By the way, this approach still has a remaining race condition. If dealing with concurrent calls to reload_config matters to you, you should use locking.

blogic:
@angt no feedback. please reopen if you still want this merged

PR #1075, opened 2017-04-28
tools/cmake: update to 3.8.0

hnyman:
* update cmake to 3.8.0
* refresh patches

Release notes:
https://cmake.org/cmake/help/v3.8/release/3.8.html

Compile-tested with ipx806x, firmware build successfully.

### DIFF
--- a/tools/cmake/Makefile
+++ b/tools/cmake/Makefile
@@ -7,12 +7,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cmake
-PKG_VERSION:=3.7.2
+PKG_VERSION:=3.8.0
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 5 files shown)


diizzyy:
More unwanted commits :-/

hnyman:
Very very interesting.
I think that somebody has force-pushed the LEDE master yesterday.
Longer explanation in https://github.com/lede-project/source/pull/1076#issuecomment-298150565

hnyman:
Rebased, as the mailing discussion has clarified the force push issue.

diizzyy:
110 Does anyone actually compile natively on FreeBSD?

120 doesn't seem necessary?
https://git.alpinelinux.org/cgit/aports/tree/main/cmake/APKBUILD

130 Needed?

150 Not needed?
```  --parallel=n            bootstrap cmake in parallel, where n is number of nodes [1] ```

hnyman:
superseded by #1089

PR #1078, opened 2017-04-30
package: Add mtd-rw kernel module

fid0did0:
This package adds a kernel module that, when loaded, makes all
read-only MTD partitions writeable. This can be used to update
the bootloader, modify bootloader environment partitions, etc.

Directly taken from openwrt packages repository

### DIFF
--- a/package/kernel/mtd-rw/Makefile
+++ b/package/kernel/mtd-rw/Makefile
@@ -1,0 +1,49 @@
+#
+# Copyright (C) 2016 Joseph C. Lehner <joseph.c.lehner@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=mtd-rw
+PKG_VERSION:=git-20160214
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://github.com/jclehner/mtd-rw.git
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_SUBDIR=$(PKG_NAME)-$(PKG_VERSION)
+PKG_SOURCE_VERSION:=7e8562067d6a366c8cbaa8084396c33b7e12986b
+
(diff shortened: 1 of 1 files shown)


obsy:
Why? This package is already in repo: https://github.com/openwrt/packages/tree/master/kernel/mtd-rw

and can be directly installed via opkg http://downloads.lede-project.org/releases/17.01.1/targets/ar71xx/generic/packages/

mkresin:
Waht obsy said! The package is already available via the packages feed which is used by OpenWrt and LEDE.

Beside of that, the package looks like a workaround for something that should be done in source. If the bootloader enviroment needs to be writeable for some reason, the read only flag should be removed in the dts/mach file.

The one who is is keen enough to upgrade the bootloader should be able to compile an image on his/her own to remove the read only flags.

fid0did0:
Dear Mathias,

thanks for your clarification.

 

My tentative target is to generate an intermediate (factory) image to replace the bootloader. But in the regular image I would keep the read only flag.

 

Shall I make two different dts?

How shall I point them in LEDE?

 

Thanks.

Bye.

 

Von: Mathias Kresin [mailto:notifications@github.com] 
Gesendet: Montag, 1. Mai 2017 09:18
An: lede-project/source <source@noreply.github.com>
Cc: Giuseppe Lippolis <giu.lippolis@gmail.com>; Author <author@noreply.github.com>
Betreff: Re: [lede-project/source] package: Add mtd-rw kernel module (#1078)

 

Waht obsy said! The package is already available via the packages feed which is used by OpenWrt and LEDE.

Beside of that, the package looks like a workaround for something that should be done in source. If the bootloader enviroment needs to be writeable for some reason, the read only flag should be removed in the dts/mach file.

The one who is is keen enough to upgrade the bootloader should be able to compile an image on his/her own to remove the read only flags.

bobafetthotmail:
There is also another person that posted on the mailing list that might need something like that to make sure that some devices with dual-partition setup always boot the right partition (the installer and maybe sysupgrade needs to manipulate a partition that is currently ro to make sure it has the right magic value for the bootloader).
http://lists.infradead.org/pipermail/lede-dev/2017-April/007120.html (see also the follow ups)

@bmork

bmork:
FWIW since I was mentioned: obsy expressed my thoughts precisely.

bobafetthotmail:
I mentioned you because you might be interested in the discussion about dealing with ro partitions temporarily during installation, not about this PR in particular.

PR #1080, opened 2017-05-02
libs/libevent2: Update to 2.1.8

diizzyy:
Update libevent2 to 2.1.8

Many thanks to Jo-Philipp Wich and Jonas Gorski for helping me out.

### DIFF
--- a/package/libs/libevent2/Makefile
+++ b/package/libs/libevent2/Makefile
@@ -8,19 +8,23 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libevent2
-PKG_VERSION:=2.0.22
+PKG_VERSION:=2.1.8
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/libevent-$(PKG_VERSION)-stable
 PKG_SOURCE:=libevent-$(PKG_VERSION)-stable.tar.gz
-PKG_SOURCE_URL:=@SF/levent
-PKG_HASH:=71c2c49f0adadacfdbe6332a372c38cf9c8b7895bb73dabeaa53cdcc1d4e1fa3
+PKG_SOURCE_URL:=https://github.com/libevent/libevent/releases/download/release-2.1.8-stable
+PKG_HASH:=965cc5a8bb46ce4199a47e9b2c9e1cae3b137e8356ffdad6d94d3b9069b71dc2
 PKG_MAINTAINER:=Jo-Philipp Wich <jo@mein.io>
 PKG_LICENSE:=BSD-3-Clause
 
 PKG_FIXUP:=autoreconf
 PKG_INSTALL:=1
 PKG_BUILD_PARALLEL:=1
+
+PKG_CONFIG_DEPENDS:= \
+	CONFIG_PACKAGE_libevent2-openssl \
+	CONFIG_PACKAGE_libevent2-pthreads
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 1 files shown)


diizzyy:
Run-tested on mips64

Supersedes #848 (runtime tested on ramips and ipq8***)

wongsyrone:
> OpenSSL and pthreads are linked to main lib in additional to the extra libs

link to openssl blindly is not the correct way to handle this.
You just need this commit: https://github.com/libevent/libevent/commit/5d18d87916728df67640bd225967c6d6df4a54f8

According to its commit message,

> (do not link ssl into every library)

diizzyy:
It's not linking blindly, https://github.com/diizzyy/source/blob/patch-27/package/libs/libevent2/Makefile#L123

wongsyrone:
Does CONFIG_PACKAGE_libevent2-openssl default to false?

diizzyy:
Yes, it does.

wongsyrone:
Have you tested these two situations?

1. compile and run test libevent2-openssl when CONFIG_PACKAGE_libevent2-openssl is false
2. the same to libevent2-pthreads

You have to be very cautious when dealing with dependencies.

diizzyy:
I'm not sure I'm following you here.

If you select libevent2-openssl, libopenssl gets enabled:
https://github.com/diizzyy/source/blob/patch-27/package/libs/libevent2/Makefile#L53
https://github.com/diizzyy/source/blob/patch-27/package/libs/libevent2/Makefile#L123

...likewise if you do the same with pthreads.
Each time openssl and/or pthreads selection changes the package system triggers a re-configure+rebuild as main lib and the extension gets linked to the external lib(s).
https://github.com/diizzyy/source/blob/patch-27/package/libs/libevent2/Makefile#L25

wongsyrone:
It works. But I still don't understand why not choose the easy way, backport a single commit, fewer modifications especially the dependency part, and you're done.

wongsyrone:
If you only take musl-libc into account, there no need to add the dependency to libpthread, it is embedded into libc.

diizzyy:
True but we still have other libraries and I'd rather have it in for consistency and correctness.

wongsyrone:
> If you enable libevent2-openssl it doesn't work on its own and the main lib libevent-2.1.so.* gets also linked to OpenSSL.

Do you mean libevent2-extra, -core, etc?

diizzyy:
tbh I only checked the main lib (libevent2) (not core, extra etc).

wongsyrone:
You can check it again with upstream commit 5d18d87 included

diizzyy:
That only affects -core which isn't the issue (that's what it says at least). The general guidelines is to stick with releases and upgrade/backport when needed. I fail to see that commit falling into the major issue-category.

diizzyy:
@blogic @jow- 
Something you can have a look at?

liudf0716:
@dizzy I don't think adding openssl and pthread dependment to all libevent library is a good idea. In fact, I have asked @azat about this question: why all libevent library need openssl, the answer was that it was a bug,here is the [issues](https://github.com/libevent/libevent/issues/473).

I have submitted a [PR](https://github.com/lede-project/source/pull/848#issuecomment-300411697) to  solve the problem, I don't know why u still submit another PR for it.

diizzyy:
That wasn't clear at the time and in that case we should just wait for a fixed release.

liudf0716:
@dizzyy for that bug, @azat,  who is maintainer of libevent,  has provided a solution, what  @wongsyrone
has done was officially recommended solution. for the bug was caused by one example, so I just commented the example to solve  it,　I thought that solution was also Ok.

diizzyy:
Let's close this PR and wait for a proper release instead...

PR #1083, opened 2017-05-04
[RFC] basefiles,procd: start process on reload [if not running]

commodo:
@dedeckeh @nbd168 

This is a continuation on change 6713694
That one changed the semantic from not doing restart on reload fail,
to defining default reload as restart.

But there is also the case where, if a service is not
running, reload could start it.

Of course, a reload hook could also implement this itself.
But, it feels like this would be a nice feature to have
for procd [able] services.

Still, if a service init script wants/needs to define
it's own `service_running` script, it can do that,
and it will be used instead of relying on procd [by default].

There are 2 base packages (system & qos-scripts) who's start & reload
calls do the same thing.
And it seems that for daemon-less init scripts, reload and start
do the same thing.

### DIFF
--- a/package/system/procd/files/procd.sh
+++ b/package/system/procd/files/procd.sh
@@ -413,6 +413,33 @@ _procd_set_config_changed() {
 	ubus call service event "$(json_dump)"
 }
 
+procd_service_running() {
+	local service="$1"
+
+	[ -n "$service" ] || return 1
+
+	json_init
+	json_add_string name "$service"
+	local result="$(ubus call service list "$(json_dump)")"
+
+	# load & process result
+	json_load "$result"
+
+	# silence warnings from here on
+	local _json_no_warning=1
+	result=1
+	json_select "$service" && \
+		json_select "instances" && \
+		json_select "instance1" && \
+		json_get_var result running && \
+		[ "$result" == "1" ] && \
+		result=0
+
+	json_cleanup
+
+	return $result
+}
+
 procd_add_mdns_service() {
 	local service proto port
 	service=$1; shift
--- a/package/base-files/files/etc/rc.common
+++ b/package/base-files/files/etc/rc.common
@@ -88,10 +88,6 @@ service_triggers() {
 	return 0
 }
 
-service_running() {
-	return 0
-}
-
 ${INIT_TRACE:+set -x}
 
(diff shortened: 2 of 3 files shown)


package/system/procd/files/procd.sh
```diff
@@ -413,6 +413,33 @@ _procd_set_config_changed() {
...
+	json_init
+	json_add_string name "$service"
+	local result="$(ubus call service list "$(json_dump)")"
+
+	# load & process result
+	json_load "$result"
+
+	# silence warnings from here on
+	local _json_no_warning=1
+	result=1
```
blogic: dont reuse the result var please it makes the code hard to read. also drop the default value assignment.

package/system/procd/files/procd.sh
```diff
@@ -413,6 +413,33 @@ _procd_set_config_changed() {
...
+	json_select "$service" && \
+		json_select "instances" && \
+		json_select "instance1" && \
+		json_get_var result running && \
+		[ "$result" == "1" ] && \
```
blogic: drop this from the && chain and call it unconditionally.

package/system/procd/files/procd.sh
commodo: ack

package/system/procd/files/procd.sh
commodo: ack

blogic:
i am not sure if this approach is the right one. in theory when sending a service signature to procd it should compare it to the existing one and then start it if changed and/or not running. if it does not do that, then we should fix it inside procd rather than add a shell wrapper. i'll look into this next week when i am near test hw again.

commodo:
I think the reload hook [if implemented by the service] does not get passed to procd.
It can be whatever you want it to be [ ex. ubus call ].

so, if a service is not running, it will just non-zero.
the discussion is more like: when calling a service reload, do we want the init.d logic to start the service ? or just let the reload call fail ?

blogic:
reload should start none running processes.

commodo:
understood ;
let's do it like this: i'll split up some of the commits from this PR into smaller PRs [ some are easy to agree upon ]
i'll re-spin the others based on comments/discussion here

PR #1084, opened 2017-05-05
ramips: make factory firmware for Tp-Link C20i/C50

heinzek:
add "strip padding from the end of the image" to reduce size of sysupgrade image.

I not set factory image to mr200 because it is not accetable by router.


mkresin:
I fail to get why this change is required. Could it be as easy that a sysupgrade.bin with metadata appended is not accepted by the TP-Link firmware? If that's the case, please add this explanation to the commit message.

Would you please split the commit into two. One commit which adds the factory images and another for the ```-s``` (strip padding?) parameter.

heinzek:
TP-Link firmware not accept sysupgrade image with append metadata. Without this firmware is accepted (if c50 have UN region is OK, but on EU still not work).

Commit is splited to two.

mkresin:
merged, thanks!

PR #1085, opened 2017-05-05
ramips: add support for Asus RT-AC51U

Redfoxmoon3:
Specification:
 - SoC: MediaTek MT7620A (580 MHz)
 - RAM: 64 MiB (Winbond W9751G6JB-25)
 - Flash: 16 MiB (Spansion S25FL128SAIF00)
 - LAN: x4 100M
 - WAN: x1 100M
 - Others: USB 2.0, reset button, wps button and 9 LEDs

Issues:
 - 5 GHz band is not functional (missing driver support)
 - No webui flashing due to custom asus TRX header

How to install:
 - Write squashfs-sysupgrade file to flash using U-boot and TFTP.

Signed-off-by: Ørjan Malde <foxyred333@gmail.com>

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -313,6 +313,10 @@ rp-n53)
 	ucidef_set_led_netdev "eth" "Network" "$board:white:back" "eth0"
 	set_wifi_led "$board:blue:wifi"
 	;;
+rt-ac51u)
+	set_wifi_led "$board:blue:wifi"
+	set_usb_led "$board:blue:usb" "1-1"
+	;;
 rt-n14u)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
 	ucidef_set_led_netdev "lan" "lan" "$board:blue:lan" eth0.1
--- a/target/linux/ramips/dts/RT-AC51U.dts
+++ b/target/linux/ramips/dts/RT-AC51U.dts
@@ -1,0 +1,126 @@
+/dts-v1/;
+
+#include "mt7620a.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "asus,rt-ac51u", "ralink,mt7620a-soc";
+	model = "Asus RT-AC51U";
+
+	gpio-leds {
+		compatible = "gpio-leds";
+				
+		power {
+			label = "rt-ac51u:blue:power";
+			gpios = <&gpio0 9 GPIO_ACTIVE_LOW>;			
+		};
+
+		usb {
+                        label = "rt-ac51u:blue:usb";
+                        gpios = <&gpio0 14 GPIO_ACTIVE_LOW>;
+                };
+
+		wifi {
+			label = "rt-ac51u:blue:wifi";
+			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+	
+		reset {
+			label = "reset";
+			gpios = <&gpio0 1 GPIO_ACTIVE_LOW>;			
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps {
+			label = "wps";
+			gpios = <&gpio0 2 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+
+	gpio_export {
+		compatible = "gpio-export";
+		#size-cells = <0>;
+
+		enable-leds {
+			gpio-export,name = "enable-leds";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 10 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&spi0 {
+	status = "okay";
+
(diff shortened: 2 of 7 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -313,6 +313,12 @@ rp-n53)
 	ucidef_set_led_netdev "eth" "Network" "$board:white:back" "eth0"
 	set_wifi_led "$board:blue:wifi"
 	;;
+rt-ac51u)
+	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
```
mkresin: Drop this LED here, it is already handled by diag.sh and should be on after boot.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -313,6 +313,12 @@ rp-n53)
...
+	ucidef_set_led_netdev "activity" "activity" "$board:blue:activity" eth0.1
+	ucidef_set_led_netdev "activity" "activity" "$board:blue:activity" eth0.2
+	ucidef_set_wifi_led "$board:blue:activity"
```
mkresin: I'm quite sure that doesn't work.

The contents of ```/sys/devices/platform/gpio-leds/leds/rt-ac51u:blue:activity/device_name``` is eth0.1, will be changed to eth0.2 and should be wlan0 in the end. As far as I know the netdev trigger doesn't have support for "monitoring" multiple devices (```echo "eth0.1 eth0.2 wlan0" > /sys/devices/platform/gpio-leds/leds/rt-ac51u:blue:activity/device_name```).

You have to decide for one netdev here.

target/linux/ramips/image/mt7620.mk
```diff
@@ -350,6 +350,14 @@ define Device/rt-n14u
 endef
 TARGET_DEVICES += rt-n14u
 
+define Device/rt-ac51u
+  DTS := RT-AC51U
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
+  DEVICE_TITLE := Asus RT-AC51U
+  DEVICE_PACKAGES := kmod-usb-core kmod-usb2 kmod-usb-ehci kmod-usb-ledtrig-usbport
```
mkresin: You can drop the ```kmod-usb-ledtrig-usbport``` here. You don't have (and don't setup) a LED which can be used for USB status indication (albeit you want to use the activity for usb).

target/linux/ramips/base-files/etc/board.d/01_leds
Redfoxmoon3: With this configuration, the LAN leds and WAN does turn on when ethernet is connected

mkresin: > With this configuration, the LAN leds and WAN does turn on when ethernet is connected

Which isn't related to this code, since you don't have any WAN or LAN leds in your dts file which can be controlled by the trigger.

target/linux/ramips/image/mt7620.mk
mkresin: Correction: According to the manual the LED is the USB LED.

target/linux/ramips/dts/RT-AC51U.dts
```diff
@@ -0,0 +1,106 @@
...
+	gpio-leds {
+		compatible = "gpio-leds";
+				
+		power {
+			label = "rt-ac51u:blue:power";
+			gpios = <&gpio0 9 GPIO_ACTIVE_LOW>;			
+		};
+
+		activity {
+			label = "rt-ac51u:blue:activity";
```
mkresin: According to the manual the LED is the USB LED. Please name the LED to usb and setup the the usb trigger in ```target/linux/ramips/base-files/etc/board.d/01_leds```

target/linux/ramips/base-files/etc/board.d/01_leds
mkresin: Please remove the ```ucidef_set_wifi_led``` here. The board has a dedicated 2.4/5GHz LED which should be used. If the "activity" led is indeed the 2.4/5GHz LED, the LED should be renamed in the dts and the usb LED needs to be added.

mkresin:
Merged into [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Thanks.

SyCoTechs:
Hi,
Still there are only 6 Leds on the Led Settings but the device actually have 9 LEDs
And 5ghz wlan dont work at all

bobafetthotmail:
> Hi,
> Still there are only 6 Leds on the Led Settings but the device actually have 9 LEDs

What are those leds for? If they are for wifi, they might not be controllable by OpenWrt opensource driver.
Some leds in devices are also not controllable at all (i.e. some power leds or others that are just wired to power lines).

> And 5ghz wlan dont work at all

**tl:dr it will be functional in the next OpenWrt stable release, due to land in a few months.**

The 5Ghz wlan chip (MediaTek MT7610EN) did not have a opensource driver in mainline Linux 
at the moment of OpenWrt 18.06 release.

Support for this chip was added to mainline Linux some time later, and now there is a package with the driver and firmware for it in OpenWrt master, see this commit https://github.com/openwrt/openwrt/commit/202ac58ca224032866f16c844cfc63396e2af928

And as you can see here, 
https://github.com/openwrt/openwrt/blob/master/target/linux/ramips/image/mt7620.mk#L588
the driver/firmware package was added to default packages for the device in OpenWrt master

This means that when you install the next OpenWrt stable release (will be out in a few months) you will be able to use 5Ghz wifi.

PR #1086, opened 2017-05-05
at91: convert boards to generic build target

BWhitten:
Needs runtime checking on the eval boards, I only have a custom board.

@claudyus Does this conversion look ok, it results in a .ubi per device instead of the generic unwrapped .squashfs and .ubifs previously.
Seemed safer in the case that a board included extra packages whilst compiling all profiles at once.

Signed-off-by: Ben Whitten <ben.whitten@gmail.com>


blogic:
LGTM, I would merge this if there are no further comments

target/linux/at91/image/Makefile
```diff
@@ -7,88 +7,29 @@
...
+  FILESYSTEMS := squashfs ubifs
+  DEVICE_DTS := $(1)
+  KERNEL_INSTALL := 1
+  KERNEL_SUFFIX := -uImage
+  KERNEL_NAME := zImage
+  KERNEL := kernel-bin | append-dtb | lzma | uImage lzma
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  SUBPAGESIZE := 512
+  MKUBIFS_OPTS := -m $$(PAGESIZE) -e 126KiB -c 2048
```
mkresin: Shouldn't this be UBIFS_OPTS?

target/linux/at91/image/legacy.mk
```diff
@@ -0,0 +1,97 @@
...
+
+define Device/at91sam9g15ek
+  DEVICE_TITLE := Atmel AT91SAM9G15-EK
+endef
+TARGET_DEVICES += at91sam9g15ek
+
+define Device/at91sam9g20ek
+  DEVICE_TITLE := Atmel AT91SAM9G20-EK
+  KERNEL := kernel-bin | lzma | uImage lzma
+  IMAGE/root.ubi := append-ubi | install-dtb
```
mkresin: This is misleading. The install-dtb build step isn't related/doesn't change the created image. Please do it this way:

```
define Device/at91sam9g20ek
  DEVICE_TITLE := Atmel AT91SAM9G20-EK
  KERNEL := kernel-bin | lzma | uImage lzma
  IMAGES += dtb
  IMAGE/dtb := install-dtb
define Device/at91sam9g20ek
```

Please do it the same way for all other images which need the dtb file.

mkresin:
Not necessarily a change request, since I'm not sure about the intended use case of the images. I wouldn't expect that the atmel target is intended for end users.

If you use the following:

```diff
diff --git a/target/linux/at91/image/Makefile b/target/linux/at91/image/Makefile
index dcebd1b..ac50c54 100644
--- a/target/linux/at91/image/Makefile
+++ b/target/linux/at91/image/Makefile
@@ -20,14 +20,12 @@ define Device/Default
   PROFILES := Default
   FILESYSTEMS := squashfs ubifs
   DEVICE_DTS := $(1)
-  KERNEL_INSTALL := 1
-  KERNEL_SUFFIX := -uImage
   KERNEL_NAME := zImage
   KERNEL := kernel-bin | append-dtb | lzma | uImage lzma
   BLOCKSIZE := 128k
   PAGESIZE := 2048
   SUBPAGESIZE := 512
-  MKUBIFS_OPTS := -m $$(PAGESIZE) -e 126KiB -c 2048
+  UBIFS_OPTS := -m $$(PAGESIZE) -e 126KiB -c 2048
   IMAGES := root.ubi
   IMAGE/root.ubi := append-ubi
 endef
diff --git a/target/linux/at91/image/legacy.mk b/target/linux/at91/image/legacy.mk
index f53f1a1..43bc77a 100644
--- a/target/linux/at91/image/legacy.mk
+++ b/target/linux/at91/image/legacy.mk
@@ -1,24 +1,36 @@
 define Device/at91sam9263ek
   DEVICE_TITLE := Atmel AT91SAM9263-EK
+  KERNEL_SIZE := 4096k
+  IMAGES := factory.bin
+  IMAGE/factory.bin := append-kernel | pad-to $$(KERNEL_SIZE) | append-ubi
 endef
 TARGET_DEVICES += at91sam9263ek
 
 define Device/at91sam9g15ek
+  KERNEL_SIZE := 6144k
   DEVICE_TITLE := Atmel AT91SAM9G15-EK
+  IMAGES := factory.bin
+  IMAGE/factory.bin := append-kernel | pad-to $$(KERNEL_SIZE) | append-ubi
 endef
 TARGET_DEVICES += at91sam9g15ek
 
 define Device/at91sam9g20ek
   DEVICE_TITLE := Atmel AT91SAM9G20-EK
   KERNEL := kernel-bin | lzma | uImage lzma
-  IMAGE/root.ubi := append-ubi | install-dtb
+  KERNEL_SIZE := 4096k
+  IMAGES := factory.bin dtb
+  IMAGE/factory.bin := append-kernel | pad-to $$(KERNEL_SIZE) | append-ubi
+  IMAGE/dtb := install-dtb
 endef
 TARGET_DEVICES += at91sam9g20ek
 
 define Device/at91sam9g20ek_2mmc
   DEVICE_TITLE := Atmel AT91SAM9G20-EK 2MMC
   KERNEL := kernel-bin | lzma | uImage lzma
-  IMAGE/root.ubi := append-ubi | install-dtb
+  KERNEL_SIZE := 4096k
+  IMAGES := factory.bin dtb
+  IMAGE/factory.bin :=  append-kernel | pad-to $$(KERNEL_SIZE) | append-ubi
+  IMAGE/dtb := install-dtb
 endef
 TARGET_DEVICES += at91sam9g20ek_2mmc
```

You can write the resulting image to the kernel (and rootfs) partition. No need to write kernel and roofs in separate steps to the flash.

Does the bootloader takes care of extending the ubi to the available flash space? If not, ubi EOF marker are required and added with `UBINIZE_OPTS := -E 5`.

target/linux/at91/image/Makefile
BWhitten: I wasn't sure here so followed the Device/Init defaults.
From https://github.com/lede-project/source/commit/9497a23ecbfb6e57d4697d7711bb2a5893208d1e it looks like MKUBIFS_OPTS will be the thing to use for per device options down the line, but only the imx6 target uses it currently.

target/linux/at91/image/legacy.mk
BWhitten: Thanks, that looks a lot cleaner will do.

target/linux/at91/image/Makefile
mkresin: ... and what you have done is completely right.

MKUBIFS_OPTS is for per device ubifs options. UBIFS_OPTS is used to set the options for the whole target/Makefile (outside of the device specific define).

Sorry for the noise.

BWhitten:
Thanks for the review @mkresin.
I think your suggestion of a factory.bin is very nice, I will put them together from the layouts in the dts'.
Most of the examples for the Atmel boards flash each component separately either from uboot or SAM-BA, having the factory.bin in addition to the components would be preferable I think.

I cant speak for the boards that use barebox, but I don't recall an issue with the ubi on boards which use uboot. How does this EOF marker work?
I thought the 'autoresize' volume flag in ubinize triggered the UBI to have the maximum possible size when UBI is first run. I could be mixing this up though.

mkresin:
> having the factory.bin in addition to the components would be preferable I think.

It was meant as replacement for the individual kernel and rootfs images. I'm not that thrilled about having additional factory.bins. It will just increase the required resources in terms of cpu time during build and storage.

> I thought the 'autoresize' volume flag in ubinize triggered the UBI to have the maximum possible size when UBI is first run. I could be mixing this up though.

That is correct. But UBI requires the PEBs to be formatted with a valid UBI header or empty
(0xFF). Depending on how many bytes you erase on your flash and whether 0xFF is written to the flash this might fail.

Long story short, due to the EOF marker, the not yet used PEBs are formatted on first boot.

Until a few month ago, the EOF marker was added by default. But it caused issues for bootloader with full UBI support. It was removed from the default image build code and added for each device which doesn't have a bootloader with native UBI support.

As long as you erase the whole to be ubi flash region with 0xFF it will work without the EOF marker. IMHO it is a nice to have to prevent issues if someone doesn't do everything the right way (or some OEM upgrade routine is used, which doesn't erase the whole flash region).

> I cant speak for the boards that use barebox, but I don't recall an issue with the ubi on boards which use uboot. How does this EOF marker work?

We  have custom patches in LEDE, which add an EOF marker to UBI Image [[1]](https://git.lede-project.org/?p=source.git;a=blob;f=tools/mtd-utils/patches/200-libubigen-add-ubigen_write_terminator-function.patch;hb=0dcc36fc7ddecf23b915621c6652dacf75006511)[[2]](https://git.lede-project.org/?p=source.git;a=blob;f=tools/mtd-utils/patches/201-ubinize-add-terminator-support.patch;h=cbb2802c5aaa4e3cc36a602d11277602117188da;hb=0dcc36fc7ddecf23b915621c6652dacf75006511).

Due to another LEDE kernel patch [[3]](https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/generic/patches-4.9/494-mtd-ubi-add-EOF-marker-support.patch;h=64a361a50aecac7e3888566279ceb2752fe09ccc;hb=HEAD), everything on the mtd blockdevice beginning at EOF marker is erased/formated and can be used afterwards.

BWhitten:
Thanks for the comments, I have adjusted it so that the evaluation boards stay in component form with no factory image whilst the end user boards just have the factory image with EOF markers added.

mkresin:
Merged to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary) with the extra changes we discussed about via e-mail.

Thanks a lot for your contribution and sorry for the delay.

PR #1087, opened 2017-05-07
dnsmasq: add dhcp-script hook for other packages

tzarc:
Adds a script which acts as a hook for when dnsmasq creates/destroys a
lease, or completes a TFTP file transfer. The hook loops through scripts
in '/etc/dnsmasq-dhcp.d', executing each one with the same arguments
supplied by dnsmasq.

Signed-off-by: Nick Brassel <nick@tzarc.org>

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77rc5
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq/release-candidates
@@ -150,10 +150,15 @@ define Package/dnsmasq/install
 	$(INSTALL_DATA) ./files/dnsmasq.conf $(1)/etc/dnsmasq.conf
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/dnsmasq.init $(1)/etc/init.d/dnsmasq
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/dhcp
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/neigh
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/ntp
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/tftp
 	$(INSTALL_DATA) ./files/dnsmasqsec.hotplug $(1)/etc/hotplug.d/ntp/25-dnsmasqsec
 	$(INSTALL_DIR) $(1)/usr/share/dnsmasq
 	$(INSTALL_DATA) ./files/rfc6761.conf $(1)/usr/share/dnsmasq/
+	$(INSTALL_DIR) $(1)/usr/lib/dnsmasq
+	$(INSTALL_BIN) ./files/dhcp-script.sh $(1)/usr/lib/dnsmasq/dhcp-script.sh
 endef
 
 Package/dnsmasq-dhcpv6/install = $(Package/dnsmasq/install)
--- a/package/network/services/dnsmasq/files/dhcp-script.sh
+++ b/package/network/services/dnsmasq/files/dhcp-script.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+case "$1" in
+	add)
+		export ACTION="add"
+		export MACADDR="$2"
+		export IPADDR="$3"
+		export HOSTNAME="$4"
+		exec /sbin/hotplug-call dhcp
+	;;
+	del)
+		export ACTION="remove"
+		export MACADDR="$2"
+		export IPADDR="$3"
+		export HOSTNAME="$4"
+		exec /sbin/hotplug-call dhcp
+	;;
+	old)
+		export ACTION="update"
+		export MACADDR="$2"
+		export IPADDR="$3"
+		export HOSTNAME="$4"
+		exec /sbin/hotplug-call dhcp
+	;;
+	arp-add)
+		export ACTION="add"
+		export MACADDR="$2"
+		export IPADDR="$3"
+		exec /sbin/hotplug-call neigh
+	;;
+	arp-del)
+		export ACTION="remove"
+		export MACADDR="$2"
+		export IPADDR="$3"
+		exec /sbin/hotplug-call neigh
+	;;
+	tftp)
+		export ACTION="add"
+		export TFTP_SIZE="$2"
+		export TFTP_ADDR="$3"
+		export TFTP_PATH="$4"
+		exec /sbin/hotplug-call tftp
+	;;
+esac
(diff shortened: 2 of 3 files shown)


dedeckeh:
Thanks for your patch; can you also increase the PKG_RELEASE in the dnsmasq package Makefile ?

tzarc:
Done, thanks!

kdarbyshirebryant:
You should squash those 2 commits into one.  This is a good idea and irrespective of the following question should go in.  What are the implications on running dnsmasq in a jail?

tzarc:
No problem, squashed.

Honestly not sure w.r.t. jails - is there some post-configuration changes for them in any way? Been using this setup for a few days now with the default configuration, and it seems like it's been functional the whole time... nothing stands out in `/etc/init.d/dnsmasq` as to enabling/changing options to use them.

At a glance, it seems it'd require adding the script + directory to the `procd_add_jail_mount` line, but that'd still be problematic if any user-specified scripts attempted to use anything else on the filesystem. Not sure how to proceed here; any suggestions about following this down the rabbit hole further?

bobafetthotmail:
can this use a dedicated subfolder in "/etc/hotplug.d" instead of "/etc/dnsmasq-dhcp.d"? (and maybe create it too on install)

There you have already subfolders for triggering scripts when block device or net or firmware or other things change.
see here for more info https://lede-project.org/docs/user-guide/hotplug_lede

tzarc:
I'm not really tied to the location; I hadn't seen the `/etc/hotplug.d` interface before, so if there's precedent already perhaps it's better to stick to convention. That said, seeing _"hotplug"_ makes me think devices, but given that there are other things like _ntp_ in there... probably doesn't matter.

The only real question I'd have is whether this would then need to be tied to `procd`, as the documentation you linked seems to state that scripts in `hotplug.d` are invoked by `procd` with special environment variables.

mkresin:
> can this use a dedicated subfolder in "/etc/hotplug.d" instead of "/etc/dnsmasq-dhcp.d"? (and maybe create it too on install)
>
> There you have already subfolders for triggering scripts when block device or net or firmware or other things change.
>see here for more info https://lede-project.org/docs/user-guide/hotplug_lede
>
> Note that this is a suggestion, what do others think about this?

Hey Alberto,

I had exactly the same in mind when I saw the PR. But the hotplug events are send by the kernel and as far as I can see, procd takes care of calling the scripts in /etc/hotplug.d. Not sure if it is a smart idea to mix it with userspace events. This should be discussed on the [lede-dev mailing list](https://lede-project.org/contact#mailing_lists).

If the script can be used for TFTP as well, it might should get a more generic name than `dhcp-script.sh`. But I fail to get form the dnsmasq man page how to distinguish between DHCP leases and TFTP file transfers.

dedeckeh:
@mkresin Not all hotplug events are generated by the kernel but also from user space; eg the ntp hotplug event is generated by the ntp busybox applet while the interface hotplug events are generated by netifd.
But indeed we should discuss this on the [lede-dev mailing list](https://lede-project.org/contact#mailing_lists) how we want to progress with this PR.

Distinction between dhcp/tftp/arp events can only be made based on the first argument passed to the script; eg for dhcp it's the string values add/del/old; for tftp it's tftp while for arp it's arp-add/arp-del

mkresin:
>  Not all hotplug events are generated by the kernel but also from user space

Thanks for the info, I wasn't aware of userspace triggered events.

> Distinction between dhcp/tftp/arp events can only be made based on the first argument passed to the script; eg for dhcp it's the string values add/del/old; for tftp it's tftp while for arp it's arp-add/arp-del

We should use the the dnsmasq dhcp-script in that case to clean this up a bit:

* add/del/old => call `/etc/hotplug/<whatever>/<dhcp lease scriptname>` ACTION=add/del/old
* arp-add/arp-del => call `/etc/hotplug/<whatever>/<arp scriptname>` ACTION=add/del

and so on.

jow-:
I'd prefer an integration with /etc/hotplug.d/dhcp/ as well, similar to how NTP hotplug events work.

An example implementation could look like:

    #!/bin/sh
    
    case "$1" in
    	add)
    		export ACTION="add"
    		export MACADDR="$2"
    		export IPADDR="$3"
    		export HOSTNAME="$4"	
    		exec /sbin/hotplug-call dhcp
    	;;
    	del)
    		export ACTION="remove"
    		export MACADDR="$2"
    		export IPADDR="$3"
    		export HOSTNAME="$4"	
    		exec /sbin/hotplug-call dhcp
    	;;
    	old)
    		export ACTION="update"
    		export MACADDR="$2"
    		export IPADDR="$3"
    		export HOSTNAME="$4"	
    		exec /sbin/hotplug-call dhcp
    	;;
    	arp-add)
    		export ACTION="add"
    		export MACADDR="$2"
    		export IPADDR="$3"
    		exec /sbin/hotplug-call neigh
    	;;
    	arp-del)
    		export ACTION="remove"
    		export MACADDR="$2"
    		export IPADDR="$3"
    		exec /sbin/hotplug-call neigh
    	;;
    	tftp)
    		export ACTION="add"
    		export TFTP_SIZE="$2"
    		export TFTP_ADDR="$3"
    		export TFTP_PATH="$4"
    		exec /sbin/hotplug-call tftp
    	;;
    esac

tzarc:
Updated with @jow-'s suggested implementation.
As for mailing list discussion... what's the process there? Do I need to raise this?

dedeckeh:
@tzarc Don't think we need to discuss this further on the mailing list as we agree on integration with /etc/hotplug.d/dhcp according to @jow's suggestion; will do some testing when I have time.
Did you test this with jailing enabled as I fear there will be issues with the current implementation ?

tzarc:
As mentioned before, I'm not sure how to enable/use the jail.

Hazarding a guess, the `procd_add_jail_mount` line in `dnsmasq.init` would need `/usr/lib/dnsmasq`, `/etc/hotplug.d`, and `/sbin/hotplug-call` added, but I'm doubtful of whether that's enough to allow hotplug to invoke things outside the jail.

package/network/services/dnsmasq/Makefile
```diff
@@ -150,10 +150,15 @@ define Package/dnsmasq/install
...
 	$(INSTALL_BIN) ./files/dnsmasq.init $(1)/etc/init.d/dnsmasq
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/dhcp
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/neigh
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/ntp
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/tftp
 	$(INSTALL_DATA) ./files/dnsmasqsec.hotplug $(1)/etc/hotplug.d/ntp/25-dnsmasqsec
 	$(INSTALL_DIR) $(1)/usr/share/dnsmasq
 	$(INSTALL_DATA) ./files/rfc6761.conf $(1)/usr/share/dnsmasq/
+	$(INSTALL_DIR) $(1)/usr/lib/dnsmasq
+	$(INSTALL_DATA) ./files/dhcp-script.sh $(1)/usr/lib/dnsmasq/dhcp-script.sh
```
dedeckeh: Use INSTALL_BIN as the dhcp-script.sh script needs to be installed with executable permission

tzarc:
Apologies for the delay - my ISP decided I no longer needed access to the internet, nor anybody else in my area.

I've made some changes setting up a per-dnsmasq-instance script, ensuring that the dhcpscript uci option is executed against the correct dnsmasq instance, before offloading to the hotplug script. There wasn't really a precedent for formatting, so functionality is present but I'm unsure if the scripts should be written/formatted in a different way.

dedeckeh:
Merged into my staging tree https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=701c0e579ca7f121c6cd6f2706be18ec59da733a
I reworked the user dhcpscript handling so the implementation is less bloated

Thanks

PR #1089, opened 2017-05-08
tools/cmake: update to 3.8.1

hnyman:
* update cmake to 3.8.1
* refresh patches

Release notes:
https://cmake.org/cmake/help/v3.8/release/3.8.html

Compile-tested with ipx806x, firmware built successfully.


nbd168:

PR #1091, opened 2017-05-09
ramips: Add support for the VoCore2 Lite.

pinney:
ramips: Add support for the VoCore2 Lite.

The VoCore2 Lite uses the same PCB as the Vocore2 with a MT7688A and 8M Flash/64M RAM
https://www.indiegogo.com/projects/vocore2-4-coin-sized-linux-computer-with-wifi#/
http://vocore.io/
http://vonger.net/

- Installing from the bootloader is recommended.
- The original firmware is LEDE/OpenWrt LuCI/sysupgrade can be used.
- However you may need to edit /sys/sysinfo/board_name and /lib/upgrade/platform.sh
- If using luci/sysupgrade use the -n option (do not keep settings)
- Reverting to the factory firmware use the bootloader if using LuCI/sysupgrade one may need to edit the same files.

Both VoCore2 devices use a common dtsi file.

VoCore2 minimal changes including :
memory registry set to 0x8000000 (128 RAM)
LED change from 'status' to 'wifi' devices have only a single LED onboard.
Remove kmod-sdhci-mt7620 from mt7628.mk
fixes mmc errors on VoCore2 basic unit (sd card not present)

https://github.com/pinney/source/commit/099c330f607264c73eec40fda59ff72ffee73525

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -336,6 +336,9 @@ vocore-16M)
 vocore2)
 	set_wifi_led "$board:fuchsia:status"
 	;;
+vocore2lite)
+	set_wifi_led "$board:green:status"
+	;;
 w502u)
 	set_usb_led "$board:blue:usb"
 	set_wifi_led "rt2800pci-phy0::radio"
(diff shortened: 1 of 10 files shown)


mkresin:
Hey Larry,

same request as on lede-dev. Please use an extra commit for any changes of the vocore2 (non lite). Otherwise it is a pain to do the review.

Please cleanup your commits. There are unrelated and merge commits.

PR #1093, opened 2017-05-10
f2fs-tools: Switch to gz tarball

diizzyy:
At some point kernel.org decided to drop xz generated tarballs, switch to gz which they still provide.


blogic:
merged, Thanks !

PR #1096, opened 2017-05-10
Add GCC 7

wongsyrone:


wongsyrone:
Please note that this version introduces new warning flags, such as -Wimplicit-fallthrough, -Wformat-truncation -Wformat-overflow, etc. If you compile with GCC 7, there are some quick "fixes" need to be done due to `-Werror`. I just ignore these warnings for testing, like https://github.com/wongsyrone/lede-1/commit/5aafc65775a935324eb26ec50295c405c82dbde3 and https://github.com/wongsyrone/lede-1/commit/b8a2c92141d3ba8051de90af58d3a642f28498d5

nbd168:
I already have gcc 7.1.0 in my staging tree. Could you please check that and see if you have any extra bits that I don't already have (aside from isl)?

wongsyrone:
Okay, now we can fix the malformed "@" symbol in the patch, it might be caused by pulling patch from mail client, although it's just a minor issue. It applies for GCC 5.4, 6 and 7

nbd168:
Merged your isl update to my staging tree and fixed patch 910 for gcc 7.1.0
Thanks

wongsyrone:
What about patch 910 in 5.4.0 and 6.3.0?

PR #1098, opened 2017-05-10
kernel: add hwmon for W83627EHF and family

pprindeville:
NCT6775 driver needs to run before W83627EHF, however, as it's the
driver of choice for NCT6775/6 (and family) chips and you want that
module to load first.

### DIFF
--- a/package/kernel/linux/modules/hwmon.mk
+++ b/package/kernel/linux/modules/hwmon.mk
@@ -97,7 +97,7 @@ define KernelPackage/hwmon-nct6775
   TITLE:=NCT6106D/6775F/6776F/6779D/6791D/6792D/6793D and compatibles monitoring support
   KCONFIG:=CONFIG_SENSORS_NCT6775
   FILES:=$(LINUX_DIR)/drivers/hwmon/nct6775.ko
-  AUTOLOAD:=$(call AutoProbe,nct6775)
+  AUTOLOAD:=$(call AutoLoad,51,nct6775)
   $(call AddDepends/hwmon,@PCI_SUPPORT @TARGET_x86 +kmod-hwmon-vid)
 endef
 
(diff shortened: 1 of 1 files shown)


package/kernel/linux/modules/hwmon.mk
```diff
@@ -313,6 +313,21 @@ endef
 $(eval $(call KernelPackage,hwmon-w83627hf))
 
 
+define KernelPackage/hwmon-w83627ehf
+  TITLE:=Winbond W83627EHF/EHG/DHG/UHG, W83667HG, NCT6775F, NCT6776F monitoring support
+  KCONFIG:=CONFIG_SENSORS_W83627EHF
+  FILES:=$(LINUX_DIR)/drivers/hwmon/w83627ehf.ko
+  AUTOLOAD:=$(call AutoLoad,52,w83627ehf)
```
blogic: Can't this be set to autoprobe ?

package/kernel/linux/modules/hwmon.mk
```diff
@@ -97,7 +97,7 @@ define KernelPackage/hwmon-nct6775
   TITLE:=NCT6106D/6775F/6776F/6779D/6791D/6792D/6793D and compatibles monitoring support
   KCONFIG:=CONFIG_SENSORS_NCT6775
   FILES:=$(LINUX_DIR)/drivers/hwmon/nct6775.ko
-  AUTOLOAD:=$(call AutoProbe,nct6775)
+  AUTOLOAD:=$(call AutoLoad,51,nct6775)
```
blogic: does w83627ehf have a dependency on nct6775 ? if that is the case then both modules should be auto probable

dangowrt: no, there is no dependency, but both drivers claim the same hardware. however, the implementation to support NCT6775 in W83627EHF is rather incomplete, hence the original nct6775 module should be preferred. This can still be achieved with AutoProbe if the w83627ehf driver is patched not to claim the nct677x hardware. Philip is currently testing this, see also
https://git.lede-project.org/?p=lede/dangole/staging.git;a=commit;h=9c01b3dbe96759a710bd24c13adf96b3be188408

pprindeville:
@dangowrt your patch works as expected.

dangowrt:
Merged with https://github.com/lede-project/source/commit/0dcc36fc7ddecf23b915621c6652dacf75006511

package/kernel/linux/modules/hwmon.mk
dangowrt: @pprindeville got back to me and confirmed that using AutoProbe with patched w83627ehf works for him on hardware with NCT6775. Hence I merged that version.

PR #1103, opened 2017-05-12
package/openvpn: add proto netifd proto handler

feckert:
I want to start a discussion why not use a openvpn proto handler for netifd?
Based on sources i found on this Page https://forum.openwrt.org/viewtopic.php?id=63522

should fixes this issue.
https://github.com/openwrt/packages/issues/3486
https://github.com/openwrt/luci/issues/845

### DIFF
--- a/package/network/services/openvpn/files/lib/netifd/openvpn-up
+++ b/package/network/services/openvpn/files/lib/netifd/openvpn-up
@@ -1,0 +1,87 @@
+#!/bin/sh
+
+#DBG=1
+
+[ -n "$INTERFACE" ] || {
+	echo "openvpn-up no INTERFACE variable appended"
+	exit 1
+}
+
+[ -n "$DBG" ] && {
+	env > /tmp/openvpn-up-${INTERFACE}-env.log
+}
+
+. /lib/netifd/netifd-proto.sh
+proto_init_update "$dev" 1 1
+proto_set_keep 1
+[ -n "$ifconfig_local" ] && proto_add_ipv4_address "$ifconfig_local" 32
+[ -n "$ifconfig_ipv6_local" ] && proto_add_ipv6_address "$ifconfig_ipv6_local" 128
+
+# set ipv4 routen
+c=1
+while true ; do
+	v=route_network_$c
+	eval "network=\$$v"
+	[ -n "$network" ] || break
+	v=route_gateway_$c
+	eval "gw=\$$v"
+	v=route_netmask_$c
+	eval "mask=\$$v"
+	v=route_metric_$c
+	eval "metric=\$$v"
+	[ -n "$metric" ] || metric="$METRIC"
+
+	proto_add_ipv4_route "$network" "$mask" "$gw" "" "$metric"
+	c=$(( $c+1 ))
+done
+
+# set default route
+[ -n "$DEFAULTROUTE" ] && {
+	proto_add_ipv4_route "0.0.0.0" "0" "$gw" "0.0.0.0/0" "$METRIC"
+}
+
+# set ipv6 routen
+c=1
+while true ; do
+	v=route_ipv6_network_$c
+	eval "network=\$$v"
+	[ -n "$network" ] || break
+	v=route_ipv6_gateway_$c
+	eval "gw=\$$v"
+
+	pref=${network%%/*}
+	preff=$pref/
+	preflen=${network#$preff}
+
+	proto_add_ipv6_route "$pref" "$preflen" "$gw"
+	c=$(( $c+1 ))
+done
+
+# set dns server
+c=1
+while true ; do
+	v=foreign_option_$c
+	eval "fo=\$$v"
+	[ -n "$fo" ] || break
+	[ "${fo%% *}" = "dhcp-option" ] && {
+		s=${fo#dhcp-option }
+		[ "${s%% *}" = "DNS" ] && {
+			dns=${s#DNS }
+			[ -n "$PEERDNS" ] && {
+				proto_add_dns_server "$dns"
+			}
+		}
+	}
+	c=$(( $c+1 ))
+done
+
+proto_send_update "$INTERFACE"
+
+[ -d /etc/openvpn/ip-up.d ] && {
+	for SCRIPT in /etc/openvpn/ip-up.d/*
+	do
+		[ -x "$SCRIPT" ] && "$SCRIPT" "$@"
+	done
+}
+
+exit 0
--- a/package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
+++ b/package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
@@ -1,0 +1,95 @@
+#!/bin/sh
+
+[ -x /usr/sbin/openvpn ] || exit 1
+
+[ -n "$INCLUDE_ONLY" ] || {
+	. /lib/functions.sh
+	. /lib/functions/network.sh
+	. ../netifd-proto.sh
+	init_proto "$@"
+}
+
+proto_openvpn_init_config() {
+	available=1
+	no_device=1
+	proto_config_add_string "ifname"
+	proto_config_add_defaults
+}
+
+get_config_param() {
+	local variable="$1"
+	local config="$2"
+	local option="$3"
+	local default="$4"
+
+	local value="$(sed -rne 's/^[ \t]*'"$option"'[ \t]+([^ \t]+)[ \t]*$/\1/p' "$config" | tail -n1)"
+	export -n "$variable=${value:-$default}"
+}
+
+proto_openvpn_setup() {
+	local cfg="$1"
+	local iface="$2"
+
+	local config dev_type opts
+
+	local ifname $PROTO_DEFAULT_OPTIONS
+	json_get_vars ifname $PROTO_DEFAULT_OPTIONS
+
+	if ! /etc/init.d/openvpn prepare "$cfg" ; then
+		echo "openvpn[$$]" "Failed to setup configuration"
+		proto_notify_error "$cfg" "SERVICE_ENABLED"
+		proto_block_restart "$cfg"
+		return 1
+	fi
+
+	if [ -f "/var/etc/openvpn-${cfg}.conf" ]; then
+		config="/var/etc/openvpn-${cfg}.conf"
+	else
+		config="/etc/openvpn/${cfg}.conf"
+	fi
+
+	[ -n "$ifname" ] || get_config_param ifname "$config" dev
+	[ -z "$ifname" ] && {
+		echo "openvpn[$$]" "Failed to get ifname"
+		proto_notify_error "$cfg" "NO_IFNAME_SET"
+		proto_block_restart "$cfg"
+		return 1
+	}
+	append opts "--dev $ifname"
+
+	get_config_param dev_type "$config" dev-type
+	case "${dev_type:-$ifname}" in
+		tap*) dev_type="tap" ;;
+		*) dev_type="tun" ;;
+	esac
+	append opts "--dev-type $dev_type"
+
+	[ -n "$metric" ] || get_config_param metric "$config" metric 0
+	append opts "--setenv METRIC ${metric}"
+
+	[ "$peerdns" = 1 ] && append opts "--setenv PEERDNS 1"
+
+	[ "$defaultroute" = 1 ] && append opts "--setenv DEFAULTROUTE 1"
+
+	proto_run_command "$cfg" /usr/sbin/openvpn \
+		--syslog "openvpn($cfg)" \
+		--status "/var/run/openvpn.${cfg}.status" \
+		--cd /etc/openvpn \
+		--config "$config" \
+		--setenv INTERFACE "$cfg" \
+		--resolv-retry infinite \
+		--route-noexec \
+		--script-security 3 \
+		--route-up /lib/netifd/openvpn-up \
+		--route-pre-down /lib/netifd/openvpn-down \
+		$opts
+}
+
+proto_openvpn_teardown() {
+	local cfg="$1"
+	proto_kill_command "$cfg"
+}
+
+[ -n "$INCLUDE_ONLY" ] || {
+	add_protocol openvpn
+}
--- a/package/network/services/openvpn/files/openvpn.init
+++ b/package/network/services/openvpn/files/openvpn.init
@@ -7,6 +7,11 @@
 START=90
 STOP=10
 
+EXTRA_COMMANDS="prepare"
+EXTRA_HELP="\
+        prepare Build config from uci\
+"
+
 USE_PROCD=1
 PROG=/usr/sbin/openvpn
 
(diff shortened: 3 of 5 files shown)


package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
```diff
@@ -0,0 +1,167 @@
...
+	proto_run_command "$cfg" /usr/sbin/openvpn \
+		--syslog "openvpn($cfg)" \
+		--status "/var/run/openvpn.${cfg}.status" \
+		--cd /etc/openvpn \
+		--config "/var/etc/openvpn-${cfg}.conf" \
+		--setenv INTERFACE $cfg \
+		--route-noexec \
+		--script-security 3 \
+		--route-up /lib/netifd/openvpn-up \
+		--route-pre-down /lib/netifd/openvpn-down
```
jow-: To keep the style similar to other protocols I'd also like if we'd support an optional `option ifname` and map it to OpenVPN's `--dev` argument.

Along with the proposed config handling above, a working approach could look like:

	# find effective config
	local config
	json_get_vars config
	
	if [ -n "$config" -a ! -f "$config" ]; then
		echo "openvpn[$$]" "User specified configuration '$config' does not exist"
		proto_setup_failed "$cfg"
		return 1
	fi
	
	if [ ! -n "$config" -a -f "/etc/openvpn/$cfg.conf" ]; then
		config="/etc/openvpn/$cfg.conf"
	fi
	
	get_config_param() {
		local variable="$1"
		local option="$2"
		local default="$3"
	
		# if a config file exists, try to find param in it
		if [ -n "$config" ]; then
			local value="$(sed -rne 's/^[ \t]*'"$option"'[ \t]+([^ \t]+)[ \t]*$/\1/p' "$config" | tail -n1)"
			export -n "$variable=${value:-$default}"
	
		# ... else try to look it up in uci
		else
			config_get value "$cfg" "${option//-/_}" "$default"
		fi
	}
	
	local ifname dev_type dev_name
	
	# get ifname from protocol config
	json_get_vars ifname
	
	# fall back to dev from config if no ifname specified
	[ -n "$ifname" ] || get_config_param ifname "$config" dev
	
	# check if device name has been set in passed or generated config
	get_config_param dev_type "$config" dev-type
	
	# guess device type, fall back to "tun"
	case "${dev_type:-${ifname:-$dev_name}}" in
		tap*) dev_type="tap" ;;
		*) dev_type="tun" ;;
	esac
	
	proto_run_command "$cfg" /usr/sbin/openvpn \
		--syslog "openvpn($cfg)" \
		--status "/var/run/openvpn.${cfg}.status" \
		--cd /etc/openvpn \
		--config "${config:-/var/etc/openvpn-${cfg}.conf}" \
		${dev_type:+--dev-type "$dev_type"} \
		${ifname:+--dev "$ifname"} \
		--setenv INTERFACE $cfg \
		--route-noexec \
		--script-security 3 \
		--route-up /lib/netifd/openvpn-up \
		--route-pre-down /lib/netifd/openvpn-down

jow-:
Another idea; instead of sharing the UCI config translation with the init script; what about simply adding a new init script action "prepare" which does all the preparation but does not actually start openvpn?

So from the protocol handler you could invoke something like

    if ! /etc/init.d/openvpn prepare "$cfg"; then
      echo "openvpn[$$]" "Failed to setup configuration"
      proto_setup_failed "$cfg"
      return 1
    fi

... and from now on you can rely on either of the following facts:

 - "/var/etc/openvpn-${cfg}.conf" exists, means the init script found a section called "$cfg" in `/etc/config/openvpn`
 - or no such config exists, yet the init script exited with code 0, means there is an `/etc/openvpn/$cfg.conf` which is not disabled.

package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
```diff
@@ -0,0 +1,167 @@
...
+
+	config_get remote "$cfg" 'remote'
+	remote=${remote%% *}
+	[ -n "$remote" ] && {
+		for ip in $(resolveip -t 5 "$remote"); do
+			#proto_add_host_dependency "$cfg" "$ip"
+			serv_addr=1
+		done
+		[ -n "$serv_addr" ] || {
+			echo "openvpn[$$]" "Could not resolve server address \"$remote\""
```
jow-: Actually something like `--resolv-retry 10` is likely okay as well and allows to propagate the failure to netifd eventually

jow-:
Here's some more options found in other netifd tunnel protocols which are worth mapping to openvpn arguments:

 * `server addr[:port]` => `--remote addr [port]`
 * `mtu N` => `--tun-mtu N`
 * `defaultroute 0` => `--route-nopull`
 * `peerdns 0` => disable DNS handling in `route-up` script
 * `username` / `password` => write to a temporary file and pass with `--auth-user-pass`
 * `ipv6 1` => `--tun-ipv6`
 * `keepalive N` => `--keepalive $N $((N * 6))`

package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
```diff
@@ -0,0 +1,167 @@
...
+	[ "${enabled}" -eq 0 ] || {
+		echo "openvpn[$$]" "Section \"$s\" is enabled in /etc/config/openvpn. It must be disabled to be started with netifd"
+		return 1
+	}
+
+	config_get remote "$cfg" 'remote'
+	remote=${remote%% *}
+	[ -n "$remote" ] && {
+		for ip in $(resolveip -t 5 "$remote"); do
+			#proto_add_host_dependency "$cfg" "$ip"
```
feckert: This was already added by the origin author. So i am nor sure how to deal with your hint. I will have a close look.

package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
feckert: Sounds good to me. Will add it to the start parameters of the proto handler.

feckert:
I Think we should only use `not enabled config section` to get started by the new openvpn proto handler. If a user upgrades the router to new lede/openwrt version the old style to start a openvpn connection by `etc/init.d/openvpn` should work as it is. If you add a new openvpn connection it should use the new proto handler style.

@jow-  thanks for comments my RFC. I will update my PR with your hints. I am not sure which config generation style is the best. But i think i will start with `/etc/init.d/openvpn prepare` first to implement

feckert:
@jow- I am not sure if it is a good user experience to share the configuration between `/etc/config/openvpn` and `/etc/config/network`. The user will not know which config will be used if it is configured twice.

The following parameters should be configured in the interface section because this are interface specific parameters which are set in the route-up scripts. 
* defaultroute
* peerdns
* metric

The following parameters should also be added as start parameters to the `proto_run_command` which are searched during `proto_openvpn_setup` function

* dev_type
* ifname

All other Configuration should be configured in the `/etc/config/openvpn` file. And should as you suggest prepared with the command `/etc/init.d/openvpn prepare $CFG`

If I want an openvpn server then I have to make my configuration in the `/etc/config/openvpn` section regardless of the proto openvpn? Is this right?

jow-:
Ideally I'd like to fade out /etc/config/openvpn entirely, most users simply want to drop their .conf file into the router and expect the corresponding openvpn instance to start, like on other Linux desktop distributions as well.

That aside, I can live with just handling the five options mentioned by you for now.

feckert:
@jow- Sorry for the question but i dont know it!

Do i need a special treatment? Or will this behavior work if the openvpn instance is configured as a openvpn server? I think the route-up/down scripts are the problems.

jow-:
Lets focus on the client part for now. Not sure about the specific `route-up` semantics in server mode.

package/network/services/openvpn/files/lib/netifd/openvpn-up
```diff
@@ -0,0 +1,88 @@
...
+	[ -n "$metric" ] || metric="$METRIC"
+
+	proto_add_ipv4_route "$network" "$mask" "$gw" "" "$metric"
+	c=$(( $c+1 ))
+done
+
+# set default route
+[ -n "$DEFAULTROUTE" ] && {
+	eval "metric=\$$METRIC"
+	proto_add_ipv4_route "0.0.0.0" "0" "$gw" "0.0.0.0/0" "$metric"
```
feckert: @jow i am not sure if this works for all situations the $gw variable is in my setup always the same.

feckert:
@jow- ping

package/network/services/openvpn/files/lib/netifd/openvpn-up
jow-: Why the `eval` indirection? Can't we just pass `$METRIC` as-is?

feckert:
@jow- yes sure i will update the PR

2017-06-23 10:31 GMT+02:00 Jo-Philipp Wich <notifications@github.com>:

feckert:
@jow- sorry for the additonal ping.
Any progress on this PR or any additional hints we could also discuss on the irc?

sbrudenell:
+1 for this.

I see some things are expecting netifd integration now (in particular the mwan3 package, which I use a lot). The lack of netifd integration in openvpn is starting to become painful. And a really long time coming, here's a post from 2012 https://forum.openwrt.org/viewtopic.php?id=39274

@feckert I'll use your branch for now and report any problems or changes I make.

package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
```diff
@@ -0,0 +1,95 @@
...
+		*) dev_type="tun" ;;
+	esac
+	append opts "--dev-type $dev_type"
+
+	[ -n "$metric" ] || get_config_param metric "$config" metric 0
+	append opts "--setenv METRIC ${metric}"
+
+	[ "$peerdns" = 1 ] && append opts "--setenv PEERDNS 1"
+
+	[ "$defaultroute" = 1 ] && append opts "--setenv DEFAULTROUTE 1"
```
sbrudenell: I vote that `defaultroute` should default to 1. It seems like most other config protocols default to applying a default route.

package/network/services/openvpn/files/openvpn.init
```diff
@@ -130,3 +135,56 @@ start_service() {
...
+
+	uci -q get openvpn.$s > /dev/null || {
+		$log error "Config $s not found"
+		exit 2
+	}
+
+	. /usr/share/openvpn/openvpn.options
+
+	config_load 'openvpn'
+	config_get enabled "$s" 'enabled' 0
```
sbrudenell: It looks like the current `/etc/init.d/openvpn` uses either `enable` or `enabled`. I still had `enable` in my config files, probably from the older OpenWRT days.

feckert:
@sbrudenell thanks for reply.
Still waiting to get this merged.
I will change your hints next week.

zorun:
@feckert I tried this and it mostly works, thanks (although I struggled a bit to understand how to configure it, it will need a documentation)

However, handling of the default route does not work properly: there is no more specific route to reach the VPN server's public IP address, so it tries to use the default route through the tunnel (obviously, this does not work).  The routing table looks like this:

```
default dev tun0 
192.168.0.0/24 dev eth0  src 192.168.0.15 
192.168.0.254 dev eth0  src 192.168.0.15 
192.168.1.0/24 dev br-lan  src 192.168.1.1 
10.0.0.0/24 dev tun0  src 10.0.0.123 
```
where `192.168.0.X` is WAN, `192.168.1.X` is LAN, and `10.0.0.X` is VPN.

Additionally when shutting down the VPN interface there is no default route anymore:

```
192.168.0.0/24 dev eth0  src 192.168.0.15 
192.168.0.254 dev eth0  src 192.168.0.15 
192.168.1.0/24 dev br-lan  src 192.168.1.1 
```

It requires a restart of the WAN interface to get the default route back.

zorun:
When using the regular method via `/etc/config/openvpn` directly and `/etc/init.d/openvpn start`, it works fine:
```
default via 10.0.0.65 dev tun0 
XX.XX.XX.28 via 192.168.0.254 dev eth0 
```
where `XX.XX.XX.28` is the public address of the VPN server.  Apparently these routes are handled by openvpn itself:

```
Mon Aug 28 11:10:24 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route add -net XX.XX.XX.28 netmask 255.255.255.255 gw 192.168.0.254
Mon Aug 28 11:10:24 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route del -net 0.0.0.0 netmask 0.0.0.0
Mon Aug 28 11:10:24 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route add -net 0.0.0.0 netmask 0.0.0.0 gw 10.0.0.65
```

Also when shutting down openvpn it puts the normal routing table back into place:
```
Mon Aug 28 11:16:00 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route del -net XX.XX.XX.28 netmask 255.255.255.255
Mon Aug 28 11:16:00 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route del -net 0.0.0.0 netmask 0.0.0.0
Mon Aug 28 11:16:00 2017 daemon.notice openvpn(ovpn)[5694]: /sbin/route add -net 0.0.0.0 netmask 0.0.0.0 gw 192.168.0.254
Mon Aug 28 11:16:00 2017 daemon.notice openvpn(ovpn)[5694]: Closing TUN/TAP interface
```

zorun:
I tried with `proto_add_host_dependency`, similarly to what is done with `xl2tpd` (it needs `resolveip` installed):

```
--- a/package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
+++ b/package/network/services/openvpn/files/lib/netifd/proto/openvpn.sh
@@ -69,7 +69,13 @@ proto_openvpn_setup() {

        [ "$peerdns" = 1 ] && append opts "--setenv PEERDNS 1"

-       [ "$defaultroute" = 1 ] && append opts "--setenv DEFAULTROUTE 1"
+       if [ "$defaultroute" = 1 ]; then
+            append opts "--setenv DEFAULTROUTE 1"
+            for ip in $(resolveip -t 5 my_vpn_server.domain.tld); do
+                logger -t "openvpn-proto" "$cfg: adding dependency for $ip"
+                ( proto_add_host_dependency "$cfg" "$ip" )
+            done
+        fi

        proto_run_command "$cfg" /usr/sbin/openvpn \
                --syslog "openvpn($cfg)" \
```

The server address is hardcoded for testing, it will need to be extracted from `/etc/config/openvpn` (which would probably be non-trivial). It works when bringing the VPN interface up:

```
default dev tun0 
192.168.0.0/24 dev eth0  src 192.168.0.15 
192.168.0.254 dev eth0  src 192.168.0.15 
192.168.1.0/24 dev br-lan  src 192.168.1.1 
10.0.0.0/24 dev tun0  src 10.0.0.123 
XX.XX.XX.28 via 192.168.0.254 dev eth0 
```

Notice the last route.  However, there is still the issue of the missing default route when bringing the interface down:

```
192.168.0.0/24 dev eth0  src 192.168.0.15 
192.168.0.254 dev eth0  src 192.168.0.15 
192.168.1.0/24 dev br-lan  src 192.168.1.1 
XX.XX.XX.28 via 192.168.0.254 dev eth0 
```

It is similar to https://github.com/openwrt/packages/issues/4597 for wireguard.  @jow- any idea?  This is on lede 17.01 btw.

feckert:
@zorun thanks for feedback

zorun:
@feckert I'm actually thinking it would be simpler to let OpenVPN handle the routes itself, like it is done now.

Another issue arises when there is IPv6 connectivity through the tunnel: even with an IPv6 default route through the tunnel, LEDE does not distribute any global IPv6 addresses on the LAN side. But in this case, it's also broken with the current OpenVPN init script :)

blogic:

PR #1104, opened 2017-05-12
ar71xx: add 8devices Rambutan developement board support

mantas-p:
Rambutan is a Wifi module based on QCA9550/9557
http://www.8devices.com/products/rambutan

This commit adds basic support for Rambutan
developement kit

Specs:
- 720/600/200 MHz (CPU/DDR/AHB)
- 128 MB of DDR2 RAM
- 128 MB of NAND Flash
- 1x 100Mbps Ethernet
- 1x 1000Mbps Ethernet (PHY on dev-kit)
- 1x Wifi radio 2x2 MIMO, dualband 2.4 and 5 GHz
- 2x U.FL connectors on module, chip antennas on dev-kit
- 1x miniPCIe slot
- 1x USB2.0 host socket + 1x USB2.0 pins on 2.54mm header

Flash instructions:
Stock firmware is OpenWRT, so use:
 sysupgrade -n /tmp/lede-ar71xx-nand-rambutan-squashfs-sysupgrade.tar
or upgarde from GUI (don't save config)
Use factory image to flash from U-Boot:
 tftpboot 80060000 lede-ar71xx-nand-rambutan-squashfs-factory.ubi
 nand erase.part ubi
 nand write 80060000 ubi ${filesize}

Signed-off-by: Mantas Pucka <mantas@8devices.com>


pepe2k:
@mantas-p, thank you for your contribution!

I have merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) with minor fixes:

- split support in uboot-envtools package into a separate commit
- fixed alphabetical order in lib/preinit/05_set_iface_mac_ar71xx
- fixed a typo in commit subject/message ("developement" -> "development").

I have also one question - are you sure about this line in `target/linux/ar71xx/image/nand.mk`:
`UBIFS_OPTS = -m 2048 -e 124KiB -c 950 -U -F`

As far as I can see, `append-ubi` uses `UBINIZE_OPTS` (see [image-commands.mk](https://github.com/lede-project/source/blob/master/include/image-commands.mk#L168)). Please, confirm/test and let me know if something needs to be changed.

Cheers,
Piotr

mantas-p:
@pepe2k, thanks for review. I've checked the  ```UBIFS_OPTS = ...``` line, it isn't actually used and should be removed.
Should I update PR or you'll just fix it your staging tree?

pepe2k:
@mantas-p,

> I've checked the UBIFS_OPTS = ... line, it isn't actually used and should be removed.

OK, just for sure, Rambutan doesn't need end-of-UBI marker (`UBINIZE_OPTS := -E 5`) in factory image?

> Should I update PR or you'll just fix it your staging tree?

I will update my staging tree.

Cheers,
Piotr

mkresin:
>> I've checked the UBIFS_OPTS = ... line, it isn't actually used and should be removed.

> OK, just for sure, Rambutan doesn't need end-of-UBI marker (UBINIZE_OPTS := -E 5) in factory image?

Which would be kind of surprising. As far as I know, the EOF marker is required to extend the ubi to use all available flash space of that partition.

pepe2k:
@mkresin

> Which would be kind of surprising. As far as I know, the EOF marker is required to extend the ubi to use all available flash space of that partition.

AFAIK, boards with UBI support in bootloader not always need this.

mantas-p:
Rambutan's U-boot always mounts UBI partition (to read kernel image), UBI gets expanded during this step if it was flashed from factory image. EOF marker only confuses U-Boot.

pepe2k:
> Rambutan's U-boot always mounts UBI partition (to read kernel image), UBI gets expanded during this step if it was flashed from factory image. EOF marker only confuses U-Boot.

Perfect, thanks for confirmation.
I have updated my staging tree.

Cheers,
Piotr

bobafetthotmail:
@mkresin also for kirkwoods it is like this (most use a custom uboot with UBI support as kernel is in UBI), only ones that need EOF marker are the two routers that must use stock uboot (with kernel on raw flash).

PR #1109, opened 2017-05-13
libs/libnl: Update to 3.3.0

diizzyy:
Update libnl to 3.3.0
Import patches to fix compilation
Source: https://git.busybox.net/buildroot/tree/package/libnl
Source: https://gitweb.gentoo.org/proj/musl.git/diff/dev-libs/libnl/files/libnl-3.3.0_rc1-musl.patch?id=48d2a287
Use more automatic toolchain logic


diizzyy:
Buildlog shows a load of these messages which seems harmless?
```OpenWrt-libtool: install: warning: `lib/libnl-route-3.la' has not been installed in `/usr/lib'```

diizzyy:
Seems to work fine on my Linksys WRT3200ACM using trunk as far as I can tell

diizzyy:
fwiw, 3.3.29 shows exactly the same errors on the buildbots so it more than likely safe to ignore.

stintel:
e4b6900fd60e43ba0527172c30290a69958b7f73

PR #1114, opened 2017-05-16
lldpd: drop specific respawn params [use system-wide]

commodo:
I think I added these respawn params [a while back],
when I did the conversion to procd init script format.


blogic:
merged, Thanks !

PR #1115, opened 2017-05-17
procd, base-files: query procd for checking the running state of a service

commodo:
For services that are handled by procd, a ubus call [to procd] can be used to query the running state of that service.

However, certain init.d scripts that set the USE_PROCD var [system, gpio_switch, firewall, qos], don't provide a comand param, so query-ing procd will return that the service is not running.
For those, I've added a dummy service_running() hook that will `return 0`.

For lldpd, I've removed the `service_running()` hook [I added that a while back], since querying procd yields the same result.

### DIFF
--- a/package/system/procd/files/procd.sh
+++ b/package/system/procd/files/procd.sh
@@ -413,6 +413,31 @@ _procd_set_config_changed() {
 	ubus call service event "$(json_dump)"
 }
 
+procd_service_running() {
+	local service="$1"
+	local running
+
+	[ -n "$service" ] || return 1
+
+	json_init
+	json_add_string name "$service"
+
+	# get information from procd via ubus call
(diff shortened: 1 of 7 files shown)


package/system/procd/files/procd.sh
```diff
@@ -413,6 +413,31 @@ _procd_set_config_changed() {
...
+	json_select "$service" && \
+		json_select "instances" && \
+		json_select "instance1" && \
+		json_get_var running running
+
+	json_cleanup
+
+	[ "$running" == "1" ]
+}
+
```
blogic: i am still not sure if i like this, might be easier to simply add a ubus call to procd allowing us to query if a service is running.

commodo: ok ; 
that's also fine from my side ;

will re-route to that approach if that's ok with you

PR #1117, opened 2017-05-17
dnsmasq: remove config file /etc/dnsmasq.conf

Jalakas:
Configfile content has been fully commented out for at least 5 years.
File is not preserved while sysupgrade. Removing it saves 1.34 KB space.

Signed-off-by: Anari Jalakas <anari.jalakas@gmail.com>


lucize:
I'm using it for UEFI tftpboot setup, and in some cases for 150+ mac address allocation that are a burden to setup and spot using the web interface
maybe it can be emptied and used for config file inclusion

Jalakas:
This commit only removes default configuration file (that most users never edit).
Dnsmasq is still parsing the /etc/dnsmasq.conf file, if it's added by user who knows what he needs.  

Wiki has very little about using this configuration:
https://lede-project.org/docs/user-guide/dhcp_configuration#using_plain_dnsmasqconf
"The dnsmasq.conf **file does not exist by default** but will be processed by dnsmasq on startup if it is present."

dedeckeh:
@Jalakas Had a discussion with @nbd168 about the removal of /etc/dnsmasq.conf; at the moment we would like to keep the file as it allows the user to specify extra dnsmasq command line arguments.
I agree the file is still processed by the dnsmasq init script but removing it will make it harder for the user to use this feature while it would also not be preserved during sysupgrades .

PR #1118, opened 2017-05-19
sunxi: add Olimex A20-OLinuXino-LIME2

lucize:
add Olimex A20-OLinuXino-LIME2

### DIFF
--- a/package/boot/uboot-sunxi/Makefile
+++ b/package/boot/uboot-sunxi/Makefile
@@ -43,6 +43,12 @@ define U-Boot/A20-OLinuXino-Lime
   BUILD_DEVICES:=sun7i-a20-olinuxino-lime
 endef
 
+define U-Boot/A20-OLinuXino-Lime2
+  BUILD_SUBTARGET:=cortexa7
+  NAME:=A20 OLinuXino LIME2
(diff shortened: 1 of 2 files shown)


hauke:
Why are you removing the regulators from the device tree file?

lucize:
It's backported from 4.6 as I saw, but is not completely, it will not start with the regulators, at least, the hdmi display.

hauke:
@lucize Is it this commit you want to backport here?
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=b43776d65a33b460925604a27fb9360af08c55ce
Then you should add it also with the lines which are getting added and with the commit comment.

lucize:
tried it before posting the PR, I think the axp209 driver is not complete, or something else is missing

adding the regulators on 4.4 will not boot at all, tried with plain 4.9/4.12(rc) and there was no hdmi output but it started, maybe updating the pachset for 4.9 will make it work, I'm not an expert in this area so I didn't tried harder because everything seems to work ok without those values

hauke:
Could you please rebase your changes on top of current master branch.

lucize:
done!

hauke:
Thank you for the patch it was merged into master

PR #1119, opened 2017-05-19
firmware-utils: support strings as soft_version

jannic:
Some TP-Link routers (C25, C59, C60) contain a version string instead of a binary structure in the soft_version partition.

Flashing LEDE from the original firmware's GUI, this version string taken from the soft_ver partition of the firmware image is written to the router's config partition.

When using tftp recovery to go back to the original Archer C25 firmware, a version check compares that version to the version of the firmware to be flashed.

Without proper contents in the config partition, reverting to the original firmware fails.

Therefore, write the string "soft_ver:1.0.0\n" to that soft_ver partition.

Signed-off-by: Jan Niehusmann <jan@gondor.com>

@heinzek, you suggested that change in https://github.com/lede-project/source/pull/1069#issuecomment-301328215. As I don't have a C59 or C60 router I can't test the patch on those devices. Could you confirm that it works?

### DIFF
--- a/tools/firmware-utils/src/tplink-safeloader.c
+++ b/tools/firmware-utils/src/tplink-safeloader.c
@@ -75,6 +75,7 @@ struct device_info {
 	const char *vendor;
 	const char *support_list;
 	char support_trail;
+	const char *soft_ver;
 	const struct flash_partition_entry partitions[MAX_PARTITIONS+1];
 	const char *first_sysupgrade_partition;
 	const char *last_sysupgrade_partition;
@@ -130,6 +131,7 @@ static struct device_info boards[] = {
 			"CPE220(TP-LINK|US|N300-2):1.1\r\n"
 			"CPE220(TP-LINK|EU|N300-2):1.1\r\n",
 		.support_trail = '\xff',
+		.soft_ver = NULL,
 
 		.partitions = {
 			{"fs-uboot", 0x00000, 0x20000},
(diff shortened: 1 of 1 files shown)


tools/firmware-utils/src/tplink-safeloader.c
```diff
@@ -683,6 +697,19 @@ static struct image_partition_entry make_soft_version(uint32_t rev) {
 	return entry;
 }
 
+static struct image_partition_entry make_soft_version_from_string(const char *soft_ver) {
+	uint32_t ver_len = strlen(soft_ver);
+	struct image_partition_entry entry = alloc_image_partition("soft-version", ver_len + 2*sizeof(uint32_t));
+
+	uint32_t *len = (uint32_t *)entry.data;
+	*len = htonl(ver_len);
+	*(len+1) = 0;
```
NeoRaider: How about writing `len[0]` and `len[1]` here?

tools/firmware-utils/src/tplink-safeloader.c
```diff
@@ -683,6 +697,19 @@ static struct image_partition_entry make_soft_version(uint32_t rev) {
...
+	memcpy(len+2, soft_ver, ver_len);
```
NeoRaider: I think it would be preferable to make the format of the soft-version partition the same as used by the stock firmware, which terminates the string a bit differently: the partition ends with two zero bytes after the newline, the first of which is included in the string length in the partition head; the second is not included in the string length, but in the partition size.

jannic:
Applied the changes suggested by @NeoRaider.
Please note that I still don't have a C59/C60 router to test the changes, so the patch is only compile-tested.

NeoRaider:
LGTM.

Could someone else take this through their staging? I probably won't be available for the rest of the week.

pepe2k:
@jannic, thank you for your contribution! I have merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) with just a minor code style fixes.

@NeoRaider, I haven't found it in your staging tree, if you have it merged locally, let me know.

Cheers,
Piotr

PR #1122, opened 2017-05-20
package/grub2: update to 2.02

alive4ever:
Update to version 2.02

Full changelog can be viewed [here](https://git.savannah.gnu.org/cgit/grub.git/log/?h=grub-2.02)


nbd168:

PR #1124, opened 2017-05-24
kirkwood: add support for Cisco ON100 

a-a:
The Cisco ON100 device is a Kirkwood based router:

- SoC: Marvell 88F6282 1600Mhz
- SDRAM memory: 512MB DDR3 1333Mhz
- Gigabit ethernet: 2x Marvell 88E1310 (over RGMII)
- Flash memory: 512MB
- 2 bi-colour status LEDs (green/red)
- 1 Reset button
- 1 USB 2.0 port (on back)

This commit adds a target profile of "Cisco Systems ON100" under the target
system "Marvell Kirkwood".

The stock u-boot can read ubi and supports booting via tftp, once "dhcp"
command issued and device has a valid lease with tftp options set, flashing
is performed as follows:

tftpboot ${loadaddr} lede-kirkwood-ciscoon100-uImage ; nand erase 0x200000
 0x600000 ; nand write ${loadaddr} 0x200000 0x600000
nand erase 0x700000 0x1f900000
tftpboot ${loadaddr} lede-kirkwood-ciscoon100-squashfs-factory.bin; nand
 erase 0x700000 ${filesize} ; nand write ${loadaddr} 0x700000 ${filesize}

LEDE is now flashed, but some configuration of environment variables is
required:

setenv lede_bootargs_root ubi.mtd=2 root=/dev/mtdblock2
setenv lede_mtdparts mtdparts=nand_mtd:0x200000@0x0(u-boot),0x400000@0x200
000(uImage),-@0x700000(data)
setenv lede_boot nand read.e \${loadaddr} 0x200000 0x200000 \; setenv
 bootargs \$(console) \$(lede_mtdparts) \${lede_bootargs_root} \; bootm
 \${loadaddr}
setenv bootcmd run lede_boot saveenv

If u-boot has been replaced (there may not seem to be a need to do this,
the stock u-boot seemed to work fine on my target), you may wish to confirm
mac addresses are still present with printenv. If they are not present or
incorrect, write your MAC addresses back with the following (use your MACs
as per the label on your device, instead of the example)

setenv ethaddr 64:00:F1:00:00:00
setenv eth1addr 64:00:F1:00:00:01
saveenv

Signed-off-by: Makoto Takeuchi <mak0@lxsys.co.uk>

### DIFF
--- a/package/boot/uboot-envtools/files/kirkwood
+++ b/package/boot/uboot-envtools/files/kirkwood
@@ -20,6 +20,9 @@ ib62x0|\
 linksys-viper|\
 nsa310b|\
 nsa325|\
+on100)
+	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000"
+	;;
 pogo_e02|\
 sheevaplug|\
 sheevaplug-esata)
--- a/target/linux/kirkwood/base-files/etc/board.d/02_network
+++ b/target/linux/kirkwood/base-files/etc/board.d/02_network
@@ -35,6 +35,9 @@ case "$board" in
 	ucidef_set_interface_lan "eth0" "dhcp"
 	ucidef_set_interface_macaddr "lan" $( mtd_get_mac_ascii uboot_env ethaddr )
 	;;
+"on100")
+	ucidef_set_interface_lan "eth0 eth1"
+	;;
 *)
 	ucidef_set_interface_lan "eth0"
 	;;
--- a/target/linux/kirkwood/base-files/etc/diag.sh
+++ b/target/linux/kirkwood/base-files/etc/diag.sh
@@ -23,6 +23,9 @@ get_status_led() {
 	nsa310b)
 		status_led="nsa310:green:sys"
 		;;
+	on100)
+		status_led="on100:green:health"
+		;;
 	esac
 }
 
--- a/target/linux/kirkwood/base-files/lib/kirkwood.sh
+++ b/target/linux/kirkwood/base-files/lib/kirkwood.sh
@@ -33,6 +33,10 @@ kirkwood_board_detect() {
 		name="ib62x0"
 		;;
 
+	"Cisco Systems ON100")
+		name="on100"
+		;;
+
 	"Cloud Engines Pogoplug E02")
 		name="pogo_e02"
 		;;
--- a/target/linux/kirkwood/image/Makefile
+++ b/target/linux/kirkwood/image/Makefile
@@ -11,7 +11,7 @@ include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/image.mk
 
 KERNEL_LOADADDR:=0x8000
-TARGET_DEVICES = linksys-audi linksys-viper dockstar goflexnet goflexhome iconnect pogo_e02 ib62x0 nsa310b nsa325
+TARGET_DEVICES = linksys-audi linksys-viper dockstar goflexnet goflexhome iconnect pogo_e02 ib62x0 nsa310b nsa325 on100
 
 UBI_OPTS := -m 2048 -p 128KiB -s 512
 UBIFS_OPTS := -m 2048 -e 126KiB -c 4096
@@ -92,6 +92,17 @@ $(Device/dockstar)
   DEVICE_TITLE := ZyXEL NSA325 (v1 and v2)
   DEVICE_DTS := kirkwood-nsa325
   DEVICE_PACKAGES := kmod-gpio-button-hotplug kmod-rtc-pcf8563 kmod-usb3
+endef
+
+define Device/on100
+  DEVICE_TITLE := Cisco Systems ON100
+  DEVICE_DTS := kirkwood-on100
+#  DEVICE_PACKAGES := kmod-mvsdio kmod-usb-storage kmod-i2c-core kmod-i2c-mv64xxx kmod-ata-core kmod-btmrvl kmod-btmrvl-sdio kmod-libertas kmod-libertas-sdio
+  DEVICE_PACKAGES := kmod-i2c-mv64xxx
+  KERNEL_SIZE := 5376k
+  IMAGES += factory.bin
+  IMAGE/factory.bin := append-kernel | pad-to $$$$(KERNEL_SIZE) | append-ubi
+  UBINIZE_OPTS := -E 5
 endef
 
 define Device/pogo_e02
--- a/target/linux/kirkwood/patches-4.4/195-on100.patch
+++ b/target/linux/kirkwood/patches-4.4/195-on100.patch
@@ -1,0 +1,173 @@
+--- /dev/null
++++ b/arch/arm/boot/dts/kirkwood-on100.dts
+@@ -0,0 +1,160 @@
++/dts-v1/;
++
++#include "kirkwood.dtsi"
++#include "kirkwood-6282.dtsi"
++
++/ {
++	model = "Cisco Systems ON100";
++	compatible = "cisco,on100", "marvell,kirkwood-88f6282", "marvell,kirkwood";
++
++	memory {
++		device_type = "memory";
++		reg = <0x00000000 0x20000000>;
++	};
++
++	aliases {
++		serial0 = &uart0;
++	};
++
++	chosen {
++		bootargs = "console=ttyS0,115200n8 earlyprintk";
++		stdout-path = "serial0:115200n8";
++	};
++
++ 	gpio_keys {
++		compatible = "gpio-keys";
++		#address-cells = <1>;
++		#size-cells = <0>;
++		pinctrl-0 = <&pmx_btn_reset>;
++		pinctrl-names = "default";
++
++		button@1 {
++			label = "Reset Button";
++			linux,code = <KEY_RESTART>;
++			gpios = <&gpio1 9 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	gpio-leds {
++		compatible = "gpio-leds";
++		pinctrl-0 = < &pmx_led_health_r &pmx_led_health_g >;
++		pinctrl-names = "default";
++
++		health-g {
++			label = "on100:green:health";
++			gpios = <&gpio1 12 GPIO_ACTIVE_LOW>;
++		};
++
++		health-r {
++			label = "on100:red:health";
++			gpios = <&gpio1 13 GPIO_ACTIVE_LOW>;
++		};
++
++		health2-g {
++			label = "on100:green:health2";
++			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
++		};
++
++		health2-r {
++			label = "on100:red:health2";
++			gpios = <&gpio1 15 GPIO_ACTIVE_LOW>;
++		};
++	};
++};
++
++&eth0 {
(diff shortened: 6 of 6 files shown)


bobafetthotmail:
Hi, is the SDcard slot working? 

If you say the uboot supports ubi, then why are you flashing separate uimage? Look at other kirkwoods to see how to set up proper nand partitions in dts for ubi.
If you set that, uboot will not need to send a command line to LEDE for booting.

Also please squash the commits into one, see here for how to do that https://lede-project.org/submitting-patches#squashing_commits

a-a:
SDcard slot is working now - i had forgotten to enable this in the DTS.

Regarding UBI, i am incorrect here, the stock bootloader does not support UBI. The image is now generated to append rootfs after kernel in a ubi, which the kernel can handle.

Updated tftp flashing instructions:
tftpboot ${loadaddr} lede-kirkwood-ciscoon100-squashfs-factory.bin; nand erase 0x200000 ${filesize}; nand write ${loadaddr} 0x200000 ${filesize}
setenv lede_boot nand read \${loadaddr} 0x200000 0x500000\; setenv bootargs ubi.mtd=2 root=/dev/mtdblock2 console=ttyS0,115200 max_loop=64\; bootm
setenv bootcmd run lede_boot
saveenv


I'll try on an SD card later, however this computer does not have a SD reader hence opting for tftp right now.

lynxis:
@a-a except the dockstar, all newer kirkwood romloader supports a serial boot aka **kwboot**. It would be nice if you could test if kwboot works.

a-a:
sure, just modifying flash layout a bit to correct that, will try testing kwboot too next on my list. thanks!

bobafetthotmail:
kwboot seems to work, some months ago a guy was trying to develop a modern uboot for it in this thread (his uboots would have ubi support, among other things) http://forum.doozan.com/read.php?2,29059,page=1
and two testers were kwbooting his test images.

it seems he made it, but that one of the testers had a device with a defective ethernet port so they lost lots of time troubleshooting that thinking it was a software issue. Looks like he did not release the sources for his uboot modifications.

if you confirm that the test image is working on your side too I will try getting the sources from him to make a ubi-aware uboot for this device too.
Can you try kwbooting his test images from this post and do the tests as he wrote (testing it both ethernet ports work from inside uboot) http://forum.doozan.com/read.php?2,29059,30236#msg-30236
this the log of the guy that was successful http://forum.doozan.com/read.php?2,29059,30257#msg-30257

a-a:
U-Boot appear to work on this, at least with image 2.18 from that link - some of the earlier ones had less success initialising the ethernet interfaces.
Note I had to set different IP on the different interfaces otherwise no pings, assume MAC table was getting upset in my router when this was not tried so it discard the packets :)

`U-Boot 2016.05-tld-2.18 (Sep 20 2016 - 13:37:24 -0700)
Cisco ON100_K9

SoC:   Kirkwood 88F6282_A1
DRAM:  512 MiB
WARNING: Caches not enabled
NAND:  512 MiB
MMC:   kwsdio: 0
*** Warning - bad CRC, using default environment

In:    serial
Out:   serial
Err:   serial
Net:   Marvell MV88E1310 PHY initialized
egiga0 Waiting for PHY auto negotiation to complete.......... done
Marvell MV88E1310 PHY initialized
egiga1 Waiting for PHY auto negotiation to complete..... done
egiga0, egiga1
Hit any key to stop autoboot:  0 
ON100_K9> setenv ethact egiga1
ON100_K9> setenv ipaddr 10.0.0.10
ON100_K9> setenv gateway 10.0.0.1
ON100_K9> setenv netmask 255.255.255.0
ON100_K9> mii device
MII devices: 'egiga0' 'egiga1' 
Current device: 'egiga0'
ON100_K9> mii info 0
PHY 0x00: OUI = 0x5043, Model = 0x29, Rev = 0x00, 1000baseT, FDX
ON100_K9> mii info 1
PHY 0x01: OUI = 0x5043, Model = 0x29, Rev = 0x00, 1000baseT, FDX
ON100_K9> ping 10.0.0.1
Using egiga1 device
host 10.0.0.1 is alive
ON100_K9> setenv ethact egiga0
ON100_K9> ping 10.0.0.1
Using egiga0 device

Abort
ping failed; host 10.0.0.1 is not alive
ON100_K9> setenv ipaddr 10.0.0.10
ON100_K9> setenv ipaddr 10.0.0.11
ON100_K9> setenv gateway 10.0.0.1
ON100_K9> setenv netmask 255.255.255.0 
ON100_K9> mii device
MII devices: 'egiga0' 'egiga1' 
Current device: 'egiga0'
ON100_K9> ping 10.0.0.1
Using egiga0 device
host 10.0.0.1 is alive
ON100_K9> 
`

a-a:
pushed out amendments to flash layout to include u-boot variables partition, and simplify boot parameters, also including console settings in dts so bootargs should no longer be required (but may need to be unset - see new bootcmd in commit message)

mkresin:
@a-a is the PR ready to merge?

@bobafetthotmail Anything left to do here?

Wouldn't it be possible to boot a dump of the stock u-boot with kwboot as well?

bobafetthotmail:
Thanks for testing that uboot. There are two things left:

1. I think there are quite a bit of useless packages added in target/linux/kirkwood/image/Makefile 

Kirkwoods have most drivers compiled in kernel (usb, sata, i2c and mmc for sure) so kmod-mmc kmod-usb2 i2c and ata are not needed.
See target/linux/kirkwood/config-4.4 for what is already integrated in kernel.
Remove the packages and test, if there is something still needed for this device (I doubt it as more or less anything should be already enabled) enable it in kernel instead.

Also, kmod-libertas kmod-libertas-sdio are for marvell wifi, kmod-btmrvl and kmod-btmrvl-sdio is for marvell bluetooth, I don't think there is a wifi radio or bluetooth here. Remove them.

2. you should set the right parameters for fw_setenv in uboot-envtools https://github.com/lede-project/source/blob/master/package/boot/uboot-envtools/files/kirkwood
(this file generates the /etc/fw_env.config and the vlaues are "Device to access      flash offset          env size" )
This is better to have as a different commit, still in the same PR.

> Wouldn't it be possible to boot a dump of the stock u-boot with kwboot as well?

afaik no, kwboot only allows you to send another uboot over serial. You can use stock uboot or a new uboot (or Debian or LEDE) to dump the flash though see here for doing it with uboot (if it supports usb or mmc commands) https://stackoverflow.com/questions/22271699/upload-firmware-from-flash-using-u-boot#22325426

Why you need a dump of stock u-boot?

mkresin:
> 2. you should set the right parameters for fw_setenv in uboot-envtools https://github.com/lede-project/source/blob/master/package/boot/uboot-envtools/files/kirkwood
(this file generates the /etc/fw_env.config and the vlaues are "Device to access --- flash offset --- env size" )
This is better to have as a different commit, still in the same PR.

Why? Access to the uboot env from userspace isn't required. Nothing - like checksum - need to be updated from userspace. To use the uboot-env utility properly, it would be necessary to remove the read-only flag from the env partition. And I prefer to keep this flag to protect the user from theirselves.

> afaik no, kwboot only allows you to send another uboot over serial. You can use stock uboot or a new uboot (or Debian or LEDE) to dump the flash though see here for doing it with uboot (if it supports usb or mmc commands) https://stackoverflow.com/questions/22271699/upload-firmware-from-flash-using-u-boot#22325426
>
>Why you need a dump of stock u-boot?

As far as I understand, kwboot is used to boot the board if for example the bootloader was erased from flash. Instead of doing the work to add support for the on100 to the upstream uboot, one can use a dump of the preinstalled bootloader to boot that board if something is/was messed up.

I'm referring here to the "-p" parameter of kwboot.

       -p     In combination with -b, patches the header in image prior to upload, to "UART boot" type.

              This  option  attempts on-the-fly conversion of some none-UART image types, such as images which were originally
              formatted to be stored in flash memory.

              Conversion is performed in memory. The contents of image will not be altered

Did I get something wrong here?

bobafetthotmail:
> Why? Access to the uboot env from userspace isn't required. Nothing - like checksum - need to be updated from userspace. To use the uboot-env utility properly, it would be necessary to remove the read-only flag from the env partition. And I prefer to keep this flag to protect the user from theirselves.

We already talked about this when I was sending patches for NSA325 and 310, don't you remember?
Kirkwood devices have uboot-envtools installed by default and can boot Debian or Arch too (this device boots Debian fine, as shown in that other thread I linked), and most have env partitions writable for this reason.

It is already like that for other kirkwoods (apart from the two routers where it does not make sense to do so even if it is technically possible), and I think we should keep that possibility open.

If you think it's wrong please fix all other kirkwoods too so we have at least a consistent situation.

@a-a this is your call, either leave the env partition writable and add that to uboot-envtools or don't. When I added some other kirkwoods the patch was merged by **nbd168** even if **mkresin** disagreed on this point.

> Did I get something wrong here?

No you are correct, kwboot can send a flash dump too. It works on my kirkwoods. 

I was mostly wondering why you were interested, I thought LEDE was not interested in keeping files for recovery.
Some rare devices have flash dumps in the wiki (usually a broken link), but that's it.
Btw, if you want to keep uboot dumps it makes sense to keep a stock env dump too (a text file is enough, also a flash dump can be used).

mkresin:
> We already talked about this when I was sending patches for NSA325 and 310, don't you remember?

Yes I do remember. And I neither get your point now nor I got it in the past. No idea how running LEDE and installing debian is related. To get LEDE working it is required to change the bootcmd via uboot prompt. Why not do it the same way for debian.

You are right, I didn't blocked your other PRs because of that. It wasn't important enough for me to start a lengthy discussion about it. And blocking your PR because of this detail would be kind of unfair.

And as usual, other devs => other opinions. All I'm saying is that **I** don't see the need and therefore do not insist on that change.

> I was mostly wondering why you were interested, I thought LEDE was not interested in keeping files for recovery.

You lost me. I'm not talking about keeping any flash dumps. I just tried to point out, that a dump of the stock u-boot should be enough to revive a bricked board and at least for this task it isn't necessary to add on100 support to the upstream uboot.

bobafetthotmail:
>All I'm saying is that I don't see the need and therefore do not insist on that change.

Sorry, I repeated only because it looked like you forgot our previous discussion, I'm not trying to convince you.

> I just tried to point out, that a dump of the stock u-boot should be enough to revive a bricked board and at least for this task it isn't necessary to add on100 support to the upstream uboot.

Ok, I missed because it was kind of obvious for me.

Yes, a modern uboot is not required for this PR, so after a-a fixed the mentioned issues this can be merged, as far as I'm concerned.

a-a:
removed unnecessary packages - i think these got put in for testing when i was struggling to get the mmc to work (i'd forgotten to enable it in the dts file, so nothing to do with modules at all!).
also added uboot-env as a seperate commit.
regarding sysupgrade, this did seem to work although i'm not exactly sure what i'm looking for when testing this. the image applied and didn't brick the box.

mkresin:
Sorry for the delay but it is now merged into my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

I've ommited the u-boot env commit. The uboot env partition is still write protected in the dts, which would make any changes from the userspace fail.

If there aren't any further comment, I'll push it during the weekend to master.

PR #1125, opened 2017-05-25
X86: include USB HID by default

lucize:
most of x86 modern hardware don't have a PS2 port


nbd168:

PR #1127, opened 2017-05-26
flex: update to 2.6.4

EricLuehrsen:
FS#809
flex 2.6.3 bug which fails code generation and compile of some packages
cc: @hnyman, @nbd168


nbd168:
Merged, thanks.

PR #1130, opened 2017-05-27
Lantiq: make possible to tweak DSL SRN from UCI

andreamerello:
This patch makes possible to tweak the downstream SNR margin on
Lantiq DSL devices.

The UCI parameter 'network.dsl.ds_snr_offset' is used to set the SNR
margin offset. It accepts values in range -50 to +50 in 0.1 dB units.

The SNR margin can thus be modified in range -5.0 to +5.0 dB in 0.1 dB
steps.

Currently this should only affect ADSL (not VDSL). It should be very
easy to make this work also on VDSL lines, but since I couldn't test
on VDSL lines this patch does not do that yet.

I have also a patch for LUCI about this, that I could submit.

Tested on FB3370 (Lantiq VR9) and Telecom Italia ADSL2+ line.

Signed-off-by: Andrea Merello <andrea.merello@gmail.com>

### DIFF
--- a/package/network/config/ltq-vdsl-app/files/dsl_control
+++ b/package/network/config/ltq-vdsl-app/files/dsl_control
@@ -96,6 +96,22 @@ tone_vdsl_b="0x1" # B43
 tone_adsl_bv="0x81" # B43 + B43c
 tone_vdsl_bv="0x5" # B43 + V43
 
+# create ADSL autoboot script. Used for SNR margin tweak
+autoboot_script() {
+    echo "[WaitForConfiguration]={
+locs 0 $1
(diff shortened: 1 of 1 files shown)


hauke:
I checked the documentation and this is supported till on the vrx200, vrx300 and probably also older chips for ADSL and VDSL.

andreamerello:
@hauke if someone can test it, or if you trust that it would work, I can easily update the PR to set SNR margin also for VDSL

hauke:
Thank you for the patch, I added the ADSL support. When I find some time I can test the VDSL version.

PR #1131, opened 2017-05-28
base-files: Allow bridge w/o physical ifname

adoal:
It's possible to configure a bridged interface without physical interface
on a running LEDE box. This is useful for devices with only one eth0 port
(and wlan0): eth0 is used as wan, and wlan0 is bridged into the empty
br-lan.

However, the `/bin/config_generate` will stop generating configuration
for an interface without physical name. This prevent users creating custom
image which will act like a out of box only-one-wired-wan to wireless-lan
gateway using imagebuilder.

It will be better, allowing generate configuration for a bridged interface
even if no wired physical ifname is attached to it. :)

A common use case is to write something like
`    ucidef_set_interfaces_lan_wan "" "eth0"`
inside `/etc/board.d/some_customization` .

Signed-off-by: Adoal Xu <adoalxu@gmail.com>

### DIFF
--- a/package/base-files/files/bin/config_generate
+++ b/package/base-files/files/bin/config_generate
@@ -65,12 +65,12 @@ generate_network() {
 		json_select ..
 	json_select ..
 
-	[ -n "$ifname" ] || return
-
 	# force bridge for multi-interface devices (and lan)
 	case "$1:$ifname" in
 		*\ * | lan:*) type="bridge" ;;
 	esac
+
+	[ -n "$ifname" -o "$type" = "bridge" ] || return
 
 	uci -q batch <<-EOF
 		delete network.$1


blogic:
why do you need a bridge if there is only one single netdev ? this looks like you are working around a bad uci configuration.

adoal:
I mean that use case:
--WAN--> [ (eth0)   router  (wlan0)] <--LAN----   [wifi Client]

In OpenWRT/LEDE, devices with only one wired port is alway configured as eth0=lan by `board_detect`, then I could:
1. login to that router through eth0(lan)
2. enable WiFi
3. login through WiFi
4. remove eth0 from br-lan
5. create wan and assign eth0 as its physical interface
till now everything works. Yes, wlan0 is bridged to logical interface lan which is a bridge called `br-lan`, but the bridge it has no physical `ethx` interface inside it. It's a legal configuration according to OpenWRT's wiki.

Since it works, I want to integrate these steps into my customized image using imagebuilder in order to not manually run the 5 steps after flashing my device.

I can enable wifi through `/etc/uci-defaults/zz-enable-wifi`. But since eth0 (which I want to run as WAN) is assign to LAN by the built-in script, I must find some way to change it. I tried `/etc/board.d/assign-eth0-as-wan` by run `ucidef_set_interfaces_lan_wan '' 'eth0'`. It should work but not. After examining `/bin/config_generate` I found it doesn't generate configuration for empty ifname. This is reasonable for real interfaces, but a bridge port can be configured initially without real interface.

The case is a bit odd, but that one-line patch is useful. Otherwise I cannot take advantage of uci's built-in solution to archive my target and must manually configure everything of LAN my self. IMHO this patch doesn't break current code, and take convenience for the special use case.

adoal:
OpenWRT/LEDE's first-boot-generated WiFi configuration always bridge wlan (maybe `wlan0` or something else) to the logical `lan` interface which is a bridge called `br-lan`, which has some physical member interfaces such as `ethx`. If I use the only `eth0` as `wan`, then where is `wlan0` bridged to? Of course still `br-lan`, but it has no wired members.

Yes this odd configuration works. I can login to the default-configured one-rj45-one-wifi LEDE box through RJ45, then change config into `one-rj45-for-wan` and `wifi-bridged-to-empty-br-lan` [CFG1]. My next step is to eject this configuration into my customized image by ImageBuilder. It's easy to enable WiFi with `/etc/uci-defaults/my-enable-wifi-script`. But assigning `eth0` as physical interface of `wan` instead of `lan` is a bit difficult. I use `ucidef_set_interfaces_lan_wan '' 'eth0'`. It should work but not. After examining `/bin/config_generate` I found the cause. 

Although this use case is odd, this one-line patch is useful. Since I can achieve [CFG1] by hand after default first-boot configuration, it's reasonable to integrate this configuration into imagebuilder'ed custom image.

Sorry for my poor English.

blogic:

PR #1133, opened 2017-05-29
scripts/download.pl: Add new alias DISTFILES

diizzyy:
Add Gentoo's distfiles repository for tarballs as alias so we can
piggyback when upstream projects doesn't provide their own mirrors.

### DIFF
--- a/scripts/download.pl
+++ b/scripts/download.pl
@@ -250,6 +250,15 @@ foreach my $mirror (@ARGV) {
 		push @mirrors, "http://ftp.belnet.be/ftp.gnome.org/sources/$1";
 		push @mirrors, "ftp://ftp.cse.buffalo.edu/pub/Gnome/sources/$1";
 		push @mirrors, "ftp://ftp.nara.wide.ad.jp/pub/X11/GNOME/sources/$1";
+    } elsif ($mirror =~ /^\@DISTFILES\/(.+)$/) {
+		push @mirrors, "https://mirror.hs-esslingen.de/Mirrors/gentoo/distfiles/$1";
+		push @mirrors, "https://mirror.yandex.ru/gentoo-distfiles/distfiles/$1";
+		push @mirrors, "https://mirror.csclub.uwaterloo.ca/gentoo-distfiles/distfiles/$1";
+		push @mirrors, "http://ftp.iij.ad.jp/pub/linux/gentoo/distfiles/$1";
+		push @mirrors, "http://mirrors.lug.mtu.edu/gentoo/distfiles/$1";
+		push @mirrors, "http://mirror.netcologne.de/gentoo/distfiles/$1";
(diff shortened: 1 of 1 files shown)


hauke:
I do not see the advantage of downloading them from gentoo's mirror instead of the lede mirror.

diizzyy:
My main intention was to provide a simple way to add availability and offload main sites and indirectly LEDE infrastructure. @DISTFILES --> Main site --> LEDE mirror

diizzyy:
Anyone?

@jow- 
I know you're a bit concerned about corp envs, this would save us quite a bit from pulling git ourselves too as we can use gentoo's tarballs. No go?

diizzyy:
Is this a no go? It's really handy when it comes to the packages tree too...

diizzyy:
Can we please have this reviewed?

diizzyy:
Any opinions? We don't have to change it as a primary source but it would be handy for a number of packages.

lynxis:
IMHO: I wouldn't like to give out the control to an other distri. We have our own mirror servers. So if you like to prefer a mirror over the git repo, that should be discussed instead.

diizzyy:
I'm giving it up, I figured much better availability and speed was of interest and but I guess not.

lynxis:
@diizzyy if you think we have speed problems, than let's address them!

diizzyy:
@lynxis 
I'm more concerned about availability (especially packages with tarballs based on git pull reqs but also packages in general where there are no mirrors defined) and piggybacking also provides faster speeds in general as the mirrors are fast and have good overall connectivity. It would also provide https, save cpu cycles and bandwidth on the current infrastructure. Worth mentioning is that both sources.lede-project.org and mirror2.openwrt.org mirrors are on the same network (Hertzner), I don't really consider the openwrt mirrors true mirrors as they don't carry most files needed. Even if we don't want to use it for packages in tree it would be a good option for the "external" openwrt package repo. Don't get me wrong here, I'm grateful to the people shelling out money for the hertzner boxes but perhaps we would use them more "wisely"?

rgzr:
I think this is useful for packages missing in the mirrors. Or maybe adding them just as general fallback mirrors

diizzyy:
Let's reopen this, we have quite a few ports that would benefit from this not to mention moving towards HTTPS downloads overall and hopefully secure ones later on.
As a side note, FreeBSD also does the for the very same reason (although manually for now in several ports). Example: https://www.freshports.org/net/rinetd/

diizzyy:
https://github.com/openwrt/packages/blob/master/net/fakeidentd/Makefile#L15
https://github.com/openwrt/packages/blob/master/libs/alsa-lib/Makefile#L16
https://github.com/openwrt/packages/blob/master/net/bitlbee/Makefile#L16
For instance already does this and there are pull requests trying to fix this issue too where there are no upstream mirrors available.

PR #1134, opened 2017-05-29
ar71xx: migrate Arduino Yùn from legacy.mk to generic.mk

camille-hack:
Migrate Arduino Yùn data from ar71xx/image/legacy.mk to ar71xx/image/generik.mk
Dismiss the 8MB version of Arduino Yùn which was only uses during its development.

Signed-off-by: Camille Bilodeau <camille.bilodeau@protonmail.com>

### DIFF
--- a/target/linux/ar71xx/image/generic.mk
+++ b/target/linux/ar71xx/image/generic.mk
@@ -790,6 +790,19 @@ define Device/bhr-4grv2
 endef
 TARGET_DEVICES += bhr-4grv2
 
+define Device/yun
+  DEVICE_TITLE := Arduino Yun
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-storage kmod-spi-gpio kmod-serial-8250 kmod-ath9k
+  BOARDNAME := Yun
+  KERNEL_SIZE = 1280k
+  IMAGE_SIZE = 16000k
+  MTDPARTS := spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,14656k(rootfs),1280k(kernel),64k(nvram),64k(art)ro,15936k@0x50000(firmware)
+  IMAGE/sysupgrade.bin = append-kernel | pad-to $$$$(KERNEL_SIZE) | append-rootfs | pad-rootfs | check-size $$$$(IMAGE_SIZE)
+  CONSOLE := ttyATH0,250000
+endef
+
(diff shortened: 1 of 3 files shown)


blogic:
Please change the email address inside the commit message to your real mail addr

target/linux/ar71xx/image/generic.mk
```diff
@@ -790,6 +790,19 @@ define Device/bhr-4grv2
 endef
 TARGET_DEVICES += bhr-4grv2
 
+define Device/yun
+  DEVICE_TITLE := Arduino Yun
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-storage kmod-spi-gpio kmod-serial-8250 kmod-ath9k
```
pepe2k: Your patch says "migrate... from legacy to generic", so this should go into a separate patch, with a proper description. Please, don't hide in your patch changes which are out of scope of the patch.

target/linux/ar71xx/image/generic.mk
```diff
@@ -790,6 +790,19 @@ define Device/bhr-4grv2
...
+  BOARDNAME := Yun
+  KERNEL_SIZE = 1280k
+  IMAGE_SIZE = 16000k
+  MTDPARTS := spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,14656k(rootfs),1280k(kernel),64k(nvram),64k(art)ro,15936k@0x50000(firmware)
+  IMAGE/sysupgrade.bin = append-kernel | pad-to $$$$(KERNEL_SIZE) | append-rootfs | pad-rootfs | check-size $$$$(IMAGE_SIZE)
```
pepe2k: This is wrong and would create not working image/brick device...

Take a look please how you defined mtd partitions in line above (or how they were defined in `legacy.mk`) and make sure that resulting image is exactly the same as before move from `legacy` to `generic` building code. Ideally, you should build and test new image before sending this patch.

target/linux/ar71xx/image/generic.mk
```diff
@@ -790,6 +790,19 @@ define Device/bhr-4grv2
...
+  CONSOLE := ttyATH0,250000
+endef
+
```
pepe2k: Drop this empty line, please.

camille-hack:
I will rewrite my commits and prepare different pull request.

PR #1136, opened 2017-05-29
dropbear: runtime limit maximum authentication attempts

kdarbyshirebryant:
Two patches here:  First introduces a new '-T' runtime option to limit the maximum number of authentication attempts per connection.  This defaults to a new compile time option DEFAULT_AUTH_TRIES which itself defaults to compile time MAX_AUTH_TRIES for backwards compatibility.   This patch has been submitted upstream.

The next patch takes makes use of the DEFAULT_AUTH_TRIES compile time option and sets it to 3.

These patches lay the groundwork necessary for implementing a uci/luci interface for run time setting of maximum authentication attempts per connection.

### DIFF
--- a/package/network/services/dropbear/Makefile
+++ b/package/network/services/dropbear/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dropbear
 PKG_VERSION:=2017.75
-PKG_RELEASE:=1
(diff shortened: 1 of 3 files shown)


diizzyy:
This should go upstream rather than downstream imho

kdarbyshirebryant:
It has already been submitted upstream for review as per description.

PR #1137, opened 2017-05-30
download.mk: export git clone opts as PKG_GIT_OPTS

wongsyrone:
Signed-off-by: Syrone Wong <wong.syrone@gmail.com>

### DIFF
--- a/include/download.mk
+++ b/include/download.mk
@@ -257,6 +257,7 @@ define Download/default
   VERSION:=$(PKG_SOURCE_VERSION)
   $(if $(PKG_MD5SUM),MD5SUM:=$(PKG_MD5SUM))
   $(if $(PKG_HASH),HASH:=$(PKG_HASH))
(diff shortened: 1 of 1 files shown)


wongsyrone:
Example usage:

`PKG_GIT_OPTS:= --depth 1`

It's equivalent to `git clone --depth 1`, which can speed up on low-speed links. To make it clear, it can be renamed to `PKG_GIT_CLONE_OPTS`

nbd168:
Who would actually put this in the package makefiles? For standard packages that are being built automatically, this is unnecessary because mirror tarballs are typically available shortly after an update has been pushed.
Do you intend to use this yourself? If so, what for?

wongsyrone:
Yes, it is for personal usage. We have low oversea bandwidth due to the GFW in China. After adding this command, we save repo clone time when doing experiments.

PR #1139, opened 2017-05-30
lantiq: export ltq_reset_once to allow drivers to reset clock domains…

sch-m:
… and ip cores

This is needed e.g. to re-enable the "reset_ppe" functionality in the
ltq-atm and ltq-ptm drivers


hauke:
Are you planning to send some patches which will use this function?
I am currently working on cleaning up the drivers for the RCU registers and get it upstream, but for the time being I am ok with this driver.

sch-m:
Yes, I will send some patches that will use this function.
In detail, I've implemented an automatic loading of the ltq-atm/ltq-ptm driver based on the sync-data within an notify script (dsl_control). This makes it possible to use the router with adsl/vdsl lines and atm/ptm xfer modes without any config-changes. The reset calls are needed to switch between atm and ptm mode without reboot. I will send this stuff next week.
I've already seen on the mailing lists that you currently working on the rcu stuff. We can port this drivers to the new api when it's released.

sch-m:
Sorry for the delayed update.
I've found some problems in my setup where after switching between ptm and atm even with the re-enabled reset_ppe() function in the ltq-atm/ltq-ptm drivers no traffic is possible. I will investigate this further and come back with a new pull request when it's fixed.
I'll also close this PR for now, because i don't know if this change will be needed anymore.

PR #1140, opened 2017-05-31
libs/lzo: Reenable unaligned access on ARM, PPC, ...

StefanOberhumer:
Due a compiler bug
 ( https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64516 )
unaligned access was disabled on targets other than i386 and x86_64.

A fix has been added to lzo-2.09 so it is not necessary to disable
unaligned access within the Makefile anymore.

### DIFF
--- a/package/libs/lzo/Makefile
+++ b/package/libs/lzo/Makefile
@@ -45,10 +45,6 @@ endef
 TARGET_CFLAGS += $(FPIC)
 MAKE_FLAGS += CFLAGS_O="$(TARGET_CFLAGS)"
 
-ifeq ($(CONFIG_i386)$(CONFIG_x86_64),)
-  TARGET_CFLAGS += -DLZO_CFG_NO_UNALIGNED=1
-endif
-
 define Build/InstallDev
 	$(INSTALL_DIR) $(1)/usr/include $(1)/usr/lib
 	$(CP) $(PKG_INSTALL_DIR)/usr/include/lzo $(1)/usr/include/


nbd168:
While technically unaligned accesses work on these architectures, they should be avoided for performance reasons. On MIPS, an unaligned load generates a kernel exception, and the exception handler needs to look at the faulting instruction and emulate it before returning back to user space.
Are you sure this is a good idea?

StefanOberhumer:
LZO determines unaligned access by "itself".
Disabling unaligned access was just a followup due the compiler bug on ARM.
See commit 061319ec3dfe9b6d14af1286a1d9979db56048d7 where the unaligned access was disabled.
Should I better "just" revert 061319ec3dfe9b6d14af1286a1d9979db56048d7 ?

nbd168:
Thanks for the explanation. The commit is missing the Signed-off-by tag, the rest is fine.

StefanOberhumer:
> ....The commit is missing the Signed-off-by tag, the rest is fine.
OK, I'll update the pull request including the Signed-off-by tag

StefanOberhumer:
I just wanted to ask if there is any open problem with this patch as it seems still not merged?
As I think the **performance gain on ARM and PPC** would be a benefit to a lot of users and packages -  since LZO is used in some packages (openvpn, uboot, ...)

nbd168:
Sorry for the delay. Merged to my staging tree now.

PR #1141, opened 2017-06-01
lantiq: lantiq_dsl.sh fixes

he-ma:
Removed bug in output of `/etc/init.d/dsl_control lucistat`, where floating values were always shown as 0.

Changed latency's text output and separated 'dB' texts with space of `/etc/init.d/dsl_control status`


hauke:
Thank you for the patches, they are added to LEDE.

PR #1143, opened 2017-06-02
mvebu: Add block device sd to default kernel config

diizzyy:
Add block device sd to kernel config otherwise AHCI/eSATA devices won't get enumerated in /dev


lynxis:
merged.

PR #1144, opened 2017-06-03
FB3370: fix hotplug script for ath9k eeprom extraction

andreamerello:
Currently, on FB3370 board, the script correctly extracts the eeprom 
from AVM EVA boot partition and it saves it as it is.

This is the format that the stock AVM firmware ath9k driver expects
(in fact /dev/wlan_eeprom1 in original AVM firmware is populated
with the same contents). On the other hand this is not the format
the in-kernel ath9k driver expects. In particular eeprom contents
are reversed (not byte-swapped, they are stored upside-down, from
last byte to the beginning).

This seems true for both a FB3370 "internation edition" and a
FB3370 for German market.

This patch makes the hotplug script still extract the verbatim
eeprom data, saving it as in a file with '.avm' suffix (e.g.
ath9k-eeprom-pci-0000:01:00.0.bin.avm) and then it fixes the eeprom
format generating the regular eeprom file e.g.
ath9k-eeprom-pci-0000:01:00.0.bin

There are a lot of hints that let me guess that this ends up in
the correct eeprom format.

- The customer data is now meaningful ("AVM3370_CAL1_V2") and
  consistent to the dmesg prints spitted out by the AVM firmware on
  boot:
  [   31.480000][1] [wlan_eeprom] Calibration data blocks found = 1
  [   31.490000][1] [wlan_eeprom] EEPROM #1, type "AR93xx/AR95xx":
  [   31.490000][1] [wlan_eeprom] Customer data="AVM3370_CAL1_V2"
  [   31.490000][1] [wlan_eeprom] Build with ART2 2.23

- The MAC address is right

- The TX and RX masks are consistent with my 3-antennas device
  (both 0b00000111)

- calFreqPier2G array contains values that are reasonable (ch 1,6,13)
  and following calibration data values have similar values on each
  chain. The same is true also for 5G calibration data.

- Some values in the 2G calibration data are equal or very close to
  those in the 5G calibration area.

  2G:
   Ant. Common Control : 0x00000110
  Ant. Common Control2 : 0x00022222
             Ant. Gain :          0
         Switch Settle :         44
            Volt Slope :          0
         txEndToXpaOff :          1
           txEndToRxOn :          2
        XPA Bias Level :          0
    txFrameToDataStart :         14
         txFrameToPaOn :         14
        txFrameToXpaOn :         14
                txClip :          3
      ADC Desired Size :        -30
              Thresh62 :         28

  5G:
   Ant. Common Control : 0x00000110
  Ant. Common Control2 : 0x00022222
             Ant. Gain :          0
         Switch Settle :         45
            Volt Slope :         15
         txEndToXpaOff :          0
           txEndToRxOn :          2
        XPA Bias Level :          0
    txFrameToDataStart :         14
         txFrameToPaOn :         14
        txFrameToXpaOn :         14
                txClip :          3
      ADC Desired Size :        -30
              Thresh62 :         28

- Some values are similar to values in other dumps found on the
  internet [1]

    Write Enable GPIO :          6
    WLAN Disable GPIO :          0
        WLAN LED GPIO :          8
  Rx Band Select GPIO :        255

[1] https://wikidevi.com/wiki/Foxconn_U98Z081.00_(HP)

Signed-off-by: Andrea Merello <andrea.merello@gmail.com>

### DIFF
--- a/target/linux/lantiq/base-files/etc/hotplug.d/firmware/12-ath9k-eeprom
+++ b/target/linux/lantiq/base-files/etc/hotplug.d/firmware/12-ath9k-eeprom
@@ -60,6 +60,14 @@ ath9k_eeprom_extract() {
 	ath9k_eeprom_extract_raw $mtd $offset $swap
 }
 
+fb3370_eeprom_fix()
+{
+    for offs in `seq 0 1089`; do
+	let in_offs=4091-$offs
+	dd if=/lib/firmware/$FIRMWARE of=/lib/firmware/$FIRMWARE_ORIG conv=notrunc bs=1 count=1 seek=$offs skip=$in_offs 2> /dev/null
+    done
+}
+
 ath9k_ubi_eeprom_extract() {
 	local part=$1
 	local offset=$2
@@ -138,7 +146,13 @@ case "$FIRMWARE" in
 				ath9k_eeprom_extract "calibration" 61440 0
 				ath9k_patch_fw_mac_crc $(macaddr_add $(mtd_get_mac_ascii uboot-env ethaddr) +2) 524
 				;;
-			FRITZ3370|FRITZ7320|FRITZ7360SL)
+			FRITZ3370)
+				FIRMWARE_ORIG=$FIRMWARE
+				FIRMWARE=${FIRMWARE}.avm
+				ath9k_eeprom_extract "urlader" 2437 0
+				fb3370_eeprom_fix
+				;;
+			FRITZ7320|FRITZ7360SL)
 				ath9k_eeprom_extract "urlader" 2437 0
 				;;
 			TDW8970|TDW8980)


mkresin:
I have the same weird reverse ath9k EEPROM on another AVM device and fixed it slightly different.

What do you thing about [reversing the bytes in memory](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=commitdiff;h=1d4ea77fe5120c8479d1cd086bcb18f7b6de025c#patch5) instead of invoking dd for each byte?

andreamerello:
IMO your way is better than mine :) It should be faster, I guess..

I'm also leaving out some bytes from the eeprom image, do you think it's worth to still save also a plain dump, just in case?

Anyway, if you plan to merge your work in LEDE tree, then I could wait for it, and then I could update my PR to take advantage of your ath9k_eeprom_extract_reverse function

mkresin:
> I'm also leaving out some bytes from the eeprom image, do you think it's worth to still save also a plain dump, just in case?

No, keeping a copy of the not reversed eeprom doesn't have any benefits.

Would you please compile an image from my staging tree and report back if the issue is fixed! I've added one more commit, which gets the boards mac addresses from the tffs partition. The former approach of using the (well known) offset in the ath9k eeprom doesn't work if the eeprom is reversed.

andreamerello:
Sure, I'll try in next days..

But in FB3370 the reversed eeprom seems to contain the right MAC address indeed. With the reversed-extracted eeprom image I'm already getting the correct MAC address for my wlan interface without doing anything else.

mkresin:
> But in FB3370 the reversed eeprom seems to contain the right MAC address indeed.

Yeah sure. But reading the WLAN mac address from flash (which is incremented/decremented to use the same LAN/WAN mac address as the AVM firmware) doesn't work [[0]](https://github.com/lede-project/source/blob/master/target/linux/lantiq/dts/FRITZ3370.dts#L218)[[1]](https://github.com/lede-project/source/blob/master/target/linux/lantiq/base-files/etc/board.d/02_network#L111), since the mac address is a) reversed (FF:EE:DD:CC:BB:AA instead of AA:BB:CC:DD:EE:FF) and b) at a different than the expected position within the ath9k eeprom on flash.

There are more odd things with the 3370. The internal switch isn't configured (no port isolation possible without) and there seams to be something wrong with the wifi button. Just in case you are interested.

andreamerello:
Ah, sorry. I didn't notice that your patch is about MAC addresses of other interfaces.. I was still thinking to wlan mac address.. I'll try it..

BTW another issue is that FB3370 international version (that with original firmware works on Annex A lines) refuses to load Annex A DSL firmware with LEDE (I've got it working with an hardware hack, but I still want to find out a SW solution). The DE-version and international version boards are actually exactly identical, and swapping original firmwares works, so it must be a matter of SW.

On the other hand the good news is that I've fixed u-boot to work on this device and I'm successfully using it. I needed also to fix a kernel bug, but now I can use UBIFS and a UHI-passed DT to boot my device; I'll try to move also the rootfs un UBIFS in next days..  I would like to do something like your installation scripts for HomeHub 5 Type A.. 

I still need to figure out the ECC NAND scheme used by AVM/EVA, and I don't know how to integrate my changes in LEDE (for example two different DT would be needed for booting from EVA and u-boot)..

PR #1145, opened 2017-06-04
ar71xx: make arduino yun work with kernel size > 1280k and use generic.mk

camille-hack:
Hello,

I gather here three commits that have been tested and work on my board.

# ar71xx: allows kernel size > 1280k
The default mtd partitionning has a "kernel" partition which is too small for new kernels. This patch let the default LEDE partitionning apply, which solve the problem.

# ar71xx: move arduino Yun 16M to generic.mk
The arduino Yun 16M settings were stored in legacy.mk and legacy-devices.mk. Here they are migrated to the new generic.mk building system.

# ar71xx: remove Arduino Yun 8 MiB prototype
The arduino Yun 8 MiB was a prototype board used internally by arduino but was never made public. I don't know what is the policy of LEDE in this regard but it would make sense to discard this board.[source](https://blog.arduino.cc/2013/08/21/updating-about-arduino-yun-and-arduino-robot/)

### DIFF
--- a/package/boot/uboot-envtools/files/ar71xx
+++ b/package/boot/uboot-envtools/files/ar71xx
@@ -75,6 +75,9 @@ rambutan)
 qihoo-c301)
 	ubootenv_add_uci_config "/dev/mtd9" "0x0" "0x10000" "0x10000"
 	;;
+arduino-yun)
+	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x10000" "0x10000"
+	;;
 esac
 
 config_load ubootenv
--- a/target/linux/ar71xx/image/generic.mk
+++ b/target/linux/ar71xx/image/generic.mk
@@ -92,6 +92,16 @@ define Device/ap90q
 endef
 TARGET_DEVICES += ap90q
 
+define Device/arduino-yun
+  BOARDNAME := Yun
+  DEVICE_TITLE := Arduino Yun
+  DEVICE_PACKAGES := kmod-usb-core kmod-usb2
+  MTDPARTS := spi0.0:256k(u-boot)ro,64k(u-boot-env),15936k(firmware),64k(nvram),64k(art)ro
+  CONSOLE = ttyATH0,250000
+  IMAGE_SIZE := 15936k
+endef
+TARGET_DEVICES += arduino-yun
+
 define Device/bsb
   DEVICE_TITLE := Smart Electronics Black Swift board
   DEVICE_PACKAGES := kmod-usb-core kmod-usb2
--- a/target/linux/ar71xx/image/legacy-devices.mk
+++ b/target/linux/ar71xx/image/legacy-devices.mk
@@ -167,12 +167,6 @@ define LegacyDevice/WLR8100
 endef
 LEGACY_DEVICES += WLR8100
 
-define LegacyDevice/YUN_16M
-  DEVICE_TITLE := Arduino Yun based on Atheros AR9331 (16MB flash)
-  DEVICE_PACKAGES := kmod-usb-core kmod-usb2
-endef
-LEGACY_DEVICES += YUN_16M
-
 define LegacyDevice/F9K1115V2
   DEVICE_TITLE := Belkin AC1750DB (F9K1115V2)
   DEVICE_PACKAGES := kmod-usb-core kmod-usb2 kmod-usb3 kmod-usb-ledtrig-usbport \
--- a/target/linux/ar71xx/image/legacy-devices.mk
+++ b/target/linux/ar71xx/image/legacy-devices.mk
@@ -167,12 +167,6 @@ define LegacyDevice/WLR8100
 endef
 LEGACY_DEVICES += WLR8100
 
-define LegacyDevice/YUN_8M
-  DEVICE_TITLE := Arduino Yun based on Atheros AR9331 (8MB flash)
-  DEVICE_PACKAGES := kmod-usb-core kmod-usb2
-endef
-LEGACY_DEVICES += YUN_8M
-
 define LegacyDevice/YUN_16M
   DEVICE_TITLE := Arduino Yun based on Atheros AR9331 (16MB flash)
   DEVICE_PACKAGES := kmod-usb-core kmod-usb2
--- a/target/linux/ar71xx/image/legacy.mk
+++ b/target/linux/ar71xx/image/legacy.mk
@@ -277,7 +277,6 @@ wndr4300_mtdlayout=mtdparts=ar934x-nfc:256k(u-boot)ro,256k(u-boot-env)ro,256k(ca
 zcn1523h_mtdlayout=mtdparts=spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,6208k(rootfs),1472k(kernel),64k(configure)ro,64k(mfg)ro,64k(art)ro,7680k@0x50000(firmware)
 mynet_rext_mtdlayout=mtdparts=spi0.0:256k(u-boot)ro,7808k(firmware),64k(nvram)ro,64k(ART)ro
 zyx_nbg6716_mtdlayout=mtdparts=spi0.0:256k(u-boot)ro,64k(env)ro,64k(RFdata)ro,-(nbu);ar934x-nfc:2048k(zyxel_rfsd),2048k(romd),1024k(header),2048k(kernel),-(ubi)
-yun_mtdlayout_16M=mtdparts=spi0.0:256k(u-boot)ro,64k(u-boot-env)ro,14656k(rootfs),1280k(kernel),64k(nvram),64k(art)ro,15936k@0x50000(firmware)
 
 define Image/BuildKernel
 	cp $(KDIR)/vmlinux.elf $(VMLINUX).elf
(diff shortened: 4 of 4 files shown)


pepe2k:
@camille-hack,

Please, use different `bootm` command in your commit message (please, test it in advance):

`setenv bootcmd "run addboard; run addtty; run addparts; run addrootfs; bootm 0x9f050000 || bootm 0x9fea0000"`

This will make it possible to load both types of firmware (kernel-rootfs and rootfs-kernel orders).

Do you maybe know if the `nvram` mtd partition is used by vendor firmware? If not, I think we can drop it and extend `firmware` partition.

Also, I prefer different commits order:
1. Drop 8M image.
2. Move 16M image to generic (squashed with different mtd organization).

We should probably also include support for this board in `uboot-envtools` package (and make `u-boot-env` mtd partition writable).

Cheers,
Piotr

camille-hack:
Thanks for your input !

Ok, I am asking the vendor Arduino about the `nvram` partition. And will reorganise the commits once everything is ready. The different `bootcmd` is working on both lininoIO and LEDE. Only the `mtdparts` U-Boot environment setting has to be changed.

Meanwhile I have two questions.

About `uboot-envtools`, I tried this:
```Diff
diff --git a/package/boot/uboot-envtools/files/ar71xx b/package/boot/uboot-envtools/files/ar71xx
index b920bb946b..c4eca901b8 100644
--- a/package/boot/uboot-envtools/files/ar71xx
+++ b/package/boot/uboot-envtools/files/ar71xx
@@ -75,6 +75,9 @@ rambutan)
 qihoo-c301)
        ubootenv_add_uci_config "/dev/mtd9" "0x0" "0x10000" "0x10000"
        ;;
+yun)
+       ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x10000" "0x10000"
+       ;;
 esac
 
 config_load ubootenv
```

The `uboot-envtools` package contains `/etc/uci-defaults/30_uboot-envtools` inside. But this file is not in the final image:
```ShellSession
root@LEDE:/# ls /etc/uci-defaults/30_uboot-envtools
ls: /etc/uci-defaults/30_uboot-envtools: No such file or directory
```
Any idea ?

Also, the final idea is to have LEDE running on arduino yun with minimal user intervention. Given that the uboot environment has to be modified, what is the best way to do that automatically ? Is something like [this upgrade script](https://github.com/lede-project/source/blob/4f272dd0323151a1390a8675bafebead0e2a69ad/target/linux/ar71xx/base-files/lib/upgrade/allnet.sh) an option ?

pepe2k:
Hello @camille-hack,

> Ok, I am asking the vendor Arduino about the nvram partition. And will reorganise the commits once everything is ready. The different bootcmd is working on both lininoIO and LEDE. Only the mtdparts U-Boot environment setting has to be changed.

In fact, our kernel cmdline hack is _somehow broken_ (on my todo/tofix list), so it's always overwritten in ar71x/generic subtarget (with `patch-cmdline` with value `CMDLINE` -> `BOARDNAME`, `MTDPARTS` and `CONSOLE`), no matter what the bootloader passes to the kernel. If you remove all that cmdline related commands from `bootcmd` and leave just the `bootm`, it will still work.

> The uboot-envtools package contains /etc/uci-defaults/30_uboot-envtools inside. But this file is not in the final image:

`uci-defaults` scripts [are run at boot and removed if exited successfully](https://github.com/lede-project/source/blob/master/package/base-files/files/etc/init.d/boot#L15). If you don't have this script after boot, it means that it did its job and got removed :smiley:

> Also, the final idea is to have LEDE running on arduino yun with minimal user intervention. Given that the uboot environment has to be modified, what is the best way to do that automatically ? Is something like this upgrade script an option ?

The problem here is that people who already have OpenWrt/LEDE/vendor firmware on Arduino Yun would need to change `bootcmd` **in advance, before** (or just after) upgrade to make it working. Basically, your patch "breaks" support/backward compatibility (which in fact doesn't exist now as the images are not generated because of too big kernel) for this device. And, based on the fact that it doesn't work now anyway, I'm fully OK with this change.

Cheers,
Piotr

artynet:
hello @pepe2k ,

my name is Arturo and I am the current maintainer for linino distro. as you mentioned I have to disable the command line hack to properly build the Linino image on LEDE...this is my personal rebase on the latest trunk :

https://github.com/artynet/LEDE

which contains all the necessary modification (made for Chaos Calmer) to build the Linino image. I already answered @camille-hack by mail giving him some useful pointers and he should be able to build a new image from my code.

Best, Arturo

pepe2k:
Hello @artynet,

> my name is Arturo and I am the current maintainer for linino distro. as you mentioned I have to disable the command line hack to properly build the Linino image on LEDE...

I didn't write that cmdline hack has to be disabled. I wrote that it's broken in the way that it doesn't allow appending (but it should, based on the original intention) parameters to those already provided/passed by bootloader. At this moment, command line is always overwritten in `ar71xx/generic` subtarget kernel which means that everything provided by U-Boot from `bootargs` variable is just ignored.

> this is my personal rebase on the latest trunk :
> 
> https://github.com/artynet/LEDE
> 
> which contains all the necessary modification (made for Chaos Calmer) to build the Linino image. I already answered @camille-hack by mail giving him some useful pointers and he should be able to build a new image from my code.

There are a lot of different changes, most of just can't be included upstream (like your custom kernel changes/configs).

In my opinion it would be better for you and end users to have a working basic support in upstream (I mean here OpenWrt/LEDE) and base your custom distribution on that, without taking care of things like kernel configuration changes, updates etc. And, to be able to do it, Yun (and probably other Arduino boards with ath79) mtd partitions organization needs to be changed in the way that the kernel size isn't limited like you have it now. I know that the `rootfs+kernel` mtd order comes from Qualcomm/Atheros SDK but it's really problematic (I was trying to not write here "stupid") and I still don't understand why manufacturers blindly follow it.

Cheers,
Piotr

camille-hack:
@artynet 

As a customer, I agree with @pepe2k and see other advantages, as LEDE would provide

- choice, I would like to have the possibility to have a minimal distribution with not so much tools by default, as I don't do electronic that much with my arduino board, and the http/javascript thing is too complicated for me;
- a bigger community, the OpenWRT/LEDE project gathering lots of people already and it gives many more people to get information and hack than the Arduino forums (who are already great);
- security, I think LEDE project has the potential to stay the dynamic linux distribution which integrate relatively quickly security fixes for my board; this point is particularly important as a lot of "IoT" projects like the Arduino Yun are getting more and more targetted;

nxhack:
Hi @pepe2k @camille-hack 
> Basically, your patch "breaks" support/backward compatibility (which in fact doesn't exist now as the images are not generated because of too big kernel) for this device.

I am an LEDE & Arduino Yun user. I am running the latest kernel by turning off 'CONFIG_KERNEL_KALLSYMS'.

However, I totally agree with this change.

camille-hack:
@nxhack I saw your wiki page which was really helpful !

Update : I reorganized the commits as requested.

Also, I changed two new files as part of this commit, as I think they are important to make the arduino Yun build again:
* kernel board name match the one from the boot loader (`linino-yun` instead of `Yun`);
* `uboot-envtools` setting added, fw_setenv is working.

Please can someone review, thank you.

pepe2k:
Hi @camille-hack,

> Also, I changed two new files as part of this commit, as I think they are important to make the arduino Yun build again:
> 
>- kernel board name match the one from the boot loader (linino-yun instead of Yun)

Please, drop this change. It's not needed and not related with what your patch actually does. What's more, your board name change is incomplete and will cause a regression. You changed the name only in kernel side code but left user space scripts untouched.

Also, new commit title is misleading, it should say something like before (`move Arduino Yun to generic building code`). In fact, your change isn't related with making this board working with newer kernel as it works with it already without any problems... the root cause of the problem here is mtd organization and size of the kernel, nothing else. Please, keep one change per commit and don't include hidden or unrelated changes.

Cheers,
Piotr

artynet:
Hi @camille-hack ,

 >- choice, I would like to have the possibility to have a minimal distribution with not so much tools by default, as I don't do electronic that much with my arduino board, and the http/javascript thing is too complicated for me;
>- a bigger community, the OpenWRT/LEDE project gathering lots of people already and it gives many more people to get information and hack than the Arduino forums (who are already great);
>- security, I think LEDE project has the potential to stay the dynamic linux distribution which integrate relatively quickly security fixes for my board; this point is particularly important as a lot of "IoT" projects like the Arduino Yun are getting more and more targeted;

you have some good points here.... Last October I met with the core developers in Berlin for  the ELC-E event and we discussed about including our boards in the main trunk code at least for provide basic support (i.e. kernel build at least).

It would be nice having a bigger kernel size compared to the current one but consider that I kept following our design guidelines which have not changed since the first release of Linino. We have never thought about it because we never had an actual need of a larger kernel partition.

As final notes, I named the board as `linino-yun` since my C file contains additional modifications suited for the new revisions of arduino Yun. For this reason, the need to differentiate between the original `arduino-yun` and `linino-yun` in the boards list and we should keep them this way.

Best, Arturo

camille-hack:
Hi @pepe2k
Ok, I changed it, thank you for your help. Please review.

pepe2k:
@camille-hack,

Thank you for your first-time contribution! I have merged your changes (with minor fixes) into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1146, opened 2017-06-04
Update mxs target to kernel v4.9

mhei:
Update and clean up Freescale/NXP i.MX 23/28 target to kernel version v4.9

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


blogic:
can you please also drop the v4.4 files in this series ?

mhei:
@blogic: Thanks for your feedback. I've added the requested commit, please review again. Thanks!

hauke:
Is it correct that this should supersedes the patches from here:
https://patchwork.ozlabs.org/patch/761360/
https://patchwork.ozlabs.org/patch/761361/

Why are you not porting the patches for mxs_power.c and mxs-regulator.c to the kernel 4.9 are they replaced by some other driver now?

mhei:
The patches in patchwork were my initial submission, then I filed this PR because I was not sure which approach is preferred by LEDE people.

I did not port the regulator and power patches from Stefan Wahren
because I talked to him and he told me that work on this is currently
stalled. And since AFAIK nothing depends on these patches and the gain to have them in is not that much, leaving them out seems reasonable (at least for my point of view). I'm not sure, why the OpenWrt platform maintainer put them in.

hauke:
Thank you for the patches, they are applied now.

PR #1149, opened 2017-06-06
mvebu: switch low-level uart to new bootloaders

Artox:
As it happens there is an old and a new method for handling low-level
serial console output:
CONFIG_DEBUG_MVEBU_UART0
CONFIG_DEBUG_MVEBU_UART0_ALTERNATE

Documentation just says that the newer should be used with newer
bootloaders. Lede follows mainline u-boot, so we can safely assume that
will hold true.

Using the previous default, old-bootloader, completely broke the
SolidRun Clearfog Pro when early-printk was enabled!!!


Artox:
This may not be fully agreeable, as I am just moving the crash from new devices to old devices.
Are there any suggestions how to deal with this properly?

Imo this kind of kernel-crashes-queitly is extremely hard to debug and wastes hours whenever it occurs!

jow-:
Ping @nbd168 @blogic

KanjiMonster:
> Lede follows mainline u-boot, so we can safely assume that will hold true.

As you noticed yourself, this is only true for boards where we do provide our own u-boot, which is exactly one, clearfog. For all other devices we don't replace the vendor one, and changing this option will likely break early-printk on those models. Therefore I don't think changing this option is a good idea.

A better idea might be to expose this option in menuconfig. You need to build your own image anyway to enable early printk. Note that there is no automatic way for choosing the right value, as you might be building for multiple devices, so the sensible default would be the one working on most devices.

Artox:
> A better idea might be to expose this option in menuconfig.

Yes, that would be fine!
At least get rid of the automatic that chooses the wrong thing by default on a defconfig level.

I'll see if I can come up with something.

PR #1150, opened 2017-06-07
base-files: Add netdev-trigger support to /lib/functions/leds.sh

sch-m:

### DIFF
--- a/package/base-files/files/lib/functions/leds.sh
+++ b/package/base-files/files/lib/functions/leds.sh
@@ -21,6 +21,12 @@ led_off() {
 	led_set_attr $1 "brightness" 0
 }
 
+led_netdev() {
+	led_set_attr $1 "trigger" "netdev"
+	led_set_attr $1 "device_name" "$2"
+	led_set_attr $1 "mode" "$3"
+}
+
 status_led_set_timer() {
 	led_timer $status_led "$1" "$2"
 	[ -n "$status_led2" ] && led_timer $status_led2 "$1" "$2"
@@ -28,6 +34,11 @@ status_led_set_timer() {
 
 status_led_set_heartbeat() {
 	led_set_attr $status_led "trigger" "heartbeat"
+}
+
+status_led_set_netdev() {
+	led_netdev $status_led "$1" "$2"
+	[ -n "$status_led2" ] && led_netdev $status_led2 "$1" "$2"
 }
 
 status_led_on() {
--- a/target/linux/lantiq/base-files/sbin/dsl_notify.sh
+++ b/target/linux/lantiq/base-files/sbin/dsl_notify.sh
@@ -22,7 +22,14 @@ if [ -n "$led" ]; then
 	case "$DSL_INTERFACE_STATUS" in
 	  "HANDSHAKE")  led_timer $led 500 500;;
 	  "TRAINING")   led_timer $led 200 200;;
-	  "UP")		led_on $led;;
+	  "UP")		config_get trigger led_dsl trigger
+			if [ "$trigger" = "netdev" ]; then
+				config_get dev led_dsl dev
+				config_get mode led_dsl mode
+				led_netdev $led $dev "$mode"
+			else
+				led_on $led
+			fi;;
 	  *)		led_off $led
 	esac
 fi


pepe2k:
@sch-m,

Personally, I don't see any reason for having more different triggers for "status LED" and prefer to keep some kind of consistency across all targets. Please, take a look also at this discussion: https://github.com/lede-project/source/pull/935#discussion_r105770604

Anyway, your commit message is missing explanation why this is needed for.

Cheers,
Piotr

sch-m:
@pepe2k ,
I want to use the netdev trigger for the dsl led in the [lantiq dsl_notify.sh script](https://github.com/lede-project/source/blob/master/target/linux/lantiq/base-files/sbin/dsl_notify.sh#L23) which already uses the functions from /lib/functions/leds.sh and thought it wouldn't hurt to extend the leds.sh with this feature. 

Regards,
Martin

pepe2k:
Hello @sch-m,

> I want to use the netdev trigger for the dsl led in the lantiq dsl_notify.sh script which already uses the functions from /lib/functions/leds.sh and thought it wouldn't hurt to extend the leds.sh with this feature.

OK, but first of all, the `status_led_*()` helper functions are used for the status LED, [the one set in `diag.sh`](https://github.com/lede-project/source/blob/master/target/linux/lantiq/base-files/etc/diag.sh#L11). I don't fully understand what/how do you want to use `status_led_set_netdev()` inside `dsl_notify.sh` as this script doesn't touch the status LED at all?

What's more, at the moment, `dsl_notify.sh` uses only `timer` and `on/off` triggers. Are you sure that `netdev` is the correct choice there? Wouldn't that break how the status of DSL connection is indicated to the user?

It would be much easier to review your idea if you included also changes inside `dsl_notify.sh`. Could you do that, please?

Cheers,
Piotr

sch-m:
Added the dsl_notify changes. 

You are right, status_led_netdev() isn't used in this case, just led_netdev().
I've also added the status_led_netdev() function for consistency only. But if that is a problem, we can simply drop that.

The idea of this all is to see rx/tx activity on a synchronized dsl (line).

mkresin:
> The idea of this all is to see rx/tx activity on a synchronized dsl (line).

... which is a really high level description of what you are trying to do here. What you are doing is to workaround the limitation that a single LED can only have a single trigger, by switching the trigger in dsl_notify.sh. Nevertheless, such explanations should go into the commit message to prevent such kind of questions.

The led name "status" is really misleading here, as it isn't _the_ status LED that you're interested in. It is either the internet or the dsl LED (depending on definition).

Wouldn't it be better to setup the netdev trigger - including the correct interface - in /e/c/system, switch to the timer trigger during handshake/training and reload the configured/saved trigger on UP?

```
config_get led led_dsl sysfs
config_get internet led_internet sysfs

...

"UP")
	if [ -n "$internet" ]; then
		# shared LED, reset to configured netdev trigger
	else
		led_on $led
	fi

...
```

It would still require some changes to the DTS files, since for boards with only one LED, this led is setup as "online/internet" LED with [a netdev trigger set](https://github.com/lede-project/source/blob/c00fbaf670494e7cd3dd05bff18cb6b407ac642d/target/linux/lantiq/base-files/etc/board.d/01_leds#L21) by default. Given your approach, not the best change I made...

And depending on who you ask, it might be confusing that the LED is switched off albeit only the ppp connection is terminated but the line is still in sync (taking into account that the line init is signalled via the same LED).

For reference an older discussion about the dsl/online led thingy, what is the intended way and what are the limitations (that I wasn't able to fix/workaround properly): https://bugs.lede-project.org/index.php?do=details&task_id=321.

sch-m:
> What you are doing is to workaround the limitation that a single LED can only have a single trigger, by switching the trigger in dsl_notify.sh.

That is already done even without my patch: led_timer(), led_on(), led_off()
My patch only extends this workaround.

> Wouldn't it be better to setup the netdev trigger - including the correct interface - in /e/c/system, switch to the timer trigger during handshake/training and reload the configured/saved trigger on UP?

OK, that's the way my patch does, isn't it?

I think this feature shouldn't be mixed with the internet led, as the thing I want to see on the dsl led here are simply the activity on the dsl. There don't have to be an "internet" on this line.

mkresin:
> OK, that's the way my patch does, isn't it?

Oh yeah, without an board actual using the code,  I failed hard to see that your patch does exactly what I asked for. My bad.

> I think this feature shouldn't be mixed with the internet led, as the thing I want to see on the dsl led here are simply the activity on the dsl. There don't have to be an "internet" on this line.

IMHO we are talking already about the same. Here an example from a HH5a default LED config, which _only_ has a broadband LED (which I'm referring to as internet/online LED):

```
config led 'led_internet'
        option name 'internet'
        option sysfs 'bthomehubv5a:blue:broadband'
        option trigger 'netdev'
        option mode 'link tx rx'
        option dev 'ptm0'
        option default '0'
```

Looks quite similar to what your code expects. Just that your change would allow to show the line init as well and expects the LED to be named "dsl".

Anyway I've chosen a slightly different approach which should do the same. Would you please confirm that the code in [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary) works for your use case as well!

I decided to keep the code in the lantiq target. In the end `status_led_set_netdev()/led_dsl_up()` is a workaround to (re)set a single LED to the stored/configured values.

I've chosen a case statement instead of a condition here, to extend the function at a later point without rewriting everything. Might be that there are more trigger interesting in future. Hopefully it doesn't look that much like the hack that it is.

Any kind of feedback is appreciated!

sch-m:
I'm not able to test your patch on real hardware, because I'm out of office for some days. But your code looks fine to me.

And about the /lib/functions/leds.sh and status_led_xxx() thing:
IMHO that wouldn't be the wrong place for led_netdev() i think, because the name /lib/functions/leds.sh suggests that it is a general LED functions "library". If this file should only be used for status_led functions, then it should be renamed to /lib/functions/status_led.sh.

mkresin:
> And about the /lib/functions/leds.sh and status_led_xxx() thing:
IMHO that wouldn't be the wrong place for led_netdev() i think, because the name /lib/functions/leds.sh suggests that it is a general LED functions "library". If this file should only be used for status_led functions, then it should be renamed to /lib/functions/status_led.sh.

The reason why i don't like to have the code in leds.sh is that I don't see a reason to configure a netdev trigger from userspace. Even you are not really interested in configuring a netdev trigger. All you are interested in, is to restore what was configured (aka stored) before on/off or timer trigger is set. We do not have a clean way to do it and therefore our hackish attempt.

I've pushed the changes to master now. Thanks a lot for your contribution.

PR #1151, opened 2017-06-07
Add missing license tags

feckert:
Add missing license tags for some packages.


mkresin:
Thanks. I've merged a squashed version to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

PR #1155, opened 2017-06-08
dnsmasq: make bind-dynamic 'non-wildcard' interfaces default

kdarbyshirebryant:
'non-wildcard' interfaces enables dnsmasq's '--bind-dynamic' mode.  This
binds to interfaces rather than wildcard addresses *and* keeps track of
interface comings/goings via a unique Linux api.

Quoting dnsmasq's author "bind-dynamic (bind individual addresses, keep
up with changes in interface config) ... On linux, there's actually no
sane reason not to use --bind-dynamic, and it's only not the default for
historical reasons."

Let's change history, well on LEDE at least, and change the default!

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.77
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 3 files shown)


dedeckeh:
@kdarbyshirebryant Increase PKG_RELEASE as well in the package Makefile

dedeckeh:
Merged into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=4e02e1347199f45d6f4587e10b3e51ecf9260ca1); thanks

PR #1156, opened 2017-06-08
ar71xx: wpj344: fix MAC address/disable eth1

yogo1212:
This way, the assigned addresses match those on the barcode labels.
Otherwise, the addresses appear to vary on boot.
```
0002e010  04 f0 21 1a 32 0e 12 1e  04 f0 21 1a 32 0f 30 30  |..!.2.....!.2.00|
0002e020  04 f0 21 1a 32 10 ec 81  04 f0 21 1a 32 11 15 c0  |..!.2.....!.2...|
```

without patch:
```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 8e:12:da:a3:a2:45 brd ff:ff:ff:ff:ff:ff
3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 1e:b4:09:7c:98:71 brd ff:ff:ff:ff:ff:ff
```

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -526,6 +526,9 @@ ar71xx_setup_macs()
 	mynet-n750)
 		wan_mac=$(mtd_get_mac_ascii devdata "wanmac")
 		;;
+	wpj344)
+		wan_mac=$(mtd_get_mac_binary u-boot 0x2e018)
+		;;
 	esac
 
 	[ -n "$lan_mac" ] && ucidef_set_interface_macaddr "lan" $lan_mac
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-wpj344.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-wpj344.c
@@ -48,8 +48,8 @@
 #define WPJ344_KEYS_POLL_INTERVAL	20	/* msecs */
 #define WPJ344_KEYS_DEBOUNCE_INTERVAL	(3 * WPJ344_KEYS_POLL_INTERVAL)
 
-#define WPJ344_MAC0_OFFSET		0
-#define WPJ344_MAC1_OFFSET		6
+#define WPJ344_MAC0_OFFSET		0x10
+#define WPJ344_MAC1_OFFSET		0x18
 #define WPJ344_WMAC_CALDATA_OFFSET	0x1000
 #define WPJ344_PCIE_CALDATA_OFFSET	0x5000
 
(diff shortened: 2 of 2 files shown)


yogo1212:
maybe some of the initialisation code can be joined with that of other compex boards

pepe2k:
Hi @yogo1212,

Based on [this post from Compex Support Forum](https://forum.compex.com.sg/160-wpj344-eth0-and-1-mac-addresses/0), wrong MAC address offset is not the only bug in support code for this board. Would you like to fix also the rest (remove eth1 as it's not used at all)?

Cheers,
Piotr

yogo1212:
Hello @pepe2k :-)

Sounds reasonable.
If it doesn't take more than a minute, you don't happen to know of a proper way to assign one of the three other addresses to eth0.2 instead of eth0.1 and eth0.2 sharing the same?

yogo1212:
don't merge yet please

yogo1212:
i'm adding this to the first commit:
```
-#define WPJ344_MAC0_OFFSET             10
-#define WPJ344_MAC1_OFFSET             18
+#define WPJ344_MAC0_OFFSET             0x10
+#define WPJ344_MAC1_OFFSET             0x18
```
not all numbers are hex..

pepe2k:
Hello @yogo1212,

> If it doesn't take more than a minute, you don't happen to know of a proper way of assigning one of the three other addresses to eth0.2 instead of eth0.1 and eth0.2 sharing the same?

Do you know how does it look with MAC addresses in vendor firmware? Why there are 4 different? I suppose that you can [add different MAC per interface (wan/lan) here](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/base-files/etc/board.d/02_network#L501).

You should also drop `ath79_register_mdio(1, 0x0);`.

Cheers,
Piotr

yogo1212:
@pepe2k ty :-)

and: will do

meanwhile, this works again:
```
$ ip l
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 04:f0:21:1a:32:0e brd ff:ff:ff:ff:ff:ff
```

yogo1212:
the addresses appear to be used for eth-lan, eth-wan, wifi0-ssid, wifi1-ssid in that order

2017-06-09-175901_1366x768_scrot
2017-06-09-180422_1366x768_scrot

yogo1212:
@pepe2k ready from my side

pepe2k:
@yogo1212,

Thank you for your first-time contribution! I have merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1159, opened 2017-06-09
hostapd: add support for chanlist option

D-Albers:
Add support for the hostapd chanlist option to the wireless configuration file.

The chanlist option is necessary to restrict the auto selection of the wlan channel to defined subset.

Signed-off-by: Daniel Albers <daniel.albers@public-files.de>


D-Albers:
Option is supported through  'list channels'.

PR #1162, opened 2017-06-09
scripts/dl_cleanup.py: add support for version format xxx-v1.2.3

D-Albers:
dl_cleanup.py: add support for version format xxx-v1.2.3 which is used for example by btrfs-progs

Signed-off-by: Daniel Albers <daniel.albers@public-files.de>


wongsyrone:
I think it should be `"(.+)[-_]v?(\d+)\.(\d+)\.(\d+)"`

wongsyrone:
In fact, you can merge this into the next line: 
`(re.compile(r"(.+)[-_](\d+)_(\d+)_(\d+)"), parseVer_123),		# xxx-1_2_3`

"v" is already optional, the only thing to do is replace `\.`

`(.+)[-_]v?(\d+)[_.](\d+)[_.](\d+)` might be what you want.

D-Albers:
@wongsyrone 

I changed the regexp to `(.+)[-_]v(\d+)\.(\d+)\.(\d+)`.

For each version string style exits one regexpression and I think we should stuck at this style.

wongsyrone:
Well, either way has its pros and cons.

jow-:
Merged to my staging tree with http://git.lede-project.org/d518477111

PR #1163, opened 2017-06-12
Netgear DM200 port

victoredwardocallaghan:
Hi,

The following patches add support for the Netgear DM200 ADSL2/VDSL2 modem. Essentially everything works, ENJOY! :)

Cheers,
Edward.

### DIFF
--- a/target/linux/lantiq/dts/EASY220.dtsi
+++ b/target/linux/lantiq/dts/EASY220.dtsi
@@ -1,0 +1,295 @@
+#include "vr9.dtsi"
+#include <dt-bindings/input/input.h>
+
+/ {
+
+	compatible = "lantiq,xway", "lantiq,vr9", "lantiq,xrx220";
+
+	chosen {
+		bootargs = "console=ttyLTQ0,115200";
+	};
+
+	aliases {
+		led-boot = &power_g;
+		led-failsafe = &power_a;
+		led-running = &power_g;
+
+		led-dsl = &dsl;
+		led-internet = &lan_g;
+	};
+
+	memory@0 {
+		reg = <0x0 0x3f00000>;
+	};
+
+	ltq_swreset {
+		compatible = "lantiq,ltq_swreset";
+		swreset_pin = <&gpio 7 GPIO_ACTIVE_HIGH>;
+		swreset_bit = <1>;
+		status = "ok";
+	};
+
+	gphy-xrx200 {
+		compatible = "lantiq,phy-xrx200";
+		firmware = "lantiq/vr9_phy22f_a2x.bin";
+		phys = [ 00 01 ];
+	};
+
+//	gpio-keys-polled {
+//		compatible = "gpio-keys-polled";
+//		#address-cells = <1>;
+//		#size-cells = <0>;
+//		poll-interval = <100>;
+/*		reset {
+			label = "reset";
+			gpios = <&gpio 7 GPIO_ACTIVE_HIGH>;
+			linux,code = <0x198>;
+		};*/
+//	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power_a: power_amber {
+			label = "power_amber";
+			gpios = <&gpio 0 GPIO_ACTIVE_HIGH>;
+		};
+		power_g: power_green {
+			label = "power_green";
+			gpios = <&gpio 1 GPIO_ACTIVE_LOW>;
+		};
+		lan_amber {
+			label = "lan_amber";
+			gpios = <&gpio 33 GPIO_ACTIVE_LOW>;
+		};
+		lan_g: lan_green {
+			label = "lan_green";
+			gpios = <&gpio 11 GPIO_ACTIVE_LOW>;
+		};
+		dsl_amber {
+			label = "dsl_amber";
+			gpios = <&gpio 8 GPIO_ACTIVE_LOW>;
+		};
+		dsl: dsl_green {
+			label = "dsl_green";
+			gpios = <&gpio 36 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&gpio {
+	pinctrl-names = "default";
+	pinctrl-0 = <&state_default>;
+
+	state_default: pinmux {
+		mdio {
+			lantiq,groups = "mdio";
+			lantiq,function = "mdio";
+		};
+		phy-rst {
+			lantiq,pins = "io42";
+			lantiq,pull = <0>;
+			lantiq,open-drain = <0>;
+			lantiq,output = <1>;
+		};
+		exin {
+			lantiq,groups = "exin3";
+			lantiq,function = "exin";
+		};
+		/*stp {
+			lantiq,groups = "stp";
+			lantiq,function = "stp";
+		};*/
+		/*nand {
+			lantiq,groups = "nand cle", "nand ale",
+					"nand rd", "nand rdy";
+			lantiq,function = "ebu";
+		};*/
+		/*pci {
+			lantiq,groups = "gnt1", "req1";
+			lantiq,function = "pci";
+		};*/
+		/*led {
+			lantiq,groups = "led";
+			lantiq,function = "led";
+		};*/
+		conf_in {
+			lantiq,pins = "io39"; /* exin3 */
+			lantiq,pull = <2>;
+		};
+		conf_out {
+			lantiq,pins =  // "io21", "io38", /* pcie */
+//					"io12", "io15", /* annexa, annexb */
+					"io0", "io1", "io8", "io11", "io33", "io36"; /* led io34, io35 */
+			lantiq,open-drain = <0>;
+			lantiq,pull = <0>;
+			lantiq,output = <1>;
+		};
+	};
+	pins_spi_default: pins_spi_default {
+		spi_in {
+			lantiq,groups = "spi_di";
+			lantiq,function = "spi";
+		};
+		spi_out {
+			lantiq,groups = "spi_do", "spi_clk", "spi_cs4";
+			lantiq,function = "spi";
+			lantiq,output = <1>;
+		};
+	};
+};
+
+&spi {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pins_spi_default>;
+
+	status = "ok";
+
+	m25p80@4 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "winbond,w25q64", "jedec,spi-nor";
+		reg = <4 0>;
+		linux,modalias = "m25p80", "w25q64";
+		linux,mtd-name = "ltq_sflash";
+		spi-max-frequency = <10000000>;
+		m25p,fast-read;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				reg = <0x0 0x20000>;
+				label = "uboot";
+				read-only;
+			};
+
+			partition@20000 {
+				reg = <0x20000 0x10000>;
+				label = "gphyfirmware";
+				read-only;
+			};
+
+			partition@30000 {
+				reg = <0x30000 0x190000>;
+				label = "kernel";
+				read-only;
+			};
+
+			partition@1c0000 {
+				reg = <0x1c0000 0x80000>;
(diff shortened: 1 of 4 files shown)


target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+		led-dsl = &dsl;
+		led-internet = &lan_g;
+	};
+
+	memory@0 {
+		reg = <0x0 0x3f00000>;
+	};
+
+	ltq_swreset {
+		compatible = "lantiq,ltq_swreset";
```
mkresin: We do not have a driver that uses this compatible string, please drop the node.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+		status = "ok";
+	};
+
+	gphy-xrx200 {
+		compatible = "lantiq,phy-xrx200";
+		firmware = "lantiq/vr9_phy22f_a2x.bin";
+		phys = [ 00 01 ];
+	};
+
+//	gpio-keys-polled {
```
mkresin: What is wrong with the reset key? Why is it comment. We usually do not keep disabled code in the dts.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+			gpios = <&gpio 7 GPIO_ACTIVE_HIGH>;
+			linux,code = <0x198>;
+		};*/
+//	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power_a: power_amber {
+			label = "power_amber";
```
mkresin: The name has to be board:coluor:function. Please follow this convention. Please have a look at the other lantiq dts files.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+			lantiq,pins = "io42";
+			lantiq,pull = <0>;
+			lantiq,open-drain = <0>;
+			lantiq,output = <1>;
+		};
+		exin {
+			lantiq,groups = "exin3";
+			lantiq,function = "exin";
+		};
+		/*stp {
```
mkresin: please drop all disabled/commented nodes here.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+			lantiq,groups = "mdio";
+			lantiq,function = "mdio";
+		};
+		phy-rst {
+			lantiq,pins = "io42";
+			lantiq,pull = <0>;
+			lantiq,open-drain = <0>;
+			lantiq,output = <1>;
+		};
+		exin {
```
mkresin: what is connected to the exin pin group?

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+			lantiq,groups = "led";
+			lantiq,function = "led";
+		};*/
+		conf_in {
+			lantiq,pins = "io39"; /* exin3 */
+			lantiq,pull = <2>;
+		};
+		conf_out {
+			lantiq,pins =  // "io21", "io38", /* pcie */
+//					"io12", "io15", /* annexa, annexb */
```
mkresin: Please drop the comments. But this one is rather interesting. Can the same board be used on annex a and annex b lines? Is annex a/b compatibility enabled by pulling these gpios?

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+	pinctrl-0 = <&pins_spi_default>;
+
+	status = "ok";
+
+	m25p80@4 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "winbond,w25q64", "jedec,spi-nor";
+		reg = <4 0>;
+		linux,modalias = "m25p80", "w25q64";
```
mkresin: drop this line, starting with kernel 4.4 it isn't used any longer.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+&spi {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pins_spi_default>;
+
+	status = "ok";
+
+	m25p80@4 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "winbond,w25q64", "jedec,spi-nor";
```
mkresin: just `compatible = "jedec,spi-nor";`

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+		reg = <4 0>;
+		linux,modalias = "m25p80", "w25q64";
+		linux,mtd-name = "ltq_sflash";
```
mkresin: this property shouldn't be required; drop it.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+				read-only;
+			};
+
+			partition@20000 {
+				reg = <0x20000 0x10000>;
+				label = "gphyfirmware";
+				read-only;
+			};
+
+			partition@30000 {
```
mkresin: please merge kernel, firmware and rootfs into a single big firmware partition. Remove the read-only property from that partition, otherwise neither the config can be saved nor an update performed.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+				read-only;
+			};
+
+			partition@240000 {
+				reg = <0x240000 0x500000>;
+				label = "rootfs";
+				read-only;
+			};
+
+			partition@imgalias {
```
mkresin: have you checked which of the partitions are really required to revert to the netgear firmware? I would like to use as much as possible flash space for LEDE. Might be that some of the following partitions are (re)created by the netgear firmware on first boot.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+		};
+	};
+};
+
+&pcie0 {
+	status = "disabled";
+};
+
+&eth0 {
+//	interrupts = <73 72 95 96>;
```
mkresin: once more, please drop the comments.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+
+	lan: interface@0 {
+		compatible = "lantiq,xrx200-pdi";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+//		reg = <2>;
+		mtd-mac-address = [ 00 11 22 33 44 55 ];
+//		mtd-mac-address = [ 00 11 22 33 44 57 ];
+//		mtd-mac-address = <&romfile 0xf100>;
```
mkresin: is it possible to get the mac address from the art partition? The art partition might contain a mac address that only needs to incremented/decremented. Please have a look at the other dts files for examples.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
+#include "vr9.dtsi"
```
mkresin: please get rid of this file and copy the content to DM200.dts

target/linux/lantiq/image/netgear.mk
```diff
@@ -0,0 +1,38 @@
...
+  IMAGES := sysupgrade.bin factory.img
+  IMAGE/default := append-kernel | append-hdr-rootfs | pad-rootfs | append-metadata
+  IMAGE/sysupgrade.bin := $$(IMAGE/default) | check-size $$$$(IMAGE_SIZE)
+  IMAGE/factory.img := $$(IMAGE/default) | netgear-dni | check-size $$$$(IMAGE_SIZE)
+endef
+
+define Device/DM200
+  $(Device/lantiqNetgear)
+  DEVICE_TITLE := Netgear DM200
+  DEVICE_PACKAGES := kmod-spi-lantiq-ssc
```
mkresin: we don't have such a package, drop it!

target/linux/lantiq/image/netgear.mk
```diff
@@ -0,0 +1,38 @@
...
+          -n 'LEDE RootFS' \
+          -d $(IMAGE_ROOTFS) $@.tmp
+  cat $@.tmp >> $@
+  rm $@.tmp
+endef
+
+define Device/lantiqNetgear
+  KERNEL := kernel-bin | append-dtb | lzma | uImage lzma | pad-offset 4 0 | pad-to 2162624
+  IMAGES := sysupgrade.bin factory.img
+  IMAGE/default := append-kernel | append-hdr-rootfs | pad-rootfs | append-metadata
```
mkresin: I'm not yet sure what append-hdr-rootfs really does, but it look similar to the netgear fake rootfs header. Does using [append-uImage-fakeroot-hdr](https://github.com/lede-project/source/commit/7d00cfe9bb693e376ac9d035e13f8ce8a5ff572c) work as well?

mkresin: Please add the  Device/lantiqNetgear code to the Device/DM200 define. For now we only have a single netgear board and it doesn't make much sense to split this code.

It might be even better to move the whole code to image/Makefile as long as we only have a single device using the code.

target/linux/lantiq/dts/EASY220.dtsi
```diff
@@ -0,0 +1,295 @@
...
+		};*/
+		/*pci {
+			lantiq,groups = "gnt1", "req1";
+			lantiq,function = "pci";
+		};*/
+		/*led {
+			lantiq,groups = "led";
+			lantiq,function = "led";
+		};*/
+		conf_in {
```
mkresin: what is connected to gpio39?

mkresin:
Please squash your commits!  Please Include in your commit message short description of the hardware and how to install LEDE on it. Have a look at the [recent additions](https://git.lede-project.org/?p=source.git&a=search&h=HEAD&st=commit&s=add+support+for) for some examples.

For more details, please have a look at https://lede-project.org/submitting-patches

zorun:
Hi,

This is duplicate work with respect to what @tomjnixon did: https://github.com/tomjnixon/lede-source/tree/add-netgear-dm200

@thomjnixon do you plan to send your patches? Before a fourth person independently reimplements support for this device :/

target/linux/lantiq/image/netgear.mk
zorun: It's actually a `netgear-dni` header.

zorun:
@mkresin : as far as I can see, the code from @tomjnixon (which I helped/reviewed and tested) addresses all your comments.  Let's wait for Tom to reply, otherwise I will send the patches myself.

tomjnixon:
I had held off on submitting a PR for the above branch, as I have had trouble with reliability of the flash, which often causes sysupgrades to fail.

This doesn't seem to have anything to do with the speed - 10MHz is conservative, and going lower doesn't make it more reliable. I've always had success writing to the flash from uboot and from the vendor firmware which implies that it's a software bug.

It wouldn't surprise me if the flash chip on my particular device has got damaged or worn (its had a hard life), so if sysupgrades work reliably for other people then we shouldn't worry too much.

The branch mentioned above does resolve a lot of these issues, I'd be happy to submit a PR if that would be helpful.

zorun:
@tomjnixon yes, please rebase and submit :)

Better to have it in, rather than keeping the current statu quo.  I haven't encountered the flash issue on my device.

mkresin:
I will go with #1170. It has way cleaner and doesn't have any of the outlined issues.

moriel5:
Hi, I wanted to point out that NETGEAR has finally released the sources for version 1.0.44.

Additionaly, I was wondering whether G.INP and vectoring  support could be added independently, or whether it is better go wait for NETGEAR to release an update (if it ever arrives) with aforementioned support.

PR #1164, opened 2017-06-12
[RFC] package/base-files: force sysupgrade file location to tmp

feckert:
If the `sysupgrade` is started with a file which is not located in the RAM a sysupgrade will brick the device.
Force user to start a `sysupgrade` with a file which is located in the `/tmp` folder

### DIFF
--- a/package/base-files/files/sbin/sysupgrade
+++ b/package/base-files/files/sbin/sysupgrade
@@ -247,6 +247,12 @@ else
 	rm -f /tmp/sysupgrade.always.overwrite.bootdisk.partmap
 fi
 
+IMAGE_FILE=$(readlink -f ${ARGV})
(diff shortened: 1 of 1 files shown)


KanjiMonster:
This is already checked and fixed up at https://github.com/lede-project/source/blob/master/package/base-files/files/sbin/sysupgrade#L193:
<pre>
IMAGE="$(readlink -f "$IMAGE")"

case "$IMAGE" in
	'')
		echo "Image file not found."
		exit 1
		;;
	/tmp/*)	;;
	*)
		v "Image not in /tmp, copying..."
		cp -f "$IMAGE" /tmp/sysupgrade.img
		IMAGE=/tmp/sysupgrade.img
		;;
esac
</pre>

feckert:
@KanjiMonster 
Sorry for the noise. Have not seen that this was merged already.

PR #1166, opened 2017-06-13
rpcd: delay brute force attacks on login failure

feckert:
delay rpcd session brute force attacks by 3 second


jow-:
NACK, this would halt the entire rpcd process. What you would need to implement is a kind of ephemeral blacklist in memory which causes login attempts to get rejected for a given session / remote ip / username after a certain amount of failures has been reached.

Your proposed implementation would introduce an easy to exploit DoS vector

feckert:
Ok thanks for the hint, i will have a to think about it.
If i do this is this worth do merge?

jow-:
Certainly, if it is configurable.

PR #1170, opened 2017-06-16
lantiq: add Netgear DM200 support

tomjnixon:
Add support for Netgear DM200; see commit for details.

This was prompted by #1163; I believe this patch addresses all the issues raised in that PR, though of course there may be additional issues spotted. Sorry for not submitting this sooner...

As mentioned in #1163, I have had issues with the flash chip on my device, which has made it hard to test. It seems to run fine, though, so I don't expect any issues.

Thanks,

Tom

### DIFF
--- a/target/linux/lantiq/dts/DM200.dts
+++ b/target/linux/lantiq/dts/DM200.dts
@@ -1,0 +1,204 @@
+/dts-v1/;
+
+#include "vr9.dtsi"
+
+#include <dt-bindings/input/input.h>
+
+/ {
+	model = "DM200 - Netgear DM200";
+
+	chosen {
+		bootargs = "console=ttyLTQ0,115200";
+	};
+
+	aliases {
+		led-boot = &power_green;
+		led-failsafe = &power_amber;
+		led-running = &power_green;
+
+		led-dsl = &dsl_green;
+	};
+
+	memory@0 {
+		reg = <0x0 0x4000000>;
+	};
+
+	fpi@10000000 {
+		gpio: pinmux@E100B10 {
+			pinctrl-names = "default";
+			pinctrl-0 = <&state_default>;
+
+			state_default: pinmux {
+				mdio {
+					lantiq,groups = "mdio";
+					lantiq,function = "mdio";
+				};
+				conf_out {
+					lantiq,pins = "io12", "io15";  // annexa, annexb
+					lantiq,open-drain;
+					lantiq,pull = <0>;
+					lantiq,output = <1>;
+				};
+			};
+
+			pins_spi_default: pins_spi_default {
+				spi_in {
+					lantiq,groups = "spi_di";
+					lantiq,function = "spi";
+				};
+				spi_out {
+					lantiq,groups = "spi_do", "spi_clk", "spi_cs4";
+					lantiq,function = "spi";
+					lantiq,output = <1>;
+				};
+			};
+		};
+
+		pcie@d900000 {
+			status = "disabled";
+		};
+	};
+
+	gphy-xrx200 {
+		compatible = "lantiq,phy-xrx200";
+		firmware = "lantiq/vr9_phy11g_a2x.bin";
+		phys = [ 01 ];
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <100>;
+		reset {
+			label = "reset";
+			gpios = <&gpio 7 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power_amber: power_amber {
+			label = "dm200:amber:power";
+			gpios = <&gpio 0 GPIO_ACTIVE_HIGH>;
+		};
+		power_green: power_green {
+			label = "dm200:green:power";
+			gpios = <&gpio 1 GPIO_ACTIVE_HIGH>;
+		};
+
+		lan_amber {
+			label = "dm200:amber:lan";
+			gpios = <&gpio 33 GPIO_ACTIVE_HIGH>;
+		};
+		lan_green {
+			label = "dm200:green:lan";
+			gpios = <&gpio 11 GPIO_ACTIVE_HIGH>;
+		};
+
+		dsl_amber {
+			label = "dm200:amber:dsl";
+			gpios = <&gpio 8 GPIO_ACTIVE_HIGH>;
+		};
+		dsl_green: dsl_green {
+			label = "dm200:green:dsl";
+			gpios = <&gpio 36 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&spi {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pins_spi_default>;
+
+	status = "ok";
+
+	m25p80@4 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <4 0>;
+		spi-max-frequency = <10000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				reg = <0x0 0x20000>;
+				label = "uboot";
+				read-only;
+			};
+
+			partition@20000 {
+				reg = <0x20000 0x10000>;
+				label = "gphyfirmware";
+				read-only;
+			};
+
+			partition@30000 {
+				reg = <0x30000 0x7b0000>;
+				label = "firmware";
+			};
+
+			partition@7e0000 {
+				reg = <0x7e0000 0x10000>;
+				label = "sysconfig";
+				read-only;
+			};
+
+			partition@7f0000 {
+				reg = <0x7f0000 0x2000>;
+				label = "ubootconfig";
+				read-only;
+			};
+
+			partition@7f2000 {
+				reg = <0x7f2000 0x1000>;
+				label = "ART";
+				read-only;
+			};
+
+			partition@7f3000 {
+				reg = <0x7f3000 0x1000>;
+				label = "pot";
+				read-only;
+			};
+
+			partition@7f4000 {
+				reg = <0x7f4000 0xc000>;
+				label = "ret";
+				read-only;
+			};
+		};
+	};
+};
+
+&eth0 {
+	lan: interface@0 {
+		compatible = "lantiq,xrx200-pdi";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+
+		ethernet@4 {
+			compatible = "lantiq,xrx200-pdi-port";
+			reg = <4>;
+			phynmode0 = "gmii";
+			phy-handle = <&phy13>;
+		};
+	};
+
+	mdio@0 {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		compatible = "lantiq,xrx200-mdio";
+		phy13: ethernet-phy@13 {
+			reg = <0x13>;
+			compatible = "lantiq,phy11g", "ethernet-phy-ieee802.3-c22";
+		};
+	};
+};
--- a/target/linux/lantiq/base-files/etc/board.d/01_leds
+++ b/target/linux/lantiq/base-files/etc/board.d/01_leds
@@ -50,6 +50,9 @@ ARV7519RW22)
 BTHOMEHUBV5A)
 	ucidef_set_led_default "dimmed" "dimmed" "dimmed" "0"
 	;;
+DM200)
+	ucidef_set_led_netdev "lan" "lan" "dm200:green:lan" "eth0"
+	;;
 FRITZ7320)
 	ucidef_set_led_netdev "wifi" "wifi" "fritz7320:green:wlan" "wlan0"
 	;;
--- a/target/linux/lantiq/base-files/etc/board.d/02_network
+++ b/target/linux/lantiq/base-files/etc/board.d/02_network
@@ -100,6 +100,12 @@ DGN3500*)
 		"0:lan:4" "1:lan:3" "2:lan:2" "3:lan:1" "5t@eth0"
 	;;
 
+DM200)
+	lan_mac=$(mtd_get_mac_binary ART 0)
+	wan_mac=$(macaddr_add "$lan_mac" 1)
+	ucidef_set_interface_lan 'eth0'
+	;;
+
 EASY80920NAND|EASY80920NOR)
 	lan_mac=$(mtd_get_mac_ascii uboot_env ethaddr)
 	wan_mac=$(macaddr_add "$lan_mac" 1)
--- a/target/linux/lantiq/image/Makefile
+++ b/target/linux/lantiq/image/Makefile
@@ -513,6 +513,17 @@ define Device/BTHOMEHUBV5A
 endef
 TARGET_DEVICES += BTHOMEHUBV5A
 
+define Device/DM200
+  IMAGES := sysupgrade.bin factory.img
+  IMAGE/factory.img := $$(IMAGE/sysupgrade.bin) | netgear-dni
+  IMAGE_SIZE := 7872k
+  DEVICE_TITLE := Netgear DM200
+  NETGEAR_BOARD_ID := DM200
+  NETGEAR_HW_ID := 29765233+8+0+64+0+0
+endef
+DEVICE_VARS += NETGEAR_BOARD_ID NETGEAR_HW_ID
+TARGET_DEVICES += DM200
+
 define Device/EASY80920NAND
   $(Device/lantiqFullImage)
   IMAGE_SIZE := 64512k


victoredwardocallaghan:
@tomjnixon this looks good, I also had issues with flash but with my partition table the issues stopped, it could be a issue in the driver with big regions I'll test your DTS tonight. Otherwise LGTM !

It would be nice to have a shout out in your commit message for others who co-authored almost equivalent ports seemingly independently at around the exact same time.

P.S., did you also test ADSL2+ as well as VDSL2+ ?
Kindest Regards, Edward.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
+/dts-v1/;
+
+#include "vr9.dtsi"
+
+#include <dt-bindings/input/input.h>
+
+/ {
+
```
mkresin: Drop this blank line.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
...
+    model = "DM200 - Netgear DM200";
+
+    chosen {
+        bootargs = "console=ttyLTQ0,115200";
+
```
mkresin: Drop this blank line.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
+/dts-v1/;
```
mkresin: Please use tabs instead of spaces in that file.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
...
+
+    };
+
+    aliases {
+        led-boot = &power_green;
+        led-failsafe = &power_amber;
+        led-running = &power_green;
+
+        led-internet = &dsl_green;
+        led-dsl = &dsl_amber;
```
mkresin: please drop this line. As far as I can see, the dsl LED is a dual colour LED and there are no distinct LEDs for internet and dsl state.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
...
+        power_amber: power_amber {
+            label = "dm200:amber:power";
+            gpios = <&gpio 0 GPIO_ACTIVE_HIGH>;
+        };
+        power_green: power_green {
+            label = "dm200:green:power";
+            gpios = <&gpio 1 GPIO_ACTIVE_HIGH>;
+        };
+
+        lan_amber: lan_amber {
```
mkresin: please drop the alias (lan_amber: and so on) from leds which are not referenced in the aliases node.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,208 @@
...
+            pinctrl-names = "default";
+            pinctrl-0 = <&state_default>;
+
+            state_default: pinmux {
+                mdio {
+                    lantiq,groups = "mdio";
+                    lantiq,function = "mdio";
+                };
+                conf_out {
+                    lantiq,pins = "io12", "io15";  // annexa, annexb
```
mkresin: Does this mux enables annex a/annex b for the board? If for example io12 is not pulled, annex a does not work?

mkresin:
What kind of issues do you have with the SPI flash chip?

What erase size is used (`cat /proc/mtd`)? If the erase size is for example 0x20000, the firmware partition has to start at a multiple of 0x20000 and the size has to be a multiple of 0x20000 (0x7A0000 for example). It is only relevant for partition to which should be written.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Done.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Done.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Done.

target/linux/lantiq/dts/DM200.dts
tomjnixon: True, done. My intention was to show both the dsl state and the internet state, but I don't think these leds are that useful anyway.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Done.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Hmm, this commend is from the vendor dts, but I'm not sure it's correct. On my connection (VDSL2, annex b), removing the io15 pin had no effect, while removing the io12 pin caused the connection to not come up.

tomjnixon:
Hi, thanks for taking a look at this.

> It would be nice to have a shout out in your commit message for others who
> co-authored almost equivalent ports seemingly independently at around the
> exact same time.

Good point, done.

> P.S., did you also test ADSL2+ as well as VDSL2+ ?

I've tested VDSL2, and @zorun tested ADSL2+. ADSL2+ works, but only if set to PTM mode, which is odd.

> What kind of issues do you have with the SPI flash chip?

Timeouts, especially while writing large amounts of data (sysupgrades, building the filesystem), leading to corruption.

I only see this when writing from LEDE (uboot and vendor firmware are fine), and slowing down the clock does not help. Given that others don't have a problem, I think it's just the flash chip on my device getting worn out.

> What erase size is used (cat /proc/mtd)?

0x10000, so the current partition layout should be fine.

Thanks,

Tom

victoredwardocallaghan:
Hey Tom,

I was able to flash this firmware just fine however I am having trouble getting ADSL2+ to work, `/etc/init.d/dsl_control status` oddly seems to suggest no signal so I will have to investigate today as it seems it should just be working.

> I've tested VDSL2, and @zorun tested ADSL2+. ADSL2+ works, but only if set to PTM mode, which is odd.

This is because `ltq_ptm_vr9` and `ltq_atm_vr9` cannot be modloaded at the same time due to a shared IRQ issue:
```
[  145.553808] genirq: Flags mismatch irq 96. 00000000 (atm_mailbox_isr) vs. 00000000 (ptm_mailbox_isr)
[  145.561614] IRQ may be occupied by other driver, please reconfig to disable it.
[  145.569488] ifxmips_atm: ATM init failed
[  145.572868] atm: probe of 10000000.fpi:ppe@E234000 failed with error -16
```

Cheers,
Edward.

tomjnixon:
Hey, thanks for trying it out.

> I was able to flash this firmware just fine however I am having trouble getting ADSL2+ to work, /etc/init.d/dsl_control status oddly seems to suggest no signal so I will have to investigate today as it seems it should just be working.

That's disappointing; does your port work? I can't see much that would make them behave differently.

> ltq_ptm_vr9 and ltq_atm_vr9 cannot be modloaded at the same time

Hmm, in theory the code should prevent this, but unloading `ltq_atm_vr9` fails. If `br2684ctl` is stopped while `dsl_control` is restarted, this appears to work properly, so it should be fixable.

victoredwardocallaghan:
Tom,

I did some further investigation today and found that in ATM mode the modem is just cycling around with:
"Line State:                               DOWN [0x1: exception]"

I have no idea exactly what kind of "exception" that is? A hw exception perhaps?

mkresin:
Whether ATM or PTM is used should be unrelated to the line sync. The transfer mode is one layer more up. I'm quite sure if something is wrong here, it is something different.

If I need to guess, I would assume that  io12/gpio 12 need to be pulled for adsl annex a and io15/gpio15 for adsl annex b. But as written, it is a wild guess.

@victoredwardocallaghan would you please try the following:

```
# get the full vdsl_cpe_control commandline
cat /proc/$(pidof vdsl_cpe_control)/cmdline | xargs -0

# stop vdsl_cpe_control
/etc/init.d/dsl_control stop
```

start `vdsl_cpe_control` manually with params you got with the above command + the --console parameter to get error messages. It _should_ work this way.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,204 @@
...
+	chosen {
+		bootargs = "console=ttyLTQ0,115200";
+	};
+
+	aliases {
+		led-boot = &power_green;
+		led-failsafe = &power_amber;
+		led-running = &power_green;
+
+		led-internet = &dsl_green;
```
mkresin: I've pushed a commit a few minutes ago, which required a change here. Please use the following:

```
led-dsl = &dsl_green;
```

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,204 @@
...
+				reg = <0x7f4000 0xc000>;
+				label = "ret";
+				read-only;
+			};
+		};
+	};
+};
+
+&eth0 {
+			lan: interface@0 {
```
mkresin: please indent this block correctly. remove two of the leading tabs for the whole block.

tomjnixon:
It might also be worth trying [this branch](https://github.com/tomjnixon/lede-source/tree/add-netgear-dm200), which is based on an older LEDE source and has been tested successfully with ADSL.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Yeah, this is what I meant to do the first time round. Done.

target/linux/lantiq/dts/DM200.dts
tomjnixon: Whoops, done.

tomjnixon:
> Would you please leave a note as soon as you are the opinion it is ready to merge.

Well, it works for me; it looks like there is an issue with ADSL, but I can't really help with that as i have no way of testing. I'd suggest either leaving this open until that is resolved, or merging now and handling that in a separate issue.

victoredwardocallaghan:
@mkresin

> DSL_CPE: Wrong combination of DSL PHY Firmware and hybrid type used! Please change one of it.

Not sure what this means as I ensured that lantiq-vrx200-a.bin points exactly to the shipped fw bin just in case..

@tomjnixon

I agree, merge this as its already had enough bike-shedding. We can fix up any corner cases as we go.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,204 @@
...
+		};
+	};
+
+	mdio@0 {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		compatible = "lantiq,xrx200-mdio";
+		phy13: ethernet-phy@13 {
+			reg = <0x13>;
+			compatible = "lantiq,phy11g", "ethernet-phy-ieee802.3-c22";
```
victoredwardocallaghan: this should be `lantiq,phy22f` btw.

target/linux/lantiq/dts/DM200.dts
```diff
@@ -0,0 +1,204 @@
...
+	lan: interface@0 {
+		compatible = "lantiq,xrx200-pdi";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+
+		ethernet@4 {
+			compatible = "lantiq,xrx200-pdi-port";
+			reg = <4>;
+			phynmode0 = "mii";
```
victoredwardocallaghan: and this as `gmii` I believe.

target/linux/lantiq/dts/DM200.dts
mkresin: NAK. phy11g is right here. But you are using the 100Mbit firmware instead of the 1Gbit firmware.

Please change your gphy-xrx200 node instead:

```
gphy-xrx200 {
	compatible = "lantiq,phy-xrx200";
	firmware = "lantiq/vr9_phy11g_a2x.bin";
	phys = [ 00 01 ];
};
```

Since the DM200 has only one LAN port exposed, `phys = [ 00 ];` should be enough.

victoredwardocallaghan:
@mkresin and I did some investigation getting ADSL2+ with ATM working.

With the following patch: https://gist.github.com/victoredwardocallaghan/fdbf30bd92db6c61ad923eb82b93396a

we were able to identify that Annex A and B are XOR. The following configuration correctly sets up the frontend for me with Annex A, ADSL2+ ATM:

```
root@LEDE:/# cat /sys/class/gpio/annexa/value 
1
root@LEDE:/# cat /sys/class/gpio/annexb/value 
0
```

from boot, running `echo 0 > /sys/class/gpio/annexb/value` gets us into this state with the above patch.

What we would like to know is, what happens when both are pulled high at the same time on VDSL?

tomjnixon:
@victoredwardocallaghan That's great news!

I did have a go with changing these gpios yesterday, and it seemed to behave opposite to the way you said, however I noticed that in your dts (and the netgear one), these pins are defined with `lantiq,open-drain = <0>`, while in my dts they are defined as `lantiq,open-drain`; I expect this explains it.

I try this out on my line and address the other issues this evening or tomorrow morning.

victoredwardocallaghan:
@tomjnixon I think the port should be merged as it is now and we can coordinate additional fixes in a further pull request. This will avoid others wasting time trying to port again as well as get more eyes on it in-tree.

tomjnixon:
The suggested changes have been made, only change was that i had to use `phys = [ 01 ]` instead of 0.

@mkresin Please merge this, we can follow up the dsl changes later.

target/linux/lantiq/dts/DM200.dts
zorun: Actually, the LAN port of the DM200 is only 100Mbit, so it makes sense to use the 100Mbit firmware.  I'm not sure it would even work with the 1Gbit firmware?

victoredwardocallaghan:
@tomjnixon can you add the following (feel free to reword as needed) to the commit message so @mkresin can merge this:

 * A more robust way to flash the factory.img image over the spi is to hold the reset button down upon first boot until uboot flashes "Upgrade mode" in minicom - set your machine with ip 192.168.0.1 and use tftp to put the image over the wire to the modem and uboot will handle the rest of the flashing process.

 * Currently ADSL2+ Annex-A is broken without pulling gpio15 low and gpio12 high manually.

zamaudio:
@tomjnixon bump^ can't wait to use this!

tomjnixon:
sorry for being slow...

@victoredwardocallaghan I added your note about ADSL, but I didn't quite understand the upgrade mode bit; do you mean that running something like `tftp something-factory.img` will load `something-factory.img` over tftp and perform the upgrade? I don't have access to my device at the moment or I'd try it out.

victoredwardocallaghan:
@mkresin time to stop bikesheding this, please merge this ASAP! I don't think people appreciate spending volumes of personal time on code that sits for months not getting merged.

tomjnixon:
As an aside, I went spelunking in the original firmware and found the code that sets the annex GPIOs. The behaviour is exactly as expected -- in annex a mode, annex_a=1 and annex_b=0, and in annex b mode, annex_a=0 and annex_b=1.

I've implemented this behaviour on this branch:

https://github.com/tomjnixon/lede-source/tree/dm200-annex-switch

This works for me (on annex b vdsl); if this works for others and the style/structure is suitable, then we could declare this to be fully working.

mkresin:
@victoredwardocallaghan I don't see any bikesheding here. I just don't like to commit something half working (upgrade from the netgear ui for example) which will cause unnecessary bug reports and/or questions in the forum. Especially if we have a working alternative. I already agreed that I'm fine with (only) documenting the GPIO thingy to get Annex A working.

Please keep in mind, that all the work on LEDE is done in our free time. No one gets paid for it. If I do not react within your expected time frame it is most likely related to the fact that something more important (reallife!) popped up.

@tomjnixon According to Tom the well known [tftp recovery](https://kb.netgear.com/22688/How-to-upload-firmware-to-a-NETGEAR-router-using-TFTP) is supported by the DM200 and doesn't show the weird spi timeout/write errors you reported.

Therefore I would like see this method mentioned/described in the commit message and a warning about doing the upgrade from the stock firmware. I guess the "install via uboot/tftp" can be dropped from the commit message afterwards.

tomjnixon:
@mkresin Thanks for looking at this again.

I'm not aware of any issue with upgrading from the netgear UI? All the times I've tested this it has worked fine, even on my dodgy flash chip. The only issue I've seen is when writing to the flash from LEDE, and only on my device; this works fine for everyone else.

Anyway, I'll document the recovery method tomorrow, as having a method that doesn't involve a serial console is useful.

mkresin:
> I'm not aware of any issue with upgrading from the netgear UI? All the times I've tested this it has worked fine, even on my dodgy flash chip. The only issue I've seen is when writing to the flash from LEDE, and only on my device; this works fine for everyone else.

Oops! I got it wrong. Sorry for the noise.

victoredwardocallaghan:
@tomjnixon hey, can you add commit e73b069 and f558156 from your `dm200-annex-switch` branch into this merge request so we can get them all merged into master please? This completes a fully working port!

zorun:
I tried the new version (with `dm200-annex-switch`) rebased on top of 7887a46f45, and it worked fine on my ADSL line (in PTM mode: it seems that my ISP supports this even for ADSL lines).  I installed the new version by sysupgrading from an old LEDE build from a few months ago.

But when I tried to sysupgrade again from the new image, it failed: the device did not boot correctly, it was simply waiting for an image to be uploaded through TFTP. Unfortunately, because of 30f61a34b4cfd2c676fea4a919e089d6a77254e9, there is not much information about what happened:

```
root@LEDE:~# sysupgrade /tmp/lede-lantiq-xrx200-DM200-squashfs-sysupgrade.bin 
Saving config files...
Commencing upgrade. All shell sessions will be closed now.
Connection to fe80::a02:8eff:fe93:bd63%eth0 closed by remote host.
```

Maybe this is the same issue with the flash as reported by @tomjnixon .  In any case it's not very hard to fix, uploading an image through TFTP works fine.  Also, I can confirm that upgrading from the stock UI works. 
 Let's merge this!

tomjnixon:
Hm, I also saw an error when sysupgrading; for me it worked successfully, but I saw these lines in the kernel log while it was building the JFFS2 filesystem:
```
[   22.712822] spi-lantiq 1e100800.spi: transmit overflow error
[   22.717107] spi-lantiq 1e100800.spi: transfer failed
[   22.722097] spi_master spi0: failed to transfer one message from queue
[   22.728688] jffs2: Write of 43 bytes at 0x00008500 failed. returned 0, retlen 0
[   22.735911] jffs2: Not marking the space at 0x00008500 as dirty because the flash driver returned retlen zero
```
This is on a brand new device...

mkresin:
>[   22.712822] spi-lantiq 1e100800.spi: transmit overflow error
>[   22.717107] spi-lantiq 1e100800.spi: transfer failed
>[   22.722097] spi_master spi0: failed to transfer one message from queue
>[   22.728688] jffs2: Write of 43 bytes at 0x00008500 failed. returned 0, retlen 0
>[   22.735911] jffs2: Not marking the space at 0x00008500 as dirty because the flash driver returned retlen zero

The only explanation I can provide is that either SPI works slightly different for the vrx220 or we have a bug in the SPI driver that is triggered by the image size.

You can try to to switch to kernel 4.9 by changing the `KERNEL_PATCHVER` in `target/linux/lantiq/Makefile` to 4.9. Kernel 4.9 contains a slightly different SPI driver.

mkresin:
Merged, thanks!

tomjnixon:
Thanks all!

@mkresin I had a go with 4.9 and haven't seen any problems, so you might be right about the different SPI drivers.

PR #1173, opened 2017-06-16
Add iptables-mod-rpfilter package

alinnastac:
Unlike /proc/sys/net/ipv4/conf/INTF/rp_filter flag, rule
`iptables -t raw -I PREROUTING -m rpfilter --invert -j DROP`
prevents conntrack table to become full when a packet flood with
randomly selected source IP addresses is received from the lan side.

Signed-off-by: Alin Nastac <alin.nastac@gmail.com>


dedeckeh:
Merged into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=638df1539dbbeb22b04373d0998e58ede1d610c0); thanks

PR #1177, opened 2017-06-18
ar71xx: Add support for D-Link DAP-1330 a1 

hitech95:
D-Link DAP-1330 rev. A1 is a wall-plug N300 Wi-Fi range extender,
based on Qualcomm/Atheros QCA9533 v2.

Short specification:

- 650/393/216 MHz (CPU/DDR/AHB)
- 1x 10/100 Mbps Ethernet
- 64 MB of RAM (DDR1)
- 8 MB of FLASH
- 2T2R 2.4 GHz
- 2x external antennas
- 6x LED (2 are bi-color), 2x button
- 4 pin on pcb (looking from eth port and from left to right tx,rx,gnd,vcc)

Flash instruction: use "factory" image directly in vendor GUI.

This device has a recovery system that assign the ip addr of env
variable "serverip" via dhcp to a pc, and the "ipaddr" value to itself.
The recovery it's triggered by a not bootable firmware,
or pressing the reset button during the bootloader startup (first 30 seconds).
The recovery uses a http page to restore the firmware, and it's checking
the firmware image header, so use the "factory" image to
restore or the original firmware.

You can access vendor firmware over serial using:
- login: root
- password: linuxrocks

Image was tested only in EU version of the device, but should work
also with the same device version sold in other countries.


target/linux/ar71xx/base-files/lib/upgrade/platform.sh
```diff
@@ -647,6 +647,7 @@ platform_check_image() {
 		return 0;
 		;;
 	# these boards use metadata images
+	dap-1330-a1|\
```
pepe2k: We don't use metadata images by default in ar71xx target, please check this.

hitech95: I moved the board identifier down so I don't have to add another return type.
If needed I can move it.

hitech95:
@pepe2k Just fixed that, moved to the standard magic check section.

stintel:
Ideally the current mkdapimg.c should be modified to support both v2 and v3 image header. IIRC it seemed doable at the time I added mkdapimg.c, but I did not have any hardware to test it on. I'm not going to block this PR for this reason, but I think it would be cleaner in a single utility.

hitech95:
@stintel The current mkdapimg is used on ramips and is different from the one you added.
Currently i have no idea of what is the revision of the header that I have.

After talking in IRC asking for a suggestion about that the suggestion waas to create a new image generator.

One of the major differences is that the header must have all the fields even though there is the size in the first parameter. Another one is the order of the fields.

If you want I could merge the two into the same one. (I'll open another PR for that) The only problem is that I don't  have the hardware to test the other header.

stintel:
@hitech95 if it was suggested on IRC then just ignore my comment. But if you feel like merging the two, go ahead and ping me, I'll test it on one of my DAP-2695s.

hitech95:
@stintel, After trying to find the problem with an i2c driver on my mt7628 device I'll merge the two and I'll ping you back.

pepe2k:
@hitech95,

> @stintel, After trying to find the problem with an i2c driver on my mt7628 device I'll merge the two and I'll ping you back.

Should we then wait with this PR for your update of the firmware-utils part?

Cheers,
Piotr

hitech95:
The idea was to merge this and open an new PR for the firmware utils. I don't have a hardware to test the old version on the header so I have to wait @stintel to test that.

This delay would lock this PR for some days. (Right now I'm a bit busy making my own hardware)

cabsandy:
Hi there
Apologies if I shouldn't be posting here but I have a fresh DAP-1330 and are happy to try out any code on it, see if it works?

Thanks.

hitech95:
@pepe2k any news?
@cabsandy you can build an image from the repository linked with this PR. (Branch dap-1330)

loicgelle:
@hitech95 I built the image successfully and installed it on my DAP-1330. However I can't figure out a way to enter the failsafe mode using physical buttons. Any ideas about that?

hitech95:
It should work as other, just press a button during boot sequence to go on
Openwrt/LEDE failsafe mode.

To flash the image you can just use the web interface.

Il 04 set 2017 10:10 PM, "loicgelle" <notifications@github.com> ha scritto:

loicgelle:
@hitech95 Thanks :-) I managed to get the repeater working on DAP-1330 with your PR.

dimischiavone:
@hitech95 i built an image yesterday and installed it succesfully on my DAP-1330 so far everything works good but opkg. If you run opkg update and then opkg list it raises a segmentation fault:
[   86.320610] do_page_fault(): sending SIGSEGV to opkg for invalid read access from 00000000
[   86.329260] epc = 7701d1cc in libc.so[76faa000+92000]
[   86.334539] ra  = 0040d6d3 in opkg[400000+1b000]
so the only way to get more packages is to build them into the image itself for now.

hitech95:
@pepe2k 
@NeoRaider
Any news?

blogic:

hitech95:
PR moved to Openwrt repository:
https://github.com/openwrt/openwrt/pull/685

PR #1178, opened 2017-06-19
ramips: update mmc cd-polling mode to work on cd-broken device.

Vonger:
For cd-polling mmc device, we can not detect card insert by cd pin.
This patch make sure cd pin is not used in cd-polling mode to avoid
sd card randomly removed by floating cd pin.

Signed-off-by: Qin Wei <me@vonger.cn>


blogic:
@Vonger could you rebase the PR on top of v4.9 please

PR #1181, opened 2017-06-20
dnsmasq: fix to make DHCP forcing of options work

nigelterry:
A small error in commit 9412fc2 caused the ability to force the sending of dhcp options to a client to stop working. This patch fixes it.

Signed-of-by: Nigel J. Terry NigelTerry@SapphireWebServices.com

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -587,7 +587,7 @@ dhcp_option_add() {
 	config_get list_len "$cfg" dhcp_option_LENGTH
 
 	if [ -n "$list_len" ]; then
-		config_list_foreach "$cfg" dhcp_option dhcp_option_append "$networkid" "$force"
+		config_list_foreach "$cfg" dhcp_option dhcp_option_append "$option" "$networkid" "$force"
 	else
 		config_get dhcp_option "$cfg" dhcp_option
 


dedeckeh:
The patch is not correct as it will break the parsing as option is passed to dhcp_option_append by the config_list_foreach function (see functions.sh).
I've tested the vendor class classifier with the force option being set in combination with dhcp_options and it works for me.
So it would help if you specify which config is not working for you.

nigelterry:
Consider the following configuration, with the last line being the issue. This forces the sending of the MTU size for the interface to 1500.
```
config dhcp 'lan'
        option interface 'lan'
        option start '100'
        option force '1'
        option limit '100'
        option ra 'server'
        option dhcpv6 'server'
        list dns '2001:4860:4860::8888 '
        list dns '2001:4860:4860::8844'
        option ra_management '1'
        option leasetime '2h'
        list dhcp_option '26,1500'
```
Without the fix, this generates:
```
dhcp-range=lan,192.168.1.100,192.168.1.199,255.255.255.0,2h
dhcp-option=lan,26,1500
no-dhcp-interface=eth0
```
As the endpoint doesn't request option 26, then the regular dhcp-option=lan,26,1500 does not send the option in the dhcp response.
With the fix, this correctly generates: 
```
dhcp-range=lan,192.168.1.100,192.168.1.199,255.255.255.0,2h
dhcp-option-force=26,1500
no-dhcp-interface=eth0
```
Changing to dhcp_option_force causes it to send that option.

The fix works for me.

jow-:
Your fix is definitely not correct. The `config_list_foreach` procedure will call the given function name with the supplied additional arguments while passing the currently iterated list item as first argument.

So given a `list dhcp_option '26,1500'` and `config_list_foreach "$cfg" dhcp_option dhcp_option_append "$networkid" "$force"`, the list item iterator would invoke `dhcp_option_append "26,1500" "$networkid" "$force"`. After your change it would invoke `dhcp_option_append "26,1500" "" "$networkid" "$force"`.

Your change maybe appears to work because the networkid is mistreated as force flag by `dhcp_option_append` but actually it breaks network id handling and causes the actual force flag to be ignored while replacing the real network id with an always empty string since `$option` is not defined in the context you're using it in.

jow-:
Looking some more into it it appears to me that you're not actually fixing a bug, but that you want to introduce a change in behaviour, mainly you want to change the existing logic to emit `dhcp-option-force` instrad of `dhcp-option`.

To properly introduce support for forcing options, we would need to implement a `list dhcp_option_force` UCI config as we cannot simply change the semantics of the existing `list dhcp_option` setting.

jow-:
I made an example commit in my staging tree which introduces a `list dhcp_option_force` in `/etc/config/dhcp` - https://git.lede-project.org/?p=lede/jow/staging.git;a=commitdiff;h=9253a7f3799790a3c168b5edd88c90970e11dc9a

nigelterry:
Ok, I'm sure you are correct. Perhaps you can tell me what my configuration should look like to force the dhcp option to be sent to the client. I read the section below and it suggested to me that setting force would force the option to be sent, i.e. dhcp-option-force passed to dnsmasq not dhcp-option. Clearly I misunderstood.

From https://lede-project.org/docs/user-guide/dhcp_configuration
<div class="table-responsive sectionedit12"><table class="inline table table-striped table-bordered table-condensed">
	<thead>
	<tr class="row0">
		<th class="col0"> Name </th><th class="col1"> Type </th><th class="col2"> Required </th><th class="col3"> Default </th><th class="col4"> Description </th>
	</tr>
	</thead>
	<tr class="row1">
		<td class="col0"> <code>dhcp_option</code> </td><td class="col1"> list of strings </td><td class="col2"> no </td><td class="col3"> <em>(none)</em> </td><td class="col4"> Additional options to be added for this network-id.  If you specify this, you also need to specify the network-id. </td>
	</tr>
	<tr class="row2">
		<td class="col0"> <code>filename</code> </td><td class="col1"> string </td><td class="col2"> yes </td><td class="col3"> <em>(none)</em> </td><td class="col4"> The filename the host should request from the boot server. </td>
	</tr>
	<tr class="row3">
		<td class="col0"> <code>networkid</code> </td><td class="col1"> string </td><td class="col2"> no </td><td class="col3"> <em>(none)</em> </td><td class="col4"> The network-id these boot options should apply to. Applies to all clients if left unspecified. </td>
	</tr>
	<tr class="row4">
		<td class="col0"> <code>serveraddress</code> </td><td class="col1"> string </td><td class="col2"> yes </td><td class="col3"> <em>(none)</em> </td><td class="col4"> The IP address of the boot server. </td>
	</tr>
	<tr class="row5">
		<td class="col0"> <code>servername</code> </td><td class="col1"> string </td><td class="col2"> yes </td><td class="col3"> <em>(none)</em> </td><td class="col4"> The hostname of the boot server. </td>
	</tr>
	<tr class="row6">
		<td class="col0"> <code>force</code> </td><td class="col1"> bool </td><td class="col2"> no </td><td class="col3"> <em>(none)</em> </td><td class="col4"> dhcp-option will always be sent, even if the client does not ask for it in the parameter request list. This is sometimes needed, for example when sending options to PXELinux. </td>
	</tr>
	<tr class="row7">
		<td class="col0"> <code>instance</code> </td><td class="col1"> dnsmasq instance </td><td class="col2"> no </td><td class="col3"> <em>(none)</em> </td><td class="col4"> Dnsmasq instance to which the boot section is bound; if not specified the section is valid for all dnsmasq instances. </td>
	</tr>
</table>

jow-:
Yeah, the table you quoted is the parameter reference for `config boot` sections. `option force '1'` in `config boot ...` indeed means to use `dhcp-option-force` instead of `dhcp-option`. For `config dhcp ...` sections though, the `force` option unfortunately has a different meaning, it enforces DHCP servicing in this case, even when another server is found in the broadcast domain.

Right now there appears to be no way to request `dhcp-option-force` in `config dnsmasq` or `config dhcp` sections, hence my linked change suggestion to implement an explicit `list dhcp_option_force`.

dedeckeh:
You're referring to the boot section documentation of  the dhcp uci config while in the example you're using a dhcp section. At the moment there's no uci option to force sending dhcp options in the dhcp section (see https://lede-project.org/docs/user-guide/dhcp_configuration#dhcp_pools)

nigelterry:
Ok, sorry to have wasted your time.

jow-:
@nigelterry - no problem. I'd be grateful if you could try the commit I linked, if it works for you we could merge it to master.

nigelterry:
@jow- Your commit works fine for my particular case.

PR #1184, opened 2017-06-21
dnsmasq: add multiple domains per DHCP range to init script

llitz:
dnsmasq supports multiple definitions of --domain/-s flag. This commit
modifies /etc/init.d/dnsmasq to support it.
generate "list domain" instead of "option domain in /etc/config/dhcp

Signed-off-by: Carlos Augusto <silvaesilva@gmail.com>


dedeckeh:
This will break existing uci configs which define domain as a config option. I'm working on a solution for both domain and local as part of FS853 to support them as list config while als keeping backwards compatibility

llitz:
Haven't thought about that as I only saw problems with luci on my env (I sent a PR over to luci as well - https://github.com/openwrt/luci/pull/1213.
Ping me if you want to share what you are working on.

blogic:

PR #1186, opened 2017-06-24
musl: fix mmap pass wrong offset to kernel

ptpt52:
musl: fix mmap pass wrong offset to kernel
	for example off_t x=0x8d9eb000, the x/4096 result is 0xfff8d9eb, not 0x8d9eb as expecting
	this happens on arm_cortex-a15 with gcc 6.3.x


hauke:
Please send this also to musl upstream.

It this related to this upstream commit:
https://git.musl-libc.org/cgit/musl/commit/?id=2577b1bc16124d0690b9dd268a9f582f80bdcd67

nbd168:
Rejected by upstream

PR #1187, opened 2017-06-24
README: Fix Windows Information

neheb:
NTFS is a case sensitive file system. It is the Win32 userspace that
turns it insensitive. Since this is controlled by a registry setting,
make sure to mention it as LEDE is potentially buildable using Cygwin or
MSYS2.

### DIFF
--- a/README
+++ b/README
@@ -17,10 +17,13 @@ Simply running "make" will build your firmware.
 It will download all sources, build the cross-compile toolchain, 
 the kernel and all choosen applications.
 
-To build your own firmware you need to have access to a Linux, BSD or MacOSX system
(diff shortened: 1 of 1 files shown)


alive4ever:
Have you tested building on Windows by yourself? Which posix subsystem works? WSL, Msys2, or Cygwin?

neheb:
WSL and Cygwin work. No idea about MSYS2. The comment was intended for
MSYS2 and Cygwin. WSL is not really a Windows platform.

alive4ever:
>WSL is not really a Windows platform.

WSL is running under Windows platform. It uses NTFS underneath, so the subsystem still suffers from NT file system overhead.

neheb:
Will resend a modified version in the future.

PR #1188, opened 2017-06-25
firewall: verify dhcp source ports

farmergreg:
dhcp servers always use 67 (ipv4) and 547 (ipv6) as the source port.


package/network/config/firewall/files/firewall.config
```diff
@@ -61,6 +62,7 @@ config rule
 	option proto		udp
 	option src_ip		fc00::/6
 	option dest_ip		fc00::/6
+	option src_port		547
```
KanjiMonster: This is known to break some DHCPv6 servers, see https://lists.openwrt.org/pipermail/openwrt-tickets/2015-August/079306.html for an explanation.

package/network/config/firewall/files/firewall.config
```diff
@@ -33,6 +33,7 @@ config rule
 	option name		Allow-DHCP-Renew
 	option src		wan
 	option proto		udp
+	option src_port		67
```
KanjiMonster: This seems reasonable.

package/network/config/firewall/files/firewall.config
farmergreg: Just dug up the RFC, and I can see where the confusion is. I'll update my patch.

> 5.2. UDP Ports
>
>   Clients listen for DHCP messages on UDP port 546.  Servers and relay
>   agents listen for DHCP messages on UDP port 547.

package/network/config/firewall/files/firewall.config
farmergreg: @KanjiMonster it looks like you're right about DHCPv6. I wonder if the same applies for DHCPv4 since the RFC says:
>DHCP uses UDP as its transport protocol.  DHCP messages from a client
>   to a server are sent to the 'DHCP server' port (67), and DHCP
>   messages from a server to a client are sent to the 'DHCP client' port
>   (68). A server with multiple network address (e.g., a multi-homed
>   host) MAY use any of its network addresses in outgoing DHCP messages.

PR #1189, opened 2017-06-25
treewide: Add DISTFILES where applicable to packages

diizzyy:
* Add Gentoo distfiles as main source and upstream as fallback
* Use tarball from distfiles instead of git where applicable
* Remove deprecated mirrors


diizzyy:
Depends on: https://github.com/lede-project/source/pull/1133

diizzyy:
Dropping it, regarded as not needed/not an issue.

PR #1190, opened 2017-06-25
scripts/download.pl: Adjust URLs

diizzyy:
Internet2 isn't considered a trusted issuer meaning that https links to rit.edu will fail.
mirror.csclub.uwaterloo.ca has a trusted SSL cert and peering is good so it can replace rit.edu without performance issues.


jow-:
Merged into my staging tree with http://git.lede-project.org/a03d9ac01c

PR #1193, opened 2017-06-26
procd: hotplug.json consistency for 0660

neheb:
Change the 0660 group to be the same as the 0666 group.

I have no idea what "return" does. Seems to work just fine without it.


neheb:
Yep it does.

PR #1194, opened 2017-06-27
Utilize HTTPS urls when downloading sources

g4jc:
Multiple source mirrors in download.pl are using insecure protocols (such as HTTP and FTP), this pull request updates all the urls to utilize HTTPS.

### DIFF
--- a/scripts/download.pl
+++ b/scripts/download.pl
@@ -190,40 +190,40 @@ foreach my $mirror (@ARGV) {
 	if ($mirror =~ /^\@SF\/(.+)$/) {
 		# give sourceforge a few more tries, because it redirects to different mirrors
 		for (1 .. 5) {
-			push @mirrors, "http://downloads.sourceforge.net/$1";
+			push @mirrors, "https://downloads.sourceforge.net/$1";
 		}
 	} elsif ($mirror =~ /^\@APACHE\/(.+)$/) {
 		push @mirrors, "https://mirror.netcologne.de/apache.org/$1";
 		push @mirrors, "https://mirror.aarnet.edu.au/pub/apache/$1";
-		push @mirrors, "http://mirror.cogentco.com/pub/apache/$1";
-		push @mirrors, "http://mirror.csclub.uwaterloo.ca/apache/$1";
-		push @mirrors, "http://mirror.navercorp.com/apache/$1";
-		push @mirrors, "http://ftp.jaist.ac.jp/pub/apache/$1";
-		push @mirrors, "ftp://apache.cs.utah.edu/apache.org/$1";
-		push @mirrors, "ftp://apache.mirrors.ovh.net/ftp.apache.org/dist/$1";
+		push @mirrors, "https://mirrors.ocf.berkeley.edu/apache/$1";
+		push @mirrors, "https://mirror.csclub.uwaterloo.ca/apache/$1";
+		push @mirrors, "https://ftp.uni-erlangen.de/apache/$1";
+		push @mirrors, "https://apache.osuosl.org/$1";
+		push @mirrors, "https://www-eu.apache.org/dist/$1";
+		push @mirrors, "https://www-us.apache.org/dist/$1";
 	} elsif ($mirror =~ /^\@GITHUB\/(.+)$/) {
 		# give github a few more tries (different mirrors)
 		for (1 .. 5) {
 			push @mirrors, "https://raw.githubusercontent.com/$1";
 		}
 	} elsif ($mirror =~ /^\@GNU\/(.+)$/) {
-		push @mirrors, "https://mirrors.rit.edu/gnu/$1";
+		push @mirrors, "https://www-uxsup.csx.cam.ac.uk/pub/gnu/$1";
 		push @mirrors, "https://mirror.netcologne.de/gnu/$1";
-		push @mirrors, "http://ftp.kddilabs.jp/GNU/gnu/$1";
-		push @mirrors, "http://www.nic.funet.fi/pub/gnu/gnu/$1";
-		push @mirrors, "http://mirror.internode.on.net/pub/gnu/$1";
-		push @mirrors, "http://mirror.navercorp.com/gnu/$1";
-		push @mirrors, "ftp://mirror.csclub.uwaterloo.ca/gnu/$1";
-		push @mirrors, "ftp://download.xs4all.nl/pub/gnu/";
+		push @mirrors, "https://mirrors.ircam.fr/pub/gnu/gnu/$1";
+		push @mirrors, "https://ca.mirror.babylon.network/gnu/$1";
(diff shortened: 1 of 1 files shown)


diizzyy:
This is a bad idea, we provide all protocols because of compatibility and all mirrors have been carefully evaluated in terms of peering, availability and where users are (in general) actually located.

alive4ever:
I prefer upgrading mirrors with ftp protocol to use http protocol instead.

Regarding use of https protocol, this will add `ca-certificates` dependency to the host to perform certificate validation.

bobafetthotmail:
I also prefer http to ftp because it's not uncommon for firewalls to block ftp while they rarely block http. 
I'd say https is irrelevant because the download is checksummed and checked anyway after download so any tampering will be detected.

> Regarding use of https protocol, this will add ca-certificates dependency to the host to perform certificate validation.

Is that even an issue? Or, are there linux distros where it isn't installed by default already?

diizzyy:
HTTPS is going to preferred over HTTP and FTP due to security, MITM attacks etc and privacy.
We're not going to drop HTTP and FTP but they do have lower priority than HTTPS.

I've been very careful to make sure that we **NOT** run into certificates issues and since we did this change several months ago I'd say that I manage to do that successfully while trying to please everyone. FWIW, there has been a lot less complaints about speeds and missing tarballs etc since the change. :-)

alive4ever:
If we want to use `https` for downloading sources, let's do it right. Let the download tools perform necessary certificate validation by removing `curl`'s `--insecure` switch and `wget`'s `--no-check-certificate`.

```
      perl

	return $have_curl
		? (qw(curl -f --connect-timeout 20 --retry 5 --location --insecure), shellwords($ENV{CURL_OPTIONS} || ''), $url)
		: (qw(wget --tries=5 --timeout=20 --no-check-certificate --output-document=-), shellwords($ENV{WGET_OPTIONS} || ''), $url)
	;
```

diizzyy:
@alive4ever 
I fully agree on that one, however I think that might be an issue on platform where packages are several years old in general (such as Debian) which may cause failure on several levels esp when it comes to the packages repo.

alive4ever:
> I think that might be an issue on platform where packages are several years old in general (such as Debian)

Debian oldstable and old-oldstable can just fetch the latest `ca-certificates` from `stable` or `testing` and install the package manually. Or even better, persuade Debian developers to add `ca-certificates` to security updates so it will be updated whenever there is a newer bundle from Mozilla.

A half-baked resource fetching method from `https` servers without  certificate verification doesn't comply with [RFC5280](https://tools.ietf.org/html/rfc5280#section-3.2).

Let's do `https` downloading right!

diizzyy:
@alive4ever 
It is valid as 2014 / 2014 is really old :-/

alive4ever:
For this change to be accepted, you must really sure that those download mirrors will keep maintaining support for https. Have you contacted the mirror maintainers on your changes regarding their commitment to support https?

>  It is valid as 2014 / 2014 is really old :-/

Debian oldoldstable is still supported, so it will work as expected.
What are recent CAs that have been added? [Amazon](https://aws.amazon.com/blogs/aws/new-aws-certificate-manager-deploy-ssltls-based-apps-on-aws/)? Lets Encrypt?

Those new CAs have their root cross-signed by other trusted CA, so that _old_ supported system such as Debian oldoldstable will be able to verify the certificate chain without any issue.

Regarding `--insecure` and `--no-check-certificate` removal, a new pull request should be made, since it's unrelated to this issue.

diizzyy:
Someone(tm) should provide a full buildlog with those switches removed, I still think it's a very bad idea to just enforce HTTPS and change mirrors to rather dubious ones.

alive4ever:
After carefully reading this pull request, the mirror ordering is representing protocol fallback.

Downloader tools will use `https` mirrors on the first try. When `https` mirror can't be reached, fallback to `http` mirrors. If `http` can't be reached, fall back to `ftp` mirrors.

I think there is no need to change the mirror ordering for now.

diizzyy:
Indeed, I also submitted a pull request to push this approach further.
https://github.com/lede-project/source/pull/1133 + https://github.com/lede-project/source/pull/1189

victoredwardocallaghan:
Given today's environment I cannot understand why this is even debatable any more. We absolutely *must* pull source over a secure channel exclusively and drop mirrors that don't provide this until they are fixed to do so.

diizzyy:
@victoredwardocallaghan 
I'm not sure if you even realize that it's not a viable option no matter how much you want it.
We have several packages where the code is hosted on non HTTPS hosts, while the pull reqs above does increase the amount of "trusted" hosts far from everything is going to be available. Apart from that doing fallback to HTTP and/or FTP is a sane tradeoff, in most cases you have at least two HTTPS hosts and if both fails then I'd say that something is broken on the client end.

bobafetthotmail:
@victoredwardocallaghan the source packages used by LEDE are checksummed and the communication is mostly one-way with no password exchange.
If there are man-in-the-middle attacks (or if the mirror is providing bogus files) the package will fail the checksum check and be discarded.

victoredwardocallaghan:
@bobafetthotmail not all hashes are made equal, beyond that, hashing is *not* a complete cryptographic validation of authenticity as one can craft payloads to collide in the hashes co-domain. The purpose of the hash is to check for file corruption not authenticity! Additionally, this isn't about passwords in transit obviously so I don't see why you even say that. The problem is malware/backdoors/<insert-bad-stuff> being crafted into src tarballs while in transit.

@diizzyy It is viable if we make steps to make it viable - reducing the attack surface is a solid start. It is clear this should be a on going project to resolve this as swiftly as possible over the coming months. Mirrors that cannot be fixed for whatever reason should have their tarballs gpg signed and checked at download time.

diizzyy:
@victoredwardocallaghan 
Yeah... I wish you the best of luck really

bobafetthotmail:
> not all hashes are made equal, beyond that, hashing is not a complete cryptographic validation of authenticity as one can craft payloads to collide in the hashes co-domain. 

I wish them luck in doing that with SHA256 hashing (which is what is being used in LEDE to check the tarballs). MD5 was phased out shortly after the LEDE split. Still available for compatibility with community package repos but not used for core packages.

diizzyy:
I think we can close this, 

* Downloads won't be HTTPS only because of practical reasons (far from all projects host their files on HTTPS capable hosts)

* Moving towards this (or at least more HTTPS enabled downloads) got rejected because it's no within the project. https://github.com/lede-project/source/pull/1133 + https://github.com/lede-project/source/pull/1189
Hosting everything within the project is just insane and will mean even more work and infrastructure.

* There are no defined host system requirements and I think there's a general desire to have as few requirements as possible as far as installed software goes meaning that old systems will fail because of outdated certificates. 

To my knowledge there are no distros who does this which would help looking at how they practically handling this?

blogic:

PR #1196, opened 2017-06-29
lantiq: support ATM/PTM auto-mode driver loading

sch-m:
This set of patches is intended to support automatically loading the correct atm or ptm driver based on the notification which will be sent by dsl_cpe_control daemon.

One Big remaining problem is, that once loaded one of the atm or ptm driver, the system must be reseted to get the other driver working. Without this reset/reboot, unloading and loading the atm/ptm drivers to switch from one tc-layer to the other ends up in no (or sometimes just not all) data packets are transmitted and received through the chain.

I've already tried to fix this by reactivating the reset_ppe() functionality in the drivers, but this doesn't really fix this problem.

Any suggestions are welcome.


lynxis:
@hauke do you have an idea how to fix the reset?

target/linux/lantiq/base-files/sbin/dsl_notify.sh
```diff
@@ -1,71 +1,35 @@
...
-			( sleep 1; ifdown "$ifc" ) &
-		else
-			json_get_var autostart autostart
-			if [ "$up" != 1 ] && [ "$autostart" = 1 ]; then
-				( sleep 1; ifdown "$ifc" ) &
-			fi
-		fi
-	fi
+# Call any available notify script(s)
+notify_dir="/etc/dsl/notify.d/*"
```
dedeckeh: I would prefer an integration in /etc/hotplug.d using dsl as subfolder. There you already have iface, ntp, dhcp, usb, etc .. allowing to trigger scripts when a network interface, ntp, dhcp or usb event is raised. More info can be found at https://lede-project.org/docs/user-guide/hotplug_lede

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
```diff
@@ -0,0 +1,51 @@
...
+
+[ "$DSL_NOTIFICATION_TYPE" = "DSL_STATUS" ] || exit 0
+
+if [ "$DSL_TC_LAYER_STATUS" = "ATM" ]; then
+
+	if grep -q "ltq_ptm_vr9" /proc/modules ; then
+		# TODO: make ltq_atm_vr9/ltq_ptm_vr9 "hot-swappable"
+		# ifconfig ptm0 down &>/dev/null
+		# rmmod "ltq_ptm_vr9"
+		logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer ATM while PTM driver is loaded. Rebooting System"
```
victoredwardocallaghan: That's interesting, in the Netgear DM200 port we had to strap two gpio pin states depending on ATM/PTM configuration.

target/linux/lantiq/base-files/sbin/dsl_notify.sh
sch-m: That's a good point. I see no problems moving the files to /etc/hotplug.d/dsl.

hauke:
I do not know what is needed to reset the dsl system correctly to switch between ATM and PTM. But it could also be that a new Firmware is needed, I think the current Lantiq shipment contains a more recent datapath firmware.

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
```diff
@@ -0,0 +1,51 @@
...
+		reboot
+		exit
+	fi
+
+	if ! grep -q "ltq_ptm_vr9" /proc/modules ; then
+		logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer EFM/PTM"
+
+		modprobe ltq_ptm_vr9
+		ip link set dev ptm0 down
+		ip link set dev ptm0 address $(printf '%012X' $((1+0x$(tr -d : </sys/class/net/eth0/address))))
```
mkresin: Depending on the configuration, this would set a different mac address than the one configured with ```ucidef_set_interface_macaddr()```. Isn't the mac address set by netifd if the new netdev is registered?

sch-m: Yes you are right. If you use ucidef_set_interface_macaddr() to specifiy the mac addresses than this code would be unneccessary or probably ends in an different mac.

In my current situation I don't want to use the ucidef_set_interface_macaddr() because this ends up with entries of the mac in /etc/config/network. But in some scenarios you want to be able to clone devices by simply loading the same config-backup from one device into other devices. Doing this with the mac in /e/c/network will result in having a lot of devices with the same macs.
So I prefer loading/calculating the correct mac addresses from the stored base mac address (which is stored in uboot-env in my case) at runtime.

I've also integrated this based on the esi command in the atm section witch is taken from the vdsl_cpe_control_wrapper. Isn't this also faulty then?

We can drop this (and consequently even the esi stuff ?) if you think so.

Any suggestions are welcome.

mkresin: > In my current situation I don't want to use the ucidef_set_interface_macaddr() because...

Well, ucidef_set_interface_macaddr() is heavily used in the lantiq target. Even if you're  not intending to use the function, you would break exists code/configs. Please test your commits for possible breakage.

> Doing this with the mac in /e/c/network will result in having a lot of devices with the same macs

I'm well aware and already hit by the same issue. We need something that applies default settings without adding them to the config files. It is a treewide problem and makes it for example unnecessary hard to rename leds. So far we are aware of the limitation but no one has started to work on this.

> I've also integrated this based on the esi command in the atm section witch is taken from the vdsl_cpe_control_wrapper. Isn't this also faulty then?

No it isn't. As far as I know, the esi command was added because of a race condition. In a dhcp over atm configuration the nas0 interface was up before netifd applied the mac address. With the result that some bogus mac address was used in the dhcp packages. The "fix" is more or less an ugly hack. Especially since the ucidef_set_interface_macaddr() mac address is ignored for atm bridges.

To my knowledge there is no real atm bridge integration in netifd. That is why br2684ctl is used. With all the drawbacks like dealing with race conditions (f6e38ec125ddb1da51780e510c7943e76f61f88d).

It might be than the linked commit already fixed the dhcp over atm race condition. Unfortunally the user with the issue isn't available for testing/doesn't have the dhcp over atm line anymore.

sch-m: OK, so I will drop the ptm mac address code to not break the existing behavior.

mkresin: Please make sure that still the configured mac address is used if the ptm module is loaded on demand. I've never checked it.

package/network/config/ltq-vdsl-app/files/dsl_control
```diff
@@ -194,34 +195,39 @@ start_service() {
 
 	eval "xtse=\"\${xtse_xdsl_$annex}\""
 
-	[ -z "${xfer_mode}" ] && xfer_mode=ptm
```
sch-m: I've updated the pullrequest/commit because i found a missed change in the dsl_control init script, which seemed to got lost when i ported this change from lede-17.01 to master.

Without this change the auto mode doesn't really work, because it always assumed ptm/vdsl should be used.
Sorry.

mkresin:
> One Big remaining problem is, that once loaded one of the atm or ptm driver, the system must be reseted to get the other driver working. Without this reset/reboot, unloading and loading the atm/ptm drivers to switch from one tc-layer to the other ends up in no (or sometimes just not all) data packets are transmitted and received through the chain.

Am I right that this issue is still present? I might got something wrong, but to my understanding the whole change will not work due to the issue.

sch-m:
No, the change is working well. The only remaining ugliness is the reboot, when the atm driver was loaded and the ptm driver is needed or vice versa.
As this situation should be really rare in real live (maybe the network provider upgrades the dsl from adsl/atm to vdsl/ptm), this wouldn't hurt really much.

Sure, I still would prefer a cleaner solution without this reboot, but the rest of this change is working really good here.

target/linux/lantiq/base-files/sbin/dsl_notify.sh
```diff
@@ -1,71 +1,35 @@
...
-		led_on $1
-		;;
-	esac
-}
-
-include /lib/network
-scan_interfaces
+# This script is called by dsl_cpe_control whenever there is a DSL event.
+
+debug=0
```
mkresin: please drop the debug code

target/linux/lantiq/base-files/sbin/dsl_notify.sh
mkresin: So please move the files to /etc/hotplug.d/dsl/ and use `exec /sbin/hotplug-call dsl` to issue a hotplug call. Have a look at package/network/ services/dnsmasq/files/dhcp-script.sh for an example.

mkresin: Have you check for sideeffects with package/network/config/ltq-adsl-app? It calls dsl_notify.sh as well. I would prefer to not trigger the atm/ptm scripts intended for vr9 systems for other lantiq subtargets.

Not sure if relevant, but esi is called in package/network/config/ltq-adsl-app/files/dsl_control as well.

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
```
mkresin: I don't really like the idea of restarting the system without user interaction. Furthermore, I would recommend to use two two hotplug scripts. One for ATM and one for PTM.

Since it is a rare case that the line mode changes, I would prefer to drop the module unload/restart for now and use something like this instead:

```
#!/bin/sh

[ "$DSL_NOTIFICATION_TYPE" = "DSL_STATUS" ] && \
[ "$DSL_TC_LAYER_STATUS" = "ATM" ] && \
! grep -q "ltq_atm_vr9" /proc/modules && exit 0

logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer ATM"

modprobe ltq_atm_vr9
esi $(printf '%012X' $((1+0x$(tr -d : </sys/class/net/eth0/address))))
```

```
#!/bin/sh

[ "$DSL_NOTIFICATION_TYPE" = "DSL_STATUS" ] && \
[ "$DSL_TC_LAYER_STATUS" = "EFM" ] && \
! grep -q "ltq_ptm_vr9" /proc/modules && exit 0

logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer EFM/PTM"

modprobe ltq_ptm_vr9
```

I'm not yet fully satisfied with the script. For instance, bringing up the atm ethernet bridge (via init script) isn't necessary if ETM/PTM is used. As far as I know, the whole atm/atm ethernet bridge stuff is exclusivly added for lantiq. I'm not aware of a nother target which supports hardware that make use of ATM.

But I don't have yet an idea how to integrate it nicely without breaking danube/ar9. Do you know by any chance how it is done that the ltq-atm module is autoloaded on danube/ar9 but not on vr9? I fail to see it.

One more concern I have is a race condition. Since ee0091ebed71254d38eef701aebd01bace5a4c1e & 45b6eec9fc4e7262865708f5cdd74dc7502b4fda the carrier flag is set at the time showtime is reached. I would expect it to fail if the netdev doesn't exists at the time showtime is reached. I gave it a quick try on a ptm line and it still seams to work (otherwise netifd wouldn't bring up ppp in my test config). But I'm not sure if it is a lucky coincident or your changes/the linux netdev subsystem takes care of such situations.

FYI due to the carrier flag change, I'm somehow sure that the esi call isn't required any longer and netifd should handle mac addresses for atm bridges perfect fine. But I wasn't able to test it. As I wrote before, the `esi` call was added to workaround a race condition in a dhcp over atm-bridge configuration, where the first dhcp package was send before the configured mac address was applied to the interface. I can only guess, but due to the move of the `esi` call we might reintroduce this race condition.

package/network/config/ltq-vdsl-app/files/dsl_control
```diff
@@ -194,34 +195,39 @@ start_service() {
...
-
 	case "${xfer_mode}" in
 	atm)
-		LOAD=ltq_atm_vr9
-		UNLOAD=ltq_ptm_vr9
+		tc_layer="1:0x1:0x1_1:0x1:0x1"
```
mkresin: Please add an explaination to the commit message what tc_layer does and what is the meaning of the magic values. Is the default case required here or could it be that `1:0x1:0x1_2:0x1:0x1` is used by default if nothing else is specified?

package/network/config/ltq-vdsl-app/files/dsl_control
```diff
@@ -194,34 +195,39 @@ start_service() {
...
 		;;
-	*)
+	vdsl)
 		mode=2
+
+		# mask out ADSL bits when VDSL is requested
+		xtse="00_00_00_00_00_00_00_${xtse##*_}"
+		;;
+	*)
+		mode=0
```
mkresin: what is the meaning of mode 0? What changes?

package/network/config/ltq-vdsl-app/files/dsl_control
```diff
@@ -194,34 +195,39 @@ start_service() {
...
-		# mask out VDSL bits when ATM is requested
+		# mask out VDSL bits when ADSL is requested
 		xtse="${xtse%_*}_00"
 		;;
-	*)
+	vdsl)
 		mode=2
+
+		# mask out ADSL bits when VDSL is requested
+		xtse="00_00_00_00_00_00_00_${xtse##*_}"
```
mkresin: Do we really need to mask out the ADSL bits here? Do you encounter any incompatibilities with DSLAMs? Does the ADSL fall back for vectoring enabled lines + a **not** vectoring xdsl firmware still works with the ADSL bits masked? Nevertheless, this change should go into an extra commit. Since it is unrelated to the commit topic.

mkresin:
I really like the idea to load the atm/ptm modules based on the dsl daemon response. One more thing that does not need to be configured => one less pitfall for users.

But shouldn't we drop the `xfer_mode` option from the default config to make use of the magic? As far as I can see, it should be enought to change [ucidef_add_vdsl_modem](https://github.com/lede-project/source/blob/master/package/base-files/files/lib/functions/uci-defaults.sh#L346), the [call of uci_add_vdsl_modem](https://github.com/lede-project/source/blob/master/target/linux/lantiq/base-files/etc/board.d/02_network#L201) and maybe in [config_generate](https://github.com/lede-project/source/blob/master/package/base-files/files/bin/config_generate#L50).

target/linux/lantiq/base-files/sbin/dsl_notify.sh
sch-m: As the atm/ptm script is part of the ltq-vdsl-app package, there shouldn't be any side-effects for ltq-adsl-app. the "split up" of the dsl_notify.sh script doesn't change anything related to the functionality.

target/linux/lantiq/base-files/sbin/dsl_notify.sh
sch-m: I intentionally inserted this debug code to give even people who don't have access to the intel/lantiq dsl-api documentation the possibility to create their own notify scripts.

without this information it is hard to know which "events" can happen and so on.

target/linux/lantiq/base-files/sbin/dsl_notify.sh
mkresin: Yeah, it is obvious. Thanks for the pointer.

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
sch-m: > Since it is a rare case that the line mode changes, I would prefer to drop the module unload/restart for now...

Doing so, a cpe won't be able to come online again if the dsl/tc_layer is changed by the network operator. This might be acceptable for private use, but for devices that are installed on places where no person is on site for the overwhelming time, this could be a real problem. So I would prefer the dirtier but finally safest way with the reboot, as this shouldn't happen to often anyway. :-)

> For instance, bringing up the atm ethernet bridge (via init script) isn't necessary if ETM/PTM is used.

This wouldn't work anyway. The current implementation (starting br2684 via init script) only works, because dsl_control (where the atm driver is loaded) is started before the br2684 init script.

When the atm driver is loaded by the notify script, the br2684 init script would be called a long time ago and the br2684 won't be setup correctly.

So with my approach you have to setup the br2684 in the notify script after loading the atm driver. But if you are right, and no one else then lantiq is using this br2684 init script, it can be dropped at all.

> But I don't have yet an idea how to integrate it nicely without breaking danube/ar9. Do you know by any chance how it is done that the ltq-atm module is autoloaded on danube/ar9 but not on vr9? I fail to see it.

As the ATM/PTM auto-mode should be also useable for ADSL only devices like danube/ar9/... what about just do it also for them?

target/linux/lantiq/base-files/sbin/dsl_notify.sh
mkresin: Please add them to the commit message introducing dsl hotplug scripts. That is the place at least I would have a look at. We generally do not carry dead/disabled code. Adding them additionally to the wiki might be a good idea as well.

package/network/config/ltq-vdsl-app/files/dsl_control
sch-m: IMHO, if I select "vdsl" as line_mode, I don't want to do anything else. If I like ADSL Fallback or something like that, I should use the "auto" line_mode.

That was the intention of this change, but you are right, it shoud go into an extra commit.

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
mkresin: > Doing so, a cpe won't be able to come online again if the dsl/tc_layer is changed by the network operator. This might be acceptable for private use, but for devices that are installed on places where no person is on site for the overwhelming time, this could be a real problem. So I would prefer the dirtier but finally safest way with the reboot, as this shouldn't happen to often anyway. :-)

Got your point but still NAK from me. Other devs might have a different opinion on this. I'm the opinion the way to go is getting the dsl subsystem reset working. Everything else is a dirty hack. Sorry.

> So with my approach you have to setup the br2684 in the notify script after loading the atm driver. But if you are right, and no one else then lantiq is using this br2684 init script, it can be dropped at all.

Yes, that was the idea I had in mind. Doing the br2684 stuff (only) in the hotplug script. No idea if it really does make sense to do so and it is kind of out of scope of the PR.

> As the ATM/PTM auto-mode should be also useable for ADSL only devices like danube/ar9/... what about just do it also for them?

I totally missed that there are ptm modules for ar/danube as well. I'm somewhat used to that ADSL is always ATM, which of course isn't right. 

Indeed, it would be required to introduce dsl hotplug scripts for ltq-adsl-app as well to move the br2684 stuff. But most likely I will not touch ltq-adsl-app. I neither have the required documentation nor an ADSL line.

sch-m:
> But shouldn't we drop the xfer_mode option from the default config to make use of the magic? As far as I can see, it should be enought to change ucidef_add_vdsl_modem, the call of uci_add_vdsl_modem and maybe in config_generate.

To hold the code as flexible as possible, I would only change the call to ucidef_add_vdsl_modem in 02_network. I would introduce there new variables:

`
annex="a"
tone="av"
xfer_mode=""
`

and change the call to

`ucidef_add_vdsl_modem "$annex" "$tone" "$xfer_mode"`

So it would be possible to set a different default xfer_mode or annex or tone for specific boards.

package/network/config/ltq-vdsl-app/files/dsl_control
mkresin: Good point. We might should drop the implicit `line_mode` setting based on the `xfer_mode`. If it is working fine in auto mode, I don't see the point in forcing a `line_mode`. Or do I miss here something?

mkresin:
>> But shouldn't we drop the xfer_mode option from the default config to make use of the magic? As far as I can see, it should be enought to change ucidef_add_vdsl_modem, the call of uci_add_vdsl_modem and maybe in config_generate.
>
>To hold the code as flexible as possible, I would only change the call to ucidef_add_vdsl_modem in 02_network. I would introduce there new variables:
>
>`
>annex="a"
>tone="av"
>xfer_mode=""
>`
>
>and change the call to
>
>`ucidef_add_vdsl_modem "$annex" "$tone" "$xfer_mode"`
>
>So it would be possible to set a different default xfer_mode or annex or tone for specific boards.

ACK. Way better than my suggestion.

package/network/config/ltq-vdsl-app/files/10_atm_ptm.sh
sch-m: Well, I have my own DSLAM here, so I have an ADSL line and I have the documentation, but I don't have any hardware.

The ltq-adsl-app uses the dsl_cpe_control-3.24.4.4 package, which also sets the DSL_TC_LAYER_STATUS environment variable for the notify script. So why shouldn't it also work here?

sch-m: > Got your point but still NAK from me. Other devs might have a different opinion on this. I'm the opinion the way to go is getting the dsl subsystem reset working. Everything else is a dirty hack. Sorry.

I can understand you, but I need an implementation where the cpe can recover itself in any situation. So this feature will land on my private patch stack. :-)

sch-m: And of course, I still would also prefer a working dsl subsystem reset.

mkresin: > So why shouldn't it also work here?

Mhh, never said that. 

>The ltq-adsl-app uses the dsl_cpe_control-3.24.4.4 package, which also sets the DSL_TC_LAYER_STATUS environment variable for the notify script. So why shouldn't it also work here?

Thanks a lot for the information!

> Well, I have my own DSLAM here, so I have an ADSL line and I have the documentation, but I don't have any hardware.

Bummer! Seams to me a loose/loose situation. But at least I was able to get why ltq-atm-vr9/ltq-ptm-vr9 isn't autoloaded. The vr9 variants have a `AUTOLOAD:=` in the Makefile, which I missed earlier.

package/network/config/ltq-vdsl-app/files/dsl_control
sch-m: > We might should drop the implicit `line_mode` setting based on the `xfer_mode`. If it is working fine in auto mode, I don't see the point in forcing a `line_mode`.

ACK. I also can't imagine anything that this is needed when the auto mode can be used.

package/network/config/ltq-vdsl-app/files/dsl_control
sch-m: The "-M" Argument of dsl_cpe_control defines the initial DSL mode for ADSL/VDSL multimode handling. (NextMode).
Possible Values: 0=API-default, 1=ADSL, 2=VDSL
Default: In case of no '-M' option is given, '0' (API-default) will be selected.

So maybe it would be nicer if we drop the '-M' option in default (auto-mode) case.

package/network/config/ltq-vdsl-app/files/dsl_control
sch-m: The '-T' argument of dsl_cpe_control defines the startup configuration for TC-Layer related parameters.
Syntax:` -T<TcADSL>:<TcCfgUsADSL>:<TcCfgDsADSL>_<TcVDSL>:<TcCfgUsVDSL>:<TcCfgDsVDSL>`
Variable definitions:
TcADSL, TcVDSL:
1=ATM, 2=PTM/EFM, 4=Auto TC-Layer

TcCfgUsADSL, TcCfgUsVDSL, TcCfgDsADSL, TcCfgDsVDSL: 
1=64/65-octet encapsulation supported
2=64/65-octet encapsulation with pre-emption
3=64/65-octet encapsulation  with short packets

Default: In case of no '-T' option is given, ADSL will be configured in ATM and VDSL in PTM/EFM: `-T1:0x1:0x1_2:0x1:0x1`

So maybe we should drop the '-T' option in default (auto-mode) case.

package/network/config/ltq-vdsl-app/files/dsl_control
mkresin: Would you please add the mode documentation to the commit message as well. Am I right that mode 0 works with adsl as well as vdsl? Just to make sure we do not introduce a regression here.

package/network/config/ltq-vdsl-app/Makefile
```diff
@@ -58,9 +58,9 @@ CONFIGURE_ARGS += \
 #CONFIGURE_ARGS += --enable-model=debug
 
 define Package/ltq-vdsl-app/install
-	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin
+	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin $(1)/etc/hotplug.d/dsl
 	$(INSTALL_BIN) ./files/dsl_control $(1)/etc/init.d/
-	$(INSTALL_BIN) ./files/vdsl_cpe_control_wrapper $(1)/sbin/
+	$(INSTALL_BIN) ./files/10_atm_ptm.sh $(1)/etc/hotplug.d/dsl
```
mkresin: Please take care of the 10_atm_ptm.sh spit into 10_atm.sh + 10_ptm.sh here as well.

package/network/config/ltq-vdsl-app/files/10_atm.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+[ "$DSL_NOTIFICATION_TYPE" = "DSL_STATUS" ] && \
+[ "$DSL_TC_LAYER_STATUS" = "ATM" ] && \
+! grep -q "ltq_atm_vr9" /proc/modules || exit 0
+
+logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer ATM"
+
+if grep -q "ltq_ptm_vr9" /proc/modules ; then
+	# TODO: make ltq_atm_vr9/ltq_ptm_vr9 "hot-swappable"
```
mkresin: please do me the favour and drop the comments. Just

```
grep -q "ltq_ptm_vr9" /proc/modules && {
  logger -p daemon.notice -t "dsl-notify" "Loading ATM driver while EFM/PTM driver is loaded is not possible. Reboot is needed."
  exit
}

File: package/network/config/ltq-vdsl-app/files/10_atm.sh [hunk 12]
```diff
@@ -0,0 +1,22 @@
...
+	# ifconfig ptm0 down &>/dev/null
+	# rmmod "ltq_ptm_vr9"
+	logger -p daemon.notice -t "dsl-notify" "Loading ATM driver while EFM/PTM driver is loaded is not possible. Reboot is needed."
+	exit
+fi
+
+modprobe ltq_atm_vr9
+esi $(printf '%012X' $((1+0x$(tr -d : </sys/class/net/eth0/address))))
+
+grep -q "br2684" /proc/modules || modprobe br2684
```
mkresin: Please drop the br2684 related lines. As long as it is handled by a daemon, we should not do it this way.

package/network/config/ltq-vdsl-app/files/10_ptm.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+[ "$DSL_NOTIFICATION_TYPE" = "DSL_STATUS" ] && \
+[ "$DSL_TC_LAYER_STATUS" = "EFM" ] && \
+! grep -q "ltq_ptm_vr9" /proc/modules || exit 0
+
+logger -p daemon.notice -t "dsl-notify" "Switching to TC-Layer EFM/PTM"
+
+if grep -q "br2684" /proc/modules ; then
```
mkresin: Please do the same changes as requested for 10_atm.sh.

package/network/config/ltq-vdsl-app/Makefile
sch-m: Oh, i missed that, sorry.

package/network/config/ltq-vdsl-app/files/10_atm.sh
sch-m: Hmmm, when i drop that, the atm part won't work anymore as i stated above:

> The current implementation (starting br2684 via init script) only works, because dsl_control (where the atm driver is loaded) is started before the br2684 init script.
>
>When the atm driver is loaded by the notify script, the br2684 init script would be called a long time ago and the br2684 won't be setup correctly.

package/network/config/ltq-vdsl-app/files/dsl_control
sch-m: OK, will add this documentation to the commit message.
And yes, you are right. mode0 also works for ADSL.
This mode settings isn't really critical anyway, because it simply defines which mode (ADSL or VDSL) should be tested at first after starting the dsl_cpe_control daemon. So, setting mode 2 (VDSL) on an ADSL line will just lead to a delay before syncing.

package/network/config/ltq-vdsl-app/files/10_atm.sh
sch-m: Long story short, the br2684ctl needs to be started after the atm driver is loaded, which isn't possible when the atm driver is loaded by the notification handler and the br2684ctl is started by a "normal boot script".

mkresin: Thanks for the pointer! Obviously I have an attention span of a dayfly... Let me thing a bit more about it.

mkresin: Would it be enough to just run ```/etc/init.d/br2684ctl start reload``` after the esi command? The br2684 module should be loaded already.  The nas0 should be upped by the dsl-deamon due to setting the carrier to 1.

sch-m: Yes you are right, when we also drop the br2684 related stuff from the 10_ptm.sh script (which doesn't really makes sense as long as we can't switch between the drivers), then a simple `/etc/init.d/br2684ctl reload` should be enough.

mkresin: Horray! Please change it this way.

sch-m: Done.

mkresin:
Thanks! Merged to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

PR #1199, opened 2017-06-30
dropbear: make ssh compression support configurable

marcin1j:
Adds config option to enable compression support. Impact on binary
size is negligible but additional 60 kB (uncompressed) is needed for
a shared zlib library.

### DIFF
--- a/package/network/services/dropbear/Config.in
+++ b/package/network/services/dropbear/Config.in
@@ -32,6 +32,15 @@ config DROPBEAR_ECC
 
 		Increases binary size by about 23 kB (MIPS).
 
+config DROPBEAR_ZLIB
+	bool "Enable compression"
+	default n
+	help
+		Enables compression using shared zlib library.
+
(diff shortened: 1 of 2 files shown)


diizzyy:
Do you have a practical example where this would help? Usually the files you transfer are already compressed anyway unless there's another use case I'm missing?

marcin1j:
My use case is a terminal session over a slow link. This really makes a difference with large text chunks.

Please bear in mind that this PR does not change default dropbear configuration. It merely gives user a choice.
I agree that most users don't need compression but they don't need optional ECC and UTMP support too. Any yet it's possible to enable ECC and UTMP support so why don't make it possible to enable compression?

dedeckeh:
@marcin1j Can you add the use case in the commit description ? This makes it clear why configurable zlib support is useful.

dedeckeh:
Thanks for the patch; added to master with a slightly adapted commit description

PR #1200, opened 2017-06-30
ar71xx: add metadata to some TP-Link images, fix Archer C60 factory image

heinzek:
First commit add metadata to TP-Link sysupgrade images created by tplink-safeloader with "magic" 2705.

Second commit fix build factory firmware TP-Link Archer C60v1. Size of partition "SupportList" is only 256 bytes, and can contain only 3 entries.

### DIFF
--- a/target/linux/ar71xx/image/tp-link.mk
+++ b/target/linux/ar71xx/image/tp-link.mk
@@ -121,7 +121,8 @@ endef
 define Device/archer-cxx
   KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
   IMAGES := sysupgrade.bin factory.bin
-  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade | \
+	append-metadata | check-size $$$$(IMAGE_SIZE)
   IMAGE/factory.bin := append-rootfs | tplink-safeloader factory
 endef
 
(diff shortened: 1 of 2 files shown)


target/linux/ar71xx/image/tp-link.mk
```diff
@@ -121,7 +121,7 @@ endef
 define Device/archer-cxx
   KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
   IMAGES := sysupgrade.bin factory.bin
-  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade | append-metadata
```
pepe2k: `tplink-safeloader` checks if kernel and rootfs fit in defined partitions in image, but as you use `append-metadata` here as a last build step and appended metadata is not stripped from sysupgrade image before writing it to flash, `check-size $$$$(IMAGE_SIZE)` is needed here.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -1093,8 +1097,9 @@ define Device/tl-wr942n-v1
   IMAGE_SIZE := 14464k
   KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
   IMAGES := sysupgrade.bin factory.bin
-  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade | append-metadata
```
pepe2k: Same here, please add `check-size $$$$(IMAGE_SIZE)` as a last build step for sysupgrade image.

heinzek:
done

pepe2k:
@heinzek,

Dzięki!

Cheers,
Piotr

PR #1202, opened 2017-07-02
fix sysctl stop processing when IO error happens

ptpt52:
this fix how to probe with sysctl
sysctl -p /path/to/file.conf will stop processing after meeting IO error line
example: xx.conf
net.core.default_qdisc=fq  #<--this line cause IO error and sysctl would stop
net.ipv4.tcp_congestion_control=htcp #<--this line would never be processed

Signed-off-by: Chen Minqiang <ptpt52@gmail.com>

### DIFF
--- a/package/base-files/files/etc/init.d/sysctl
+++ b/package/base-files/files/etc/init.d/sysctl
@@ -19,7 +19,5 @@ set_vm_min_free() {
 
 start() {
 	set_vm_min_free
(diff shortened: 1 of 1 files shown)


jow-:
This introduces a lot of overhead by spawning one `sysctl` process per configuration line.
It makes more sense to fix this is `busybox` instead.

PR #1203, opened 2017-07-02
ramips: add support for TP-Link RE350

eamaclean:
These commits add support for the TP-Link RE350.

The TP-Link RE350 is a wall-wart AC1200 range extender/access point with
a single gigabit ethernet port and two non-detachable antennas, based on
the MT7621A SoC with MT7603E and MT7612E radios.

Firmware wise it is very similar to the QCA based RE450, and the
pre-existing support for such eased this port.

SoC: MediaTek MT7621A (880MHz)
Flash: 8MiB (Winbond W25Q64)
RAM:   64MiB (DDR2)
Ethernet: 1x 1Gbit
Wireless: 2T2R 2.4Ghz (MT7603E) and 5GHz (MT7612E)
LEDs: Power, 2.4G, 5G (blue), WPS (red and blue), ethernet link/act
(green)
Buttons: On/off, LED, reset, WPS

Serial header at J1, 57600 8n1:
Pin 1 TX
Pin 2 RX
Pin 3 GND
Pin 4 3.3V

Factory image can be uploaded directly through the stock UI.


pepe2k:
Hi @eamaclean,

It seems that we were working on the same changes related to `firmware-utils` and `mktplinkfw-kernel`: https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary :dizzy_face: (I needed that for TL-WA820RE v1 which is still in WIP).

Anyway, I'm pretty sure that your `sysupgrade` is broken for this board. Did you test it at all? For `safeloader` image types you shouldn't use mtd splitter - please take a look how it's done for all other boards with `safeloader` image type (including the `RE450` under `ar71xx`) - they use statically defined `kernel`, `rootfs` and `firmware` partitions. You should also read carefully what `mktplinkfw-kernel` tool exactly does - it applies `stripped down version of TP-Link v1` header only, so there is no information about `rootfs` size and offset which TP-Link mtd firmware splitter relies on.

Cheers,
Piotr

target/linux/ramips/base-files/lib/upgrade/platform.sh
```diff
@@ -255,6 +255,13 @@ platform_check_image() {
 		# these boards use metadata images
 		return 0
 		;;
+	re350)
+		[ "$magic" != "01000000" ] && {
+			echo "Invalid image type."
+			return 1
+		}
+		return 0
+		;;
```
heinzek: Use "magic" only to verify image it's not safe. Better use metadata.

tools/firmware-utils/src/mktplinkfw-kernel.c
```diff
@@ -305,7 +313,7 @@ int main(int argc, char *argv[])
 	while ( 1 ) {
 		int c;
 
-		c = getopt(argc, argv, "H:E:L:V:N:k:o:h");
+		c = getopt(argc, argv, "H:E:L:V:N:k:o:eh");
```
heinzek: missing colon

tools/firmware-utils/src/tplink-safeloader.c
```diff
@@ -709,6 +709,47 @@ static struct device_info boards[] = {
...
+		.soft_ver = NULL,
+
+		/**
+			The original os-image partition is too small,
+			so we enlarge it to 1.5M, this is just large enough
+		*/
+		.partitions = {
+			{"fs-uboot", 0x00000, 0x20000},
+			{"os-image", 0x20000, 0x180000},
+			{"file-system", 0x200000, 0x400000},
```
heinzek: Wrong calculate. os-image end at 0x1A0000, file-system starts at 0x200000. You waste 512KB

tools/firmware-utils/src/mktplinkfw-kernel.c
pepe2k: It's OK, -e and -h don't take additional argument.

eamaclean:
Pushed out updates which I think resolve all issues mentioned here and on IRC. This should probably wait now until Piotr pushes his version of the mktplinkfw-kernel changes.

pepe2k:
> Pushed out updates which I think resolve all issues mentioned here and on IRC. This should probably wait now until Piotr pushes his version of the mktplinkfw-kernel changes.

Give me some time, I want to clean that mktplinkfw-kernel/mktplinkfw duplication. We don't need separate tool for the same job.

Cheers,
Piotr

eamaclean:
> Give me some time, I want to clean that mktplinkfw-kernel/mktplinkfw duplication. We don't need separate tool for the same job.

Agreed. Take your time, gives me the opportunity for little fixes.. Like I just pushed, checking sysupgrade image size after appending metadata.

eamaclean:
Updated to use tplink-v1-header, fixed a typo in config-4.9, and future-proofed in case TP-LINK want to go to v14 with these, too. Sysupgrade and factory images both tested and working.

mkresin:
Merged to my staging tree. Thanks!

PR #1204, opened 2017-07-03
ar71xx: fix netgear wndr3700 v4 switch port mapping

sotux:
Fix switch port mapping for netgear wndr3700 v4. This model has the same board as wndr4300 v1.


pepe2k:
@sotux,

Cheers,
Piotr

PR #1209, opened 2017-07-09
ramips: fix PCI init on MT7620 with Linux 4.9+

K900:
So, this is kind of complicated. This has been upstream for a while,
imported from OpenWRT/LEDE with some cleanups. LEDE ramips has stayed
on linux-4.4 this whole time, with the old(er) version of the patch
that had correct behavior[0], while upstream got changed[1].

When LEDE updated to kernel 4.9, the older version of the code from
the patch got replaced with the upstream version containing the bug.

The original behavior, however, seems to be correct here, as the
official programming guide[2] indicates that bit 31 (PDRV_SW_SET)
in register PPLL_CFG1 is reserved, but bit 23 (added as PPLL_LD)
is the PPLL lock state (which also happens to line up with the
error message).

The original confusion probably comes from the double definition
of PDRV_SW_SET[3, 4] in the upstream code, with one correct definition
(31) and one incorrect one (23).

I've also used the opportunity to clean up the error message a bit -
it's still not really helpful to anyone who doesn't already know what
the PPLL is, but at least it's slightly more readable now.

This will probably need to be upstreamed as well, since with the way
it's currently set up, it's unlikely PCI ever worked for anyone who's
running an upstream kernel on that SoC.

[0]: https://github.com/lede-project/source/blob/05d6e92594c507dcd1f4be6c1bcb2282fe1abe1f/target/linux/ramips/patches-4.4/0009-PCI-MIPS-adds-mt7620a-pcie-driver.patch#L259
[1]: https://github.com/torvalds/linux/blob/026d15f6b9878794fae1f794cae881ccd65052e5/arch/mips/pci/pci-mt7620.c#L246
[2]: http://www.anz.ru/files/mediatek/MT7620_ProgrammingGuide.pdf
[3]: https://github.com/torvalds/linux/blob/026d15f6b9878794fae1f794cae881ccd65052e5/arch/mips/pci/pci-mt7620.c#L36
[4]: https://github.com/torvalds/linux/blob/026d15f6b9878794fae1f794cae881ccd65052e5/arch/mips/pci/pci-mt7620.c#L39

Signed-off-by: Ilya Katsnelson <me@0upti.me>

---

This is my first contribution, so please tell me where I messed up :)
The patch has been tested on my Xiaomi Mi WiFi mini, and everything works fine (the PCIe 5GHz wireless module goes up and is functional, unlike on LEDE master).

### DIFF
--- a/target/linux/ramips/patches-4.9/999-fix-pci-init-mt7620.patch
+++ b/target/linux/ramips/patches-4.9/999-fix-pci-init-mt7620.patch
@@ -1,0 +1,23 @@
+Index: linux-4.9.34/arch/mips/pci/pci-mt7620.c
+===================================================================
+--- linux-4.9.34.orig/arch/mips/pci/pci-mt7620.c
++++ linux-4.9.34/arch/mips/pci/pci-mt7620.c
+@@ -35,6 +35,7 @@
+ #define PPLL_CFG1			0x9c
+ 
+ #define PPLL_DRV			0xa0
++#define PPLL_LD			(1<<23)
+ #define PDRV_SW_SET			(1<<31)
+ #define LC_CKDRVPD			(1<<19)
+ #define LC_CKDRVOHZ			(1<<18)
+@@ -242,8 +243,8 @@ static int mt7620_pci_hw_init(struct pla
+ 	rt_sysc_m32(0, RALINK_PCIE0_CLK_EN, RALINK_CLKCFG1);
+ 	mdelay(100);
+ 
+-	if (!(rt_sysc_r32(PPLL_CFG1) & PDRV_SW_SET)) {
+-		dev_err(&pdev->dev, "MT7620 PPLL unlock\n");
++	if (!(rt_sysc_r32(PPLL_CFG1) & PPLL_LD)) {
++		dev_err(&pdev->dev, "MT7620 PPLL is unlocked, aborting init\n");
+ 		reset_control_assert(rstpcie0);
+ 		rt_sysc_m32(RALINK_PCIE0_CLK_EN, 0, RALINK_CLKCFG1);
+ 		return -1;


K900:
Also, the patch probably needs to be upstreamed once we know it works for more than one person. I'm not really familiar with the process, so any help would be appreciated.

airend:
This change does fix PCIe initialization; it would be nice to get confirmation on hardware beyond Xiaomi's, but merging something like this is definitely necessary, especially that upstreaming might take a while.

K900:
The driver is credited to @blogic upstream, so upstreaming hopefully shouldn't be a big issue. Getting into the LTS branch, on the other hand, is not something I know anything about.

K900:
Sorry for the spam (if any), I tried to rebase my changes on top of master and messed up history for a bit. Either way, it's all good now. Can someone please merge this or tell me why it can't be merged in its current state?

WordlessEcho:
Seems that this problem also in Lenovo Y1 (Newifi Mini).
Here is my feedback: [5G Wi-Fi doesn’t working in snapshot of Lenovo Y1](https://forum.lede-project.org/t/5g-wi-fi-doesnt-working-in-snapshot-of-lenovo-y1/5218/4)
I will have a try on Lenovo Y1.

WordlessEcho:
I have built a version for Lenovo Y1, here you can get it: https://gongjianwei.com/lede/
I will try to flash it tomorrow.

zorun:
@blogic will have a look since he's the maintainer for ramips.

I think you should renumber your patch to 03xx, see https://wiki.openwrt.org/doc/devel/patches#naming_patches

Also, you have an alignment issue in your define for `PPLL_LD` (possibly a tab vs. space issue?)

K900:
I've fixed the alignment issue for now, thanks for pointing that out - it was tabs all along :)

Also, a better name for the patch is probably 01xx - this should definitely go upstream at some point.

WordlessEcho:
Now I can sure that problem fixed, 5G Wi-Fi working fine now on my router.
But still have some ERROR log about PCI:
```log
Mon Jul 24 01:35:08 2017 kern.err kernel: [    0.560615] pci 0000:00:00.0: card - bus=0x0, slot = 0x0 irq=0
Mon Jul 24 01:35:08 2017 kern.err kernel: [    0.566524] pci 0000:01:00.0: card - bus=0x1, slot = 0x0 irq=4
```

simontretter:
Same here.

namidairo:
The errors aren't a regression, the 7260 PCI driver always prints that it seems. @blogic might want to change that upstream.

https://github.com/torvalds/linux/blame/1001354ca34179f3db924eb66672442a173147dc/arch/mips/pci/pci-mt7620.c#L382-L383

f00b4r0:
Works on EX3700/EX3800 (MT7620A).

blogic:
merged, sorry for the delay.

K900:
Thanks dude :) Now, should we submit it upstream?

blogic:
@K900 feel free and Cc: me so i can add my Acked-by

PR #1210, opened 2017-07-09
mvebu: add support for Turris Omnia

rmounce:
Adds support for the Turris Omnia and builds an image in its native
'medkit' format. This is a rootfs tar, initial flash is via built-in
recovery which formats the internal eMMC storage as btrfs and extracts the
image. sysupgrade is not currently implemented. It is planned to use the
same 'medkit' image with metadata for sysupgrade.

armada-385-turris-omnia.dts was added to the mainline kernel after 4.9.
It has been backported as-is and patched separately so that the dts file
may be removed for later kernels.

Notable hardware support omissions are support for switching between SFP
cage and copper PHY, and RGB LED control.

Specifications:
- Marvell Armada 385 1.6GHz dual-core ARMv7 CPU
- 1GB DDR3 RAM
- 8GB eMMC Flash
- 5x Gigabit LAN via Marvell 88E6176 Switch (2x RGMII CPU ports)
- 1x switchable RJ45 (88E1514 PHY) / SFP SGMII WAN
- 2x USB 3.0
- 12x dimmable RGB LEDs controlled by independent microcontroller
- 3x Mini PCIe slots
- Compex WLE200N2 Mini PCIe AR9287 2x2 802.11b/g/n (2.4GHz)
- Compex WLE900VX Mini PCIe QCA9880 3x3 802.11ac (2.4 / 5GHz)

Flash instructions:
(in a state of flux depending on image approach taken)


danrl:
Wow! Thank you so much for this!

target/linux/mvebu/config-4.9
```diff
@@ -60,6 +60,8 @@ CONFIG_BLK_DEV_LOOP=y
 CONFIG_BLK_DEV_SD=y
 CONFIG_BLK_MQ_PCI=y
 CONFIG_BOUNCE=y
+CONFIG_BTRFS_FS=y
```
diizzyy: No, this shouldn't be enabled by default for mvebu. Use the package instead if you need support however I don't see why this should be in by default.

target/linux/mvebu/image/Makefile
```diff
@@ -212,4 +225,17 @@ define Device/globalscale-mirabox
 endef
 TARGET_DEVICES += globalscale-mirabox
 
+define Device/turris-omnia
+  $(Device/NAND-256K)
+  KERNEL_INSTALL := 1
+  KERNEL := dtb | kernel-bin
+  DEVICE_DTS := armada-385-turris-omnia
+  DEVICE_TITLE := Turris Omnia
+  DEVICE_PACKAGES := swconfig wpad-mini kmod-ath9k kmod-ath10k ath10k-firmware-qca988x
```
diizzyy: Not sure if we should include wifi drivers by default as not all(?) devices shipped with Atheros/QCA wifi?

target/linux/mvebu/config-4.9
rmounce: btrfs is used for the rootfs, it can't be distributed as a module. The only way around this is to use a different filesystem which will complicate the flashing process somewhat.

ext4 and F2FS are already in there just for the ClearFog, so there is some precedent. I will test the impact this has on the kernel image size.

target/linux/mvebu/image/Makefile
rmounce: Case for exclusion: It increases the size of images and take some space for users that don't have WiFi modules.

Case for inclusion: The Omnia has a very large amount of storage and I assume that a majority shipped with the WiFi modules. <1MB out of 8GB is negligible.

I think it is more likely that space-conscious users will be building their own images, while those downloading images with the default configuration will be expecting some level of functionality out of the box.

There, I have expressed my opinion. Hopefully it's settled faster than the last time I was involved in this discussion. #931

danrl: Maybe @oerdnj can give some numbers of how many Omnia shipped without wifi so we have some data added to the discussion.

target/linux/mvebu/config-4.9
rmounce: Quick comparison
ext4 + F2FS + btrfs: 2.426 MiB
ext4 + F2FS (current image): 2.103 MiB
(without ClearFog support): 1.927 MiB

I had hoped to maintain a simple flashing process, however seeing this I will look at the possibility of using a ClearFog-like eMMC layout and performing a somewhat more manual initial flash via built-in OpenWrt-based Omnia recovery partition.

bobafetthotmail: > I would at least like to avoid requiring use of a serial cable for this process (though at least the pin header is populated in the factory!)

Does the bootloader have an automated tftp recovery mode? If it does you can just send it an initramfs image and then you can ssh in and do a sysupgrade (which will reformat if needed).
If you have a sysupgrade working, anyway.

rmounce: There is no automated tftp recovery mode. Serial is required to access u-boot for tftp and also as it turns out to use the built-in OpenWrt recovery image.

https://www.turris.cz/doc/en/howto/omnia_factory_reset

The only way I can think to get the Omnia to do anything other than boot from internal btrfs without opening it/serial is to use fw_setenv from the factory OS to set tftp for next boot.

(thinking out loud) It may also be possible to write a script to run on Turris OS that prepares fw_setenv for booting a ClearFog-like image, downloads the LEDE eMMC image to /tmp/ and then dd's it on top of the mounted btrfs filesystem. A dirty hack, but it would avoid having to tftp from another machine.

I quite liked the concept of using the Omnia's built-in recovery and rollback built on btrfs, but writing sysupgrade scripts (and probably also btrfs userland tools) is increasingly seeming like it will be a nuisance all while adding bloat to the mvebu kernel for the sake of bending to one device's factory partition layout.

I will experiment over the coming days with tftp booting my Omnia, flashing FAT/squashfs image using ClearFog sysupgrade scripts and then u-booting from this, and test flashing from a running btrfs environment.

package/boot/uboot-envtools/files/mvebu
```diff
@@ -26,6 +26,9 @@ armada-385-linksys-rango)
 armada-xp-linksys-mamba)
 	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x40000" "0x20000"
 	;;
+armada-385-turris-omnia)
```
stintel: Please maintain alphabetical order.

target/linux/mvebu/config-4.9
stintel: If the rootfs is BTRFS this is fine. Anything else does not make sense.

target/linux/mvebu/base-files/etc/board.d/02_network
```diff
@@ -39,6 +39,11 @@ armada-388-clearfog-*)
 		ucidef_add_switch "switch0" \
 			"0:lan:5" "1:lan:4" "2:lan:3" "3:lan:2" "4:lan:1" "5u@eth1" "6:lan:6"
 	;;
+armada-385-turris-omnia)
```
stintel: Same here. One could argue that it currently isn't ordered alphabetically, but let's not make things worse. Ideally, fix the order in a separate commit first.

target/linux/mvebu/image/Makefile
stintel: Enough storage, so let's not block the PR over this.

target/linux/mvebu/base-files/etc/board.d/02_network
rmounce: Fixed + added separate commit to fix initial order.

neheb:
Another issue I have is the permissions are off. Only root has permission for anything, making LuCI unusable.

Also, does sysupgrade work with the medkit file?

rmounce:
LuCI appears to work fine here. Permissions are preserved, only owner/group is changed to root. Or am I missing something else?

As noted in the commit message sysupgrade is not currently implemented, but it is possible. The medkit file is already a metadata image, I plan to add upgrade scripts and btrfs userland tools in a later commit.

neheb:
All the permissions under /www are X00 where X is 6 or 7.

rmounce:
Where are you seeing this? Permissions appear fine in the medkit tar and the same after installation.
```
drwxr-xr-x 0/0               0 2017-07-10 00:51 ./www/
drwxr-xr-x 0/0               0 2017-07-10 00:51 ./www/cgi-bin/
-rwxr-xr-x 0/0             135 2017-07-10 00:51 ./www/cgi-bin/luci
-rw-r--r-- 0/0             495 2017-07-10 00:51 ./www/index.html
drwxr-xr-x 0/0               0 2017-07-10 00:51 ./www/luci-static/
drwxr-xr-x 0/0               0 2017-07-10 00:51 ./www/luci-static/bootstrap/
-rw-r--r-- 0/0           50527 2017-07-10 00:51 ./www/luci-static/bootstrap/cascade.css
```

neheb:
I compile my own build with luci built in. I don't use opkg if that  helps.
It's the medkit file that comes with the bad permissions.

rmounce:
Perhaps there is some issue with your build environment? The above listing is from my own medkit build with LuCI.

I built the above image on Debian 9 (unprivileged LXC container under Proxmox 5). I am building now on macOS to confirm that this is also fine.

target/linux/mvebu/config-4.9
diizzyy: It's not fine as it bloats all other targets and btrfs isn't really a great idea on nand either.

rmounce: This is not raw NAND where UBIFS should be used, eMMC devices have a flash translation layer. F2FS may be optimised for this case, but that does not really make Btrfs (or ext4) a bad idea.

I would prefer to reach some consensus on this matter before doing any further work towards the current (bloat-inducing) medkit/Btrfs approach or alternative (and more complicated for end users) TFTP flash, u-boot environment update, ClearFog-like sysupgrade approach.

neheb:
I'm running on fedora 25. I happen to be building on an NTFS drive but oddly enough, i've never encountered this issue.

KanjiMonster:
I also started poking at the Turris Omnia (before I was aware of your work), and went the modified u-boot bootargs / resuse of the clearfog code way (which does work fine). Obviously this is no good solution, as it requires serial access.

My best idea so far to get both easy installation and a LEDE layout (readonly-fs + overlay) is to create a second stage u-boot and the following partition layout:
<pre>
p0: btrfs
+ boot/zimage - second stage u-boot
+ boot/dtb - so original u-boot doesn't complain (if actually needed)
p1: vfat (or any other FS already supported by the kernel)
+ zImage - actual kernel - could also very well be a fit image
+ dtb 
p2: squashfs
p3: f2fs (overlay)
</pre>

So the sd-card image would span p0 to p2. The second stage u-boot in p0 is in this case responsible for booting the kernel with a modified commandline to make sure the rootfs points to the right partition.

For easy flashing, my idea is to include the sdcard image in the turris-medkit*.tar.gz, and let the second stage u-boot check for it and write the image to the eMMC if it finds it.

That way we don't need to modify the u-boot env, don't need serial access, can easily implement sysupgrade and don't need btrfs in the kernel.

Of course writing a custom u-boot is easier said then done, as the omnia support isn't upstream. Then we also need to create a host-build for the btrfs-tools to create the btrfs partition. But mkfs.btrfs does allow creating file system images with custom contents, so it should be possible to create a sd-card image that boots with the original u-boot commandline.

Comments welcome, especially those that show how we don't need a second stage u-boot ;-)

neheb:
FWIW the omnia is geared towards relatively advanced users. Requiring serial access is not that much of a stretch.

BTW, u-boot args can be modified in userspace using fw_setenv.

tohojo:
neheb <notifications@github.com> writes:

> FWIW the omnia is geared towards relatively advanced users. Requiring
> serial access is not that much of a stretch.

Even for advanced users, requiring serial access to flash is a major
PITA and a barrier for adoption. I would absolutely hate to have to
bring a laptop and a serial cable to where my router is just to be able
to flash it; even if it's just the one time...

neheb:
I was pitching the idea of having netconsole in the U-boot to avoid that but they argued it's a security risk since there's no switch driver in U-boot.

In any case, fw_setenv should accomplish the goal

diizzyy:
Question is if we really want to pollute the tree for a handful of users or just tell them that serial access is needed. I'd vote for the serial access route in that case.

localnet:
I only can say that I prefer any method no requiring open the router and use serial console. But you are the developers, only is my opinion as Turris Omnia user.

JamesT42:
The current way is implemented and working, it is only lacking the failsafe/overlay-fs functions. But that is no problem when we are able to use the btrfs snapshots. In Turris-OS you are able to rollback snapshots, mount the "broken" snapshot, mend your config and then activate the repaired snapshot state. So I think there is no need to fiddle with serial consoles, additional u-boots etc. because it has no advantage.

rmounce:
I think I have come up with a relatively clean and safe approach that avoids serial console and TFTP. Inspiration comes from the Debian wiki, just hold the reset button until 6 LEDs are illuminated to boot the LEDE ramdisk. From there you can SSH in and download/flash a sysupgrade image. 

```
# run within Turris OS (tested on omnia-medkit-201706290957-minimal.tar.gz)
# first copy lede-zimage and lede-dtb to /boot/

echo '/dev/mtd0 0xC0000 0x10000 0x40000' > /etc/fw_env.config
fw_setenv tmpbootcmd 'btrload mmc 0 0x01000000 boot/lede-zimage @; btrload mmc 0 0x02000000 boot/lede-dtb @; bootz 0x01000000 - 0x02000000'
fw_setenv bootcmd 'i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; if test $rescue -ge 2; then echo BOOT eMMC LEDE initramfs; run tmpbootcmd; else if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC TurrisOS FS; run mmcboot; fi; fi'
```

Obviously the variables will have to be changed again after flashing LEDE.

It suppose it would also be possible to create a kernel+initramfs that includes btrfs support and a script to copy the sysupgrade image from Btrfs /boot/ to /tmp/, unmount Btrfs, flash sysupgrade and set uboot variables automatically. At that point the user experience could be a matter of copying some files into Turris OS /boot/, executing a script, waiting a few minutes and then LEDE is fully flashed.

I am sorry to those who enjoy Omnia recovery features with Btrfs but this is looking like the path of least resistance to please those that want to keep the kernel lean, use established LEDE partition layout+sysupgrade, and avoid packaging extra Btrfs userland stuff.

JamesT42:
So ext4 is ok for clearfog but btrfs is not ok for omnia? Is that the correct interpretation?

bobafetthotmail:
> So ext4 is ok for clearfog but btrfs is not ok for omnia?

ext4 is compiled in kernel in many other devices that don't have space issues (like say kirkwoods) as it is a commonly used filesystem for external storage.

f2fs is used already in LEDE images for sdcard or for x86 images (that usually go on sdcard or usb flash storage), so again it is not uncommon.

btrfs is a kind of white elephant here, and is not truly required.

Then again, I don't see such hard space issues in mvebu target anyway, there is only one board with a kernel size of 3072k (linksys-wrt1900ac), while others have 6 MB or 8 MB or even no space constraints at all.
A kernel with btrfs is 2.4 Mb as said above.

jow-:
I have no clear preference. On one hand I agree that there are no hard kernel space constraints on mvebu so we can likely affort the few hundred KB for btrfs support.

On the other hand, btrfs is a white elephant here, no other target uses it and it introduces even more variations into the already quite diverse image format zoo.

I think I would have a slight preference on consistency, means ext4/f2fs or squash+overlay for this target.

mkresin:
> I think I would have a slight preference on consistency, means ext4/f2fs or squash+overlay for this target.

Same here

bobafetthotmail:
> means ext4/f2fs or squash+overlay for this target.

I think ext4/f2fs does not allow easy "reset to default" or failsafe so I would prefer squash+overlay (with overlay being ext4 or f2fs as this is emmc, not raw flash) between these two.

rmounce:
I have updated my patch to re-use ClearFog image building and sysupgrade, squash+f2fs overlay. Flashing process feels like a bit of a hack to me, but it works.

bobafetthotmail:
it's better if you write installing instructions in the commit description. Also because afaik the LEDE wiki does not have pages for device-specific instructions, only for hardware specs.

rmounce:
I began doing that but the commands are too long to fit in the commit without wrapping, unless the 75-char limit only applies to prose.

I am also going to have to have a quick look now and see if there is an easy way to get initramfs and dtb as build artifacts.

rmounce:
Install process simplified slightly by appending DTB to ramdisk, and enabling ramdisk image generation by default for mvebu (so that users can download ramdisk from LEDE mirror rather than infected image from someone's Github releases).

Install process documented in commit message.

KanjiMonster:
To make the installation easier, how about we just wrap the ramdisk zImage in a turris-mediakit*.tar.gz so you can use the factory recovery flash method to boot it.

After that you can sysupgrade to a full sdcard image. This way you don't need to touch the u-boot environment manually at all.

And for the "new" boot scripts, my suggestion would be something like this (only the LEDE part and switch on failure tested; I didn't try to flash & boot a TurrisOS image again):
<pre>
bootargs_lede=earlyprintk console=ttyS0,115200 root=/dev/mmcblk0p2 rootfstype=auto rootwait
bootargs_turris=earlyprintk console=ttyS0,115200 rootfstype=btrfs rootdelay=2 root=b301 rootflags=subvol=@,commit=5 rw
mmcload_lede=setenv bootargs "$bootargs_lede cfg80211.freg=$regdomain"; fatload mmc 0 0x01000000 zImage @ && fatload mmc 0 0x02000000 armada-385-turris-omnia.dtb @
mmcload_turris=setenv bootargs "$bootargs_turris cfg80211.freg=$regdomain"; btrload mmc 0 0x01000000 boot/zImage @ && btrload mmc 0 0x02000000 boot/dtb @
mmcboot=run mmcload_lede || run mmcload_turris; bootz 0x01000000 - 0x02000000
</pre>

The idea is to use the return value of fatload to dynamically detect whether we boot LEDE (FAT) or turrisOS (btrfs), and allow going back to TurrisOS easily without ever having to touch the environment again.

target/linux/mvebu/config-4.9
```diff
@@ -246,8 +249,11 @@ CONFIG_HWMON=y
 CONFIG_HZ_FIXED=0
 CONFIG_HZ_PERIODIC=y
 CONFIG_I2C=y
+CONFIG_I2C_ARB_GPIO_CHALLENGE=y
 CONFIG_I2C_BOARDINFO=y
 CONFIG_I2C_CHARDEV=y
+CONFIG_I2C_MUX=y
+CONFIG_I2C_MUX_GPIO=y
```
KanjiMonster: why do we need this? I don't see this used anywhere

target/linux/mvebu/config-4.9
```diff
@@ -279,6 +285,8 @@ CONFIG_MAGIC_SYSRQ=y
 CONFIG_MANGLE_BOOTARGS=y
 CONFIG_MARVELL_PHY=y
 CONFIG_MDIO_BOARDINFO=y
+CONFIG_MDIO_BUS_MUX=y
+CONFIG_MDIO_BUS_MUX_GPIO=y
```
KanjiMonster: why do we need this? I don't see this used anywhere

bobafetthotmail:
> bootargs_lede=earlyprintk console=ttyS0,115200 root=/dev/mmcblk0p2 rootfstype=auto rootwait

Why aren't the LEDE bootargs set in the dts as common in other boards?

Then you would just have a 
load_lede=setenv bootargs "cfg80211.freg=$regdomain"; fatload mmc 0 0x01000000 zImage @ && fatload mmc 0 0x02000000 armada-385-turris-omnia.dtb @

KanjiMonster:
> Why aren't the LEDE bootargs set in the dts as common in other boards?

Because device tree aware u-boots dynamically update the dtb with the bootargs from their environment (as well as potentially a few other things like the memory size).

stintel:
I really don't see why we should not allow btrfs. It has some really nice features that make sense to have in LEDE. Reverting to a previous snapshot when there's a problem during a sysupgrade, is an idea I have been playing with for a longer time. I would really love to have this on my x86 devices. And once we start supporting it for a platform, we can no longer call it the white elephant anymore.

And people who claim btrfs is dodgy really need to stop spreading FUD. I am using btrfs as my default filesystem on all my new systems for a couple of years now, and it works amazingly well. Yes, you can run into problems when you completely fill up the fs, but that is due to the nature of the filesystem. There is a reason why it is recommended to keep ~10% free space on a ZFS filesystem as well. Just RTFM and don't use it in raid56 mode.

And don't get me wrong, I am all for consistency, but we shouldn't allow that to block innovation either.

KanjiMonster:
> I really don't see why we should not allow btrfs. 

How would failsafe work with btrfs? How do you e.g. ensure you can login passwordless over ssh? How do you ensure that you cannot break failsafe by modifying the wrong files?

diizzyy:
@stintel 
No, it bloats and there's no need for btrfs and as a counterpart there are lots of reports stories about btrfs etc which we can and want to avoid. Not to mention that ARM is far from the main platform for btrfs testing so it's relatively untested compared to AMD64/x86-64 for instance.

neheb:
I remember on the 3.18 kernel (OpenWRT DD) the btrfs driver would crash
within 10 seconds of me mounting the drive. Probably does not happen now
but I still remember that. I think others have similar memories.

I personally want to see btrfs in LEDE. Recall that one if the goals of
LEDE compared to OpenWRT is to scale up beyond simple routers. That happens
to be a goal of the Omnia as well.

At the end of the day, we're arguing about very capable hardware.

localnet:
@KanjiMonster 
About of snapshots, I can be wrong. But taking a snapshot at moment to flashing the device it will haven't root password and by default snapshot is read-only, but yes it can be mounted as read-write. In the case of Turris Omnia for me seems secure (https://www.turris.cz/doc/en/howto/omnia_factory_reset) on other cases I don't know.

diizzyy:
@neheb 
I'm all for having it as a module, not in the kernel by default. We want as similar kernel configs as possible across all supported platforms afaik.

stintel:
>> I really don't see why we should not allow btrfs.

> How would failsafe work with btrfs? How do you e.g. ensure you can login passwordless over ssh? How do you ensure that you cannot break failsafe by modifying the wrong files?

As already suggested, you can set subvolumes (snapshots) as read-only. Making the rootfs subvolume read-only and using a snapshot of it as overlay comes to mind. I actually think we could support current functionality with btrfs only, and we could drop squashfs, jffs/f2fs and even overlayfs support from the kernel.

neheb:
Speaking of which, btrfs also natively supports compression. Could save some space I think.

bobafetthotmail:
>  there are lots of reports stories about btrfs etc which we can and want to avoid.

f2fs had a bumpy ride too. Recently on MIPS its fsck hosed the f2fs partition, for example. (fixed with last f2fs-tools update)

stintel:
>> there are lots of reports stories about btrfs etc which we can and want to avoid.

> f2fs had a bumpy ride too. Recently on MIPS its fsck hosed the f2fs partition, for example. (fixed with last f2fs-tools update)

And exactly that fix, [broke it on Octeon (MIPS64)](https://bugs.lede-project.org/index.php?do=details&task_id=749&string=f2fs&search_name=&type%5B0%5D=&sev%5B0%5D=&pri%5B0%5D=&due%5B0%5D=&reported%5B0%5D=&cat%5B0%5D=&status%5B0%5D=open&percent%5B0%5D=&opened=&dev=&closed=&duedatefrom=&duedateto=&changedfrom=&changedto=&openedfrom=&openedto=&closedfrom=&closedto=)...

diizzyy:
@stintel 
Yet it works fine on trunk/master ;-)

stintel:
>  Yet it works fine on trunk/master ;-)

But it doesn't in our stable release. And guess what end users prefer over bleeding edge? Stability.

target/linux/mvebu/config-4.9
rmounce: These are left over from the patch set I used as my base, I probably should have tidied up earlier.

I assumed that MDIO bus mux had been added for good reason and was somehow necessary to control both the 88E6176 switch and 88E1514 PHY on the same MDIO bus, but both seem to work fine on an image compiled without.

I am away from my Omnia now, but I will do some testing later to test all of these options and remove as many as possible. I suspect some will return in a future patch to add support for the Omnia's I2C-attached microcontroller for RGB LED control, but of course it is better to logically group options with the code that requires them rather than letting cruft build for 'just in case' selections.

rmounce:
> To make the installation easier, how about we just wrap the ramdisk zImage in a turris-mediakit*.tar.gz so you can use the factory recovery flash method to boot it.

Good idea. TFTP aficionados can still untar the medkit. I will test this + your uboot commands and update documentation.

> About of snapshots, I can be wrong. But taking a snapshot at moment to flashing the device it will haven't root password and by default snapshot is read-only,

This is what the Omnia's recovery does for us when flashing :)

I would also like to see Btrfs in LEDE, but I don't think this PR is the right place to be breaking new ground in that regard and to be honest I am not currently prepared to do the work for userland tools and making sure sysupgrade, failsafe etc work. Perhaps this can be brought up later in a different target for ClearFog + Omnia.

target/linux/mvebu/image/Makefile
oerdnj: I have notified the Turris team about this PR, hopefully someone will pop-up and comment.

danrl:
With the latest omnia-related u-boot commits, we should now be able to flash the bootloader and format the omnia to ext4, right?

neheb:
Any progress on this? I can't seem to flash a build with the new commits

JamesT42:
I only tested the btrfs branch, which works very nicely in combination with the "schnapps" tool from the turris team. I can take snapshots, mount them, revert to earlier snapshots etc. with no trouble.

rmounce:
Not forgotten, but I have been very busy recently.

mvebu (even without Omnia patches) does not build on my main development machine (macOS 10.13 beta) which is another setback. My Linux build box is slow so at the moment I am putting the time I do get into fixing the macOS build then I will resume Omnia cleanup.

wackejohn:
@rmounce I've tested the sysupgrade. there are some issues 
1.while sysupgrade,the uboot env won't change.
And I found that in ```target/linux/mvebu/base-files/lib/upgrade/clearfog.sh``` the ```local board=$(mvebu_board_name)``` should change to ```local board=$(board_name)```,and then the uboot env changed successfully while sysupgrade.
2.the led of eth2 (wan port,network working ok) can't blink,
3.the rtc seems broken:
```root@HOME-Router:~# hwclock --debug --show
hwclock from util-linux 2.29.2
Using the /dev interface to the clock.
Last drift adjustment done at 1502863469 seconds after 1969
Last calibration done at 1502863469 seconds after 1969
Hardware clock is on UTC time
Assuming hardware clock is kept in UTC time.
Waiting for clock tick...
select() to /dev/rtc0 to wait for clock tick timed out...synchronization failed
```

rmounce:
1. sysupgrade is WIP, I don’t consider the patchset ready in its current state.
2. I have no intention of addressing LEDs in this PR, the ones that do work are just using their default function. Xiche has a fairly complete looking patch for this that I may rebase and submit separately.
3. I had not looked at the RTC, in fact I’ve never had a LEDE device with one and forgot it was there. If someone identifies an easy fix then great, otherwise I do not see this as a big issue for a permanently internet connected device with NTP enabled by default.

diizzyy:
@stintel @rmounce 
I agree with the above that btrfs is a no go both for inclusion and/or consistency.  That said, memory usage is also a concern.

rmounce:
> With the latest omnia-related u-boot commits, we should now be able to flash the bootloader and format the omnia to ext4, right?

We can use ext4/f2fs as the overlay FS without any u-boot support, by booting the kernel+initramfs from FAT32 like the ClearFog. This being the case, I do not see any reason to build our own u-boot as this will just make it more difficult to revert to the factory Turris OS.

@diizzyy I thought we all had moved past that. I would like to see btrfs and its features used in LEDE, but it does not belong with a hardware support patch. Concerns about consistency, value and effort aside, you appear to be the only participant in this discussion that has real 'beef' with btrfs itself. Does it have a <a href="http://www.baltimoremick.com/blog/wp-content/uploads/2008/07/reiser-wife.jpg">ReiserFS-inspired feature</a> we don't know about?

diizzyy:
@rmounce 
There are others and I fail to see why less than like 5% (given the amount of omnia units shipped and possible LEDE users) of total mvebu users should be the reason bloat the kernel config by quite a bit as the rest of all mvebu devices doesn't need it or the additional kernel modules.

rmounce:
@diizzyy I fail to see why we are still having this discussion when consensus has already been reached for this PR. You 'won'.

rmounce:
Got around to making some planned revisions. Thanks to KanjiMonster's suggestion I am now satisfied with the flashing process. I still need to update the commit message and do a final review of things with a clear head, however straight from my notes:

- copy lede-mvebu-turris-omnia-sysupgrade.img.gz and omnia-medkit-lede-initramfs.tar.gz to root of fat32 flash drive
- connect flash drive and power on omnia
- hold down reset button until 4 leds are lit
- wait about 2 minutes for initramfs to be flashed and booted
- connect to lan port
- ssh root@192.168.1.1
- mount /dev/sda1 /mnt
- sysupgrade /mnt/lede-mvebu-turris-omnia-sysupgrade.img.gz
- wait another minute
- done!

The current blocker from my perspective is a new bug whereby the f2fs partition isn't formatted on first boot and it just hangs. I was able to work around it once by entering failsafe mode on a reboot and running mount_root, however this is also hanging now. Another day!

danrl:
Quick question: I booted the `lede-mvebu-turris-omnia-initramfs-kernel.bin` via TFTP at `0x01000000`. It says "Starting kernel ..." and then, after a while, it reboots and I am back where I have started. I think I screwed up my bootcmd during earlier experiments with earlier commits. Any idea how to un-brick my router so that I can test the latest commits in this branch?

````
Method 1 - TFTP w/ serial
- Download lede-mvebu-turris-omnia-initramfs-kernel.bin to TFTP server
- Connect Turris Omnia WAN port to DHCP-enabled network with TFTP server
- Connect serial console and interrupt U-Boot
=> dhcp
=> setenv serverip <tftp_server_ip_here>
=> tftpboot 0x01000000 lede-mvebu-turris-omnia-initramfs-kernel.bin
=> bootz 0x01000000
````

**Comment by danrl** [comment]
Update: I am able to boot the zImage from https://repo.turris.cz/omnia/nor_fw/ but I cannot get the zImage from this PR to boot on my omnia. I'll keep trying a few other things and will report back.
Let me know how I can help making this work!

**Comment by wackejohn** [comment]
@danrl You should reset the uboot env to factory default,and the follow this page to do factory teset 
https://www.turris.cz/doc/en/howto/omnia_factory_reset
And then get start again...

**Comment by rmounce** [comment]
I have updated flashing instructions in the commit message, sorry for any confusion. It is only necessary to reset the U-Boot environment, and restoring the official Turris OS medkit images does not touch U-Boot.

My testing is still blocked on an unrelated sysupgrade issue that has also been reported here: http://lists.infradead.org/pipermail/lede-dev/2017-August/008707.html

**Comment by rmounce** [comment]
As far as I'm concerned this is only blocked by https://github.com/lede-project/source/pull/1330

Could someone please review and merge the mvsw61xx patch, 02_network sorting and ClearFog sysupgrade fix commits as these are independent. Might make this PR look a bit less intimidating :)

File: target/linux/mvebu/files/arch/arm/boot/dts/armada-385-turris-omnia.dts [hunk 7]
```diff
@@ -0,0 +1,392 @@
...
+			usb3@f0000 {
+				status = "okay";
+			};
+
+			usb3@f8000 {
+				status = "okay";
+			};
+		};
+
+		pcie {
```
teslamint: It should be `pcie-controller` or Wi-Fi/LTE card will not detected.

rmounce: You are correct. This was changed in the updated device tree I pulled from the mainline kernel a few days ago, but it has broken under 4.9 as you've found.

https://github.com/torvalds/linux/commit/28fbb9c539e2192fbb3fdb6f5c3ea6756bc3c3d6

I have reverted to the older device tree.

neheb:
Just installed this. Settings are not kept between reboots.

Ran df. Only two tmpfs entries are shown, /tmp and /dev. Totally broken.

rmounce:
Now that the util-linux issue has been resolved (thanks nbd) I do not have any outstanding concerns affecting this PR. Instead I will just wait for others to point them out :)

neheb:
I'll try. Since this is a new board, is there a chance of getting DSA supported and kicking swconfig out? From what I can tell, the driver currently being used is an old one based on swconfig whereas the DSA one lives in the kernel. Suspicious that the armada-xp-gp has 4 separate interfaces... might be using DSA.

rmounce:
DSA can't use multiple CPU ports for a single switch yet. Other than that, DSA works for the Turris Omnia using the device tree in the mainline kernel and also included here.

LEDE uses swconfig everywhere, so if you want to 'kick it out' then you have quite a bit of work on your hands. Not going to happen.

neheb:
Such a patch kicking swconfig out exists for the R7800 though ;)

https://github.com/dissent1/r7800/commit/51e4e56469b944cfae8e4b894b4d499856505fbf

bobafetthotmail:
>LEDE uses swconfig everywhere, so if you want to 'kick it out' then you have quite a bit of work on your hands. Not going to happen.

While probably not top priority, I've seen at least @mkresin that would welcome DSA driver (it was in a PR about updating kirkwood boards to kernel 4.9).

Afaik one of the main stoppers for a tree-wide conversion is that so far is no script in place to convert switchconfig uci configurations to DSA when sysupgrading current boards, but for new boards there is no such issue.

JamesT42:
Could this please be only about hardware support, not swconfig or btrfs or whatever? Seems to me those belong somewhere else.

bobafetthotmail:
> Could this please be only about hardware support, not swconfig or btrfs or whatever? Seems to me those belong somewhere else.

Well, root filesystem and switch driver _are_ hardware support decisions.

Here the main issue is that no core dev picked this PR up for review.

neheb:
found a fatal issue. No idea how the mounts are set up, but they show up under LuCI > Mount Points. When adding a new device, it unmounts everything (expected) including /dev/mmcblk0p1, 2,  and /dev/loop0.

Those three are not supposed to be there.

rmounce:
The same happens on x86, it is beyond the scope of this PR.

neheb:
Any idea what USB controller is used in this? driver shows xhci-hcd and lsusb shows f10f8000 as an ID, which seems bogus.

rmounce:
I don’t use USB but the controller is internal to the Marvell SoC and consumes at least one SERDES lane. It is not PCIe attached.

JamesT42:
@stintel You requested changes, Could you please review the latest code so that this PR can finally be finished?

skydiablo:
hmmm... i have brick my omnia :( just follow this instruction here https://github.com/lede-project/source/pull/1210#issuecomment-324375106 ... after `sysupgrade` the following output via serial-console:

```
U-Boot SPL 2015.10-rc2 (Nov 17 2017 - 15:12:26)
High speed PHY - Version: 2.0
SERDES0 card detect: PEX

Initialize Turris board topology
Detected Device ID 6820
board SerDes lanes topology details:
 | Lane #  | Speed |  Type       |
 --------------------------------
 |   0    |  5   |  PCIe0       |
 |   1    |  5   |  USB3 HOST0  |
 |   2    |  5   |  PCIe1       |
 |   3    |  5   |  USB3 HOST1  |
 |   4    |  5   |  PCIe2       |
 |   5    |  0   |  SGMII2      |
 --------------------------------
:** Link is Gen1, check the EP capability
PCIe, Idx 0: remains Gen1
:** Link is Gen1, check the EP capability
PCIe, Idx 1: remains Gen1
PCIe, Idx 2: detected no link
High speed PHY - Ended Successfully
DDR3 Training Sequence - Ver TIP-1.29.0
Memory config in EEPROM: 0x01
DDR3 Training Sequence - Switching XBAR Window to FastPath Window
DDR3 Training Sequence - Ended Successfully


U-Boot 2015.10-rc2 (Nov 17 2017 - 15:12:26 +0100), Build: jenkins-omnia-test-1313

SoC:   MV88F6820-A0
       Watchdog enabled
I2C:   ready
SPI:   ready
DRAM:  1 GiB (ECC not enabled)
Enabling Armada 385 watchdog.
Disabling MCU startup watchdog.
Regdomain set to **
MMC:   mv_sdh: 0
SF: Detected S25FL164K with page size 256 Bytes, erase size 64 KiB, total 8 MiB
PCI:
  00:01.0     - 168c:002e - Network controller
PCI:
  01:00.0     - 11ab:6820 - Memory controller
  01:01.0     - 168c:003c - Network controller
Model: Marvell Armada 385 GP
Board: Turris Omnia SN: 0000000B0001391D
Regdomain set to **
SCSI:  MVEBU SATA INIT
SATA link 0 timeout.
SATA link 1 timeout.
AHCI 0001.0000 32 slots 2 ports 6 Gbps 0x3 impl SATA mode
flags: 64bit ncq led only pmp fbss pio slum part sxs
Net:   neta2
Hit any key to stop autoboot:  0
neta2 Waiting for PHY auto negotiation to complete......... done
BOOTP broadcast 1
BOOTP broadcast 2
BOOTP broadcast 3
BOOTP broadcast 4
*** Unhandled DHCP Option in OFFER/ACK: 2
*** Unhandled DHCP Option in OFFER/ACK: 224
*** Unhandled DHCP Option in OFFER/ACK: 2
*** Unhandled DHCP Option in OFFER/ACK: 224
DHCP client bound to address 192.168.100.103 (3425 ms)
*** Warning: no boot file name; using 'C0A86467.img'
Using neta2 device
TFTP from server 0.0.0.0; our IP address is 192.168.100.103; sending through gateway 192.168.100.1
Filename 'C0A86467.img'.
Load address: 0x800000
Loading: *
TFTP error: 'File not found' (1)
Not retrying...
Wrong Image Format for bootm command
ERROR: can't get kernel image!
```
there is no way to boot `lede-mvebu-turris-omnia-initramfs-kernel.bin`, loaded via tftp... sometime i get to load the `zimage` from here: https://repo.turris.cz/omnia/nor_fw/

what can i do to flash an running kernel image?

rmounce:
The latest instructions are in the commit message, or at https://github.com/rmounce/lede-source/releases.

It looks like the problem is that you are connecting to an existing LAN which has the boot file option set on its DHCP server, it is trying to boot this and then failing. Please disconnect from the rest of your LAN and only connect a computer acting as a DHCP client (or static 192.168.1.1xx) to a LAN port on the Omnia throughout the flashing process.

You should be able to restore either the official medkit image or my 'bootstrap' medkit image at any point from USB via the reset button technique, unless the boot process has changed with the latest uboot update.

Even in this worst case scenario, you should always be able to restore your Omnia to a factory state from serial console by

- Disconnect everything from the Omnia except serial console
- Interrupt uboot and run
- `env default -a`
- `saveenv`
- Download the official medkit image to USB
- Follow the official reset button medkit recovery process

skydiablo:
hmmm... ok, based on given link, release steps and downloads i get this result:

```
root@LEDE:/tmp# sysupgrade /tmp/lede-mvebu-turris-omnia-sysupgrade.img.gz
Cannot save config while running from ramdisk.
Commencing upgrade. All shell sessions will be closed now.
Watchdog handover: fd=3
- watchdog -
killall: telnetd: no process killed
Sending TERM to remaining processes ... logd rpcd netifd odhcpd uhttpd ntpd dnsmasq ubusd
Sending KILL to remaining processes ...
Performing system upgrade...
dd: writing '/dev/mmcblk0': No space left on device
0+65 records in
0+64 records out
Upgrade completed
Rebooting system...
umount: can't unmount /dev: Resource busy
umount: can't unmount /tmp: Resource busy
umount: can't unmount /: Invalid[  173.927866] reboot: Restarting system

U-Boot SPL 2015.10-rc2 (Nov 17 2017 - 15:12:26)
High speed PHY - Version: 2.0
SERDES0 card detect: PEX

Initialize Turris board topology
Detected Device ID 6820
board SerDes lanes topology details:
 | Lane #  | Speed |  Type       |
 --------------------------------
 |   0    |  5   |  PCIe0       |
 |   1    |  5   |  USB3 HOST0  |
 |   2    |  5   |  PCIe1       |
 |   3    |  5   |  USB3 HOST1  |
 |   4    |  5   |  PCIe2       |
 |   5    |  0   |  SGMII2      |
 --------------------------------
:** Link is Gen1, check the EP capability
PCIe, Idx 0: remains Gen1
:** Link is Gen1, check the EP capability
PCIe, Idx 1: remains Gen1
PCIe, Idx 2: detected no link
High speed PHY - Ended Successfully
DDR3 Training Sequence - Ver TIP-1.29.0
Memory config in EEPROM: 0x01
DDR3 Training Sequence - Switching XBAR Window to FastPath Window
DDR3 Training Sequence - Ended Successfully


U-Boot 2015.10-rc2 (Nov 17 2017 - 15:12:26 +0100), Build: jenkins-omnia-test-1313

SoC:   MV88F6820-A0
       Watchdog enabled
I2C:   ready
SPI:   ready
DRAM:  1 GiB (ECC not enabled)
Enabling Armada 385 watchdog.
Disabling MCU startup watchdog.
Regdomain set to **
MMC:   mv_sdh: 0
SF: Detected S25FL164K with page size 256 Bytes, erase size 64 KiB, total 8 MiB
PCI:
  00:01.0     - 168c:002e - Network controller
PCI:
  01:00.0     - 11ab:6820 - Memory controller
  01:01.0     - 168c:003c - Network controller
Model: Marvell Armada 385 GP
Board: Turris Omnia SN: 0000000B0001391D
Regdomain set to **
SCSI:  MVEBU SATA INIT
SATA link 0 timeout.
SATA link 1 timeout.
AHCI 0001.0000 32 slots 2 ports 6 Gbps 0x3 impl SATA mode
flags: 64bit ncq led only pmp fbss pio slum part sxs
Net:   neta2
Hit any key to stop autoboot:  0
Setting bus to 1
BOOT eMMC FS
** Bad device mmc 0 **
** Bad device mmc 0 **
** Bad device mmc 0 **
** Bad device mmc 0 **
Kernel image @ 0x1000000 [ 0x000000 - 0x48bdb8 ]
## Flattened Device Tree blob at 02000000
   Booting using the fdt blob at 0x2000000
   Loading Device Tree to 0fff8000, end 0ffff2e3 ... OK

Starting kernel ...
```
the only way to flash, is via serial... connection via LAN is impossible (no DHCP response and with static client IP there is no connection possible - timeout)... for now, there is an reboot loop... in paralell, i'm compiling an new firmware, based on your "mine" branch... we will see...

PS:
```
Performing system upgrade...
dd: writing '/dev/mmcblk0': No space left on device
0+65 records in
0+64 records out
Upgrade completed
```
looks not so good ????

localnet:
I too had troubles with method 1, but I thought was my fault. After use the method 2 all seems work ok since 1 month ago.

PS: I used @rmounce branch "mine".

rmounce:
@skydiablo I think there is a hardware problem with the eMMC flash on your Omnia based on that dd output, and would also explain everything else :(

skydiablo:
ok, dont ask how, but after some flash and reflash and tftp and bootloops and factory recover steps and downgrade u-boot and so on... (it is insane!) ... for now, it is currenty running, but the boot output is not "the yellow from the egg" (german proverb):

```
Setting bus to 1
BOOT eMMC FS
** Unrecognized filesystem type **
** Unrecognized filesystem type **
BtrFS init: Max node or leaf size 4096
7743056 bytes read in 367 ms (20.1 MiB/s)
BtrFS init: Max node or leaf size 4096
17124 bytes read in 78 ms (213.9 KiB/s)
Kernel image @ 0x1000000 [ 0x000000 - 0x762650 ]
## Flattened Device Tree blob at 02000000
   Booting using the fdt blob at 0x2000000
   Loading Device Tree to 0fff8000, end 0ffff2e3 ... OK

Starting kernel ...

[    0.000000] Booting Linux on physical CPU 0x0
[    0.000000] Linux version 4.9.58 ....
```

is this the default output, in point of `** Unrecognized filesystem type **` ???

neheb:
Sounds like you need an updated bootloader

rmounce:
Disagree, if anything the updated bootloader has changed behaviour and caused issues.

localnet:
@rmounce, can any be done to work the RTC?

```
# date
Fri Nov 24 19:49:40 CET 2017
# hwclock -r -f /dev/rtc0 
Sat Jan 13 19:13:45 2018  0.000000 seconds
# hwclock -w -u -f /dev/rtc0 
# hwclock -r -f /dev/rtc0 
Sat Jan 13 19:13:45 2018  0.000000 seconds
```

rmounce:
- My Omnia is in production so I’m hesitant to test things
- I have no need for RTC
- I don’t know what’s missing nor how to fix

I’m happy to review and incorporate someone else’s patch to fix the RTC, but nothing will happen here if it is left to me alone.

wackejohn:
@rmounce 
the linksys wrt1200ac/wrt1900ac have the same issue...
see https://bugs.lede-project.org/index.php?do=details&task_id=831&order=id&sort=desc&pagenum=4

localnet:
My Omnia too is in production, thanks to you @rmounce so thanks very much. I hope this pull request can be merged soon.

About the RTC I have only found [this](https://wiki.debian.org/InstallingDebianOn/TurrisOmnia#RTC) reference relative to U-Boot:
```
mw 0xf10184a0 0xfd4d4cfa
```
But I don't have experience with U-Boot and after read [this](https://www.denx.de/wiki/view/DULG/UBootCmdGroupMemory#Section_5.9.2.8.) I don't understand if change is permanent or temporal and if it can affect to the correct Omnia boot.

bobafetthotmail:
> About the RTC I have only found this reference relative to U-Boot:

It is manually writing a magic value (a number) in a hardware register, which is a dirty workaround to initialize something when the uboot didn't do that on its own.
To know what that numbers mean you need to have the SoC hardware manual, which I think is under NDA for any Marvell SoC newer than Kirkwood.

It does not survive a reboot, but you can have the uboot do that action automatically every boot (so it becomes "permanent" in practice).

That line must be added to uboot's **bootcmd** line. 
uboot configuration works like a script, all you can do manually with uboot commandline you can write down in its configuration and have it do that on its own.
The point where it starts to read/execute boot configuration "script" is the **bootcmd** "variable".

This line (from the link you posted above) 
`bootcmd=i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi`

should become 
`bootcmd=mw 0xf10184a0 0xfd4d4cfa; i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi`

adding that line from uboot's serial console:
```
setenv bootcmd 'mw 0xf10184a0 0xfd4d4cfa; i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi'
saveenv
```

adding that line from inside the system with fw_setenv
`fw_setenv bootcmd 'mw 0xf10184a0 0xfd4d4cfa; i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi'`


I know of these tricks due to other devices that required similar manual register writing to disable or enable stuff manually, I don't know the Omnia specifically but I'm sure uboot works the same.

wackejohn:
@localnet @bobafetthotmail 
I've tried `mw 0xf10184a0 0xfd4d4cfa` on my omnia,but nothing happend...
And my omnia can't get the rtc time at all:
`root@HOME-Router:~# hwclock -D -r -f /dev/rtc0
hwclock from util-linux 2.30.2
Using the rtc interface to the clock.
Assuming hardware clock is kept in UTC time.
Waiting for clock tick...
select() to /dev/rtc0 to wait for clock tick timed out...synchronization failed
root@HOME-Router:~# date
Sat Nov 25 22:56:41 CST 2017
root@HOME-Router:~# `

localnet:
@bobafetthotmail thank you, very good explication.

@wackejohn for me it's working:
```
# date
Sat Nov 25 22:10:21 CET 2017
# hwclock -r /dev/rtc0 
Sat Nov 25 21:10:22 2017  0.000000 seconds
```

My current U-Boot bootcmd:
```
bootcmd=i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; mw 0xf10184a0 0xfd4d4cfa; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi
```

I copied it and only added `mw 0xf10184a0 0xfd4d4cfa`. I too tried to do `date reset` in U-Boot as suggests [here](https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/armada-38x-solidrun-microsom.dtsi):
```
rtc@a3800 {
	/*
	 * If the rtc doesn't work, run "date reset"
	 * twice in u-boot.
	 */
	status = "okay";
};
```
But my U-Boot does not has the command `date`. Because, I don't know.

wackejohn:
@localnet 
For my omnia,that command doesn't working at all...and my u-boot also doesn't have the `date` command,it's in the mainline u-boot,but not in the omnia u-boot...

localnet:
@wackejohn 
My Turris Omnia is running LEDE, custom build of @rmounce branch mine.

wackejohn:
@localnet 
My omnia also running LEDE,that's strange....hardware problem?

localnet:
I don't know, but my BusyBox `hwclock` doesn't have the option `-D` that you used higher:
```
# hwclock --help
BusyBox v1.27.2 () multi-call binary.

Usage: hwclock [-r] [-s] [-w] [-t] [-l] [-u] [-f FILE]

Query and set hardware clock (RTC)

        -r      Show hardware clock time
        -s      Set system time from hardware clock
        -w      Set hardware clock from system time
        -t      Set in-kernel timezone, correct system time
                if hardware clock is in local time
        -u      Assume hardware clock is kept in UTC
        -l      Assume hardware clock is kept in local time
        -f FILE Use specified device (e.g. /dev/rtc2)
```
I used:
```
# hwclock -w -u -f /dev/rtc0
# hwclock -r -u -f /dev/rtc0 
Sun Nov 26 13:55:27 2017  0.000000 seconds

```
Can be interesting know if it work in TurrisOS...

wackejohn:
@localnet 
My omnia didn't use the busybox built in hwclock,I'll try the busybox built in version ....

bobafetthotmail:
> But my U-Boot does not has the command date. Because, I don't know.

the uboot in turris omnia is customized (for example mainline uboot does not support booting from btrfs), and in general most uboots in embedded devices don't have many commands to save space or whatever.

turris omnia uboot is opensourced here https://gitlab.labs.nic.cz/turris/turris-omnia-uboot

mibastian:
> I’m happy to review and incorporate someone else’s patch to fix the RTC, but nothing will happen here if it is left to me alone.

I wish the remaining issues(are there any?) with this commit would be finally resolved.

wackejohn:
I found that someone's wrt1900acs also have the rtc issue,see:
https://forum.openwrt.org/viewtopic.php?pid=320725#p320725
My omnia output:
`root@HOME-Router:~# cat /proc/interrupts | grep rtc
 42:          0          0     GIC-0  53 Level     f10a3800.rtc`

localnet:
Same here:
```
# cat /proc/interrupts | grep rtc
 42:          0          0     GIC-0  53 Level     f10a3800.rtc
```
I did a reboot at monday:
```
# dmesg | grep -i rtc
[    1.270379] armada38x-rtc f10a3800.rtc: rtc core: registered f10a3800.rtc as rtc0
[    1.458483] armada38x-rtc f10a3800.rtc: setting system clock to 2017-11-27 18:43:41 UTC (1511808221)
```
The date time seems continue ok:
```
# hwclock -r -u
Wed Nov 29 18:54:13 2017  0.000000 seconds
```

wackejohn:
@localnet 
That's strange....Can you  upload your LEDE image and also your uboot env?
 I want to try your image to see if there is a hardware problem.Thanks.

localnet:
@wackejohn
Here is my LEDE image: https://www.dropbox.com/sh/k3ezfjrswhr6l0w/AADjdP2f0q4sfKW_aUZUAviwa?dl=0

My U-Boot environment is most complicated... Shutdown, open the router, connect the serial console... I have few time and my Omnia is in production. The only change I did was:
```
setenv bootcmd 'i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; mw 0xf10184a0 0xfd4d4cfa; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi'
saveenv
```

Try my LEDE image and if the RTC continue without working for you, I will try to get the U-Boot environment. But I will need time...

bobafetthotmail:
> My U-Boot environment is most complicated... Shutdown, open the router, connect the serial console

I was under the impression that Turris Omnia had **fw_printenv** tool you can use from SSH.

localnet:
You are right, it is selected by `make defconfig`. I thought I needed select it at build time.

```
baudrate=115200
bootargs=earlyprintk console=ttyS0,115200 rootfstype=btrfs rootdelay=2 root=b301 rootflags=subvol=@,commit=5 rw
bootcmd=i2c dev 1; i2c read 0x2a 0x9 1 0x00FFFFF0; setexpr.b rescue *0x00FFFFF0; mw 0xf10184a0 0xfd4d4cfa; if test $rescue -ge 1; then echo BOOT RESCUE; run rescueboot; else echo BOOT eMMC FS; run mmcboot; fi
bootdelay=3
ethact=neta2
ethaddr=xx:xx:xx:xx:xx:xx
factory_mmcload=setenv bootargs "$bootargs cfg80211.freg=$regdomain"; btrload mmc 0 0x01000000 boot/zImage @; btrload mmc 0 0x02000000 boot/dtb @
fdt_high=0x10000000
initrd_high=0x10000000
lede_bootargs=earlyprintk console=ttyS0,115200 root=/dev/mmcblk0p2 rootfstype=auto rootwait
lede_mmcload=setenv bootargs "$lede_bootargs cfg80211.freg=$regdomain"; fatload mmc 0 0x01000000 zImage; fatload mmc 0 0x02000000 armada-385-turris-omnia.dtb
mmcboot=run lede_mmcload || run factory_mmcload; bootz 0x01000000 - 0x02000000
regdomain=**
rescueboot=i2c mw 0x2a.1 0x3 0x1c 1; i2c mw 0x2a.1 0x4 0x1c 1; mw.l 0x01000000 0x00ff000c; i2c write 0x01000000 0x2a.1 0x5 4 -s; setenv bootargs "$bootargs omniarescue=$rescue"; sf probe; sf read 0x1000000 0x100000 0x700000; bootz 0x1000000
stderr=serial@12000
stdin=serial@12000
stdout=serial@12000
```

PS: I have edited `ethaddr`

wackejohn:
@localnet 
Thank you for your LEDE image,I've tried it,and my omnia's rtc is working properly with your LEDE image.
But still not working on my own build LEDE image.
```
root@HOME-Router:/# hwclock
Fri Dec  1 10:30:31 2017  0.000000 seconds
root@HOME-Router:/# hwclock
Fri Dec  1 10:30:31 2017  0.000000 seconds
```
I think there were something missconfigured by myself,can you upload your `.config` and `config-4.9` file?Thanks a lot.

localnet:
@wackejohn
Here are:
https://www.dropbox.com/sh/k3ezfjrswhr6l0w/AADjdP2f0q4sfKW_aUZUAviwa?dl=0

I has done changes only in LEDE packages: https://github.com/openwrt/luci/pull/1422 and https://github.com/openwrt/packages/pull/5040

But I haven't done changes in LEDE core:
```
$ git status
On branch mine
Your branch is up-to-date with 'origin/mine'.
nothing to commit, working tree clean
```

Can be more easy if you discard your changes and you do:
```
$ curl -L -O https://gist.github.com/localnet/ea4c8873b245288eaf6557d5a90ac4d3/raw/6ea5d63d91adfead31f6b2928fc9a9e582876542/turris-omnia
$ mv turris-omnia .config
$ make defconfig
$ make -j 9
```

PS: Edit turris-omnia, add/remove packages, as you want.

wackejohn:
@localnet 
Thank you for your reply,I've fixed this issue by using back the busybox bulit in hwclock...
Now my omnia's rtc is working properly,thanks again.

rmounce:
I've successfully tested the RTC fix, included it in the u-boot environment that is set during sysupgrade, and updated this in the PR.

So that's one minor issue fixed (not that I considered it a blocker)...  it would be nice to get this merged in 2017 :)

diizzyy:
A bit late to the party but SolidRun just added RTC stuff to their repo.

https://github.com/SolidRun/lede-project/commit/2fae43771d71fdd77cf547ce6c5ca6d53103c7f9
https://github.com/SolidRun/lede-project/commit/5371b7adfe0f1ba58cdfb404d74e9be455d0be09

tohojo:
I have also managed to compile and flash an image based on this pull request. The USB-stick flashing procedure seemed to work, but the router didn't boot afterwards. Retrying with the tftp/serial approach worked. It may have had something to do with the uboot environment as I also needed to reset that to get the image to boot over tftp.

Tested-by: Toke Høiland-Jørgensen <toke@toke.dk>

hauke:
There are some open questions mostly from @KanjiMonster could someone please take care about them, otherwise this looks ok to me.

target/linux/mvebu/config-4.9
tohojo: The SFP autodetect script shipped with the omnia uses i2c to probe the EEPROM of inserted SPF modules, as well as some GPIO pins to detect changes. Not sure if this option is related to that.

If it is, I don't think there's any reason to keep it enabled for now: the image I built from this pull request seems to be lacking driver support for some of the features that script is using anyway, so updates will have to be made to add the SFP switch support...

tohojo:
One issue I've seen since I flashed this is that having receive offloads (GRO) enabled on the wan interface breaks my he.net 6in4 tunnel (I will only get around 1mbps of by throughput). Turning off GRO immediately unbreaks things.

I've seen this same behaviour before, but I don't recall if it was on the stock Omnia firmware or if it was on a different device...

blogic:

tohojo:
I took the liberty of fixing the few outstanding nits and reopening this against the openwrt repo. See https://github.com/openwrt/openwrt/pull/684

Not trying to take credit, just trying to speed this along. @rmounce, let me know if you object :)

PR #1215, opened 2017-07-11
utils/util-linux: Update to 2.30.1

diizzyy:
utils/util-linux: Update to 2.30.1
Refresh patch

### DIFF
--- a/package/utils/util-linux/Makefile
+++ b/package/utils/util-linux/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=util-linux
-PKG_VERSION:=2.29.2
+PKG_VERSION:=2.30.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
-PKG_SOURCE_URL:=@KERNEL/linux/utils/$(PKG_NAME)/v2.29
-PKG_HASH:=accea4d678209f97f634f40a93b7e9fcad5915d1f4749f6c47bee6bf110fe8e3
+PKG_SOURCE_URL:=@KERNEL/linux/utils/$(PKG_NAME)/v2.30
+PKG_HASH:=1be4363a91ac428c9e43fc04dc6d2c66a19ec1e36f1105bd4b481540be13b841
 
 PKG_LICENSE:=GPL-2.0
 PKG_LICENSE_FILES:=	COPYING					\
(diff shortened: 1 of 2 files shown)


wongsyrone:
Based on my test, libncurses should be added to fdisk's dependency list. Have you tested this?

diizzyy:
I did add that
https://github.com/lede-project/source/pull/1215/files#diff-2e01364f97cf4e95b275547c354e89ffR175
Not sure what you're asking

wongsyrone:
Just wondering what's the difference between `+libncursesw` and `+libncurses`

commodo:
`libncursesw` supports unicode/wide-char

wongsyrone:
Okay.

diizzyy:
Compile/Run-tested on mvebu, Linksys WRT3200ACM, trunk

package/utils/util-linux/Makefile
```diff
@@ -172,7 +172,7 @@ endef
 define Package/fdisk
 $(call Package/util-linux/Default)
   TITLE:=manipulate disk partition table
-  DEPENDS:= +libblkid +libsmartcols +libfdisk
+  DEPENDS:= +libblkid +libsmartcols +libfdisk +libncursesw
```
pprindeville: Does it also build with just `libncurses`?  Why does it require the wide version?  Is NLS turned on?

diizzyy: https://github.com/lede-project/source/blob/master/package/libs/ncurses/Makefile#L42
It's an alias and I think it's worth preparing packages for unicode, for now it's just a cosmetic thing.

pprindeville: Given that `fdisk` is rarely (if ever used), it's not clear that it needs wide-byte support or NLS. How much smaller is it with NLS turned off?  Because libncursesw.so is about 8.4MB long...

But I imagine there are a lot of other things already pulling it in, so maybe it's a moot point.

diizzyy: We don't really have the manpower to maintain custom patches to left and right. Looking at most outdated packages those carry custom patches and I highly doubt it's worth it. Also, there's no libncursesw.so for now ;-)

pprindeville: Wasn't suggesting a patch.  Maybe just adding:

`DISABLE_NLS:=--disable-nls`

so that it's always built without NLS even if `CONFIG_BUILD_NLS=y` is set.

hauke:
I merged this, but would be nice to have a patch which adds "DISABLE_NLS:=--disable-nls" I think we do not need NLS support in fdisk.

pprindeville:
Ask and ye shall receive.  PR #1307.

PR #1219, opened 2017-07-16
kernel: update kernel 4.9 to version 4.9.38

anomeome:
update kernel, check if any refresh required

compile/test on master for target mvebu/mamba


zorun:
This duplicates https://patchwork.ozlabs.org/patch/789577/

anomeome:
I would suggest that the patch sent to the [ML](http://lists.infradead.org/pipermail/lede-dev/2017-July/008294.html) on Monday the 17th duplicates this PR submitted on Sunday the 16th. But chronology is greatly overrated, and I also prefer non-linear time progressions.

zorun:
I only meant to suggest that the two patches were duplicates of each other, without any precedence implication :)

PR #1220, opened 2017-07-16
build: add PKG_SOURCE_URL_FILE

champtar:
This will allow us to dowload from github without having name conflicts
for exemple:
```
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://github.com/<project>/<repo>/archive/
PKG_SOURCE_URL_FILE:=$(PKG_VERSION).tar.gz
```


diizzyy:
I don't think that'll help as they're rolling tarballs afaik?
https://bugs.lede-project.org/index.php?do=details&task_id=374&order=id&sort=desc

zorun:
As Stijn said in this bug report, you can freely choose the filename by just appending it to the URL, like this:

```
PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
PKG_SOURCE_URL:=https://github.com/<project>/<repo>/archive/$(PKG_VERSION)/$(PKG_NAME)-$(PKG_VERSION).tar.gz
```

diizzyy:
Again, are we sure that these aren't rolling tarballs?

zorun:
It depends on the version.  If `PKG_VERSION` is the name of a git tag, the tarball shouldn't change (unless upstream later changes the tag to point to a different commit, which does happen but is not common).  If it's the name of a branch, obviously it will change each time the branch changes.  If it's a commit ID, then it should never change.

As an example:

- tarball from git tag: https://github.com/savoirfairelinux/opendht/archive/1.3.4/opendht-1.3.4.tar.gz
- tarball from git branch: https://github.com/savoirfairelinux/opendht/archive/master/opendht-master.tar.gz
- tarball from git commit: https://github.com/savoirfairelinux/opendht/archive/83d1cb97c877f9af51e058721b98f91893c2a244/opendht-83d1cb97.tar.gz

diizzyy:
So in short, there's no way we can guarantee.
I think it's better to use git directly and/or piggyback gentoo distfiles or whatever when possible.

zorun:
I don't think you read me correctly?  Pointing to a tag yields a stable tarball unless upstream performs black magic, and pointing to a commit always yields a stable tarball.

champtar:
@zorun that's a good new
just tested from the US and Europe and archive have the same hash
we can now use https archive instead of git \o/

NeoRaider:
@champtar I think it's generally preferable to download over git via HTTPS instead of using the tar archives when there are no official tarballs. Even if the contents of the tars should never change for tag and commit tarballs, they are generated on the fly, so checksums may change if Github updates their tar or gzip software.

diizzyy:
So pretty much what nbd said in the bug report.
@NeoRaider 
Do you think piggybacking is a good alternative as we already have mirrors by doing so (in most cases) and it's more corp fw friendly?

PR #344, opened 2016-09-28
[patch v6]layerscape: add 64b/32b target for ls1043ardb device

fsl-jyt:
Add support for NXP layerscape ls1043ardb 64b/32b Dev board.

LS1043a is an SoC with 4x64-bit up to 1.6 GHz ARMv8 A53 cores.
ls1043ardb support features as: 2GB DDR4, 128MB NOR/512MB NAND, USB3.0, eSDHC,
I2C, GPIO, PCIe/Mini-PCIe, 6x1G/1x10G network port, etc.

64b/32b ls1043ardb target is using 4.4 kernel, and rcw/u-boot/fman images from
NXP QorIQ SDK release.

All of 4.4 kernel patches porting from SDK release or upstream.

QorIQ SDK ISOs can be downloaded from this location:
http://www.nxp.com/products/software-and-tools/run-time-software/linux-sdk/linux-sdk-for-qoriq-processors:SDKLINUX

Signed-off-by: Yutang Jiang <yutang.jiang@nxp.com>

PR #1221, opened 2017-07-17
layerscape: align layerscape source code to layerscape lsdk

yangbolu1991:
This patch-set is to update LEDE layerscape source code with NXP Layerscape LSDK.
https://lsdk.github.io
https://github.com/qoriq-open-source

Attached verification logs for current 5 layerscape platforms.

[layerscape-log.zip](https://github.com/lede-project/source/files/1152515/layerscape-log.zip)


yangbolu1991:
Any comments on this patch-set?
:)

package/boot/uboot-layerscape-armv8_32b/Makefile
```diff
@@ -8,18 +8,18 @@
...
-PKG_VERSION:=v1.0
+PKG_VERSION:=2017.03
 PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
-PKG_MIRROR_HASH:=a29edbeddde4b3007ac349c2a5ee434719049ac4753e459dbb1bab21e1c585da
-PKG_SOURCE_URL:=https://github.com/fsl-jyt/uboot-ls-32b.git
-PKG_SOURCE_VERSION:=6476195551ea83193afd06d0c441483c04729c8d
+PKG_MIRROR_HASH:=e8a3294cae1aa0b47fbe159f75051060ab2a81e0a52f7609f5306bd65428f636
+PKG_SOURCE_URL:=git@github.com:yangbolu1991/u-boot-lede.git
```
nbd168: Please fix the source URL, it should refer to a public one that does not require a github account.

nbd168: What I mean is:
`https://github.com/yangbolu1991/u-boot-lede.git` instead of `git@github.com:yangbolu1991/u-boot-lede.git`.
The first one allows public download, the second one requires a ssh key for github access.

package/boot/uboot-layerscape-armv8_32b/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+define Build/Compile
+
+endef
+
+define Build/InstallDev
+	$(INSTALL_DIR) $(STAGING_DIR_IMAGE)
+	$(CP) $(PKG_BUILD_DIR)/ls1043ardb-uboot.bin $(STAGING_DIR_IMAGE)/ls1043ardb-armv8_32b-uboot.bin
+	$(CP) $(PKG_BUILD_DIR)/ls1046ardb-uboot.bin $(STAGING_DIR_IMAGE)/ls1046ardb-armv8_32b-uboot.bin
+	$(CP) $(PKG_BUILD_DIR)/ls1012ardb-uboot.bin $(STAGING_DIR_IMAGE)/ls1012ardb-armv8_32b-uboot.bin
```
hauke: Why don't you build these binaries from source code inside of LEDE?
For GPL compliance you anyway have to provide the source code.

package/firmware/fman-ucode/Makefile
```diff
@@ -9,13 +9,13 @@ include $(TOPDIR)/rules.mk
...
 
 PKG_SOURCE_PROTO:=git
-PKG_MIRROR_HASH:=209c60e4efaa3ce9c9a954266a7210fa0047a2de47ce00a412aa0318d94d0d88
-PKG_SOURCE_URL:=https://github.com/fsl-jyt/fm-ucode.git
-PKG_SOURCE_VERSION:=b19c645821941493fbef32e616b5a16404259976
+PKG_MIRROR_HASH:=80d8d2454063b7f730ef49d7a44be8f8870711ee5109c53a0881b62026037a37
+PKG_SOURCE_URL:=https://github.com/qoriq-open-source/fm-ucode.git
+PKG_SOURCE_VERSION:=c275e91392e2adab1ed22f3867b8269ca3c54014
 
 PKG_BUILD_DIR=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)-$(BUILD_VARIANT)/$(PKG_NAME)-$(PKG_VERSION)
```
hauke: The License tag is wrong:
PKG_LICENSE:=GPL-2.0 GPL-2.0+This is not the GPL: https://github.com/qoriq-open-source/fm-ucode/blob/integration/Freescale-Binary-EULA

package/firmware/rcw/Makefile
```diff
@@ -9,13 +9,13 @@ include $(TOPDIR)/rules.mk
...
 
 PKG_SOURCE_PROTO:=git
-PKG_MIRROR_HASH:=8082979dbdef57e36a3096986aa426ef678c152403eb1dbd132da28f01490b6d
-PKG_SOURCE_URL:=https://github.com/fsl-jyt/rcw.git
-PKG_SOURCE_VERSION:=17e09e19bd5cf94e5ebc84916b4607743c29723e
+PKG_MIRROR_HASH:=8fcaed3bc0230a84863eca333a1cd0fc700db783b236eaa35c1f25060ca85a09
+PKG_SOURCE_URL:=https://github.com/qoriq-open-source/rcw.git
+PKG_SOURCE_VERSION:=6eed8fdfc07920178b70fba199fa20fb42adb20b
 
 PKG_BUILD_DIR=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)-$(BUILD_VARIANT)/$(PKG_NAME)-$(PKG_VERSION)
```
hauke: This package is licensed under some BSD license and not GPL:
https://github.com/qoriq-open-source/rcw/blob/integration/LICENSE

package/firmware/rcw/Makefile
```diff
@@ -78,6 +78,7 @@ define Build/Configure
 endef
 
 define Build/Compile
+	$(MAKE) -C $(PKG_BUILD_DIR)
 endef
```
hauke: you can remove the "define Build/Compile" block completely, this is more or less what "define Build/Compile/Default" does.

target/linux/layerscape/armv8_32b/target.mk
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (C) 2016 Jiang Yutang <jiangyutang1978@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+ARCH:=arm
+BOARDNAME:=ARMv8 32-bit based boards
+CPU_TYPE:=cortex-a9
```
hauke: What type of CPUs are supported by this target?
Is it a cortex A9 or some other CPU like a cortex A53?

target/linux/layerscape/armv8_64b/config-default
```diff
@@ -0,0 +1,4667 @@
+#
```
hauke: Please run "make kernel_oldconfig CONFIG_TARGET=subtarget" when this subtarget is selected to reduce the size of this configuration file.

target/linux/layerscape/armv8_32b/config-default
```diff
@@ -0,0 +1,6337 @@
+#
```
hauke: Please run "make kernel_oldconfig CONFIG_TARGET=subtarget" when this subtarget is selected to reduce the size of this configuration file.

target/linux/layerscape/armv8_64b/target.mk
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (C) 2016 Jiang Yutang <jiangyutang1978@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+ARCH:=aarch64
+BOARDNAME:=ARMv8 64-bit based boards
+CFLAGS:=-Os -pipe -fno-caller-saves
```
hauke: Is the CFLAGS setting really needed? LEDE should use similar compiler options by default.

target/linux/layerscape/armv8_64b/target.mk
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (C) 2016 Jiang Yutang <jiangyutang1978@gmail.com>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+ARCH:=aarch64
```
hauke: If the SoC supported by this target all have an cortex A53 CPU, please add:
CPU_TYPE:=cortex-a53

target/linux/layerscape/patches-4.4/0001-arch-Introduce-post-init-read-only-memory.patch
```diff
@@ -0,0 +1,114 @@
+From 7ece66e296972744ec471d521b8ca07990e2494f Mon Sep 17 00:00:00 2001
```
hauke: This commit and probably many more are already in kernel 4.9, could you please update this target to kernel 4.9.
The next LEDE release will be based on kernel 4.9 anyway so you have to update this target when the layerscape target should be included in the next release.

package/boot/uboot-layerscape-armv8_32b/Makefile
yangbolu1991: Oh... Sorry for my misunderstanding. I will update the link in next version.

package/boot/uboot-layerscape-armv8_32b/Makefile
yangbolu1991: We always used 64b u-boot image to support both 32b and 64b kernel and rootfs. It's convenient.
So this uboot-layerscape-armv8_32b here used same image/code with uboot-layerscape-armv8_64b.
The reason we used image is that LEDE could only provide 32b toolchain for 32b target.

package/firmware/rcw/Makefile
yangbolu1991: Thanks. Will remove this in next version.

target/linux/layerscape/armv8_64b/target.mk
yangbolu1991: Actually, ls1012a/ls1088a/ls1043a is cortex A53 cpu while ls1046a/ls2088a is cortex A57 cpu.

target/linux/layerscape/armv8_64b/target.mk
yangbolu1991: I just tried to compile without that. There was no problem. Let me remove it in next version. Thanks.

target/linux/layerscape/armv8_32b/target.mk
yangbolu1991: Layerscape CPU is cortex A53 or cortex A72 now, but we want it to run on 32-bit mode which is compatible with ARMv7.
So we set CPU_TYPE as cortex-a9 to select 32b toolchain.

target/linux/layerscape/patches-4.4/0001-arch-Introduce-post-init-read-only-memory.patch
yangbolu1991: NXP LSDK will release kernel v4.9 on September 2017. We will update our targets to v4.9 in LEDE on September.
But currently customers still use v4.4, this patch-set is still needed.

May I know whether LEDE will still maintain v4.4 after v4.9 is released?

target/linux/layerscape/armv8_64b/config-default
yangbolu1991: Thanks a lot for letting me know this method.
No problem.

target/linux/layerscape/armv8_32b/config-default
yangbolu1991: Thanks a lot for letting me know this method.
No problem.

package/boot/uboot-layerscape-armv8_32b/Makefile
yangbolu1991: Any suggestion here? Or should I add a note to explain in Makefile?

package/firmware/fman-ucode/Makefile
yangbolu1991: Oh, you're right... These're binaries developed by Freescale/NXP.
Let me remove the error PKG_LICENSE.

May I know whether PKG_LICENSE is required in Makefile? Could I just ignore it.
fman-ucode is neither GPL or BSD, it's just Freescale-Binary-EULA.

package/firmware/rcw/Makefile
yangbolu1991: I will correct it in next version. :)

target/linux/layerscape/armv8_64b/target.mk
fsl-jyt: Hi @yangbolu1991 ,
Please double check the libnl-tiny-0.1
Without that CFLAGS, it will arise compile error, please have a look: 
https://github.com/lede-project/source/pull/344

nbd168: What kind of compile error? I don't see an error log anywhere in there...

yangbolu1991: @fsl-jyt Thanks for providing the information. 
I just double checked the compiling(make distclean and make, for NXP target) and no problem. Maybe the old issue had been fixed.

So I think there should be no problem to remove it.

yangbolu1991:
@hauke @nbd168 @fsl-jyt 
I updated the patch-set with below changes to address your comments.
Please help to review. Thanks a lot :)

layerscape: take over maintainership
Changes for v2:
	- None

layerscape: rename subtargets and update makefile files
Changes for v2:
	- None

layerscape: update packages with LSDK git trees
Changes for v2:
	- Fixed u-boot-layerscape-armv8_32b url with public url.
	- Added note to explain why used u-boot images for 32b targets, and removed wrong license info.
	- Removed wrong license info of fman_ucode.
	- Corrected rcw license info with BSD.
	- Removed useless Build/Compile definition for rcw.
	
layerscape: update memory layout according to LSDK
Changes for v2:
	- None

layerscape: update linux to LSDK linux
Changes for v2:
	- Reduced size of kernel config for 32b/64b targets.
	- Removed useless CFLAGS for 64b targets.

yangbolu1991:
Any comments? 
Thanks a lot.

target/linux/layerscape/armv8_32b/target.mk
yangbolu1991: @chunkeey 
Sorry for my late response. I just come back to office today from holiday.
I didn't insist A9 ... In current layerscape platforms, some are with A53 cpus and some are with A72 cpus. More details,
http://www.nxp.com/products/microcontrollers-and-processors/arm-processors/qoriq-layerscape-arm-processors:QORIQ-ARM

I don't know which CPU_FLAGS is proper for *all* of them and doesn't introduce any potential issue, but at least platforms worked fine with current A9.
You suggested to use A7 or A15, which one exactly should I use in your view?
Should I also add CPU_SUBTYPE:=neon-vfpv4 ?
I really would like to fix it with proper value.

Regarding to BOARDNAME, I think it seems proper because they are indeed ARMv8 32-bit.

@hauke @nbd168 @fsl-jyt 
Do you have any suggestion on CPU_TYPE?
Thanks a lot :)

yangbolu1991:
Just rebased this patch-set.

yangbolu1991:
Could we consider to merge this pull request if there is no other change request?
This patch-set is mainly to update all packages code and kernel patches to Layerscape SDK.
Regarding to CPU_TYPE discussion, we can still continue or start a email thread after this PR was closed, and I can send a seperate patch for that if need.

Thanks a lot :)

target/linux/layerscape/armv8_32b/target.mk
hauke: As the layerscape 32 biot target is not so popular I would suggest choosing something which is already used by some other target, see here: https://downloads.lede-project.org/snapshots/packages/

I would suggest this:
CPU_TYPE:=cortex-a15
CPU_SUBTYPE:=neon-vfpv4 
This is probably a better fit than the A9

target/linux/layerscape/armv8_64b/target.mk
hauke: yes please remove this

target/linux/layerscape/armv8_64b/target.mk
hauke: I am ok with this, then you can share the packages with the armvirt target.

package/boot/uboot-layerscape-armv8_32b/Makefile
hauke: this is ok with me

package/firmware/fman-ucode/Makefile
hauke: This looks ok to me now

target/linux/layerscape/patches-4.4/0001-arch-Introduce-post-init-read-only-memory.patch
hauke: I do not like all these patches being added you are ending up with 835 patches on top of kernel 4.4. Please do the migration to kernel 4.9 then you can probably get rid of many patches.

target/linux/layerscape/armv8_32b/target.mk
yangbolu1991: Thanks a lot for @chunkeey @hauke  suggestion! 
It's ok for me. I will fix this in next version.

target/linux/layerscape/patches-4.4/0001-arch-Introduce-post-init-read-only-memory.patch
yangbolu1991: @hauke , 
Almost all these patches are for layerscape IP modules. We are also working on kernel 4.9 now. Even we migrate to kernel 4.9, I'm afraid the number of patches would not shrunk so much. 

The large number patches mainly comes from dts (155 patches), mc-bus (139 patches), sec (138 patches), qspi (80 patches). What I can do is to merge each IP module patches into one. We can provide only one patch for each module/component. Then we will only have dozens of patches finally.

It's also easy for maintaining, becasue we have each module as a feature branch when backport patches.

Do you think is it ok? Let me generate a new version for your reveiwing.
Thank you very much.

yangbolu1991:
I have updated the patch-set with v3 version.
See below changes for v3,

layerscape: take over maintainership
 Changes for v3:
 - None

layerscape: rename subtargets and update makefile files
 Changes for v3:
 - None

layerscape: update packages with LSDK git trees
 Changes for v3:
 - None

layerscape: update memory layout according to LSDK
 Changes for v3:
 - None

layerscape: update linux to LSDK linux
 Changes for v3:
 - fixed CPU_TYPE issue for 32-bit.
 - merged each IP module kernel patches into one.

yangbolu1991:
@hauke 
v3 patch-set fixed the CPU_TPYE problem, and merge each IP module patches into one. Now there are only 36 patches.
Do you think is it ok?

yangbolu1991:
Hi @hauke ,
Sorry for disturbing.  Did you have a chance to have a look at this?
Thanks.

yangbolu1991:
@hauke @chunkeey @nbd168 @fsl-jyt 
Hi all,

Rebased this patch-set again since Hauke's below patch introduced conflicts. Actually this changes had already in this patch-set as we discussed.
a5d26d7 layerscape: do not add custom CFLAGS.

Could we speed up the reviewing? I think all current comments had been addressed, right?
Thanks a lot.

yangbolu1991:
I splited patch "layerscape: update linux to LSDK linux" into 5 patches.
I think it will be easier for you to review.

See https://github.com/lede-project/source/pull/1327

PR #1224, opened 2017-07-19
[badvpn, sshpass, corkscrew, httping] - Add some new packages

hillz2:
Add some new packages, they all are related to networking. The reason why I want to submit these packages is so that I wouldn't have to compile each package every time I install a new version of LEDE or OpenWrt when they are finally merged.

> Signed-off-by: Galih Prastowo Aji <galih.prastowo@gmail.com>


mkresin:
These packages should go into the [packages feed](https://github.com/openwrt/packages/). Do not get confused by the fact that it is in the OpenWrt repository. The packages feed is compatible to and used by OpenWrt as well as LEDE.

diizzyy:
Also, should be split into separate pull requests.

PR #1227, opened 2017-07-21
kernel: switch to BBR congestion algorithm

dissent1:
BBR is a perspective TCP congestion algorithm developed by google. It provides better congestion handling thus better bandwidth and less bufferbloat.

Info about BBR: http://queue.acm.org/detail.cfm?id=3022184

The more ppl switch to it the better internet will be :)


anomeome:
Everything I have read regarding BBR has always indicated it to be server side deployment protocol. Not a protocol for a device utilized _in the link_ as on a router.

neheb:
This is correct.

FYI, BBR also requires TCP pacing support. Only sch_fq and cake support this. Without TCP pacing results are pathetic.

Having said that, BBR could be useful if running servers on the router like OPEN ON or BitTorrent.

dissent1:
A router is a server for a home network and effective congestion algorithm helps to deal with low wifi signal as a main benefit. Also it helps in peer to peer connections, upload bandwidth.

Cubic is outdated, westwood or yeah perform much better in all cases. Have met a lot of situations when switching off cubic helped to deal with voip/video over ip connection issues. 
In my experience bbr is definitely not worse than those 2.

I am not that theory expert I just operate on what I see.
Some throughout bench can be found here https://github.com/osmc/osmc/pull/375

If you think opposite, well, then I'm mistaken obviously.

dissent1:
And btw, it maybe really strange but on one ISP switching off cubic drastically (up to 2 times) increased torrent **downloading** (uploading remained the same). I've retested for cpl of days, the result was affected directly.

anomeome:
For _some people_ their router also acts a a server, but that is not necessarily the standard scenario, nor the standard use case of LEDE OOTB. There are many with hundreds of installs, and have no interest in devices behaving as one. I would argue anyone wanting such, should build an image with _special configuration settings_ themselves.

Also, I don't know how many bytes this may add to kernel size, but it might push _less resourceful_ devices over the edge; the very kind of devices that tend be be deployed in those large install bases.

neheb:
what's the size difference between CUBIC, BBR, and Westwood?

dissent1:
I might have explained incorrectly, I mean that router is always a server for home network, a server that routes traffic and _transmits_ it to local devices. Low signal wifi clients benefit from a good congestion control very much.

I actually don't understand what do you have against, if you are against bbr in particular - ok with me, but cubic is a no go against more advanced congestion algorithms. I haven't seen any recent research where cubic outperforms westwood, yeah, bbr.

I don't know what's the size impact actually, but for low flash targets it cab be switched back to cubic or whatever.

neheb:
If the size increase is negligible, I don't see why not.

I am interested in replacing CUBIC since in my experience it tends to work poorly compared to Westwood (what I used to use with DD-WRT.).

wongsyrone:
For new kernels, we have pacing support integrated.

https://github.com/torvalds/linux/commit/218af599fa635b107cfe10acf3249c4dfe5e4123

kdarbyshirebryant:
I long ago remember the clamour to get tcp vegas into 'tomato' - it wasn't worthwhile and was admitted as a fools errand because in a typical router/wireless AP environment very little traffic originates/terminates at the router.  I see this as a similar scenario.  It's TCP only, it's endpoint-to-endpoint only (and thus no intervening router can directly tell BBR TCP to speed up/slow down). There is still need to shape traffic travelling *through* the router (think UDP etc) and there are plenty of qdiscs available to help/hinder that.  I suspect that the vast majority of home/small installations (some using the router as a file server) accessed over wifi would benefit far more from things such as airtime fairness (in ath9k & mt76 I think) and improved rate selection - that's where I'd invest my money/time.

dissent1:
I clearly understand statements made by network experts that congestion control has nothing to do with home setup router, but my practice tells different and that's not placebo, I've made a lot of tests back in the time engaging different algorithms (YEaH was the best in may environment). And each of them performed better than cubic in low signal area - i could skype without issues, most of them helped me increase throughput.
I don't know why, maybe stars are setting up in a proper order each time I switch but it has an effect (not as a server in your definition).

By the way, seems that internal pace is easily portable and is comming after this one https://github.com/torvalds/linux/commit/3f4888adae7c1619b990d98a9b967536f71822b8 that has a nice effect on tcp performance and is also easily portable.

I cane make a PR for that one if needed.

zorun:
This is useless in general for LEDE, because as said before, the congestion control algorithm of TCP is an endpoint property.  It would only affect TCP connections terminated on the LEDE device.  By the way, Bittorrent clients today mostly use UDP with LEDBAT, not TCP.  I don't know how many people run TCP-based services on their LEDE devices, but I would guess not a lot.

Also, from my tests, BBR is not so great on wireless.  I think this is because capacity evolves too fast for BBR to adapt (it only probes for more capacity every ~10 seconds if I remember correctly).  So this could actually hurt the few use-cases where this has an effect (TCP service running on the device, with TCP client behind wireless).

dissent1:
Okay, to sum things up:
1. We know that router being a server will benefit from it and it's not that what common users do.
2. No matter that it has a positive effect in practice, statement that wifi behaves better in low signal area is not accepted by networking theorists also :)

Let's go from the other side - what's the drawback of switching from cubic?

champtar:
@dissent1 what size are we talking about ?
Also as said by @neheb ```bbr``` need ```fq``` or ```cake```, which I think are not the default,
so your commit is not acceptable as is ...

neheb:
More info on that: https://news.ycombinator.com/item?id=14814530

A fair amount of testing is needed. I switched to Westwood and found that WiFi performance dropped massively on my R7800.

dissent1:
Compiled bbr.o is of 50kb in size, the same as cubic.o.
I've already ported that internal pace along with tx throughput commit into my working copy, so that's not a problem to adjust the PR.

I'm not blindly proposing to switch to bbr. If there're drawbacks for mass market for real, then I'm washing my hands :)

Would be really meaningful if devs put their opinion on it :)

tohojo:
As @neheb mentioned above, BBR **requires** TCP pacing to work properly; which means it only works with sch_fq installed as a qdisc (cake does **not** have pacing support). Using sch_fq is not appropriate for a router, especially not a small device, since it will allocate memory for *every* flow that it sees, leading to potential overflows.

The benefit of BBR on a LEDE device is probably also going to be low; only services running **on the router** will benefit, and those are likely to be servicing clients over the local network, where BBRs benefits are not as great (especially not over WiFi).

I believe there has been some talk about adding pacing support to FQ-CoDel upstream; when that lands, running BBR on the router may be practical. Until then, I would advise against it except perhaps as an advanced option for users who want to experiment (and presumably know what they are doing).

dissent1:
> As @neheb mentioned above, BBR requires TCP pacing to work properly; which means it only works with sch_fq installed as a qdisc (cake does not have pacing support). Using sch_fq is not appropriate for a router, especially not a small device, since it will allocate memory for every flow that it sees, leading to potential overflows.

There's internal pacing support in k4.13-rc1 https://github.com/torvalds/linux/commit/218af599fa635b107cfe10acf3249c4dfe5e4123 that is portable into k4.9, so no need for fq_codel.
As a side note there a nice coincidence happened: when a guy wanted to pull Shortcut FE driver from my working tree, he merged my whole tree with switching to BBR. His device is ipq8065 based, so that Shortcut FE driver should have had no effect for him as it already does up to 1gbit wired and 500-600 wireless at 80mhz. Alas here's the result https://forum.lede-project.org/t/qualcomm-fast-path-for-lede/349
Excluding Shortcut FE it's only switch to BBR could provide such effect in my tree.

tohojo:
Ah, yes, that was the pacing support I meant, I guess. Wasn't aware that it had landed already (and apologies for missing the link in the thread above).

So, backporting that should make it **possible** to run BBR on the router. I'm still sceptical that it will have much of an impact, though. As I read the post you linked to, that user is seeing an improvement for traffic *going through* the router; which is most surely not causes by BBR.

I would have no issue with backporting the pacing commit and building BBR as a kmod package for users to install; making it the default is probably premature...

neheb:
To be fair, I switched from CUBIC to Westwood and found the wireless almost unusable. The CC algorithm has some kind of effect for some unexplained reason. Maybe something to do with the switch...

I'm currently testing BBR and so far have not noticed any negative effects. I don't want to say positive since it's probably placebo, but nothing negative.

stintel:
> I would have no issue with backporting the pacing commit and building BBR as a kmod package for users to install; making it the default is probably premature...

I agree. While at it, maybe also backport torvalds/linux@2660bfa84e9236016f3a4f21b7864431d9663338 and torvalds/linux@9a568de4818dea9a05af141046bd3e589245ab83?

dissent1:
@stintel 
I suppose this one https://github.com/torvalds/linux/commit/2660bfa84e9236016f3a4f21b7864431d9663338 doesn't provide any effect
I've also tried backporting https://github.com/torvalds/linux/commit/9a568de4818dea9a05af141046bd3e589245ab83 before, but it has so many conflicts while applying so I've got scared with going further, not sure if it's easily backportable.
Anyway nbd168 asked me to do some throughput tests. I'll be doing it by the weekend.

dissent1:
Status update:
one of these upstream kernel commits lead to a severe performance degradation when testing tx with iperf originated from the router - drops from 450mbits to 10-15mbits. 

Maybe something is missing.

dissent1:
FYI, here's tx throughput with iperf3 default packet size. Server - amd e450 based, client - Netgear R7800, ipq8065 based.
Comparison with https://github.com/lede-project/source/pull/1227/commits/14ed6233944c579c54a117c885d8c99cdbac38b7 applied and without

With commit applied
-----------------------------------------------------------
Accepted connection from 192.168.1.1, port 35966
[  5] local 192.168.1.2 port 5201 connected to 192.168.1.1 port 35968
[ ID] Interval           Transfer     Bandwidth
[  5]   0.00-1.00   sec  26.3 MBytes   221 Mbits/sec
[  5]   1.00-2.00   sec  57.3 MBytes   481 Mbits/sec
[  5]   2.00-3.00   sec  57.2 MBytes   480 Mbits/sec
[  5]   3.00-4.00   sec  57.0 MBytes   478 Mbits/sec
[  5]   4.00-5.00   sec  57.3 MBytes   481 Mbits/sec
[  5]   5.00-6.00   sec  57.1 MBytes   479 Mbits/sec
[  5]   6.00-7.00   sec  56.9 MBytes   477 Mbits/sec
[  5]   7.00-8.00   sec  57.1 MBytes   479 Mbits/sec
[  5]   8.00-9.00   sec  57.1 MBytes   479 Mbits/sec
[  5]   9.00-10.00  sec  56.5 MBytes   474 Mbits/sec
[  5]  10.00-10.03  sec  1.83 MBytes   477 Mbits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth
[  5]   0.00-10.03  sec  0.00 Bytes  0.00 bits/sec                  sender
[  5]   0.00-10.03  sec   542 MBytes   453 Mbits/sec                  receiver
-----------------------------------------------------------

Without the commit
-----------------------------------------------------------
Accepted connection from 192.168.1.1, port 53756
[  5] local 192.168.1.2 port 5201 connected to 192.168.1.1 port 53758
[ ID] Interval           Transfer     Bandwidth
[  5]   0.00-1.00   sec  13.3 MBytes   112 Mbits/sec
[  5]   1.00-2.00   sec  46.8 MBytes   393 Mbits/sec
[  5]   2.00-3.00   sec  50.7 MBytes   425 Mbits/sec
[  5]   3.00-4.00   sec  52.4 MBytes   439 Mbits/sec
[  5]   4.00-5.00   sec  52.2 MBytes   438 Mbits/sec
[  5]   5.00-6.00   sec  52.4 MBytes   440 Mbits/sec
[  5]   6.00-7.00   sec  52.3 MBytes   439 Mbits/sec
[  5]   7.00-8.00   sec  52.3 MBytes   439 Mbits/sec
[  5]   8.00-9.00   sec  52.4 MBytes   440 Mbits/sec
[  5]   9.00-10.00  sec  52.3 MBytes   439 Mbits/sec
[  5]  10.00-10.03  sec  1.58 MBytes   439 Mbits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth
[  5]   0.00-10.03  sec  0.00 Bytes  0.00 bits/sec                  sender
[  5]   0.00-10.03  sec   479 MBytes   400 Mbits/sec                  receiver
-----------------------------------------------------------

The result is pretty steady and repeats over the time.

The pacing commit breaks throughput totally.

montvid:
Yeah, I would test it as a module if it is possible to make a module for this. Thanks for your work @dissent1.

dissent1:
I was wrong about enabling BBR as is and I was not able to backport internal implementation of pacing properly - it caused a huge regression.

I’ve opened a new PR#1333 with cherry-picking tx perf patch + 1 more

ohsorry:
bbr is really useful, i use LEDE as proxy router to bypass GFW.
[https://forum.lede-project.org/t/tcp-bbr-support/1168/13](url)

ohsorry:
A 'proxy router' , i mean an  [openwrt-shadwosocks proxy client](https://github.com/shadowsocks/openwrt-shadowsocks) running on LEDE router.
without tcp_bbr the tcp speed is only below 20KB/s.

jack338c:
@dissent1 
hello,I have applyed your patch to my router
but I 'm not sure it is working or not?
Could you please tell me how to verify on my openwrt router
thanks

PR #1228, opened 2017-07-21
samba36: Remove legacy options

neheb:
browseable is now set through LuCI per share, so remove it. Same with writeable (inverted synonym for read only). domain master and preferred master seem to be legacy settings for Windows 9x. encrypt passwords defaults to yes. Probably should not be disabled either.


jow-:
Merged with 9dcb3fe7eb06b42c4f026e6cc5e89b2a18932782, thank you.

PR #1230, opened 2017-07-22
    ramips: add support for the GnuBee Personal Cloud One

pinney:
The GnuBee Personal Cloud One crowdfunded on https://www.crowdsupply.com
    It is a low-cost, low-power, network-attached storage device.

    Specifications:

    - SoC: MediaTek MT7621AT
    - RAM: DDR3 512 MB
    - Flash: 32 MB
    - Six SATA ports for 2.5" Drives
    - One micro SDcard
    - One USB 3.0
    - Two USB 2.0
    - Gigabit Ethernet: 1 x WAN and 1 x LAN
    - UART 3.5mm Audio Jack or 3 pin header - 57600 8N1
    - Four GPIOs available on a pin header

    Flash instructions:

    The GnuBee Personal Cloud One ships with libreCMC installed.
    libreCMC is a Free Software Foundation approved fork of LEDE/OpenWrt.
    As such one can upgrade using the webinterface or sysupgrade.

    Das U-Boot has multiple options for recovery or updates including :

    - USB
    - http
    - tftp

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -183,6 +183,10 @@ fonera20n)
 	set_usb_led "$board:orange:usb"
 	set_wifi_led "$board:orange:wifi"
 	;;
+gb-pc1)
+	ucidef_set_led_netdev "lan1" "lan1" "$board:green:lan1" "eth0.1"
+	ucidef_set_led_netdev "lan2" "lan2" "$board:green:lan2" "eth0.2"
+	;;
 gl-mt300a|\
 gl-mt300n|\
 gl-mt750)
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -183,6 +183,10 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "0:wan" "6@eth0"
 		;;
+	gb-pc1)
+		ucidef_add_switch "switch0" \
+			"0:lan" "4:lan" "6@eth0"
+		;;
 	gl-mt300n-v2)
 		ucidef_add_switch "switch0" \
 			"1:lan" "0:wan" "6@eth0"
--- a/target/linux/ramips/dts/GB-PC1.dts
+++ b/target/linux/ramips/dts/GB-PC1.dts
@@ -1,0 +1,120 @@
+/dts-v1/;
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "gnubee,gb-pc1", "mediatek,mt7621-soc";
+	model = "GB-PC1";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x1c000000>, <0x20000000 0x4000000>;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,57600";
+	};
+
+	palmbus: palmbus@1E000000 {
+		i2c@900 {
+			status = "okay";
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio0 18 GPIO_ACTIVE_HIGH>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		system {
+			label = "gb-pc1:green:system";
+			gpios = <&gpio0 6 GPIO_ACTIVE_LOW>;
+		};
+
+		status {
+			label = "gb-pc1:green:status";
+			gpios = <&gpio0 8 GPIO_ACTIVE_LOW>;
+		};
+
+		lan {
+			label = "gb-pc1:green:lan";
+			gpios = <&gpio0 24 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			label = "gb-pc1:green:wan";
+			gpios = <&gpio0 25 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&sdhci {
+	status = "okay";
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,chunked-io = <32>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x30000>;
+			read-only;
+		};
+
+		partition@30000 {
+			label = "u-boot-env";
+			reg = <0x30000 0x10000>;
+			read-only;
+		};
+
+		factory: partition@40000 {
+			label = "factory";
+			reg = <0x40000 0x10000>;
+			read-only;
+		};
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0x1fb0000>;
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xe000>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "wdt", "rgmii2", "uart3", "ge2";
+			ralink,function = "gpio";
+		};
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -73,6 +73,14 @@ define Device/firewrt
 endef
 TARGET_DEVICES += firewrt
 
+define Device/gb-pc1
+  DTS := GB-PC1
+  DEVICE_TITLE := GnuBee Personal Cloud One
+  DEVICE_PACKAGES := kmod-ata-core kmod-ata-ahci kmod-usb3 kmod-sdhci-mt7620
+  IMAGE_SIZE := $(ralink_default_fw_size_32M)
+endef
+TARGET_DEVICES += gb-pc1
+
 define Device/hc5962
   DTS := HC5962
   BLOCKSIZE := 128k
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -77,6 +77,7 @@ get_status_led() {
 	dir-620-a1|\
 	dir-620-d1|\
 	dwr-512-b|\
+	gb-pc1|\
 	hpm|\
 	hw550-3g|\
 	mac1200rv2|\
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -208,6 +208,9 @@ ramips_board_detect() {
 	*"FreeStation5")
 		name="freestation5"
 		;;
+	*"GB-PC1")
+		name="gb-pc1"
+		;;
 	*"GL-MT300A")
 		name="gl-mt300a"
 		;;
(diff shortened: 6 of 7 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -183,6 +183,9 @@ fonera20n)
 	set_usb_led "$board:orange:usb"
 	set_wifi_led "$board:orange:wifi"
 	;;
+gb-pc1)
```
mkresin: drop this, the led is handled by diag.sh.

target/linux/ramips/dts/GB-PC1.dts
```diff
@@ -0,0 +1,157 @@
+/dts-v1/;
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	model = "GB-PC1";
```
mkresin: please add a compatible string here. check the recent ramips board support commits for examples.

target/linux/ramips/dts/GB-PC1.dts
```diff
@@ -0,0 +1,157 @@
...
+			gpios = <&gpio0 24 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			label = "gb-pc1:green:wan";
+			gpios = <&gpio0 25 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-export {
```
mkresin: What is the purpose of the whole gpio export block? We are usually using it to enable usb power and so on.

target/linux/ramips/dts/GB-PC1.dts
```diff
@@ -0,0 +1,157 @@
...
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x30000>;
+			read-only;
+		};
+
+		partition@30000 {
+			label = "u-boot-env";
+			reg = <0x30000 0x10000>;
+		};
```
mkresin: add a read-only here as well.

target/linux/ramips/dts/GB-PC1.dts
```diff
@@ -0,0 +1,157 @@
...
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0x1fb0000>;
+		};
+	};
+};
+
+&cpuclock {
```
mkresin: Why do you need clocks different to the ones defined in mt7621.dtsi? As far as I can see, no other mt7621 board has these.

target/linux/ramips/dts/GB-PC1.dts
```diff
@@ -0,0 +1,157 @@
...
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xe000>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "wdt", "rgmii2", "mdio";
```
mkresin: As far as I can see, none of the mdio group gpios are used. Instead the uart3 gpios are used for the system and status led.

target/linux/ramips/image/mt7621.mk
```diff
@@ -73,6 +73,15 @@ define Device/firewrt
 endef
 TARGET_DEVICES += firewrt
 
+define Device/gb-pc1
+  DTS := GB-PC1
+  DEVICE_TITLE := GnuBee Personal Cloud One
+  DEVICE_PACKAGES := kmod-ata-core kmod-ata-ahci
+  IMAGE_SIZE := $(ralink_default_fw_size_32M)
+endef
+
```
mkresin: please drop this blank line.

target/linux/ramips/base-files/etc/board.d/01_leds
pinney: OK no problem.
Perhaps I should define the lan/wan LEDs?

target/linux/ramips/dts/GB-PC1.dts
pinney: Sure I can add one ... But is a "compatible string" really required ?
Seems to work OK without one.

target/linux/ramips/dts/GB-PC1.dts
pinney: If we mark this read-only users cannot use the uboot-envtools package properly.

target/linux/ramips/dts/GB-PC1.dts
pinney: 880MHz is boring ... This SoC can and will run much faster.
This device shipped with this clock setting.
I've been running my ZBT-WG2626 & WE1326 @ much higher clock rates.
BogoMIPS                : 2129.92
compared to the stock figure of :
BogoMIPS                : 583.68

target/linux/ramips/dts/GB-PC1.dts
pinney: Remove "mdio" OK

target/linux/ramips/image/mt7621.mk
pinney: No Problem.

target/linux/ramips/dts/GB-PC1.dts
pinney: The purpose here is to enable the Four GPIOs available on a pin header

target/linux/ramips/base-files/etc/board.d/01_leds
mkresin: If they are not controlled by the hardware, yes.

target/linux/ramips/dts/GB-PC1.dts
mkresin: Yes it is a formal requirement to dts files. Try to upstream a dts without board specific compatible string and see what happens.

target/linux/ramips/dts/GB-PC1.dts
mkresin: Please drop it. A dts should describe the hardware and should not implement userspace stuff. As I already outlined, we are using the gpio-export only in specific cases where an GPIO need the be pulled to get something working.

target/linux/ramips/dts/GB-PC1.dts
mkresin: I'm quite sure we had this discussion before. If there is no need to change the env (to updated checksums or similar) it should be marked as read only. So please mark it as read only.

target/linux/ramips/dts/GB-PC1.dts
mkresin: Some as above. The dts should describe the hardware. Beside of that I'm not keen of doing overclocking by default. Please drop the nodes.

target/linux/ramips/dts/GB-PC1.dts
mkresin: ....and add uart3

target/linux/ramips/dts/GB-PC1.dts
pinney: I checked the dts(i) files for MT7621 devices.
I did not find a single example of a "compatible string"
Would you like me to add one for every MT7621 device?

target/linux/ramips/dts/GB-PC1.dts
pinney: The pins are part of the hardware and installed at the factory.
Why can I not export these pins like is done on the VOCORE device?

target/linux/ramips/dts/GB-PC1.dts
pinney: The SamKnows Whitebox 8 has it's "u-boot-env" set writable.
Can I set the "u-boot-env" writable on my device as well?

target/linux/ramips/dts/GB-PC1.dts
pinney: The .dts does describe the hardware...but not the Mediatek default clock setting meant to insure functionality with DDR2 RAM and other design factors.
I spent many hours and much money during the design and testing of this hardware.
That includes days of running the prototypes and other MT7621 devices at a higher clock rate and running tests such as 'stress'.
The ODM/OEM (me) shipped the devices using these clock settings.
Do all MT7621 devices from the "factory" have the clock set @ 880MHz?
And if so why can I not set my clock @ 2GHz ?

target/linux/ramips/dts/GB-PC1.dts
bobafetthotmail: Can this device boot any other OS that isn't LEDE/OpenWRT (or its stock that is also the same thing)? Because if it can't boot anything else anyway....

Not that it matters much, **mkresin** does not like uboot env access even on kirkwood devices where it makes sense because they can boot mainline linux (Debian/Arch ARM). So for this thing either you do as he asks or you wait for another developer to merge this.

pinney: @bobafetthotmail Yes it will run Debian using the "OpenWrt" kernel.

target/linux/ramips/dts/GB-PC1.dts
pinney: Checked more thoroughly and found two examples : 
The "mt7621-eval-board" ... and your recent commit for the Phicomm K2P by Chuanhong Guo
Would you like me to add compatible string to the other MT7621 devices?

target/linux/ramips/dts/GB-PC1.dts
bobafetthotmail: @pinney yeah but that does not require changing uboot envs right? you just plop in the SDcard. Besides, if your uboot supports uEnv.txt files (reading envs from a text file in a boot partition, sunxi and cubieboard uboots do that) you can just have them read-only to appease people here and not lose the ability to boot whatever without attaching a serial.

mkresin:
Thanks a lot for doing all the requested changes.

> Checked more thoroughly and found two examples :
The "mt7621-eval-board" ... and your recent commit for the Phicomm K2P by Chuanhong Guo
Would you like me to add compatible string to the other MT7621 devices?

If you have time, yes please.

> Perhaps I should define the lan/wan LEDs?

I'm quite sure these were defined in the boardfile the last time I had a look at it. But now they are gone. Personally, I would setup both RJ45 ports by default as LAN. A WAN port on a NAS is somehow unusual.

If you decide to setup led triggers, please add ```CONFIG_SWCONFIG_LEDS=y``` to the kernel config and use something like:

```
ucidef_set_led_switch "lan" "lan" "$board:blue:lan" "switch0" "0x01"
```

The LED will be triggered by the corresponding switch port (0x01 == 000001 = switch port 0) this way, which works perfect for your use case, since you have a LED for each RJ45 port.

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -78,6 +78,7 @@ ramips_setup_interfaces()
 	dir-320-b1|\
 	dir-610-a1|\
 	dir-615-h1|\
+	gb-pc1|\
```
mkresin: Albeit it will work this way, it is wrong. The gb-pc1 has only two ports attached to the switch according to your commit message. The case block is for a five port config.

pinney: Sorry about that one.
Can you please fix that one ... my github is out of wack.
I think it related to being in China and behind the firewall.

I'll start on the MT7621 "compatible strings" tonight.

And Thanks!

mkresin: I can not fix it without knowing to which switch ports the RJ45 jacks are connected.

pinney: RJ45 jacks are connected to 0 and 4 ...

I pushed this :

	zbt-wa05)
		ucidef_add_switch "switch0" \
			"1:lan" "2:lan" "3:lan" "4:lan" "0:wan" "6@eth0"
		;;
	gb-pc1)
		ucidef_add_switch "switch0" \
			"0:lan" "4:lan" "6@eth0"
		;;
	gl-mt300n-v2)
		ucidef_add_switch "switch0" \
			"1:lan" "0:wan" "6@eth0"
		;;

target/linux/ramips/base-files/etc/board.d/01_leds
pinney: and I pushed this too ... still not sure what happened.
gb-pc1)
	ucidef_set_led_netdev "lan" "lan" "$board:green:lan" "eth0.1"
	ucidef_set_led_netdev "wan" "wan" "$board:green:wan" "eth0.2"
	;;

pinney: So now that network changed ... this need changed as well

mkresin:
I've applied the remaining fixes regarding the lan leds to the version in my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

Would you please test my changes on real hardware and ACK the changes.

PR #1231, opened 2017-07-22
Sta63

ghost:


etomm:
Has this been merged elsewhere? Will it merge inside Lede?

jow-:
It is not even clear what this PR is about so it will definitely not get merged.

etomm:
It's about bridge and nat offloading, but @dissent1 should explain in detail this.

PR #1240, opened 2017-07-24
ar71xx: lzma loader

j-d-r:
Two commits for lzma loader of ar71xx:

- The first one: speed up boot time on some boards. On my board, u-boot leaves caches in a very (very very very) slow mode. Kernel decompression takes 33 seconds on mips 74kc@775Mhz. Good cache coherency reduce it to less than 1 second. Code is taken from linux kernel (https://github.com/torvalds/linux/blob/5924bbecd0267d87c24110cbe2041b5075173a25/arch/mips/include/asm/mach-ath79/kernel-entry-init.h)
- The second: Use LTO when building. You may not want to merge it, as it only saves 270 bytes, but the changes are only in Makefile. Code optimisation might be better too. To handle LTO: use gcc for linking instead of ld, and change all ld flags to gcc flags.


hauke:
This was merged by @nbd168  some days ago.

PR #1241, opened 2017-07-24
ramips: fixed wps gpio in DWR-512

fid0did0:
this patch fix the gpio pin number in the DWR-512 device tree

### DIFF
--- a/target/linux/ramips/dts/DWR-512-B.dts
+++ b/target/linux/ramips/dts/DWR-512-B.dts
@@ -17,7 +17,7 @@
 
 		wps {
 			label = "wps";
-			gpios = <&gpio0 0 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio0 7 GPIO_ACTIVE_LOW>;
 			linux,code = <KEY_WPS_BUTTON>;
 		};
 


mkresin:
Thanks! I merged a version with an improved commit message to my staging tree. Please let me know if you do not agree with the commit message.

fid0did0:
Hi Mathias,

the commit message is perfect.

 

By the way I have an additional issue where I need your advice.

On this router I found that if the pin 10 is not set as an output to ‘0’ there is a conflict on the dataout pin of the flash.

 

Currently I can see only two option:

1)      Export the pin in the dts as an output driving ‘0’

2)      Hide this pin from the dts and keep the correct configuration from the bootloader

 

What is your recommendation?

 

Bye.

 

 

Von: Mathias Kresin [mailto:notifications@github.com] 
Gesendet: Dienstag, 25. Juli 2017 10:26
An: lede-project/source <source@noreply.github.com>
Cc: Giuseppe Lippolis <giu.lippolis@gmail.com>; Author <author@noreply.github.com>
Betreff: Re: [lede-project/source] ramips: fixed wps gpio in DWR-512 (#1241)

 

Thanks! I merged a version with an improved commit message to my staging tree. Please let me know if you do not agree with the commit message.

mkresin:
> On this router I found that if the pin 10 is not set as an output to ‘0’ there is a conflict on the dataout pin of the flash. 

I don't get it. Why does the pin need to be output with an value of 0?

fid0did0:
I do not have a justification (because I do not have the schematics), but this evidence:

pic001

As you can see the first bit of the output transaction have a lower voltage level. This typically means that there are multiple driver on that net.

If I set the pin 10 to '0' everything work as expected.

Bye.
Bye.

mkresin:
To be honest, I can not provide much input. All I can say is that Pin#10 is part of the `uartf` group (Pin#7-Pin#14) which which we switch to GPIO mode. Might be the explanation for what you are seeing.

PR #1244, opened 2017-07-25
ramips: DIR-860L-B1 fix switch port numbering

f00b4r0:
Luci shows switch ports in inverted order on that device.
This patch fixes switch port numbering and matches them to the device
silkscreen.


mkresin:
merged, thanks!

PR #1245, opened 2017-07-25
kernel: update kernel 4.4 to version 4.4.78

artynet:
Refresh patches.
Compile-tested on ar71xx, generic.

Signed-off-by: Arturo Rinaldi <arty.net2@gmail.com>


nbd168:
This commit seems to randomly delete some patches without explanation. Looks broken to me.

nbd168:
There's a better patch for this on the mailing list anyway, so I'll close this pull req.

artynet:
thank you @nbd168, I'll take a look at that.

PR #1249, opened 2017-07-27
ramips: add a compatible string to MT7621 dts files

pinney:
Add a missing compatible string to the MT7621 dts files without one.
    

mkresin:
merged, thanks!

PR #1250, opened 2017-07-28
lantiq/xrx200-net: fix initial link state

sch-m:
Fix a bug which was added with the multi-phy support patch where the
stored initial link state port->link was not equal with the state of
phydev->link, which is preassigned with 1 in phy_device_create().

This leads to the problem, that the carrier of the netdevice was
always 1 (up) until a cable was connected and removed to an related
ethernet port.

### DIFF
--- a/target/linux/lantiq/patches-4.4/0025-NET-MIPS-lantiq-adds-xrx200-net.patch
+++ b/target/linux/lantiq/patches-4.4/0025-NET-MIPS-lantiq-adds-xrx200-net.patch
@@ -209,7 +209,7 @@ Subject: [PATCH 25/36] NET: MIPS: lantiq: adds xrx200-net
 +};
 --- /dev/null
 +++ b/drivers/net/ethernet/lantiq_xrx200.c
-@@ -0,0 +1,1853 @@
+@@ -0,0 +1,1856 @@
 +/*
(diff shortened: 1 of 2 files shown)


blogic:
@sch-m might be better to fix the multi phy support patch ?

sch-m:
The multi phy support patch was merged into the xrx200net patch here: 172eebbd283d0b4b0ef25f65cc0f39ee162a24b7, so I thought this should be ok.

mkresin:
While hunting a similar bug on ramips, I was able to reproduce the issue on lantiq:

1. boot without cable: eth0 link state UNKNOWN but carrier
2. attach cable: eth0 state UNKNOWN and carrier
3. remove cable: eth0 state DOWN and no carrier
4. attach cable: eth0 state UP and carrier

<hr>

1. boot with cable: eth0 state UP and carrier

<hr>

With the patch applied:

1. boot without cable: eth0 state DOWN and no carrier
2. boot with cable: eth0 state UP and carrier

Most of the time the issue does not cause any harm. But as soon as the two interface mode of the switch driver or a configuration without vlan config is used + a netdev led trigger for one of the interfaces, it is pretty obvious that something is wrong.

Long story short: Thanks for the patch, it is merged.

PR #1251, opened 2017-07-28
Archer c50v1: support both EU and US versions

f00b4r0:
This patch series enable support of both the US and EU versions of the Archer C50v1

mktplinkfw2 is modified to allow commandline override of two header fields.

The build makefile is modified to generate a US and EU factory image for both versions. The US factory image is identical to the current "factory.bin" image. the EU factory image has the correct field to pass validation on EU hardware.

The device is renamed "Archer C50v1" to differentiate from the currently unsupported v2 and v3 hardware.

The factory-eu.bin has been successfully tested on EU hardware.

Furthermore, board.d/02_network is updated so that switch port numbering matches device physical order.

The DTS is also fixed with correct GPIO levels for the LEDs.

### DIFF
--- a/target/linux/ramips/image/mt7620.mk
+++ b/target/linux/ramips/image/mt7620.mk
@@ -42,15 +42,17 @@ define Device/ArcherC20i
 endef
 TARGET_DEVICES += ArcherC20i
 
-define Device/ArcherC50
+define Device/ArcherC50v1
   $(Device/Archer)
   DTS := ArcherC50
   SUPPORTED_DEVICES := c50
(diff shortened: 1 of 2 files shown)


mkresin:
I've cherry-picked the switch port fix as well as the led active fixes. Please rebase your PR on master.

f00b4r0:
done.

mkresin:
merged, thanks!

rotanid:
wouldn't it make sense to backport this to the lede-17.01 branch? thanks in advance!

PR #1252, opened 2017-07-28
tegra: add new target with support for CompuLab TrimSlice

tmn505:
This PR adds:
new target: tegra,
new tools: cbootimage, cbootimage-configs,
new package: uboot-tegra.
````
New target introduces initial support for NVIDIA Tegra SoC based devices.
It focuses on Tegra 2 CPUs, for successors supporting NEON instruction set
the target should be split in two subtargets.

This adds also support for CompuLab TrimSlice. It is small nettop with rich
amount of expansion ports. Some hardware specs and supported features in
this commit:

CPU: NVIDIA Tegra 2 @ 1GHz
RAM: 1GB DDR2-667
Storage: SDHC card slot
         µSDHC card slot
         USB to SATA bridge (depends on model)
         1MB SPI NOR flash for bootloader (single partition)
LAN: RTL8111DL GbE
WIFI: RT3070 b/g/n with external antenna (depends on model)
RTC: EM3027 (mapped as rtc0; with battery backup)
     Tegra 2 buil-in (mapped as rtc1)
Sound: Analog/Digital (TLV320AIC23b; S/PDIF not tested)
Connectors: 4x USB 2.0
            RS232 (mini serial)
Other: power button with green led (not functional for early revisions
       without programmed PMIC)
       2x configurable green led

TrimSlice uses U-Boot placed in NOR flash. Boots from any media connected
to USB, SATA or SD card inserted in slot. Can also boot from TFTP (didn't
test). To run LEDE/OpenWrt image one need to update U-Boot to fairly recent
version (the versions, predts/dts provided by CompuLab won't suffice):

 1. Boot TrimSlice into Your current linux distro,
 2. Download trimslice-spi.img from u-boot-trimslice subdir,
 3. Install mtd-utils,
 4. Run following commands:
     flash_erase /dev/mtd0 0 256
     nandwrite /dev/mtd0 trimslice-spi.img
 5. Poweroff, insert SD card with LEDE/OpenWrt image, boot and enjoy.

If by some obstacle You can't follow those instructions, it is possible to
flash U-Boot using serial console.

 1. Interrupt boot process to run U-Boot command line,
 2. Insert FAT or EXT3 formatted SD card with trimslice-spi.img,
 3. Run following commands:
     ${fs}load mmc 0 0x04080000 /trimslice-spi.img
     sf probe 0
     sf erase 0 0x100000
     sf write 0x04080000 0x0 ${filesize}
     reset
 4. Poweroff, insert SD card with LEDE/OpenWrt image, boot and enjoy.

If something went wrong follow instructions on
http://compulab.co.il/utilite-computer/wiki/index.php/Firmware_Update
and after, repeat previous steps.
````

### CODE REVIEW

**Comment by hauke** [comment]
This looks good to me. Is this board useful with LEDE?

**Comment by tmn505** [comment]
Thank You for looking at my PR.
I use this board as an access point with airplay, mpd and DVB-T streaming device. It is comparable with Iomega iConnect. With ethernet connected to pcie lines it much better in comparison with Raspberry PI.
BTW
I added to the PR ``79_move_config`` because after sysupgrade the saved configs weren't copied to their places.

File: target/linux/tegra/base-files/lib/tegra.sh [hunk 1]
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
mkresin: Drop this file. /tmp/sysinfo/model should be filed with the model name from the device tree source file by default.

tmn505: Yes there is valid model in /tmp/sysinfo/model, but during creation of image the metadata: board is set to tegra. When I did sysupgrade it failed the metadata check (it is not mandatory, so it did proceed further). That is why this script is sourced only on sysupgrade.

mkresin: > Yes there is valid model in /tmp/sysinfo/model, but during creation of image the metadata: board is set to tegra.

The image metadata boardname is validated against  /tmp/sysinfo/board_name, which is the compatible string from the dts. If you want to support multiple boards with one image you need to add all of them (space separated) to the `SUPPORTED_DEVICES`.

Either I get something wrong or it never worked as intended.

> When I did sysupgrade it failed the metadata check (it is not mandatory, so it did proceed further).

Please make it mandatory. Add `REQUIRE_IMAGE_METADATA=1` to the top of `target/linux/tegra/base-files/lib/upgrade/platform.sh`.

tmn505: > Please make it mandatory. Add ``REQUIRE_IMAGE_METADATA=1`` to the top of ``target/linux/tegra/base-files/lib/upgrade/platform.sh``.

Done.

> The image metadata boardname is validated against /tmp/sysinfo/board_name, which is the compatible string from the dts. If you want to support multiple boards with one image you need to add all of them (space separated) to the ``SUPPORTED_DEVICES``.
> 
> Either I get something wrong or it never worked as intended.

Yes that is right, but seems it also validates contents of ``/tmp/sysinfo/model`` (in my case ``Compulab TrimSlice board``) with string ``"board": "tegra"`` in metadata, and thus this fails. The sunxi images have similar structure, maybe someone can test it invoking ``sysupgrade -v <lede-sunxi-image>``.

diizzyy:
While this is really cool work I'd be a bit worried that this is going to very susceptive to bit/code-rot given that this device is really rare in general.

tmn505:
@mkresin Sorry, You were right all along. I did clean rebuild, and it did proceed the sysupgrade without modification of files inside ``/tmp/sysinfo``, probably had some leftovers on sdcard with the image.

@diizzyy Well I can agree with the statement of how rare is this device. That is why I hope owners of other tegra boards would add their devices too LEDE tree. If You mean bitrot on kernel side, this is where I have to disagree. The only out of tree patch is the leds support, and the backports are pretty recent additions. The bitrot on LEDE side, well I can't 100% assure it won't also I can't test all corner-cases , but I'll try to keep it up in good shape, with highest kernel version available in LEDE.

tmn505:
Bumped u-boot from 2017-07 to 2017-09 to remove backported patch.

tmn505:
Fixed few misspellings and changed cbootimage to git checkout to avoid possible name conflict of downloaded tarballs. Which in result allowed to remove backported patches.

blogic:

PR #1255, opened 2017-07-30
    ramips: rt305x: update device tree source files

pinney:
       rt305x device tree source files.
    
### DIFF
--- a/target/linux/ramips/dts/3G-6200N.dts
+++ b/target/linux/ramips/dts/3G-6200N.dts
@@ -2,10 +2,11 @@
 
 #include "rt3050.dtsi"
 
+#include <dt-bindings/gpio/gpio.h>
 #include <dt-bindings/input/input.h>
 
 / {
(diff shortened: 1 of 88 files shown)


pinney:
This one doesn't include the rt305x-legacy targets.
It can be added to this or done later.

blogic:
looks good but this will break board detection inside  target/linux/ramips/base-files/lib/ramips.sh

mkresin:
Nah, it shouldn't break anything. Board detection on ramips is done via model name from device tree and none of the model names is touched by this PR.

blogic:
true, LGTM in that case

mkresin:
Merged to my staging tree. I added the compat strings/GPIO macros for the legacy boards, some missing regular boards and the usual minor changes/fixes.

@pinney would you please have a look at the commits in my staging tree to ensure that I did not introduced a regression or have done something silly.

pinney:
Looks OK

PR #1260, opened 2017-07-30
ramips: cleanup EX2700 and WN3000RPv3 LEDs

f00b4r0:
This patch cleans up the WN3000RPv3 and EX2700 setup, bringing it in line
with other similar devices:

The power led is a bicolor one. The bootloader brings the red side on at
powerup.

Instead of blinking the red side in diag.sh and need to forcefully turn it off
in 01_leds, this patch simplifies the setup by setting the default state of the
red side to off in the DTS and blinking the green side as with other devices.


mkresin:
merged, thanks!

PR #1262, opened 2017-08-01
ramips: rt305x-legacy: update device tree source files

pinney:
   rt305x-legacy device tree source files.


mkresin:
I've amended these changes already to your rt305x patch.

PR #1265, opened 2017-08-01
ramips: mt76x8: update device tree source files

pinney:
   mt76x8 device tree source files.


mkresin:
merged to my staging tree, thanks!

PR #1266, opened 2017-08-01
ramips: fix cd-poll sd card remove randomly

Vonger:
Fix when add 'mediatek,cd-poll' to dts cause the sd card be removed randomly.
Special for the device without card-detect pin.


mkresin:
Isn't it the same as your #1178?

Vonger:
Yes, but this is submitted for Linux 4.9

mkresin:
#1178 applies changes to target/linux/ramips/**patches-4.9**/0046-mmc-MIPS-ralink-add-sdhci-for-mt7620a-SoC.patch as well. If this PR replaces #1178, close #1178.

Vonger:
Ah, ye, my mistake, closed :)

blogic:
@Vonger merged, Thanks !

PR #1267, opened 2017-08-01
package/base-file: suppress uci not found output in login.sh

feckert:
If `ttylogin` is not set in `/etc/config/system`, the we get on the tty the output `uci: Entry not found`.
This PR will fix this issue.


mkresin:
Please fix the empty commit message. The prefix is just "base-files:".

feckert:
@mkresin thanks fixed

blogic:
merged, Thanks !

PR #1270, opened 2017-08-02
RFC: Firewall hook for /sbin/hotplug-call firewall

feckert:
The use case is:
If the firewall changed then some package could hook into the firewall to do some ACTION per `/etc/hotplug.d/firewall`.

My use case is:
I need this for mwan3, because if the firewall is restarted then the firewall configuration for mwan3 get flushed as well. Because mwan3 is not a service mwan3 gets no notification that something happens. 
To fix this if the firewall gets restarted a callback script under `/etc/hotplug.d/firewall` with the `ACTION=post-restart `could restart mwan3 to set the firewall rules again.

I am not sure if a similar PR was already made? I did not find any PR or mails on the mailing list.

### DIFF
--- a/package/network/config/firewall/files/firewall.init
+++ b/package/network/config/firewall/files/firewall.init
@@ -29,7 +29,7 @@ validate_firewall_rule()
 }
 
 service_triggers() {
-	procd_add_reload_trigger firewall	
+	procd_add_reload_trigger firewall
 
 	procd_open_validate
 	validate_firewall_redirect
--- a/package/network/config/firewall/files/firewall.init
+++ b/package/network/config/firewall/files/firewall.init
@@ -38,19 +38,27 @@ service_triggers() {
 }
 
 restart() {
+	ACTION="pre-restart" /sbin/hotplug-call firewall
 	fw3 restart
+	ACTION="post-restart" /sbin/hotplug-call firewall
 }
 
 start_service() {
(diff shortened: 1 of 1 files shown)


jow-:
The firewall already emits hotplug actions for zone add/remove events, can you hook into that?

feckert:
I know that this hooks are done for each interface on start/stop/restart/reload.
But i need a hook which is called only once `/usr/sbin/mwan3 restart`, because the restart of mwan3 will call the remove/add hooks on the firewall again.

feckert:
I am sure if this ever worked, but i have found an additional example that will do the same
https://github.com/openwrt/packages/blob/master/net/vpnbypass/files/vpnbypass.hotplug

feckert:
This is the following up PR #1311

feckert:
@jow- i have played around with some solution but every time i get into a loop! Especially on mwan3.
If i do an `/sbin/mwan3 restart`, then every time a zone interface add/remove event is generated. If i add a fw3 hotplug.d reload event then this will generate another reload event with `/etc/hotplug.d/iface/20-firewall`on interface up/down. And so on....

I am not sure how to get this solved. Maybe adding an explicit `fw3 interface $DEVICE reload` in `/etc/hotplug.d/iface/20-firewall` and not a general `fw3 reload` could solve the loop.

blogic:

PR #1271, opened 2017-08-02
ipq806x: pci QSDK fixes

dissent1:
Fix pci settings in accordance to QSDK and OEM gpl tarballs.

Additional info in commit messages.

### DIFF
--- a/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-ap148.dts
+++ b/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-ap148.dts
@@ -167,6 +167,7 @@
 		pcie1: pci@1b700000 {
 			status = "ok";
 			phy-tx0-term-offset = <7>;
+			force_gen1 = <1>;
 		};
 
 		nand@1ac00000 {
--- a/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-c2600.dts
+++ b/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-c2600.dts
@@ -352,6 +352,7 @@
 		pcie1: pci@1b700000 {
 			status = "ok";
 			phy-tx0-term-offset = <7>;
+			force_gen1 = <1>;
 		};
 
 		mdio0: mdio {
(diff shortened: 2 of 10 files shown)


mkresin:
I miss an information what issue get fixed by the changes.

dissent1:
No noticeable fixes I'm aware of, but there may be some that some users experience. 
It's synchronization with Qualcomm's recomendations and OEM gpl. 

According to OEM gpl ap148 based devices' 2nd pci slot should operate in gen1 mode
 https://github.com/paul-chambers/netgear-r7800/blob/eeac2e10190f6f45e32e4c7012c4babc351898d8/git_home/linux.git/sourcecode/arch/arm/mach-msm/board-ipq806x.c#L1668
The same requirement in other devices' OEM gpls.

The 2nd commit sets MRRS and MPS to 256B like the pci buffer size.
https://github.com/lede-project/source/blob/master/target/linux/ipq806x/patches-4.9/0071-pcie-qcom-fixes.patch#L174

mkresin:
>No noticeable fixes I'm aware of, but there may be some that some users experience.
It's synchronization with Qualcomm's recomendations and OEM gpl.

I might be wrong, but I would expect that the QSDK code base and the upstream kernel codebase is a different one. I'm in doubt that blindly porting patches from QSDK has any benefit. Please send the patch upstream for inclusion and feel free to backport the changes if they are accepted.

So long a NAK from me.

dissent1:
> I might be wrong, but I would expect that the QSDK code base and the upstream kernel codebase is a different one.

Yep you are wrong on this one, they have cherry-picked upstream qcom-pci driver and base their changes on top of it.
QSDK https://source.codeaurora.org/quic/qsdk/oss/kernel/linux-msm/commit/drivers/pci/host/pcie-qcom.c?h=eggplant&id=6a7a5bc8c13c19af92a79648636f4e6d9eeb2d29

Kernel 4.9 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/drivers/pci/host/pcie-qcom.c?h=v4.9.40&id=82a823833f4e3769e82cdb4df1bc2234bc65b16c

Actually @nbd 's 071 patch corresponds to following qsdk changes to support ipq806x, so these commits are further steps.

dissent1:
superseded by #1559

PR #1272, opened 2017-08-02
lede-17.01: ramips: mt7620 do not pad sysupgrade Archer images

f00b4r0:
The current makefile unnecessarily pads sysupgrade image for Archer devices.

This has two implications:
1. higher risk of OOM when uploading the binary image to the device
2. much slower upgrade due to time wasted erasing and writing padding

This is already fixed in master, albeit in a completely different way (the
whole target have been reworked)


mkresin:
> 1. higher risk of OOM when uploading the binary image to the device
> 2. much slower upgrade due to time wasted erasing and writing padding

tbh, nothing that i consider as critical enough to commit the changes to the stable branch. The mem argument is kind of invalid, given the amount of ram the mt7620 boxes have.

NAK from me, but maybe someone else is willing to commit this.

mkresin:
Add a few Fixes tags and my NAK turns into an ACK ;-).

Merged, thanks!

PR #1273, opened 2017-08-02
ar71xx: add metadata to wpj344 and wpj558

HenryGiraldo:
This commit add metadata to wpj344 and wpj558 to prevent
load a firmware of wpj344 in a wpj558 and vice versa.
This until now was possible and break the units and had to
be recovered from the uboot.

Signed-off-by: Enrique Giraldo <enrique.giraldo@galgus.net>

### DIFF
--- a/target/linux/ar71xx/image/generic.mk
+++ b/target/linux/ar71xx/image/generic.mk
@@ -852,6 +852,9 @@ define Device/wpj344
   $(Device/wpj-16m)
   DEVICE_TITLE := Compex WPJ344 (16MB flash)
   BOARDNAME := WPJ344
+  SUPPORTED_DEVICES := wpj344
+  IMAGES := sysupgrade.bin
+  IMAGE/sysupgrade.bin := append-kernel | pad-to $$$$(BLOCKSIZE) | append-rootfs | pad-rootfs | append-metadata | check-size $$$$(IMAGE_SIZE)
 endef
 
 define Device/wpj531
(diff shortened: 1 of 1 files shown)


target/linux/ar71xx/image/Makefile
```diff
@@ -58,7 +58,7 @@ define Device/Default
   KERNEL := kernel-bin | patch-cmdline | lzma | uImage lzma
   COMPILE :=
   IMAGES := sysupgrade.bin
-  IMAGE/sysupgrade.bin = append-kernel | pad-to $$$$(BLOCKSIZE) | append-rootfs | pad-rootfs | check-size $$$$(IMAGE_SIZE)
+  IMAGE/sysupgrade.bin = append-kernel | pad-to $$$$(BLOCKSIZE) | append-rootfs | pad-rootfs | append-metadata | check-size $$$$(IMAGE_SIZE)
```
pepe2k: This will probably break* almost all images under whole ar71xx target, please include `append-metadata` step per device/image, not here as default. We will (someday) make this default but for now, only few devices under ar71xx target uses sysupgrade images with embedded metadata.

* Variable `SUPPORTED_DEVICES` isn't reset/cleared anywhere here which might end up with sysupgrade images with wrong/unrelated metadata embedded.

HenryGiraldo:
I moved append-metadata to devices. Fully working 👍

pepe2k:
@HenryGiraldo,

Thank you for your contribution! I have merged your changes (with minor changes) into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1274, opened 2017-08-03
modules: definition should include explict params

pprindeville:
Some modules build out-of-the-box with defaults which are useful for end-systems, desktops, etc. but suboptimal for network devices (routers) or small footprint embedded systems.

Allow the module packager to specify saner parameters.

Fixes FS#935.

### DIFF
--- a/include/kernel.mk
+++ b/include/kernel.mk
@@ -133,11 +133,13 @@ define ModuleAutoLoad
 	probe_module() { \
 		local mods="$$$$$$$$1"; \
 		local boot="$$$$$$$$2"; \
-		local mod; \
 		shift 2; \
+		local mod; \
 		for mod in $$$$$$$$mods; do \
+			local params="$$$$$$$$1"; \
+			shift; \
 			mkdir -p $(2)/etc/modules.d; \
-			echo "$$$$$$$$mod" >> $(2)/etc/modules.d/$(1); \
+			echo "$$$$$$$$mod $$$$$$$$params" >> $(2)/etc/modules.d/$(1); \
 		done; \
 		if [ -e $(2)/etc/modules.d/$(1) ]; then \
 			if [ "$$$$$$$$boot" = "1" -a ! -e $(2)/etc/modules-boot.d/$(1) ]; then \
@@ -151,11 +153,13 @@ define ModuleAutoLoad
 		local priority="$$$$$$$$1"; \
 		local mods="$$$$$$$$2"; \
 		local boot="$$$$$$$$3"; \
-		local mod; \
 		shift 3; \
+		local mod; \
 		for mod in $$$$$$$$mods; do \
+			local params="$$$$$$$$1"; \
+			shift; \
 			mkdir -p $(2)/etc/modules.d; \
-			echo "$$$$$$$$mod" >> $(2)/etc/modules.d/$$$$$$$$priority-$(1); \
+			echo "$$$$$$$$mod $$$$$$$$params" >> $(2)/etc/modules.d/$$$$$$$$priority-$(1); \
 		done; \
 		if [ -e $(2)/etc/modules.d/$$$$$$$$priority-$(1) ]; then \
 			if [ "$$$$$$$$boot" = "1" -a ! -e $(2)/etc/modules-boot.d/$$$$$$$$priority-$(1) ]; then \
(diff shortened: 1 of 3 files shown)


pprindeville:
Based on @ffainelli suggestions and the comments in FS#935.  Fixes FS#935.

pprindeville:
@pepe2k can you please assign reviewers?

pepe2k:
Hi @pprindeville,

> @pepe2k can you please assign reviewers?

I'm sorry but it's not my decision who is going to do the review.
I can only add "needs reviewer" label for you.

Cheers,
Piotr

ffainelli:
If there is only one module specified in AutoLoad or AutoProbe, it would be more compact to avoid writing <module>.<parameter> and have the ability to use <parameter> alone

pprindeville:
> If there is only one module specified in AutoLoad or AutoProbe, it would be more compact to avoid writing . and have the ability to use alone

@ffainelli And that's exactly what the 3rd pattern in the `case` is for.

ffainelli:
Fair enough then, can you post a patch on lede-dev@lists.infradead.org so I can add this to my staging tree and later merge this into the master branch?

NeoRaider:
@ffainelli You can simply add `.patch` to the PR URL (or the individual commit URLs) if you don't want to pull directy: https://github.com/lede-project/source/pull/1274.patch

pprindeville:
@NeoRaider Thanks Matthias, learned something new today!

lynxis:
`git fetch https://github.com/lede-project/source pull/1274/head` is my favourite

jow-:
I do not like the proposed solution, I think the format is hard to read and introduces even more duplicated multi-level escaped shell code.

Also the need to escape commas and inability to preserve spaces makes this a rather clunky solution overall.

What about a syntax like that:

    define KernelPackage/e1000e
      SUBMENU:=$(NETWORK_DEVICES_MENU)
      TITLE:=Intel(R) PRO/1000 PCIe cards kernel support
      DEPENDS:=@PCIE_SUPPORT +kmod-ptp
      KCONFIG:=CONFIG_E1000E
      FILES:=$(LINUX_DIR)/drivers/net/ethernet/intel/e1000e/e1000e.ko
      AUTOLOAD:=$(call AutoProbe,e1000e)
      MODPARAMS.e1000e:= \
        IntMode=1 \
        InterruptThrottleRate=4,4,4,4,4,4,4,4
    endef

pprindeville:
@jow- Okay, I'm trying to figure out what that might look like, but the macros and scripts are a little hairy.  Got time to discuss it online?

pprindeville:
Got something sort of done, but there's an issue.

I need to know the list of modules so I can expand out the list of arguments (possibly empty args like `""`) for each module in `$(FILES)`.  I do that here:

```
define AutoLoad
  add_module "$(1)" "$(call version_filter,$(2))" "$(3)" $$(foreach mod,$$(wildcard $$(FILES)),$$(mod)))" "$$(MODPARAMS.$$(basename $$(notdir $$(mod))))");
endef
```

with the `$$(foreach ...)` sequence.

I have to use the `$(wildcard ...)` because some definitions include wildcards:

```
ARP_MODULES = arp_tables arpt_mangle arptable_filter
define KernelPackage/arptables
  SUBMENU:=$(NF_MENU)
  TITLE:=ARP firewalling modules
  DEPENDS:=+kmod-ipt-core
  FILES:=$(LINUX_DIR)/net/ipv4/netfilter/arp*.ko
  KCONFIG:=CONFIG_IP_NF_ARPTABLES \
    CONFIG_IP_NF_ARPFILTER \
    CONFIG_IP_NF_ARP_MANGLE
  AUTOLOAD:=$(call AutoProbe,$(ARP_MODULES))
endef
```

I could put the `$(wildcard ...)` directly as:

`  FILES:=$(wildcard $(LINUX_DIR)/net/ipv4/netfilter/arp*.ko)`

and fix that...  It will become obvious why this is needed later.

Otherwise, we have some definitions for modules that may or may not be "built-in", like here:

```
define KernelPackage/scsi-core
  SUBMENU:=$(BLOCK_MENU)
  TITLE:=SCSI device support
  KCONFIG:= \
        CONFIG_SCSI \
        CONFIG_BLK_DEV_SD
  FILES:= \
        $(LINUX_DIR)/drivers/scsi/scsi_mod.ko \
        $(LINUX_DIR)/drivers/scsi/sd_mod.ko
  AUTOLOAD:=$(call AutoLoad,40,scsi_mod sd_mod,1)
endef
```

And those files may not exist if they're built-in to the kernel.  That's problematic for the `$(wildcard ...)` above, because the names will be replaced with nothingness if they aren't present.  Which then results in not enough `""`'s being generated as the arguments for `probe_module()` or `load_module()` when the `shift` happens while parsing the `$params` argument to each module.

So, how to workaround this?

pprindeville:
By the way, is there a faster way to retest changes than:

```
% make target/linux/clean
% make -j1 world V=s
```

if I want to see what's being built?

Is there a way to force just a single module to be repackaged?  i.e. `kmod-e1000e` since that's my guinea pig?

pprindeville:
Trying a different approach as:

```
                       local name="param_$$$$$$$$mod"; \
                       local params="$$$$$$$$$${!name}"; \
...
define AutoLoad
  $$(foreach mod,$$(FILES),param_$$(basename $$(notdir $$(mod)))="$$(MODPARAMS.$$(basename $$(notdir $$(mod))))";) add_module "$(1)" "$(call version_filter,$(2))" "$(3)"
endef
```

but this has 2 problems: (1) we're not running `bash` for some reason so `${!name}` results in `Bad substitution` (2) I got a complaint about the `$(foreach ...` not having a closing paren.

pprindeville:
Fix by @jow- with commit 2b6facc8d4b33c04ae0b544933d1489adf347bfc.

PR #1275, opened 2017-08-03
tools: patch various gnu tools for macOS 10.13

rmounce:
These host tools compile but crash at runtime when calling gnulib's vasnprintf on
macOS 10.13 (High Sierra). Backport upstream gnulib patch until new
releases of affected tools.

https://lists.gnu.org/archive/html/bug-gnulib/2017-07/msg00056.html
https://git.savannah.gnu.org/cgit/gnulib.git/commit/?id=c41f233c4c38e84023a16339782ee306f03e7f59


blogic:
merged, Thanks !

PR #1277, opened 2017-08-04
Implement get_port_stats() for several switch drivers

f00b4r0:
This patch series implements get_port_stats() for several switch devices.

This enables the swconfig_leds trigger to correctly blink selected LED when port(s) traffic occurs.

This change has been successfully tested on mt7620, but **needs testing** for all other devices.

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/adm6996.c
+++ b/target/linux/generic/files/drivers/net/phy/adm6996.c
@@ -112,6 +112,9 @@ static const struct adm6996_mib_desc adm6996_mibs[] = {
 	MIB_DESC(ADM_CL30, "Error"),
 };
 
+#define ADM6996_MIB_RXB_ID	1
+#define ADM6996_MIB_TXB_ID	3
+
 static inline u16
 r16(struct adm6996_priv *priv, enum admreg reg)
 {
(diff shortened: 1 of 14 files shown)


eamaclean:
Tested and working as expected on MT7621A on a single port.

mkresin:
Tested and working on AR8216.

blogic:
merged, Thanks !

zorun:
@f00b4r0 This apparently causes high load on some system, see https://bugs.lede-project.org/index.php?do=details&task_id=1004

See also a similar attempt with a similar issue from March, by @VittGam : http://lists.infradead.org/pipermail/lede-dev/2017-March/006816.html

Any idea about what is wrong?

f00b4r0:
@zorun the Archer C2600 appears to have a QCA8337 switch and so would indeed be affected by b6198e5

I've just taken a quick look at the ar8216.c code and here's a preliminary answer:

It appears that the implementation sleeps inside locks, like so:
https://github.com/lede-project/source/blob/53839da46e6fb21e68b3878b4a2187c7d32d688e/target/linux/generic/files/drivers/net/phy/ar8216.c#L275

This is bound to cause lock contention, and I suspect this is what's happening with the new code added: the switch led trigger (which is fairly inefficient code as well) wants access to data that's locked by the _sleeping_ switch driver from a different thread and this triggers high load and performance degradation.

I am not familiar with this hardware so I do not know if it is possible to implement this without sleeping inside the lock, someone with knowledge of this hardware should probably chime in.

It's also possible that my analysis is incorrect, of course ;)

Meanwhile, to confirm that the switch led trigger code is involved, one can simply select a different led trigger for leds currently driven by the switch led trigger and see if it improves the situation.

f00b4r0:
@zorun the Archer C2600 appears to have a QCA8337 switch and so would indeed be affected by b6198e5

I've just taken a quick look at the ar8216.c code and here's a preliminary answer:

It appears that the implementation has a significant busy-wait delay inside locks, like so:
https://github.com/lede-project/source/blob/53839da46e6fb21e68b3878b4a2187c7d32d688e/target/linux/generic/files/drivers/net/phy/ar8216.c#L275

This expands to a 5us delay here:
https://github.com/lede-project/source/blob/53839da46e6fb21e68b3878b4a2187c7d32d688e/target/linux/generic/files/drivers/net/phy/ar8216.h#L643

This is bound to cause lock contention, and I suspect this is what's happening with the new code added: the switch led trigger (which is fairly inefficient code as well) wants access to data that's locked by the _busy waiting_ (in delay loop) switch driver from a different thread and this triggers high load and performance degradation.

I am not familiar with this hardware so I do not know if it is possible to implement this differently, someone with knowledge of this hardware should probably chime in.

It's also possible that my analysis is incorrect, of course ;)

Meanwhile, to confirm that the switch led trigger code is involved, one can simply select a different led trigger for leds currently driven by the switch led trigger and see if it improves the situation.

zorun:
@f00b4r0 thanks for the nice explanation! I think you mean 4d8a66d9346373c2a7fcac5bdae3f662a9dbd9df (but it's identical to the commit you pointed out).

This driver seems to have been worked on by @nbd168 and @blogic , and more recently by @dissent1 .

Let's move this discussion to https://bugs.lede-project.org/index.php?do=details&task_id=1004

f00b4r0:
@zorun I am in fact no longer certain my explanation is correct. I looked into the code some more and the driver uses a mixture of sleeps and delays in many places, but the mutex implementation _should_ make it behave nicely anyway (my understanding is that mutex waiting isn't busy).

I do not have the hardware and I don't know the operation of this driver so I'm mostly speculating here.

TBH I see nothing incorrect with the code I proposed, and similar implementations are working fine on other platforms so until proven wrong I suspect the current problem is specific to the ar8xxx implementation, and simply "highlighted" by the switch led trigger system.

Zsub:
> Meanwhile, to confirm that the switch led trigger code is involved, one can simply select a different led trigger for leds currently driven by the switch led trigger and see if it improves the situation.

I submitted the FS bug. I'll try changing the LED trigger later this week.

PR #1285, opened 2017-08-07
ramips: add support for the VoCore2 Lite

pinney:
ramips: add support for the VoCore2 Lite

The VoCore2 Lite uses the same PCB as the Vocore2.
This patch adds support for the VoCore2 Lite.
It also moves the common VoCore2 parts into dtsi and
removes memory node in the device tree source file.
Memory is detected automatically.

http://vocore.io/    http://vonger.net/    http://vonger.cn/
    
    Specifications:
    - SoC: MediaTek MT7688AN
    - RAM: 64MB DDR2 EtronTech EM68B16CWQH-25H
    - Flash: 8MB NOR SPI Flash GigaDevice GD25Q64CWIG
    - Wireless: Built into MT7688AN with onboard IPEX connector
    
    Firmware installation:
    - VoCore2-Lite ships with firmware forked from OpenWrt.
    - Installation from the bootloader is recommended.
    - If using luci/sysupgrade use the -n option (do not keep settings)
      original firmware uses a modified proprietary MediaTek wireless driver.
    - The wireless is disabled by default in LEDE.
    - If reverting to factory firmware using the bootloader is recommended.
    
    Tested by: Noble Pepper <noblepepper@gmail.com>

### DIFF
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -240,6 +240,9 @@ get_status_led() {
 	vocore2)
 		status_led="$board:fuchsia:status"
 		;;
+	vocore2lite)
+		status_led="$board:green:status"
+		;;
 	w306r-v20|\
 	witi|\
 	zbt-wr8305rt)
--- a/target/linux/ramips/dts/VOCORE2.dtsi
+++ b/target/linux/ramips/dts/VOCORE2.dtsi
@@ -1,0 +1,49 @@
+#include "mt7628an.dtsi"
+
+/ {
+	compatible = "vocore,vocore2", "mediatek,mt7628an-soc";
+
+	chosen {
+		bootargs = "console=ttyS2,115200";
+	};
+};
+
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "wled_an", "perst", "wdt";
+			ralink,function = "gpio";
+		};
+	};
+};
+
+&wmac {
+	status = "okay";
+	ralink,mtd-eeprom = <&factory 0x4>;
+};
+
+&esw {
+	mediatek,portmap = <0x7>;
+	mediatek,portdisable = <0x3a>;
+};
+
+&i2s {
+	status = "okay";
+};
+
+&gdma {
+	status = "okay";
+};
+
+&pwm {
+	status = "okay";
+};
+
+&uart1 {
+	status = "okay";
+};
+
+&uart2 {
+	status = "okay";
+};
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -361,6 +361,9 @@ vocore-16M)
 vocore2)
 	set_wifi_led "$board:fuchsia:status"
 	;;
+vocore2lite)
+	set_wifi_led "$board:green:status"
+	;;
 w502u)
 	set_usb_led "$board:blue:usb"
 	set_wifi_led "rt2800pci-phy0::radio"
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -302,7 +302,8 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "5:wan" "0@eth0"
 		;;
-	vocore2)
+	vocore2|\
+	vocore2lite)
 		ucidef_add_switch "switch0" \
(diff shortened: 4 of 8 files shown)


target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -240,6 +240,9 @@ get_status_led() {
 	vocore2)
 		status_led="$board:fuchsia:status"
 		;;
+	vocore2l)
```
pepe2k: Please, combine this line with block above, where the same settings are used.

target/linux/ramips/dts/VOCORE2.dtsi
```diff
@@ -0,0 +1,79 @@
...
+
+	chosen {
+		bootargs = "console=ttyS2,115200";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		wifi_led: wifi {
+		gpios = <&gpio1 12 GPIO_ACTIVE_HIGH>;
```
pepe2k: Missing tab.

mkresin: Please move the gpio-leds node into the dts files. In the past I've done the same as you are doing here, but it has turned out to be not practical.

target/linux/ramips/dts/VOCORE2.dts
```diff
@@ -1,45 +1,18 @@
 /dts-v1/;
 
-#include "mt7628an.dtsi"
-
-#include <dt-bindings/gpio/gpio.h>
-#include <dt-bindings/input/input.h>
+#include "VOCORE2.dtsi"
 
 / {
-	compatible = "vocore,vocore2", "mediatek,mt7628an-soc";
```
mkresin: please keep the compatible string here. I know it is inherit from the dts, but I feel we should have it anyway.

target/linux/ramips/dts/VOCORE2L.dts
```diff
@@ -0,0 +1,52 @@
+/dts-v1/;
+
+#include "VOCORE2.dtsi"
+
+/ {
+	model = "VoCore2-Lite";
```
mkresin: please add a compatible string here: `compatible = "vocore,vocore2lite", "vocore,vocore2", "mediatek,mt7628an-soc";`.

Would you please rename the file to VOCORE2LITE. It is often hard to get to which board a dts belongs due to the abbreviations.

target/linux/ramips/dts/VOCORE2L.dts
```diff
@@ -0,0 +1,52 @@
...
+
+	memory@0 {
```
mkresin: One only slightly related question: Is the correct amount of memory (auto)detected without a memory node?

pinney:
@mkresin I have not tested memory detection without using a memory node.
please note these two devices only have one LED used for both status and wifi.
Question does this seem correct or should "something" else be done?
+		status {
+			label = "vocore2lite:green:status";
+			gpios = <&gpio1 12 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&wifi_led {
+	label = "vocore2lite:green:wifi";
+};

mkresin:
> Question does this seem correct or should "something" else be done?

My gut feeling tells me it is wrong. If there is no wireless LED, we shouldn't use a different one for this purpose. But I'm not sure if there is a supposed usage of the LED.

It is up to you to decide it. I've already seen this kind of double use for other boards. But in that case it was to compensate a missing boot status led.

pinney:
Built and Tested on both devices ... I think this looks OK.

mkresin:
Hey Larry,

I started to fix the remaining issues on my own but it is to much, sorry.

- please word wrap your commit message at 72 chars
- please add a more verbose hardware description and installation guide to the commit message. use the VoCore2 commit as template (3f31029b190417e12f4d19d9999cd52ec76d0f85)
- please add or add back the led colours to the led label (fuchsia for VoCore2 and green for the VoCore2Lite)
- keep the former used name "status" for the LEDs
- the firmware partition for the Lite is to big (0x50000 + 0x7b0000 == 0x1000000 == 16 MByte)
- drop the input binding include, there is no button
- move the gpio binding include to the dts files (the place where they are used the first time)
- the compat string for the vocore2.dtsi and vocore2.dts is `compatible = "vocore,vocore2", "mediatek,mt7628an-soc";`
- the compat string for the vocore2lite.dts is `compatible = "vocore,vocore2lite", "vocore,vocore2", "mediatek,mt7628an-soc";`

pinney:
Hi Mathias : 

If this looks OK ... or just needs a simple fix-up please pull it in.
I would like to move on to the pinmux fixes then add support for the VoCore2-Ultimate.

Thanks!

mkresin:
Merged, thanks.

PR #1289, opened 2017-08-11
tools/mkimage: fix musl build

hanetzer:
Taken from: http://git.denx.de/?p=u-boot.git;a=commit;h=26e355d131a6b56ea78a156c1cee4f6ba0500b37

Signed-off-by: Marty Plummer <ntzrmtthihu777@gmail.com>


diizzyy:
I wonder if it's not worth updating u-boot (mkimage) instead of adding (more?) upstreamed patches.

blogic:
merged, Thanks !
I replaced the patch with the original upstream version and gave it a different number

PR #1294, opened 2017-08-13
samba36-net: new package

Ansuel:
Samba could also be usefull for sending commands to windows pc (like shoutdown command). This new package add the bin to include this kind of command to the samba package.

### DIFF
--- a/package/network/services/samba36/Makefile
+++ b/package/network/services/samba36/Makefile
@@ -46,12 +46,19 @@ define Package/samba36-client
   DEPENDS:=+libreadline +libncurses
 endef
 
+define Package/samba36-net
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Samba 3.6 SMB/CIFS net commands
+  URL:=https://www.samba.org/
+  DEPENDS:=+libreadline +libncurses
+endef
+
 define Package/samba36-server/config
 	config PACKAGE_SAMBA_MAX_DEBUG_LEVEL
 		int "Maximum level of compiled-in debug messages"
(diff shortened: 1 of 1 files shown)


obsy:
Better make it as separate package?

Ansuel:
Ummm... Yes it would be better. Will update this pr

diizzyy:
Does this even work on newer platforms given the age of Samba 3.x and that there are newer versions of SMB after Windows 7 (which is ~in line with 3.x)? Just make it an option as I would imagine that you pretty much need everything that smbd requires anyway.

Ansuel:
@dizzy net commands are in a separate bin... I think there isn't any problem with creating a new package and flag samba server as a dependency of it

diizzyy:
A new package would mean a new compilation which is just a waste in this case. Why do we need this in the first place? :-)

Ansuel:
@diizzyy with a new package people can install it... with the option people needs to compile samba.

stintel:
@Ansuel did you check all Samba packages for size changes after this change? If there is no impact on existing packages, I have no problem accepting this PR.

Ansuel:
@stintel i think the best way to support this is by adding a subpackage for samba (as net is a separate bin) What do you think?

hauke:
You can create a package in the Makefile like samba36-client and samba36-server are two packages, you could add samba36-net.

Ansuel:
@hauke i know that was my idea... so do this way is good?

Ansuel:
Ok done

blogic:
merged, Thanks !

PR #1295, opened 2017-08-14
rules.mk: add missing CPP configuration argument

hbl0307106015:
Some build system (such as waf [https://waf.io](url)) requires CPP variable during configuration phase, for the compatibility, I think this should be pre-set in rules.mk.


blogic:
merged, Thanks !

PR #1297, opened 2017-08-15
ca-certificates: Update to 20170717

chris5560:
Update to 20170717


diizzyy:
Out of curiosity, why are we not using certificates from nss (mozilla) instead?

blogic:
merged, Thanks !

PR #1298, opened 2017-08-16
base-file: add pkgcheck script

feckert:
This script shows in a easy way all packages installed into the overlay
by diff the two opkg dir "usr/lib/opkg/info" and "/rom/usr/lib/opkg/info".

Usefull if you change/reflash your sytem to get the name of the
package which need to get re/installed on a blank box.

Signed-off-by: Florian Eckert <fe@dev.tdt.de>

### DIFF
--- a/package/base-files/files/bin/pkgcheck.sh
+++ b/package/base-files/files/bin/pkgcheck.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+OPKG_INFO_DIR=/usr/lib/opkg/info/
+OPKG_ROM_INFO_DIR=/rom/usr/lib/opkg/info/
+
+[ ! -d "${OPKG_ROM_INFO_DIR}" ] && {
+	echo "No ROM filesystem found!"
+	exit 1
+}
+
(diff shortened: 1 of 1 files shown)


obsy:
diff is not enabled by default. cmp is.

feckert:
I am not sure if `cmp` is an option. It will increase the script.
It was already enabled on this commit till. 
https://github.com/lede-project/source/commit/5ad114bf478168692fdb7768880ed5ee01d94ee6
Could we not enable this again?

blogic:
having the script without diff installed makes no sense, adding diff to the default config does not sound like a good idea.

feckert:
@blogic At first is this script useful for you? If yes i could check if cmp command could do the same.
Thanks

blogic:
i dont really see the added value. it would end up being one of those undocumented features that no one uses. in general its not a bad idea but i dont really see the added value

feckert:
I patched my luci to show the changes. Maybe i should add this to the luci feed
Thank you for your comment

PR #1299, opened 2017-08-17
ramips: add Xiaomi Mi Router 3G support

dissent1:
This commit adds support for Xiaomi Mi WiFi Router 3G.

Short specification:
 - MT7621AT + MT7603EN + 7612EN
 - 256MB DDR3 RAM
 - 128MB NAND flash
 - 1+2 x 1000M Ethernet
 - 1x USB 3.0 port
 - reset button
 - yellow, blue, red leds

Installation through telnet/ssh:
- copy lede-ramips-mt7621-mir3g-squashfs-kernel1.bin and lede-ramips-mt7621-mir3g-squashfs-rootfs0.bin to usb disk or wget it from LEDE download site to /tmp
- switch to /extdisks/sda1/ (if copied to USB drive) or to /tmp if wgetted from LEDE download site
- run: mtd write lede-ramips-mt7621-mir3g-squashfs-kernel1.bin kernel1
- run: mtd write lede-ramips-mt7621-mir3g-squashfs-rootfs0.bin rootfs0
- run: mtd erase kernel0
- run: reboot

Originally stock firmware has following partitions:
 - ...
 - kernel0 (primary kernel image)
 - kernel1 (secondary kernel image, used by u-boot in failsafe routine)
 - rootfs0 (primary rootfs)
 - rootfs1 (secondary rootfs in case primary fails)
 - overlay (used as ubi overlay)

This commit squashes rootfs0, rootfs1 and overlay partitions into 1, so it can be used by LEDE fully for package installation, resulting in 117,5MiB.

This device lacks hw watchdog, so adding softdog instead (stock does the same).

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -242,6 +242,10 @@ miniembplug)
 	set_wifi_led "$board:red:wlan"
 	set_usb_led "$board:green:mobile"
 	;;
+mir3g)
+	ucidef_set_led_netdev "eth" "Ethernet" "$board:red:wan" "eth0"
+	set_usb_led "$board:blue:usb"
+	;;
 miwifi-mini)
 	ucidef_set_led_default "power" "power" "$board:red:status" "1"
 	;;
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -122,6 +122,10 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "6@eth0"
 		;;
+	mir3g)
+		ucidef_add_switch "switch0" \
+			"1:wan" "2:lan:2" "3:lan:1" "6@eth0"
+		;;
 	psg1218b)
 		ucidef_add_switch "switch0" \
 			"0:lan:3" "1:lan:2" "2:lan:1" "3:wan" "6@eth0"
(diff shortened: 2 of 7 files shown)


robimarko:
This looks really similar to Xiaomi Mi router R3
https://wiki.openwrt.org/toh/xiaomi/mir3
If it is similar support for R3 also would be great

dissent1:
@robimarko actually these devices are pretty different, The twin brother to mi3 is https://wiki.openwrt.org/toh/xiaomi/mini

robimarko:
@dissent1 Yeah, I was pretty tired and did not check before writing.
Only difference between mini and R3 is that R3 has 128MB of NAND flash instead of 16MB

robimarko:
@dissent1 I am looking into your PR and I am wondering what did you do to kernel0 partition.
In DTS there is comment that you deleted kernel0 partition and renamed it to kernel_erase but it looks like you did not erase the whole partition since the original kernel0 partition ends in 0x600000 and you have deleted it until 0x400000

dissent1:
As ppl keep worrying, I'll put the answer here as well:
In DTS I've just renamed kernel0 into kernel_erase for others to remind that it should be erased with 'mtd erase kernel0' while on stock firmware.
0x400000 is the size of the partition, not an end address

das1969:
May I ask how you got telnet/ssh access? Is this through UART?

dissent1:
Changed wifi nodes - added specific vendor,device ids instead of plain pci device.
https://github.com/dissent1/r7800/commit/40946324786f351b5f8ac71c1445ea68bd7d3342

blogic:
merged, Thanks !

PR #1302, opened 2017-08-18
lantiq: sync kernel config target and subtarget (lede-17.01)

feckert:
I Have synced the kernel config for:

- target(lantiq)
`make kernel_menuconfig CONFIG_TARGET=target`

- subtarget(xrx200)
`make kernel_menuconfig CONFIG_TARGET=subtarget`


target/linux/lantiq/config-4.4
```diff
@@ -178,7 +179,6 @@ CONFIG_RTL8366_SMI=y
 CONFIG_SCHED_HRTICK=y
 # CONFIG_SCHED_INFO is not set
 # CONFIG_SCSI_DMA is not set
-# CONFIG_SENSORS_LTQ_CPUTEMP is not set
```
hauke: I think removing this will cause build error on danube and other not xrx200 platforms.
When these platforms select CONFIG_HWMON the build system does not know what to do with CONFIG_SENSORS_LTQ_CPUTEMP. on xrx200 this is no problem because it is selected in its kernel config.

feckert: Ok agree, but this should not happend! If i do `make kernel_menuconfig CONFIG_TARGET=target` then the system should prepare the whole tree for lantiq independently of the subtarget. The command should set the missing config param in target if it is not set in the subtarget.
How do you do this kind of changes? If it is working as it is then it is error prone.

feckert: @hauke I revert my statement.

CONFIG_SENSORS_LTQ_CPUTEMP depend on CONFIG_HWMON.

This is not enabled on target lantiq. Only in some of there subtargets for example xrx200. So CONFIG_SENSORS_LTQ_CPUTEMP could not be set with 
`make kernel_menuconfig CONFIG_TARGET=target ` only if i enable CONFIG_HWMON for all subtargets.

If you select xway in `make menuconfig` there are also some CONFIG parameters which will be removed. I would update this as well could you verify if this is ok? I have no subtarget xway.

subtarget falcon and xway_lecancy are ok no update is needed

hauke: The problem is that "make kernel_menuconfig" only takes the the kernel configuration without the selected packages. 
When you remove CONFIG_SENSORS_LTQ_CPUTEMP from the config-4.4 it is ok for the xway target as long as the hwmon package is not selected. If that gets selected it misses a configuration option for CONFIG_SENSORS_LTQ_CPUTEMP. Normally these options are added to the config in generic/config-4.4, but this driver is only available for lantiq. 
I would suggest to manually add this to the config-4.4 file like it was done before.

feckert: Ok i can do that but i am not convinced. I think if some changes the subtarget the changer should also update the target. I think the PC is smarter then the human and should do the right.

- Add the missing CONFIG options
- update the config lexical.

So if some updates the CONFIG like me he did not know that CONFIG_SENSORS_LTQ_CPUTEMP must be added if in the future someone add CONFIG_HWMON in the other targets.

mkresin:
I don't get the point of such a change in the 17.01 branch. AFAIK nothing is broken with the kernel config. Feel free to do the cleanup in master.

This is a NAK from me.

feckert:
@mkresin no problem i will close this PR

PR #1306, opened 2017-08-21
uboot-mvebu: fix build on macOS

rmounce:
macOS no longer provides OpenSSL/libssl which is required to build u-boot for mvebu. LibreSSL is built to work around this elsewhere and u-boot uses pkg-config to include libssl, however a workaround is required to use the correct host staging directories for pkg-config.

### DIFF
--- a/package/boot/uboot-mvebu/Makefile
+++ b/package/boot/uboot-mvebu/Makefile
@@ -16,7 +16,9 @@ PKG_HASH:=f54baf3f9325bf444c7905f3a5b6f83680edb1e6e1a4d5f8a5ad80abe885113f
 include $(INCLUDE_DIR)/u-boot.mk
 include $(INCLUDE_DIR)/package.mk
 
-UBOOT_MAKE_FLAGS:=
+UBOOT_MAKE_FLAGS:= \
+	PKG_CONFIG_PATH="$(STAGING_DIR_HOST)/lib/pkgconfig:$(PKG_CONFIG_PATH)" \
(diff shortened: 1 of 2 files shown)


blogic:
imho this only works around a problem. it looks like pkgconfig is failing and this patch just hides that fact.could you try and figure out why pkgconfig fails ?

rmounce:
Well yes, it does say 'workaround' in the commit message :)

Without the PKG_CONFIG_PATH  / PKG_CONFIG_LIBDIR overrides, the host library directories where libssl.pc exists are not searched by pkg-config at all so it is not found. u-boot's Makefile falls back to a naive -lssl include and it is not found on the host (macOS in my case). Build fails.

If these variables are overridden to include the host then pkg-config does find libssl.pc, however the pkg-config wrapper overrides the prefix and exec_prefix variables. (indeed, that is all the wrapper does). pkg-config returns the invalid overridden path to the compiler instructing it to find libssl in the target staging directory rather than the host staging directory. Build fails.

pkg-config is behaving as designed, the wrapper in LEDE is triggering the issue. I am working around the workaround..? I can't see why it is not sufficient for the build system to pass appropriately scoped PKG_CONFIG_PATH  / PKG_CONFIG_LIBDIR environment variables to pkg-config.

I don't have a sufficiently comprehensive understanding of the build system to come up with a cleaner solution. To be honest I don't even need uboot-mvebu, it is just built unconditionally for mvebu (even though only needed for ClearFog) and getting in my way when it failed.

blogic:

PR #1307, opened 2017-08-21
util-linux: don't need to build NLS support

pprindeville:
Per @hauke's https://github.com/lede-project/source/pull/1215#issuecomment-323836838


blogic:
merged, Thanks !

PR #1311, opened 2017-08-23
package/firewall: add start/stop/reload hotplug events

feckert:
I have added the following hotplug events into fw3
- start
- stop
- reload

This is a replacement for PR #1270


feckert:
Will close this PR seen some problem with this solution!

PR #1312, opened 2017-08-23
ath10k: fix rx ring buffer memory leak and spectral scan on some boards

dissent1:
- ath10k: fix ring buffer memory leak
Backporting ath.git commit that moves rx ring buffer reorder flag earlier in the sequence thus avoiding race condition and preventing memory leak.

Fixes FS#801 - wifi crash caused by rx ring buffer corruption.

- ath10k: fix spectral scan
Backporting ath.git commit that fixes spectral scan for qca9984, qca9888, ipq4019 because of different spectral data length provided by the board.

### DIFF
--- a/package/kernel/mac80211/patches/332-ath10k-fix-rx-ring-buffer-memory-leak.patch
+++ b/package/kernel/mac80211/patches/332-ath10k-fix-rx-ring-buffer-memory-leak.patch
@@ -1,0 +1,58 @@
+From f35a7f91f66af528b3ee1921de16bea31d347ab0 Mon Sep 17 00:00:00 2001
+From: Rakesh Pillai <pillair@qti.qualcomm.com>
+Date: Wed, 2 Aug 2017 16:03:37 +0530
+Subject: ath10k: fix memory leak in rx ring buffer allocation
+
+The rx ring buffers are added to a hash table if
+firmware support full rx reorder. If the full rx
+reorder support flag is not set before allocating
+the rx ring buffers, none of the buffers are added
+to the hash table.
+
+There is a race condition between rx ring refill and
+rx buffer replenish from napi poll. The interrupts are
+enabled in hif start, before the rx ring is refilled during init.
+We replenish buffers from napi poll due to the interrupts which
+get enabled after hif start. Hence before the entire rx ring is
+refilled during the init, the napi poll replenishes a few buffers
+in steps of 100 buffers per attempt. During this rx ring replenish
+from napi poll, the rx reorder flag has not been set due to which
+the replenished buffers are not added to the hash table
+
+Set the rx full reorder support flag before we allocate
+the rx ring buffer to avoid the memory leak.
+
+Signed-off-by: Rakesh Pillai <pillair@qti.qualcomm.com>
+Signed-off-by: Kalle Valo <kvalo@qca.qualcomm.com>
+---
+ drivers/net/wireless/ath/ath10k/core.c | 12 ++++++------
+ 1 file changed, 6 insertions(+), 6 deletions(-)
+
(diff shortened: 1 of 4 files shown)


chunkeey:
I've been running this for two days now. Let's hope that it does indeed fix the elusive "wifi dies after 24/48 Hours" on the QCA9984. As for the other performance issues: I think @hauke is planning to update mac80211 package in the near future. You could point the people with affected systems to his tree: https://git.lede-project.org/?p=lede/hauke/staging.git;a=shortlog;h=refs/heads/mac80211

dissent1:
вс, 27 авг. 2017 г. в 1:32, Christian Lamparter <notifications@github.com>:

> I've been running this for two days now. Let's hope that it does indeed
> fix the elusive "wifi dies after 24/48 Hours" on the QCA9984. As for the
> other
>
Yep, fortunately it seems to fix the issue


performance issues: I think @hauke <https://github.com/hauke> is planning
> to update mac80211 package in the near future. You could point the people
> with affected systems to his tree:
> https://git.lede-project.org/?p=lede/hauke/staging.git;a=shortlog;h=refs/heads/mac80211
>
Unfortunately 2.4ghz performance in my case is not fixed with this. In
contrary it bugs very much. Seems that Hauke is missing some more
compatibility commits, because WiFi stucks after the 1st disassociate and
it’s not possible to auth anymore. No relative logs in dmesg when this
happens.
802.11w breaks AP as well.
More than that, for some reason it crashes the system while sysupgrading,
so I had to tftp the recovery image in order to switch.


<https://git.lede-project.org/?p=lede/hauke/staging.git;a=shortlog;h=refs/heads/mac80211>

chunkeey:
https://forum.lede-project.org/t/netgear-r7800-exploration-ipq8065-qca9984/285/470 

looks like it didn't fix it fully. Even with the "reduced rx buffers" patch undone, the hardware still crashed.
Looking at the crashdump:
`ath10k_pci 0000:01:00.0: [00]: 0x0004a000 3735928559 3735928559 3735928559 3735928559`
the 3735928559 value sticks out. It's the decimal value for the famous 0xdeadbeef magic.

https://github.com/torvalds/linux/blob/master/drivers/net/wireless/ath/ath10k/pci.h#L200
"The device has a special powersave-oriented register. When device is considered asleep it drains less power and driver is forbidden from accessing most MMIO registers. If host were to access them without waking up the device might scribble over host memory or return 0xdeadbeef readouts."

Could it be that the chip is overheating? The custom board-2.bin could help in this situation, since the ath10k-firmware's board-2.bin are for QCA9984 reference cards in dev boards... And these usually don't have an enclosure.

dissent1:
These changes are staged in @hauke tree within new k4.13.1 backports.
Closing the PR.

hauke:
The problems with disassociating are fixed in my patches based on 4.13.1 and later.
Would it make sense to add these patches to lede 17.01?

dissent1:
@hauke yes, testing it since your 4.13.1 update, it doesn’t seem to cause more issues compared to current master version on ipq8065/qca9984.

As for that rx ring commit in this PR, after 16 days the rx ring buffer corruption (on current master) has occurred again, although it’s better than 1-2 days, it’s not directly related to the mentioned commit. It seems like smth else in my tree in combination with this commit workarounded the issue to operate for a longer period of time. But this commit was a trigger part. I can’t figure out the combination yet...

So while it’s good to port it to 17.01, it’s not enough to fix the issue.
@chunkeey

hauke:
It probably makes sense to involve the upstream ath10k developers, when you also have the problems with the backports based on 4.14-rc1, then it is probably a problem in upstream.

PR #1313, opened 2017-08-24
package/firewall: add common hotplug events 

feckert:
Following common hotplug events are called if option "-e" in the fw3 command is set

- start
- stop
- reload
- restart
- flush

This is a replacement for PR #1270

### DIFF
--- a/package/network/config/firewall/files/firewall.init
+++ b/package/network/config/firewall/files/firewall.init
@@ -29,7 +29,7 @@ validate_firewall_rule()
 }
 
 service_triggers() {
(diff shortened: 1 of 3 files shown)


hauke:
Could you please send your patches for the firewall3 project directly to the lede mailing list and base them on top of the frirewall3 repository at: https://git.lede-project.org/?p=project/firewall3.git;a=summary

feckert:
@hauke is on the way

blogic:
moved to mailing list

PR #1315, opened 2017-08-25
ar71xx: add support for TL-WR1041N(v2) LAN/WAN LEDs.

laughterk:
Current implemntation only implemented WPS/Power/WLAN LEDs attached
to cpu (ar9342), the LAN/WAN LEDs attached to ar8327 are not supported.
The commit add support for the LAN/WAN LEDs.

PS: copied from TL-WR-1043ND (v2) implementation with LED orders modified.
PPS: WAN/LAN LEDs still not works when connected to 100M ethernet (the ar8327 driver problem?).

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1041n-v2.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1041n-v2.c
@@ -74,6 +74,22 @@ static struct gpio_keys_button tl_wr1041nv2_gpio_keys[] __initdata = {
 	}
 };
 
+static const struct ar8327_led_info tl_wr1041n_leds_ar8327[] = {
+	AR8327_LED_INFO(PHY0_0, HW, "tp-link:green:wan"),
+	AR8327_LED_INFO(PHY1_0, HW, "tp-link:green:lan1"),
+	AR8327_LED_INFO(PHY2_0, HW, "tp-link:green:lan2"),
+	AR8327_LED_INFO(PHY3_0, HW, "tp-link:green:lan3"),
+	AR8327_LED_INFO(PHY4_0, HW, "tp-link:green:lan4"),
+};
+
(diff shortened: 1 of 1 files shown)


target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1041n-v2.c
```diff
@@ -90,7 +106,10 @@ static struct ar8327_platform_data db120_ar8327_data = {
 		.duplex = 1,
 		.txpause = 1,
 		.rxpause = 1,
-	}
+	},
+	.led_cfg = &wr1041n_v2_ar8327_led_cfg,
+	.num_leds = ARRAY_SIZE(tl_wr1041n_leds_ar8327),
+	.leds = tl_wr1041n_leds_ar8327
```
hauke: I think the indention is wrong here

hauke:
Please use ar71xx as prefix instead of kernel

laughterk:
> Please use ar71xx as prefix instead of kernel
Updated

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1041n-v2.c
laughterk: Sorry, I don't get it quite well, the indention seems to be OK.

laughterk:
> PPS: WAN/LAN LEDs still not works when connected to 100M ethernet (the ar8327 driver problem?).

Checked AR8327 datasheet, I think I find why WAN/LAN LEDs are not working when connected to 100M ethernet, will modify and try it later.

laughterk:
Now the LAN/WAN LEDs on TL-WR1041N are fully functional.

blogic:
please fold both patches into one single commit

laughterk:
> please fold both patches into one single commit

done, please check commit 564d5bf5f8b80fe2160031128bf6dd5fa2e1460b

blogic:
merged, Thanks !

PR #1290, opened 2017-08-12
treewide: Remove CyaSSL (wolfSSL) from tree

diizzyy:
Remove wolfSSL (formely CyaSSL) from the tree. No maintainer and it's always
out of date which is bad from a security point of view. Having that in mind
it also indicates that there are very few/no users of this at all given the
above.

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>

PR #1319, opened 2017-08-27
libs/wolfssl: rename from cyassl; update to 3.12.0 ; add myself as maintainer

commodo:
This PR continues on the basis on this PR https://github.com/lede-project/source/pull/1290

Tested with `curl` ; worked fine.
Tested on x86_64 trunk LEDE hash: https://github.com/lede-project/source/commit/fc48aebdc1d97c73ec8a8428d981bdaa850568d2


commodo:
@kaleb-himes
see commit https://github.com/lede-project/source/pull/1319/commits/a0d4dc85aa4baf2c48bb887e8b55cd772fccf57e

there's a small issue with the hardening option;
admittedly, it's only a warning, but depending on the CFLAGS of the package, that could fail the build (as the case in `libwebsockets-cyassl`)
the issue itself does not seem too bad

as a general note to this PR: maybe later I should see about adding a smarter way to configure algo/build options for this lib;

package/libs/wolfssl/Makefile
```diff
@@ -56,58 +56,58 @@ CONFIGURE_ARGS += \
         --enable-ipv6
 endif
 
-ifeq ($(CONFIG_WOLFSSL_HAS_AES_CCM),y)
+ifeq ($(CONFIG_WOLFSSL_HAS_AES_CCM),y) # default disabled
 CONFIGURE_ARGS += \
```
blogic: please drop the comments at the end end of the lines

blogic:
are there packages in the packages feed that would need to be fixed aswell ?

commodo:
yep ; `libwebsockets-cyassl` and `coova-chilli` will need adjusting 
see: https://github.com/commodo/packages/commit/2a85cd1b8c867d876c7f38c23f5d86902fa554c8
and  https://github.com/commodo/packages/commit/f7d51da20139a9aa35a0648aa47189b9bc1a7bc9

FWIW: both these builds seem broken with LEDE even without updating from 3.10 to 3.12; maybe they were building with an older version of `cyassl` [3.9.6 is in OpenWrt atm ]
but, just with those patches, it seems that even with 3.10 [in LEDE trunk already] both packages build

i'm feeling that the old CYASSL name [which needs a re-direct to WOLFSSL] is not very well handled in `wolfssl`

commodo:
@blogic 
do you want me to spin up a PR [ or 2 ] with those patches that fix libs ?
or shall I just notify the maintainers ?

commodo:
also re-spinned with update from comment

kaleb-himes:
@commodo 
> see commit a0d4dc8
> 
> there's a small issue with the hardening option;
> admittedly, it's only a warning, but depending on the CFLAGS of the package, that could fail the build (as the case in libwebsockets-cyassl)
> the issue itself does not seem too bad
>
> as a general note to this PR: maybe later I should see about adding a smarter way to configure > algo/build options for this lib;

Sorry for the delay, I was traveling on business and have just recently returned to the office. I'll take a look this week and see if I can find a solution to the warning.

- KH

commodo:
@kaleb-himes 
no worries ; i've worked around it;
i've also sent 2 PRs to fix 2 packages with the libcyassl/wolfssh build
https://github.com/openwrt/packages/pull/4788
https://github.com/openwrt/packages/pull/4789

i think after a while we may need to do more renames from cya to wolf
to avoid that redir layer of headers

hauke:

PR #1323, opened 2017-09-01
comgt-ncm: add support for specifying modem manufacturer

marcin1j:
Add `manufacturer` option for manual modem manufacturer selection.

On Huawei E3372 manufacturer string cannot be reliably determined using
`AT+CGMI` due to:
 1. modem mixing periodic (unsolicited) status messages with reply to
    AT commands,
 2. the lack of prefix in `AT+CGMI` response which makes distinguishing
    the reply from random status messages impossible.

What's worse when auto-detection returns invalid (random) manufacturer
name, interface is in FAILED state and it's impossible to bring it up
even manually without touching uci configuration.

A real-life example of `AT+CGMI` output from Huawei E3372 below. In this
case, autodetection returns `a` as a manufacturer:

```
A
^RSSI:20

^HCSQ:"LTE",49,39,121,20
T+CGMI
huawei

OK
```

### DIFF
--- a/package/network/utils/comgt/Makefile
+++ b/package/network/utils/comgt/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=comgt
 PKG_VERSION:=0.32
-PKG_RELEASE:=29
+PKG_RELEASE:=30
 
 PKG_SOURCE:=$(PKG_NAME).$(PKG_VERSION).tgz
 PKG_SOURCE_URL:=@SF/comgt
--- a/package/network/utils/comgt/files/ncm.sh
+++ b/package/network/utils/comgt/files/ncm.sh
@@ -10,6 +10,7 @@ proto_ncm_init_config() {
 	no_device=1
 	available=1
 	proto_config_add_string "device:device"
+	proto_config_add_string manufacturer
 	proto_config_add_string apn
 	proto_config_add_string auth
 	proto_config_add_string username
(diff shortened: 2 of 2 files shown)


john-tho:
Hi,
I came across this problem with my E3372 in NCM mode.
In a different approach, I use the command `AT^CURC=0` at the start of the modern communication to turn off these messages. [AT command details on StackOverflow](https://stackoverflow.com/questions/11074808/at-command-for-disable-radio-signal-strength-indication-and-alike#11184909)

`--- /etc/gcom/getcardinfo.gcom.original 2017-09-21 11:07:44.0000000
0 -1000
+++ /etc/gcom/getcardinfo.gcom  2017-09-21 10:49:32.000000000 -1000
@@ -6,6 +6,8 @@
  flash 0.1

 :start
+ send "AT^^CURC=0^m"
+ get 1 "" $s
  send "AT+CGMI^m"
  get 1 "" $s
  print $s`

Using only this `AT^CURC=0` change works for me, but I don't know how other modems would interact with this command.

marcin1j:
You cannot use `AT^CURC` prior to `AT+CGMI` (manufacturer detection). It's Huawei proprietary command and you don't know modem manufacturer at that moment.
By doing so you're effectively hardcoding Huawei as the only supported modem manufacturer.

Had the problem occurred after manufacturer is detected, we could solve it by adding `AT^CURC=0` to Huawei-specific initialization command list. Unfortunately manufacturer detection itself is affected so there are only two ways of fixing it:
1. change manufacturer detection routine (an alternative is to use `ATI` command which has been used some time ago),
2. let the user override auto-detection which is less intrusive (works like before unless user sets manufacturer manually).

blogic:

PR #1324, opened 2017-09-02
Update ARV7519RW22.dts

Mandrake-Lee:
Enable PCIe bus and therefore access to the wifi chipset for this platform

### DIFF
--- a/target/linux/lantiq/dts/ARV7519RW22.dts
+++ b/target/linux/lantiq/dts/ARV7519RW22.dts
@@ -91,7 +91,8 @@
 		};
 
 		pcie@d900000 {
-			status = "disabled";
+			status = "okay";
+			gpio-reset = <&gpio 21 GPIO_ACTIVE_HIGH>;
 		};
 	};
 


mkresin:
Please have a look at https://lede-project.org/submitting-patches. Your commit has the following formal issues:

- missing prefix
- missing Signed-off-by

Furthermore, the commit title should mention what is done (here: enable ARV7519RW22 PCIe bus). The commit message should explain why it is done. In your case it should explain why you need access to an unsupported wireless chip.

Are you sure the wireless chip is connected to PCIe and not to PCI as it is for the TD-W8980? Does lspci shows the Wave300?

Mandrake-Lee:
Hi Mathias,
Sorry for the formal issues as this is the first time I ask for a commit.

Coming to the real thing, the PCIe bus needs to be reset after boot via gpio for this platform, router Livebox 2.1 in my case.

I can confirm that the wifi WAVE300 is hardwired to PCIe, not PCI. This can be easily seen with lspci.

I agree that WAVE300 driver is unsupported as per today although there's ongoing work (see https://github.com/Mandrake-Lee/WAVE300).

We could wait until the driver is functional to enable the PCIe but I think it's definetely good to include now the gpio reset i.e.:
`gpio-reset = <&gpio 21 GPIO_ACTIVE_HIGH>;`

It is a description of the hardware which is missing.

Please let me know if I have to drop this commit and start a new one just with the line above.

mkresin:
> We could wait until the driver is functional to enable the PCIe

No need to wait, I've done some already for a different board which has the WAVE300. So, fine with me.

> Please let me know if I have to drop this commit and start a new one just with the line above.

Just update the commit to fix the formal issues and I'm fine to commit the patch. Have a look at https://lede-project.org/submitting-patches#squashing_commits. It is more or less the same you have to do to update this PR.

mkresin:
Please have a look at 31b5069a502e6b2e5755fbc728c400719cede77f for the commit message.

Please use ```git commit --amend``` to change the commit message. Use ```git push --force``` to overwrite your github repository/branch. The PR will afterwards update automatically. Please don't close this PR and open a new one.

Mandrake-Lee:
Sorry but I didn't manage to amend the PR.

I've created a new one #1341 

You can drop this one.

PR #1326, opened 2017-09-04
packages: add ifrename util from the wireless-tools

sch-m:
This tool makes it possible to simply rename netdevices.

example: ifrename -i ptm0 -n dsl0

### DIFF
--- a/package/network/utils/wireless-tools/Makefile
+++ b/package/network/utils/wireless-tools/Makefile
@@ -52,6 +52,19 @@ define Package/libiw/description
  "Linux Wireless Extensions".
 endef
 
+define Package/ifrename
+$(call Package/wireless-tools/Default)
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=Tool to rename network interfaces
+  DEPENDS:= +libiw
+endef
+
+define Package/ifrename/description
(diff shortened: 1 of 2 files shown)


mkresin:
What about using the already included ip command:

```
ip link set dev ptm0 down
ip link set ptm0 name dsl0
ip link set dev dsl0 up
```

I know the ppp interface is renamed already on up. I wasn't able to spot the code where it happens and whether the iface can be renamed while it is up.

Depending on your usecase, it might be easier to change the ltq-atm/ptm source to use dslN as iface name.

sch-m:
Yes, you are right.
I've totally overseen the iproute2 renaming feature.
So this PR is obsolete.

About the usecase:
For PTM, it would be a change in the ltq-ptm driver, but for ATM you have to modify the br2684ctl to use another name than nasN.
As you know I am currently working on the lantiq xrx200 platform and I don't want to break other targets, so changing the default br2684 naming would be a really hot modification.

So I thought renaming the interfaces in the 10_atm_ptm.sh script would be the safest variant.

mkresin:
>  As you know I am currently working on the lantiq xrx200 platform and I don't want to break other targets, so changing the default br2684 naming would be a really hot modification.

Oh yeah it would be. I was a bit naive to assume it will be that easy.

PR #1331, opened 2017-09-06
uboot-envtools: add Xiaomi Mi Router 3G support

dissent1:
I've found out, that stock kernel image contains usb recovery script, so preserving kernel0 partition allows restoring back stock fw easily through usb recovery.

So since now there is no need to erase kernel0 partition.
To boot LEDE while on stock fw you have to:
mtd write lede-ramips-mt7621-mir3g-squashfs-kernel1.bin kernel1
mtd write lede-ramips-mt7621-mir3g-squashfs-rootfs0.bin rootfs0
nvram set flag_last_success=1
nvram commit
reboot

If you want to revert back to stock fw while on LEDE:
insert usb stick (FAT/FAT32) with stock fw renamed to miwifi.bin
fw_setenv flag_last_success 0
power off the device
hold the reset button and power on the device, keep holding the reset button until yellow led starts blinking.

@mkresin


mkresin:
Please split the commits into two. One adding u-boot env support for the MIR and one changing the MIR dts + package selection.

I still don't get how it works. You told about the md5sum in the u-boot env that is validated on boot. Isn't it relevant any longer? What happens to the installed LEDE if flag_last_success is 0 on u-boot start but no usb stick is attached?

How does the stock firmware recovery works if the root partition for the stock kernels are dropped as it is now?

There is at least some documentation about the behaviour required.

Is it possible to convince u-boot to set flag_last_success to 0 on boot? Means, if we do not switch it back to 1 the LEDE system is broken and we need recovery.


 Isn't it validated anymore

dissent1:
Okay, no problem, I think I know how to make it more failsafe. Need some tests.

> I still don't get how it works. 

Me neither :)

> You told about the md5sum in the u-boot env that is validated on boot. Isn't it relevant any longer?

Yes it’s still in place. But now it looks like that when you set that flag to 1 it boots the 2nd partition straight away in any case, no matter if md5 is correct. When this flag is left at 0 - it leads to the default behavior as I’ve explained you in the irc channel.

> What happens to the installed LEDE if flag_last_success is 0 on u-boot start but no usb stick is attached?

If stock rootfs0 and rootfs1 are erased, then stock kernel0 fails to boot, and uboot switches to kernel1 (partition flashed with lede). The flag that 1st partition has failed is set after that:
flag_try_sys1_failed=1   - kernel0 (stock)
flag_try_sys2_failed=0  - kernel1 (lede)

> How does the stock firmware recovery works if the root partition for the stock kernels are dropped as it is now?

Because they implemented it in kernel partition (not in uboot or rootfs)... interesting enough that to recover the fw you have to have a semi-working firmware.

> Is it possible to convince u-boot to set flag_last_success to 0 on boot? Means, if we do not switch it back to 1 the LEDE system is broken and we need recovery.

That is the tricky part that requires more investigation.

It’s = 0 by default. So to flash lede from factory (if we don’t set the flag to 1) we have to erase the 2nd rootfs1, because stock rootfs ds contain partition copy script that will overwrite all partitions if stock kernel manages to boot.

But it leads to the situation that this flag is set by uboot, because stock kernel fails flag_try_sys1_failed=1.
Further behavior requires testing actually.

ldir-EDB0:
"insert usb stick (FAT/FAT32) with stock fw renamed to miwifi.bin" does not actually appear to be correct - the stock firmware USB recovery looks for any file "*.bin"
[    5.570000] Check for USB recovery...
[    5.580000] Found usb storage, wait for ready
[    5.880000] scsi 0:0:0:0: Direct-Access     Generic  STORAGE DEVICE   9407 PQ: 0 ANSI: 0
[    6.010000] sd 0:0:0:0: [sda] 3970048 512-byte logical blocks: (2.03 GB/1.89 GiB)
[    6.020000] sd 0:0:0:0: [sda] Write Protect is off
[    6.030000] sd 0:0:0:0: [sda] No Caching mode page found
[    6.030000] sd 0:0:0:0: [sda] Assuming drive cache: write through
[    6.040000] sd 0:0:0:0: [sda] No Caching mode page found
[    6.050000] sd 0:0:0:0: [sda] Assuming drive cache: write through
[    6.050000]  sda: sda1
[    6.060000] sd 0:0:0:0: [sda] No Caching mode page found
[    6.060000] sd 0:0:0:0: [sda] Assuming drive cache: write through
[    6.070000] sd 0:0:0:0: [sda] Attached SCSI removable disk
starting pid 425, tty '': '/etc/rcS'
[    6.640000] led=8, on=1, off=4000, blinks,=1, reset=1, time=1
[    6.650000] led=10, on=10, off=10, blinks,=1, reset=1, time=4000
[    6.660000] led=6, on=1, off=4000, blinks,=1, reset=1, time=1
[   16.700000] FAT-fs (sda): utf8 is not a recommended IO charset for FAT filesystems, filesystem will be case sensitive!
mount: mounting /dev/sda on /mnt/sda failed: Invalid argument
[   16.730000] FAT-fs (sda1): utf8 is not a recommended IO charset for FAT filesystems, filesystem will be case sensitive!
System recovery [   20.240000] Burning uImage.bin to kernel0 ...is in progress...
Unlocking kernel0 ...

Writing from uImage.bin to kernel0 ...
Done
[   21.920000] Burning uImage.bin to kernel1 ...Unlocking kernel1 ...

Writing from uImage.bin to kernel1 ...
Done
[   23.600000] Burning root.ubi to rootfs0 ...ubiformat: mtd10 (nand), size 33554432 bytes (32.0 MiB), 256 eraseblocks of 131072 bytes (128.0 KiB), min. I/O size 2048 bytes
libscan: scanning eraseblock 255 -- 100 % complete
ubiformat: 256 eraseblocks have valid erase counter, mean value is 0
ubiformat: flashing eraseblock 155 -- 100 % complete
ubiformat: formatting eraseblocDone7 -- 72 % complete
ubiformat: formatting eraseblock 255 -- 100 % complete     34.600000] Burning root.ubi to rootfs1 ...
ubiformat: mtd11 (nand), size 33554432 bytes (32.0 MiB), 256 eraseblocks of 131072 bytes (128.0 KiB), min. I/O size 2048 bytes
libscan: scanning eraseblock 255 -- 100 % complete
ubiformat: 256 eraseblocks have valid erase counter, mean value is 0
ubiformat: flashing eraseblock 155 -- 100 % complete
ubiformat: formatting eraseblocDone7 -- 72 % complete
ubiformat: formatting eraseblock 255 -- 100 % complete
Unlocking overlay ...
Erasing overlay ...
[   47.200000] Restarting system.

A USB device without not containing *.bin produces "
ls: /mnt/sda1/*.bin: No such file or directory" and a reboot.

dissent1:
@ldir-EDB0 
It doesn’t mean that it flashes any bin file, probably it just lists all bins available and flashes according to its name. There’s also a miwifi-ssh.bin, a small binary, that enables ssh on stock developer firmware.

dissent1:
@mkresin actually I’m confused now. Now I think we can preserve fully functional stock kernel and avoid booting it by setting flag_try_sys1_failed=1. But I don’t completely trust uboot wether it tries to boot stock fw at some point with fully functional lede, thus overwriting it.

mkresin:
I would prefer to wait till @ldir-EDB0 has finished his analysis of what u-boot really does.

mkresin:
superseded by f2107fc328ff7f9817fe9ca64f84bba9e32abfc6

PR #1333, opened 2017-09-07
generic: backport upstream kernel performance patches

dissent1:
Backporting several upstream patches, further info is in commit/patch messages.

Compile and run tested on ipq8065.


neheb:
Ack on the second. I have not tested the first.

MartB:
Tested both on octeon, compiles and runs.

lucize:
using this pull on x86_64 since 5 days ago, nothing unusual since then

haruhitamako:
how to used it ?should i only patch e1cd083.patch and make. or anything should i choose?thank you

lucize:
I see no options to be selected in the diff files so it should be straightforward

haruhitamako:
oh! thanks

stintel:
Going to pull this into my staging tree and run some tests with this. Since all patches were accepted upstream I do not see why we shouldn't backport them to our codebase. Thanks!

stintel:
Pulled into my staging tree: https://git.lede-project.org/?p=lede/stintel/staging.git

PR #1334, opened 2017-09-07
utils/e2fsprogs: Update to 1.43.6

diizzyy:
Update e2fsprogs to 1.43.6
Disable compilation of fuse2fs (we don't package it)
Disable thread support (only affects fuse2fs)
Enable linking with libblkid instead of using private (included) version.

### DIFF
--- a/package/utils/e2fsprogs/Makefile
+++ b/package/utils/e2fsprogs/Makefile
@@ -8,8 +8,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=e2fsprogs
-PKG_VERSION:=1.43.5
-PKG_HASH:=261f3d9ade383fbf032a19140c9c25e998cc0f71a1ae686614fb3ae0eb955a17
+PKG_VERSION:=1.43.6
+PKG_HASH:=c9b226234ee878a8d89951f3b155cec7416dd0ec09c932c855f1b1df8ad93402
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


diizzyy:
Compile tested: mvebu, Linksys WRT3200ACM, LEDE trunk
Run tested: mvebu, Linksys WRT3200ACM, LEDE trunk

diizzyy:
To be clear, the ```ac_cv_lib_pthread_sem_init=no``` entry does disable threads but using configure arg is less of a "hackish" way of doing it.

hauke:
The libblkid.so.1.1.0 from util-linux is about 210 KBytes and the version from e2fsprogs is about 50 Kbytes, When statically linking it the binaries are getting about 25 Kbytes bigger each. I think this change will make the overall system bigger.

diizzyy:
Since we have external packages pulling libblkid I figured it was a good tradeoff. Also, I'm not misreading the Makefile e2fsprogs still pulls libblkid externally as libext2fs has it as a dependency (which e2fsprogs depends on) so that logic seems to been a bit broken (?). Also, as a block-mount user you probably want the full lib as the tiny lib doesn't recognize certain types of filesystems.

diizzyy:
As a side note I also think it's good idea to try to be inline with other linux distros such as Alpine Linux and Arch Linux which both use the external libblkid library making troubleshooting easier but I do understand the concern about size.

hauke:

PR #1335, opened 2017-09-08
tools/e2fsprogs: Update to 1.43.6

diizzyy:
Update e2fsprogs to 1.43.6
Remove FreeBSD patch as it's not needed
Remove libmagic patch, RHEL 5 is EoL (End of Production Phase) since
March 31, 2017.


hauke:
Are you sure the freebsd patch is not needed any more? It was added in this commit:
https://github.com/lede-project/source/commit/1f77bfff9608f97996bffdb28dabec4e787631f6

diizzyy:
* 9.1 is EOL since December 31, 2014 (the entire 9.X branch is EOL now)
https://lists.freebsd.org/pipermail/freebsd-announce/2014-October/001586.html
* The FreeBSD ports tree does patch e2fsprogs however not the part we patch
https://svnweb.freebsd.org/ports/head/sysutils/e2fsprogs/files/
* I did manually compile e2fsprogs on 11.1 and it compiles just fine as long as you use g(nu)make (which is used by default and also used by ports) https://svnweb.freebsd.org/ports/head/sysutils/e2fsprogs/Makefile?revision=449579&view=markup#l21

hauke:

PR #1336, opened 2017-09-08
ramips: add support for TP-Link Archer C20 v1

Linaro1985:
ramips: add support for TP-Link Archer C20 v1

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -86,6 +86,12 @@ c108)
 	ucidef_set_led_netdev "lan" "lan" "$board:green:lan" "eth0"
 	ucidef_set_led_netdev "modem" "modem" "$board:green:modem" "wwan0"
 	;;
+c20)
+	ucidef_set_led_switch "lan" "lan" "$board:blue:lan" "switch0" "0x1e"
+	ucidef_set_led_switch "wan" "wan" "$board:blue:wan" "switch0" "0x01"
+	set_usb_led "$board:blue:usb"
+	ucidef_set_led_netdev "wlan2g" "wlan2g" "$board:blue:wlan2g" "wlan0"
+	;;
 c20i)
 	ucidef_set_led_switch "lan" "lan" "$board:blue:lan" "switch0" "0x1e"
 	ucidef_set_led_switch "wan" "wan" "$board:blue:wan" "switch0" "0x01"
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -118,6 +118,7 @@ get_status_led() {
 	w502u)
 		status_led="$board:blue:wps"
 		;;
+	c20|\
 	d240|\
 	dap-1350|\
 	na930|\
--- a/target/linux/ramips/dts/ArcherC20.dts
+++ b/target/linux/ramips/dts/ArcherC20.dts
@@ -1,0 +1,189 @@
+/dts-v1/;
+
+#include "mt7620a.dtsi"
+
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+
+/ {
+	compatible = "tplink,c20", "ralink,mt7620a-soc";
+	model = "TP-Link Archer C20";
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		lan {
+			label = "c20:blue:lan";
+			gpios = <&gpio0 1 GPIO_ACTIVE_LOW>;
+		};
+
+		power {
+			label = "c20:blue:power";
+			gpios = <&gpio0 7 GPIO_ACTIVE_LOW>;
+		};
+
+		usb {
+			label = "c20:blue:usb";
+			gpios = <&gpio2 4 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			label = "c20:blue:wan";
+			gpios = <&gpio0 11 GPIO_ACTIVE_HIGH>;
+		};
+
+		wan_orange {
+			label = "c20:orange:wan";
+			gpios = <&gpio0 9 GPIO_ACTIVE_HIGH>;
+		};
+
+		wlan5g {
+			label = "c20:blue:wlan5g";
+			gpios = <&gpio0 17 GPIO_ACTIVE_LOW>;
+		};
+
+		wlan2g {
+			label = "c20:blue:wlan2g";
+			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
+		};
+
+		wps {
+			label = "c20:blue:wps";
+			gpios = <&gpio1 15 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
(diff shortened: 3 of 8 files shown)


mkresin:
Your commit message is empty. Please follow the advice from the [Submitting patches](https://lede-project.org/submitting-patches) guide:

> If you add support for new hardware: Include in your commit message short description of the hardware and how to install LEDE on it. Have a look at the [recent additions](https://git.lede-project.org/?p=source.git&a=search&h=HEAD&st=commit&s=add+support+for) for some examples.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -82,6 +82,13 @@ broadway)
 	set_usb_led "$board:red:diskmounted"
 	set_wifi_led "$board:red:wps_active"
 	;;
+c20)
+	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
```
mkresin: Drop this line, the LED is handled by the diag.sh script and should be on after boot by default. If the LED is off after boot, the GPIO_ACTIVE value for the LED is wrong in the device tree source file.

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -128,6 +128,7 @@ get_status_led() {
 	wl-330n|\
 	wl-330n3g|\
 	wli-tx4-ag300n|\
+	c20|\
```
mkresin: please keep alphabetical order. You might need to move whole block to keep the overall order.

target/linux/ramips/dts/ArcherC20.dts
```diff
@@ -0,0 +1,189 @@
...
+			reg = <0x7f0000 0x10000>;
+			read-only;
+		};
+	};
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "i2c", "uartf", "rgmii1", "rgmii2", "wled", "ephy", "spi refclk", "mdio", "wdt", "nd_sd";
```
mkresin: Remove the mdio, rgmii11, nd_sd and rgmii2 groups from the pinmux. None of the GPIOs covered by these groups are used.

Linaro1985:
> Please add a bootlog (serial console or dmesg) to PR!
1) Which one? Stock or lede bootlog?
2) Where I can place it?

Thanks!

target/linux/ramips/base-files/etc/board.d/01_leds
Linaro1985: Thanks. I will check it.

mkresin:
> Which one? Stock or lede bootlog?

LEDE

>  Where I can place it? Can I attach here?

Yes, attach it to a comment here.

Please rebase you PR in master. It has already merge conflicts due to a recent commit.

Linaro1985:
I completed changes in code) Thanks!
[lede_bootlog.txt](https://github.com/lede-project/source/files/1291567/lede_bootlog.txt)

mkresin:
Thanks a lot for the update.

The fact that only the wan leds are active on high looks strange. Would please test if a brightness value of 255 switches the leds to on and a brightness value of 0 switches the leds off.

The command for the orange wan led would be `echo 255 > "/sys/class/leds/c20:orange:wan/brightness"`.

Linaro1985:
> echo 255 > "/sys/class/leds/c20:orange:wan/brightness"
Orange led is on after this command.

P.S. I tested all leds via web interface before pull request and all works fine.

mkresin:
@Linaro1985 Please use your real name in the Signed-off-by.

> We are able to prepare factory firwmare file which is accepted and
(almost) correctly flashed from the vendor GUI. As it turned out, it
accepts files without U-Boot image with second header at the beginning
but due to some kind of bug in upgrade routine, flashed image gets
corrupted before it's written to flash. So, to flash this device we must
to prepare image using original firmware from tp-link site with uboot.

And u-boots tftp doesn't accept an image without U-Boot image with second header at the beginning, like the lantiq based TD-W8980v1 (using `tplink-v2-image -s -V "ver. 1.0" `)?

@pepe2k I remember that you faced the same bug with tp-link GUI update routine. As far as I remember a bit of the image is overwritten before flashing. What was your workaround?

Linaro1985:
> @Linaro1985 Please use your real name in the Signed-off-by.
Done.

> And u-boots tftp doesn't accept an image without U-Boot image with second header at the beginning, like the lantiq based TD-W8980v1 (using tplink-v2-image -s -V "ver. 1.0")?

u-boot has hardcoded update command:
`tftp 0x80060000 ArcherC20V1_tp_recovery.bin;erase tplink 0x00000 0x7c0000;cp.b 0x80060000 0x00000 0x7c0000`

In this command the image loads from the tftp server. Next flash memory will be erased from 0x00000 to 0x7c0000. After that the loaded image is flashed from the address 0x00000. So we can't use image without u-boot binary otherwise the uboot area will be wrong and we will receive a brick.

mkresin:
> u-boot has hardcoded update command:
tftp 0x80060000 ArcherC20V1_tp_recovery.bin;erase tplink 0x00000 0x7c0000;cp.b 0x80060000 0x00000 0x7c0000

> To access U-Boot CLI, keep pressed "4" key during boot.

What about running your own flash command from the U-Boot CLI:

```
tftp 0x80060000 lede-image-name.bin;erase tplink 0x20000 0x7a0000;cp.b 0x80060000 0x20000 0x7a0000
```

I'm not sure about the erase command syntax. Usually it is `erase 0x20000 +0x7a0000`. Might be that TP-Link changed the syntax of the command.

Linaro1985:
> tftp 0x80060000 lede-image-name.bin;erase tplink 0x20000 0x7a0000;cp.b 0x80060000 0x20000 0x7a0000
Yes, I think it will work. But not all have an opportunity to be connected to UART port of the device! U-boot hardcoded command automatically called when we press reset button during 7-8 seconds. And in this case the device shouldn't be disassembled.

mkresin:
> tftp 0x80060000 lede-image-name.bin;erase tplink 0x20000 0x7a0000;cp.b 0x80060000 0x20000 0x7a0000
> Yes, I think it will work. But not all have an opportunity to be connected to UART port of the device! U-boot hardcoded command automatically called when we press reset button during 7-8 seconds. And in this case the device shouldn't be disassembled.
> 
Yeah got it and already pointed out this disadvantage on my own in my edit. Unfortunately it isn't uncommon that access to the serial console is required to flash LEDE.

Nevertheless, I would like to see a tested method documented in the commit message which doesn't require any further image modification or external files. No idea if the C20 v1 firmware file is still provided by tp-link in 5 years.

Linaro1985:
>  tftp 0x80060000 lede-image-name.bin;erase tplink 0x20000 0x7a0000;cp.b 0x80060000 0x20000 0x7a0000

> Nevertheless, I would like to see a tested method documented in the commit message which doesn't require any further image modification or external files. 

Right now I have no opportunity to check command because of there is no device nearby. Tomorrow I will check it and if everything is OK I'll add this method in the commit message if isn't success then it is necessary to look for spi flash programmer:)

Linaro1985:
I have tested the second way and have added it to the instruction.
Thanks!

mkresin:
Thank for your contribution. I've merged your PR to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I've reorder the installation instructions to have the one which doesn't require any external files first.

PR #1339, opened 2017-09-09
ramips: fix hg255d support

yangfl:
Fix LED and MAC support for hg255d

Signed-off-by: Yangfl <mmyangfl@gmail.com>

### DIFF
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -24,6 +24,7 @@ get_status_led() {
 	ex2700|\
 	ex3700|\
 	fonera20n|\
+	hg255d|\
 	kn|\
 	kn_rc|\
 	kn_rf|\


mkresin:
Thanks for your contribution. Looks good so far.

> Signed-off-by: Yangfl <mmyangfl@gmail.com>

We have a realname policy for Signed-off-by. "Yangfl" doesn't look like a full name to me.

target/linux/ramips/base-files/lib/ramips.sh
```diff
@@ -235,7 +235,7 @@ ramips_board_detect() {
 	*"HC5962")
 		name="hc5962"
 		;;
-	*"HG255D")
```
mkresin: Drop this change. It should make no difference since the "HuaWei" is already covered by the wildcard "*".

mkresin:
I've merged your commit to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I intend to backport this change to the lede-17.01 (stable) branch as well, since the diag LED is essential for the first troubleshooting.

PR #1342, opened 2017-09-10
hostapd: Don't implicitly add WPA2-PSK-SHA256 if PMF is enabled.

neheb:
Some older g clients (Notably Apple iDevices) have an issue where if WPA2-PSK-SHA256 is advertised, it mistakenly uses that instead of SHA1.

While both PMF and WPA2-PSK-SHA256 come from the same standard (802.11w), there's no reason for them to be enabled together unless explicitly desired.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

PSK-SHA256 should probably be configurable through UCI.


stintel:
NAK. This patch just rips out the -SHA256 algorithms, without adding a possibility to enable them.

hauke:
NACK: please make this configurable, and use the current setting as the default.

mkresin:
Please reopen the PR if you have done the requested changes.

PR #1343, opened 2017-09-11
Bash syntax errors in packages

BigNerd95:
Fix multiple syntax errors in bash files (of packages only)
These errors were causing many conditions to not working properly

Signed-off-by: Lorenzo Santina <lorenzo.santina@edu.unito.it>

### DIFF
--- a/package/network/services/hostapd/files/hostapd.sh
+++ b/package/network/services/hostapd/files/hostapd.sh
@@ -627,7 +627,7 @@ wpa_supplicant_add_network() {
 		scan_ssid=""
 	}
 
-	[[ "$_w_mode" = "adhoc" -o "$_w_mode" = "mesh" ]] && append network_data "$_w_modestr" "$N$T"
+	[ "$_w_mode" = "adhoc" -o "$_w_mode" = "mesh" ] && append network_data "$_w_modestr" "$N$T"
 
 	case "$auth_type" in
 		none) ;;
--- a/package/network/utils/comgt/files/directip.sh
+++ b/package/network/utils/comgt/files/directip.sh
@@ -44,8 +44,7 @@ proto_directip_setup() {
 		return 1
 	}
 
-	cardinfo=$(gcom -d "$device" -s /etc/gcom/getcardinfo.gcom)
-	[ -n $(echo "$cardinfo" | grep -q "Sierra Wireless") ] || {
+	$(gcom -d "$device" -s /etc/gcom/getcardinfo.gcom | grep -q "Sierra Wireless") || {
 		proto_notify_error "$interface" BAD_DEVICE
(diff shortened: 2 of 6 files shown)


package/network/services/hostapd/files/hostapd.sh
```diff
@@ -627,7 +627,7 @@ wpa_supplicant_add_network() {
 		scan_ssid=""
 	}
 
-	[[ "$_w_mode" = "adhoc" -o "$_w_mode" = "mesh" ]] && append network_data "$_w_modestr" "$N$T"
+	[[ "$_w_mode" = "adhoc" || "$_w_mode" = "mesh" ]] && append network_data "$_w_modestr" "$N$T"
```
mkresin: Would you please move this change into an own commit!

This is the only change related to bash compatibility (LEDE uses busybox ash by default). All other changes are fixing common syntax/typos/logic errors.

I would suggest to use the following commit titles or similar:

- treewide: fix shellscript syntax errors/typos
- hostapd: fix bash compatibility

package/network/utils/comgt/files/directip.sh
```diff
@@ -45,7 +45,7 @@ proto_directip_setup() {
 	}
 
 	cardinfo=$(gcom -d "$device" -s /etc/gcom/getcardinfo.gcom)
-	[ -n $(echo "$cardinfo" | grep -q "Sierra Wireless") ] || {
+	$(echo "$cardinfo" | grep -q "Sierra Wireless") || {
```
mkresin: Shouldn't we change the whole line to something like

`gcom -d "$device" -s /etc/gcom/getcardinfo.gcom | grep -q "Sierra Wireless" || {`

I mean the cardinfo variable is never set to local as it should be and not really used

dedeckeh:
Indeed excellent catch; could you also increase the PKG_RELEASE in the relevant package Makefiles (gre/comgt/hostapd/mac80211)  ?

package/network/services/hostapd/files/hostapd.sh
BigNerd95: I decided to remove useless [ ] instead of replace -o with ||

mkresin:
I added the missing PKG_RELEASE increments as requested by Hans, dropped the unnecessary command substitution from directip.sh, backported the commit to the stable lede-17.01 branch and merged everything to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

@dedeckeh Would you please have a look especially at the PKG_RELEASE increments. I have to admit, I didn't took care of these till now.

@BigNerd95 Fine with my changes?

BigNerd95:
Yes, I think they are perfect

PR #1345, opened 2017-09-12
generic: drop support for get_port_stats() on ar8xxx

f00b4r0:
The implementation is not efficient on ar8xxx switches. It triggers high
CPU load and degrades device performance, see FS#1004.

This commit removes the get_port_stats() code introduced in 4d8a66d and
leaves a note for future hacker's beware.

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/ar8216.c
+++ b/target/linux/generic/files/drivers/net/phy/ar8216.c
@@ -49,12 +49,6 @@ extern const struct ar8xxx_chip ar8337_chip;
 		.name = (_n),	\
 	}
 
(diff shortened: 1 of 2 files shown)


neheb:
FWIW it might be worth retesting and changing the usleep_range calls to udelay or mdelay. I've seen too many examples already where usleep_range has been causing issues in kernel development.

mkresin:
I've merged cour commit to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

I've changed the commit message and the comment in the code slightly and disabled the get_port_stats op in ar8327.c as well.

Are you fine with my changes?

f00b4r0:
LGTM, thanks

PR #1347, opened 2017-09-13
ramips: Correct WAN MAC address for Phicomm K2P

jwang-link:
For example, if the factory partition contains:
000e000 7d74 be24 f287 7d74 be24 f187 ffff ffff

In stock firmware the WAN/LAN MAC addresses are
74-7D-24-BE-87-F1/2
while in LEDE the WAN/LAN MAC addresses are
74-7D-24-BE-87-F3/2

This commit correct it.

Signed-off-by: Jiawei Wang <me@jwang.link>

### DIFF
--- a/target/linux/ramips/dts/K2P.dts
+++ b/target/linux/ramips/dts/K2P.dts
@@ -106,7 +106,7 @@
 };
 
 &ethernet {
-	mtd-mac-address = <&factory 0xe000>;
+	mtd-mac-address = <&factory 0xe006>;
 };
 
 &pinctrl {


mkresin:
Please change the `mtd-mac-address` property of the ethernet node in the K2P.dts to `<&factory 0xe004>;`

jwang-link:
Thank you for reviewing this PR.
Sorry I do not understand why the mtd-mac-address property should be changed to <&factory 0xe004>.
And it results in a random generation for the LAN MAC address.

mkresin:
At the moment you have `mtd-mac-address = <&factory 0xe000>` in the dts, which has to be read as use the mac address (next 4bytes) at offset 0xe000 for eth0. The switch driver uses the eth0 mac address for the first vlan (eth0.1 - lan) and increments the mac address for the second vlan (eth0.2 -wan). 

You have two mac addresses in the factory partition:
- offset 0xe000: 7d74 be24 f287 == 74-7D-24-BE-87-F2
- offset 0xe004: 7d74 be24 f187 == 74-7D-24-BE-87-F1

If you take the switch driver mac address increment magic into account, you need to use the mac address at offset 0xe004 of the factory partition by default.

Changing `mtd-mac-address = <&factory 0xe000>` to `mtd-mac-address = <&factory 0xe004>` in [K2P.dts line 109](https://github.com/lede-project/source/blob/master/target/linux/ramips/dts/K2P.dts#L109) should exactly do this.

jwang-link:
The commit has been updated.
However, the WAN/LAN MAC addresses are swapped, compared with the stock firmware.
Please check if it's an issue for the LEDE project.

mkresin:
> However, the WAN/LAN MAC addresses are swapped, compared with the stock firmware.
If it's LEDE's style to always set WAN = LAN + 1, I think we are done.

It isn't really LEDE style. We try to use the same MAC addresses for the same interfaces as the stock firmware.

But pointing to a MAC address for a specific vlan interface via device tree isn't possible in LEDE at the moment. For that reason the switch driver of this particular target has this auto increment magic.

Sure, it can be fixed in user space, but that will cause other issues like mac adresses in the config backup, which will cause duplicate MAC address issues if you restore the backup from one device to another. The userspace approach is usally used if the MAC addresses can not be read from the (raw) flash (because it is stored in u-boot env, ubi volume and so on).

As long as we do not use MAC addresses which are not intended/reserved for a board, I'm fine with having the MAC addresses assigned to the wrong interfaces.

Anyway, I've rewritten the complete commit message and merged the commit to [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Are you fine with the new commit message?

jwang-link:
I get it. Thank you for rewriting the commit message. I am fine with it.

mkresin:
Merged, thanks!

PR #1348, opened 2017-09-14
samba4x: add new package

hbl0307106015:
This pull request adds the initial support for latest stable samba project. Since the samba36 does really fit to embedded system which flash size is limited, I think making samba4x and samba36 coexist is a better chioce. By the way, new samba4x seems to require at least 32MB flash size :-).

Maintainer: nbd
Compile tested: x86_64, x86_64 relevant machine, LEDE latest
Compile tested via C Library: both musl and glibc. :-)

Signed-off-by: BangLang Huang banglang.huang@foxmail.com

### DIFF
--- a/package/network/services/samba4x/Makefile
+++ b/package/network/services/samba4x/Makefile
@@ -1,0 +1,134 @@
+#
+# Copyright (C) 2007-2017 OpenWrt
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=samba4x
+PKG_VERSION:=4.7.0
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/samba-team/samba.git
+PKG_SOURCE_VERSION:=d1e69845e28c20a491c4cd60c712b46ddfcb9dc0
+PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
+PKG_MIRROR_HASH:=97e6fa7c0bc951ab19265fb2065695a2bef05c1fee94d74319124a96a3879f07
+PKG_LICENSE:=GPL-3.0
+
+PKG_MAINTAINER:=Felix Fietkau <nbd@nbd.name>
+PKG_INSTALL:=1
+
+include $(INCLUDE_DIR)/package.mk
+# for $(LINUX_VERSION)
+include $(INCLUDE_DIR)/kernel.mk
+# for $(VERSION_DIST)
+include $(INCLUDE_DIR)/version.mk
+
+define Package/samba4x-server
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Samba $(PKG_VERSION) SMB/CIFS server
+  URL:=http://www.samba.org/
+  DEPENDS:=+zlib +libpopt +libgnutls +tdb
+endef
+
+define Package/samba4x-server/description
+  The Samba software suite is a collection of programs that implements the
+  SMB protocol for UNIX systems, allowing you to serve files and printers to
+  Windows, NT, OS/2 and DOS clients. This protocol is sometimes also referred
+  to as the LanManager or Netbios protocol.
+endef
+
+CONFIGURE_VARS += \
+	CPP="$(TARGET_CROSS)cpp" \
+	python_LDFLAGS="" \
+	python_LIBDIR=""
+
+CONFIGURE_CMD = ./buildtools/bin/waf
+
+CONFIGURE_PREFIX = /usr
+
+CONFIGURE_ARGS = \
+		--target=$(GNU_TARGET_NAME) \
+		--program-prefix="" \
+		--prefix=$(CONFIGURE_PREFIX) \
+		--exec-prefix=$(CONFIGURE_PREFIX) \
+		--bindir=$(CONFIGURE_PREFIX)/bin \
+		--sbindir=$(CONFIGURE_PREFIX)/sbin \
+		--libexecdir=$(CONFIGURE_PREFIX)/lib \
+		--sysconfdir=/etc \
+		--datadir=$(CONFIGURE_PREFIX)/share \
+		--localstatedir=/var \
+		--mandir=$(CONFIGURE_PREFIX)/man \
+		--infodir=$(CONFIGURE_PREFIX)/info
+
+CONFIGURE_ARGS += \
+		--hostcc="$(HOSTCC)" \
+		--cross-compile \
+		--cross-answers="$(PKG_BUILD_DIR)/cache.txt" \
+		--disable-avahi \
+		--disable-cephfs \
+		--disable-cups \
+		--disable-fault-handling \
+		--disable-glusterfs \
+		--disable-iprint \
+		--disable-python \
+		--disable-rpath \
+		--disable-rpath-install \
+		--disable-rpath-private-install \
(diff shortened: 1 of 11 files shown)


diizzyy:
There's a branch here which makes it more embedded friendly
https://github.com/wongsyrone/lede-1/tree/master/package/external/samba44related/samba44

hbl0307106015:
@diizzyy  In comparation with the branch you referenced, I prefer to use internal heimdal module rather than external. It seems that using internal module is more stable and acceptable, like NTFS-3g using internal fuse library as default.
On the other hand, I just want to make samba to be OOTB (out of the box).   :-) .

diizzyy:
Well, it slims down Samba to ~5-6mbyte compared to yours which is 4-5x more?

hbl0307106015:
@diizzyy  Yes, I tried my best.  Is there any suggestions or tips ?

lucize:
You could add an option to use the external or the internal module and make the smallest one the default

diizzyy:
@hbl0307106015 
While I think the other package is much more suitable I don't like the huge amount of patching and there's something wonky with heimdal too as it will fail during rebuilds I think it's a good way to move forward and perhaps adapt a few things such as the multicall binary stuff.

buliaoyin:
The lastest samba4 beta version added --disable-python built option, maybe this option would revive the package size?

hbl0307106015:
@buliaoyin thank you, this is a good news, it would be better if the option released with 4.7.0-stable version.

hauke:
Samba 4.7 should get release on September 21 2017
How big is the samba 4.X package now and how big would it be with the patches which @diizzyy linked?

diizzyy:
Samba 4.4.14 - 4.9M on ar71xx and 5M on mvebu

ar71xx: 5073284 bytes
mvebu: 5184299 bytes

wongsyrone:
Good to see you guys are taking care of it. I'm not using it for a long time but I would like to help if any information is needed. Please check these two issues to get more info.

https://github.com/wongsyrone/lede-1/issues/25
https://github.com/wongsyrone/lede-1/issues/26

hbl0307106015:
banglang@banglang-PC:~/work/lede/x86$ ls -lh bin/packages/x86_64/base/samba4x-server_4.6.7-1_x86_64.ipk 
-rw-r--r-- 1 banglang banglang **6.6M** Sep 21 16:38 bin/packages/x86_64/base/samba4x-server_4.6.7-1_x86_64.ipk

hbl0307106015:
@lucize  All the options which marked 'default=yes' and 'enable' are turned to 'no' and 'disable'. So this is the smallest package I could make.

diizzyy:
@hbl0307106015 
On a fresh checkout of trunk it fails :-/

```
Checking for custom code                                                          : Could not find the python development headers
```

Adding 
```TARGET_CFLAGS+=-I$(STAGING_DIR)/usr/include/python2.7 $(TARGET_CFLAGS) ```
Fixes that issue

However:
```
[ 244/3645] Linking default/source4/heimdal_build/compile_et
/usr/bin/ld: default/lib/replace/replace_1.o: Relocations in generic ELF (EM: 40)
default/lib/replace/replace_1.o: error adding symbols: File in wrong format
/usr/bin/ld: default/lib/replace/replace_1.o: Relocations in generic ELF (EM: 40)
default/lib/replace/replace_1.o: error adding symbols: File in wrong format
```
And it fails

hbl0307106015:
@diizzyy Python module is disabled now, try it again,
Let's bump to samba 4.7.0 stable :-)...

wongsyrone:
Just some random guesses since I don't have my computer at hand.

> Python module is disabled now

The python module you referred might be those get called at runtime, the most important thing here is we want to modify waf, the build system in python, to get all it needs for cross compiling.

hbl0307106015:
Maybe the issue is that I canceled the dependence on python-light (DEPENDS=+python-light). All right, I will recheck it from a fresh check out of LEDE.

wongsyrone:
If you add python-light to DEPENDS, it means runtime dependency.

wongsyrone:
> PKG_BUILD_DEPENDS:=python-light

This is build dependency, just to satisfy waf and we don't need it at runtime.

hbl0307106015:
@wongsyrone thanks for your explanation :-). By the way, are u chinese as well ?

diizzyy:
Fails at the exact some thing as before (arm)

hbl0307106015:
@diizzyy how about after adding the line PKG_BUILD_DEPENDS = python-light ??

diizzyy:
That's no it, Samba seems to compile for the wrong arch. Try arm yourself and you'll see the issue.

KanjiMonster:
Python light is part of the external packages feed, not of LEDE itself. Base packages may not depend on packages from external feeds.

Therefore this pull request needs to be done against the OpenWrt packages feed.  In addition, I seriously doubt @nbd agreed being the maintainer. Please add yourself or find someone willing to be it.

diizzyy:
Was going to address that once we all got it compiling but oh well ;-)

hbl0307106015:
ls -lh bin/packages/mipsel_24kc/base/samba4x-server_4.7.0-1_mipsel_24kc.ipk 
-rw-r--r-- 1 banglang banglang **5.2M** Sep 22 17:55 bin/packages/mipsel_24kc/base/samba4x-server_4.7.0-1_mipsel_24kc.ipk

diizzyy:
that's mips, not arm/mvebu

hbl0307106015:
@diizzyy This is for you, :-)...
ls -lh bin/packages/arm_cortex-a9_vfpv3/base/samba4x-server_4.7.0-1_arm_cortex-a9_vfpv3.ipk 
-rw-r--r-- 1 banglang banglang **6.0M** Sep 22 20:09 bin/packages/arm_cortex-a9_vfpv3/base/samba4x-server_4.7.0-1_arm_cortex-a9_vfpv3.ipk

hbl0307106015:
Hi @KanjiMonster , thanks for your reminding but I made it, the new samba4x package does not need external package(including python-light) feeds any more. So could you reopen this pull request ? :-)

diizzyy:
No, python-light isn't in LEDE it's in the openwrt package repo. This is also true for tdb, gnutls for instance.

Andy2244:
Oh nice, someone else is working on current samba 4.7 as well. Was running 4.6 with my patches (based on wongsyrone 4.x), but will try your version as well.

hbl0307106015:
@diizzyy Sorry for my poor English :-( ...  I just wanted to say the samba4x does not need python-light any more.

Andy2244:
If anyone is interested here is my latest samba 4.7 feed, based on @hbl0307106015 and @wongsyrone versions.

My version adds some patches to reduce nvram wear and does a full installation, so we can use vfs modules and nearly all extra samba cmd tools, since its just a few 100kb larger, than manually selecting the bins/libs.
I also still use wongsyrone external Heimdal, since hbl0307106015 big internal heimdal cross-compile patch, was to large to review properly and the external Heimdal version worked fine for me since 2 years.

I also adapted the init script to allow more samba share options, like the vfs modules, not all are exposed to the luci-ui. wongsyrone version also lowers the process priority of smbd, nmbd, which i think is usefully, so i still use this init logic.

[myfeed.zip](https://github.com/lede-project/source/files/1327317/myfeed.zip)

avdarie:
Hi, 

Have you prevision when the package will update in official Lede package list ? Or will be only unfficial. I m interested to install in my Lamobo BananaPi R1. 

Thanks, 
Andrei Darie

hbl0307106015:
@avdarie  here it is https://github.com/openwrt/packages/pull/4875, I pushed it to openwrt-packages repository.

Andy2244:
cool thx to hbl0307106015 again, will see if i get time next week to add my patches/changes

PR #1349, opened 2017-09-15
dnsmasq: fix dhcp "ignore" option on wwan interfaces

marcin1j:
Init script won't append --no-dhcp-interface option if interface
protocol is one of: ncm, directip, qmi, mbim.
This is caused by IP address assigned to dynamically created netifd
interfaces. As a result there's no netmask assigned to the main
interface and dhcp_add() function returns prematurely.

By moving network subnet check we can ensure that --no-dhcp-interface is
properly generated for wwan interfaces.

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -457,7 +457,6 @@ dhcp_add() {
 	config_get networkid "$cfg" networkid
 	[ -n "$networkid" ] || networkid="$net"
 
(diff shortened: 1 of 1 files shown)


dedeckeh:
Thanks; merged into my staging tree with some small adaptations (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=ff7bd8afe52fc8ea5272168e5b09d4e8a6cbecf6)
Please also increase PKG_RELEASE for future patches

PR #1350, opened 2017-09-15
kernel: add packaging for Xeon iTCO watchdog timer

pprindeville:
Third (I believe) generation and later Xeon chipsets included the iTCO watchdog timer on the processor.  Add module support for this.


pprindeville:
Don't know if this has to do with this being a boot-time module or not, but going into the file `/etc/modules.d/50-itco_wdt' and changing the line:

`iTCO_vendor_support`

to:

`iTCO_vendor_support vendorsupport=911`

and rebooting does not seem to change the value of this parameter:

`[    4.145868] iTCO_vendor_support: vendor-support=0`

This might be a broader problem related to FS #935.

target/linux/x86/64/config-default
```diff
@@ -201,7 +201,7 @@ CONFIG_INTEL_IDLE=y
 # CONFIG_IOMMU_DEBUG is not set
 # CONFIG_ISCSI_IBFT_FIND is not set
 # CONFIG_ITCO_VENDOR_SUPPORT is not set
-CONFIG_ITCO_WDT=y
+# CONFIG_ITCO_WDT is not set
```
hauke: Why should this not be build into the kernel any more for this target?

pprindeville: Because (1) it's not applicable to Pentium, Atom, or any flavor or AMD processor (k8-64 or k10), (2) because watchdogs are traditionally built as kmods, and (3) because you might want to configure the parameters that it's loaded with (without having to diddle the grub parameters).

hauke:

pprindeville:
Thanks!

diizzyy:
@pprindeville 
Can you provide a style fix for this, all other modules are in alphabetical order as far as I can tell.

pprindeville:
I don't think they are. Shouldn't they be sorted by type and then by name alphabetically within that if we were organizing them?

diizzyy:
Yeah, nm... I'm not sure what the order is looking at it again :-)

PR #1353, opened 2017-09-17
kernel: Fix commit a5922f6

diizzyy:
Add kmod-mmc as dependency and only build this if platform is mvebu

### DIFF
--- a/package/kernel/linux/modules/other.mk
+++ b/package/kernel/linux/modules/other.mk
@@ -117,7 +117,7 @@ $(eval $(call KernelPackage,bluetooth_6lowpan))
 define KernelPackage/btmrvl
   SUBMENU:=$(OTHER_MENU)
   TITLE:=Marvell Bluetooth Kernel Module support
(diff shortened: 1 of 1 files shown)


diizzyy:
@hauke  @heinzek 
This should fix the buildbots

mkresin:
Please fix you commit title. `kernel: btmrvl: add mmc dependency` describes what you are doing. Reference the commit you fix in the commit message.

IMHO this module should not have a dependency on mvebu. The kernel helptext says:

> This driver is required if you want to support Marvell Bluetooth devices, such as 8688/8787/8797/8887/8897/8997.

At least the 88W8897 is used on a  Azurewave AW-CB178NF, which is a M.2 module and can be used on more that the mvebu platform.

mkresin:
@hauke made the requested changes and merged the commit. Thanks!

diizzyy:
@mkresin 
In theory indeed but that's probably unlikely to occur in the end? It was just a quick attempt to avoid building it for several platforms where it's never going to be used. Either way, we have working buildbots now. :-)

PR #1355, opened 2017-09-18
sunxi: add Olimex A20-OLinuXino-LIME2-eMMC

lucize:
sunxi: add Olimex A20-OLinuXino-LIME2-eMMC


hauke:
Thank you for the patch, it was applied with some small modifications.

PR #1356, opened 2017-09-19
ramips: reduce napi_weight in the ethernet driver.

neheb:
Currently dmsg a weight of 128 which is above the kernel limit of 64. Silence the warning.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


blogic:
merged, Thanks !

PR #1357, opened 2017-09-19
tools: flex: fix segmentation fault with gcc 7.2.0

ratkaj:
Fix segmentation fault caused by implicit declaration of function 'reallocarray'. Added patch will enable reallocarray() prototype in glibc 2.26+ on Linux systems. This fix will be included in flex 2.6.5.

Affected systems:
Arch Linux and other systems with glibc 2.26+

Closes issue:
https://bugs.lede-project.org/index.php?do=details&task_id=1003

### DIFF
--- a/tools/flex/patches/200-build-AC_USE_SYSTEM_EXTENSIONS-in-configure.ac.patch
+++ b/tools/flex/patches/200-build-AC_USE_SYSTEM_EXTENSIONS-in-configure.ac.patch
@@ -1,0 +1,27 @@
+From 24fd0551333e7eded87b64dd36062da3df2f6380 Mon Sep 17 00:00:00 2001
+From: Explorer09 <explorer09@gmail.com>
+Date: Mon, 4 Sep 2017 10:47:33 +0800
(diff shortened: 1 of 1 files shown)


blogic:
add full annotation explaining what problem the patch fixes and how please. also send the fix upstream pleaseplease

zorun:
@ratkaj basically, add the full upstream commit (with its commit message) in the patch.  You can do that with `git format-patch -1 24fd0551333e7eded87b64dd36062da3df2f6380` (if I'm not mistaken and this is indeed https://github.com/westes/flex/commit/24fd0551333e7eded87b64dd36062da3df2f6380 )

Also, since the upstream commit message does not mention anything about a segfault or gcc 7.2, you need to elaborate on the fix (what it does and how it fixes the problem).

Lastly, please mention that it fixes FS#1003 in your commit message (not just in the pull request message).

ratkaj:
Updated pull request by adding full upstream commit and correcting my commit message.
Updated pull request message since issue was caused bu glibc 2.26 not gcc 7.2

hauke:

PR #1358, opened 2017-09-19
uboot-mvebu: add missing OpenSSL compat patches

ratkaj:
Add missing OpenSSL compatibility patches. Fixes the following build
issue: "undefined reference to `EVP_MD_CTX_create'".

Add missing UBOOT_MAKE_FLAGS variable.
Fixes "/bin/sh: HOSTCPPFLAGS: command not found" build errors

Fix "SETEXPR" redefinition warning

### DIFF
--- a/package/boot/uboot-mvebu/Makefile
+++ b/package/boot/uboot-mvebu/Makefile
@@ -16,6 +16,8 @@ PKG_HASH:=f54baf3f9325bf444c7905f3a5b6f83680edb1e6e1a4d5f8a5ad80abe885113f
 include $(INCLUDE_DIR)/u-boot.mk
 include $(INCLUDE_DIR)/package.mk
 
+UBOOT_MAKE_FLAGS:=
+
 define U-Boot/Default
   BUILD_TARGET:=mvebu
   HIDDEN:=1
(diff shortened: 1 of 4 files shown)


diizzyy:
I'm probably missing something here but we don't have (at least yet) OpenSSL 1.1+ in tree?

package/boot/uboot-mvebu/Makefile
```diff
@@ -16,6 +16,8 @@ PKG_HASH:=f54baf3f9325bf444c7905f3a5b6f83680edb1e6e1a4d5f8a5ad80abe885113f
 include $(INCLUDE_DIR)/u-boot.mk
 include $(INCLUDE_DIR)/package.mk
 
+UBOOT_MAKE_FLAGS:=
+
```
hauke: Why is this needed?

hauke:
It would make sense if you split this into multiple patches.
One patch with the backports needed to fix the openssl 1.1.x problems
One patch for the SETEXPR fix
and the for the other problem.

package/boot/uboot-mvebu/Makefile
ratkaj: Douring build we get more than 60 errors:
/bin/sh: HOSTCPPFLAGS: command not found

Above change will remove those errors. Several other uboot Makefiles use this, examples are uboot-ar71xx and uboot-lantiq

ratkaj:
@diizzyy sry for the late response
we are using 1.0.2L and "L" is where the problems started. These patches will add compatibility with 1.1.x and fix the issues. Maybe we can make make a smaller patch just for 1.0.2l but i dont have time to debug that at the moment

ratkaj:
@hauke splitted into 3 patches

hauke:
Thank you for your patches, they are applied to master now.

PR #1359, opened 2017-09-19
usbutils: avoid duplicating the git revision

pprindeville:
The same revision (SHA) was being used in 2 places, and one of them was a macro definition. So, use that macro in place of the 2nd instance.


blogic:
merged, Thanks !

PR #1361, opened 2017-09-19
build: remove @ as it's causing an error

pprindeville:
Since `$(DownloadMethod/unknown)` is being invoked in the expansion of
`$(call locked ...)` anyway, you can't have an `@` because the shell
doesn't know what to do with it.


blogic:
merged, Thanks !

PR #1362, opened 2017-09-19
ar71xx: specify full UBI size for WNDR4300

aleho:
This allows the system to use the full ubi size.

`/proc/mtd`:
```
mtd0: 00040000 00020000 "u-boot"
mtd1: 00040000 00020000 "u-boot-env"
mtd2: 00040000 00020000 "caldata"
mtd3: 00080000 00020000 "pot"
mtd4: 00200000 00020000 "language"
mtd5: 00080000 00020000 "config"
mtd6: 00300000 00020000 "traffic_meter"
mtd7: 00200000 00020000 "kernel"
mtd8: 07700000 00020000 "ubi"
mtd9: 07900000 00020000 "firmware"
mtd10: 00040000 00020000 "caldata_backup"
```

`/proc/partitions`:
```
  31        0        256 mtdblock0
  31        1        256 mtdblock1
  31        2        256 mtdblock2
  31        3        512 mtdblock3
  31        4       2048 mtdblock4
  31        5        512 mtdblock5
  31        6       3072 mtdblock6
  31        7       2048 mtdblock7
  31        8     121856 mtdblock8
  31        9     123904 mtdblock9
  31       10        256 mtdblock10
 254        0       3596 ubiblock0_0
```


mkresin:
NAK!

Such an approach was already discussed in [FS#146](https://bugs.lede-project.org/index.php?do=details&task_id=146) and there are plenty of arguments why it isn't the right way.

You should have at least mentioned that the caldata_backup partition moves due to this change and therefore will not contain the same values as before. No idea what breaks by doing this.

The correct way of fixing the issue is to find a way to create something like a (virtual) mtd block device that spans over two existing ones.

PR #1363, opened 2017-09-19
kernel: don't scrimp on memory on big iron

pprindeville:
x86_64 platforms typically don't lack memory, so don't needlessly
economize memory if fq_codel on capable platforms.


hauke:
Thank you for the patch, it was applied with some small modifications.

PR #1364, opened 2017-09-20
ipq-wifi: fix missing define of PKG_NAME

ptpt52:
make clean need PKG_NAME
rm -rf build_dir/target...../$PKG_NAME is invoked
missing PKG_NAME cause parent directory being removed


blogic:
merged, Thanks !

PR #1365, opened 2017-09-20
kernel: kmod-usb-storage-uas

jcadduono:
This will allow you to build and package the uas.ko module.
With more routers supporting USB 3.0 host this could help
speed up activities like DLNA and Samba, as well as reduce
CPU utilization over BOT mass storage drivers.

Signed-off-by: James Christopher Adduono <jc@adduono.com>

### DIFF
--- a/package/kernel/linux/modules/usb.mk
+++ b/package/kernel/linux/modules/usb.mk
@@ -895,6 +895,25 @@ endef
 $(eval $(call KernelPackage,usb-storage-extras))
 
 
+define KernelPackage/usb-storage-uas
+  SUBMENU:=$(USB_MENU)
+  TITLE:=USB Attached SCSI (UASP) support
+  DEPENDS:=+kmod-usb-storage
+  KCONFIG:=CONFIG_USB_UAS
+  FILES:=$(LINUX_DIR)/drivers/usb/storage/uas.ko
+  AUTOLOAD:=$(call AutoProbe,uas)
+endef
+
+define KernelPackage/usb-storage-uas/description
+ Say Y here if you want to include support for
+ USB Attached SCSI (UAS/UASP), a higher
+ performance protocol available on many
+ newer USB 3.0 storage devices
+endef
+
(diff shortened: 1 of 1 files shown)


jcadduono:
using on my personal builds, i guess sometime during kernel upgrades this module went unnoticed

diizzyy:
Good work, unfortunately it's only a few controllers such as ASMedia that will benefit :/

borkra:
@diizzyy That is not quiet correct. The limited list of controllers you are listing are for the device side. On the host side any USB 3 controller is UAS capable. And as routers generally implement host side, so it is beneficial for any router. The only thing required is that newer/supporting SSD or SD card reader used.

hauke:

diizzyy:
@borkra 
That doesn't seem correct, http://linux-sunxi.org/USB/UAS
I've also seen several software solutions on Windows trying to address the above.

borkra:
Well, I do know I've been using UASP on my portable SSD with all my Intel Linux boxes for the last year and a half. And it does work. So this article is probably outdated/incomplete.

UASP in-general is the software thing, it does not affect USB hardware layer, just another device profile.

Even the article you reference mentions as much on the first line:

"It was introduced as an optional part of the USB 3.0 specs but can also be used with USB 2.0 host controllers when hardware/drivers are compatible"

On the device side though things are more complicated as to reduce cost USB higher level functions are jammed into ASIC, so it supports what it supports.

diizzyy:
@borkra 
Well, the majority of hw used isn't Intel so I don't know how that would be valid as reference, but anyway... let's try it on various ARM SoCs etc. :-)

hauke:
To my understanding this only needs a software implementation on the USB host side like most (all?) other USB profiles, otherwise it would be strange why someone would make a USB 2.0 hardware supporting this instead of a USB 3.0 hardware. USB 3.0 sells better than UAS. It could be that other operating systems like Windows need this host software implementation in the driver as long as the operating system does not support this natively. Anyway this is an optional feature as long as you do not install it, it will not harm you.

PR #1366, opened 2017-09-20
kernel: Remove Joliet CDROM extensions

diizzyy:
Remove Joliet CDROM extensions from the kernel


diizzyy:
compile tested on ar71xx

hauke:
NACK: This config option depends on CONFIG_ISO9660_FS which is only activated on the x86 and uml target and with the kmod-fs-isofs package.
Is alos only changes the isofs file system so it does only affect the kmod-fs-isofs package or the targets named above.
This will probably also cause compile problems on the x86 and uml target.

diizzyy:
Ahh, ok. Thanks

PR #1370, opened 2017-09-23
octeon: support the UniFi Security Gateway 3P

MartB:
Added support for the Ubiquiti Unifi Security Gateway

The default_packages edit is due to a failure to create the overlay filesystem because of a missing mkfs.f2fs binary.

The devicetree related edits are mandatory as theres no other established way (known to me) to load the proper devicetree on bootup on these devices and the USG3P has 2 leds (blue, white) i wanted to support.

If theres another way to do the device tree related changes and achieve the same result please tell me!

Compile and run tested on an USG3P.
Would be great if someone could test this on an edge router lite.

### DIFF
--- a/target/linux/octeon/patches-4.9/103-ubnt_usg3p_support.patch
+++ b/target/linux/octeon/patches-4.9/103-ubnt_usg3p_support.patch
@@ -1,0 +1,73 @@
+--- a/arch/mips/include/asm/octeon/cvmx-bootinfo.h
++++ b/arch/mips/include/asm/octeon/cvmx-bootinfo.h
+@@ -295,6 +295,7 @@ enum cvmx_board_types_enum {
+	 */
+	CVMX_BOARD_TYPE_CUST_PRIVATE_MIN = 20001,
+	CVMX_BOARD_TYPE_UBNT_E100 = 20002,
++	CVMX_BOARD_TYPE_UBNT_E120 = 20004,
+	CVMX_BOARD_TYPE_UBNT_E200 = 20003,
+	CVMX_BOARD_TYPE_UBNT_E220 = 20005,
+	CVMX_BOARD_TYPE_CUST_DSR1000N = 20006,
+@@ -398,6 +399,7 @@ static inline const char *cvmx_board_type_to_string(enum
+		    /* Customer private range */
+		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_CUST_PRIVATE_MIN)
+		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_UBNT_E100)
++		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_UBNT_E120)
+		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_UBNT_E200)
+		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_UBNT_E220)
+		ENUM_BRD_TYPE_CASE(CVMX_BOARD_TYPE_CUST_DSR1000N)
+--- a/arch/mips/cavium-octeon/executive/cvmx-helper-board.c
++++ b/arch/mips/cavium-octeon/executive/cvmx-helper-board.c
+@@ -169,6 +169,7 @@ int cvmx_helper_board_get_mii_address(int ipd_port)
+		else
+			return -1;
+	case CVMX_BOARD_TYPE_UBNT_E100:
++	case CVMX_BOARD_TYPE_UBNT_E120:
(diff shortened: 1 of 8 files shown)


target/linux/octeon/Makefile
```diff
@@ -21,6 +21,6 @@ endef
 
 include $(INCLUDE_DIR)/target.mk
 
-DEFAULT_PACKAGES += mkf2fs e2fsprogs
+DEFAULT_PACKAGES += mkf2fs e2fsprogs f2fs-tools
```
stintel: This is unrelated and should go in a different commit, if it should go in at all.

MartB: Tbh im not sure if it should go in but its the only way i got this working because the overlay filesystem kept failing.

stintel: That's strange. I'm using the squashfs + f2fs image on my ERLite and it works fine. I assume you are testing on master, as the PR is against master? Can you give more info on how it fails?

MartB: When i leave that package out it says "overlay filesystem has not been fully initialized yet." and then it switches to a temporary one that does hold the data until next reboot. I traced it down to the missing mkfs binary for f2fs. As soon as it was available it worked.

stintel: But f2fs-tools doesn't contain the mkfs utility:
```
root@ar1:~# opkg files f2fs-tools
Package f2fs-tools (1.8.0-3) is installed on root and has the following files:
/usr/sbin/f2fstat
/usr/sbin/parse.f2fs
/usr/sbin/fibmap.f2fs
```

stintel: Maybe you just didn't have mkf2fs enabled in your config initially? Having something in DEFAULT_PACKAGES does not guarantee it will be enabled in the .config. I suggest to disable f2fs-tools in menuconfig again and try a new build. After seeing what files are in there, it definitely does not belong in DEFAULT_PACKAGES.

I'll gladly test on my ERL when you update the PR.

MartB: I will do everything you mentioned above and update this pr.
I will also check if the mkfs.f2fs gets compiled and is available to the system.
I verified it got enabled in my configs.

MartB:
Do not review at the moment im using this git repo on my build system and didnt branch out sorry!
I will remove the WIP as soon as it compiles.

MartB:
I split the patches on 60bc430 into two separate files to get the smallest diff possible that does not contain the whole source from the linux kernel i hope its okay like this.

stintel:
Not yet. I would like to see the patches exactly as how they were sent upstream by the original author, including the headers. E.g. https://github.com/lede-project/source/blob/master/target/linux/generic/backport-4.9/010-Kbuild-don-t-hardcode-path-to-awk-in-scripts-ld-vers.patch

target/linux/octeon/patches-4.9/103-ubnt_usg3p_support.patch
```diff
@@ -0,0 +1,73 @@
...
++extern const char __dtb_ubnt_e100_begin;
++extern const char __dtb_ubnt_e120_begin;
+ void __init device_tree_init(void)
+ {
+	const void *fdt;
+@@ -1206,6 +1208,14 @@ void __init device_tree_init(void)
+		fdt = &__dtb_octeon_68xx_begin;
+		do_prune = true;
+		fill_mac = true;
++	} else if (octeon_bootinfo->board_type == CVMX_BOARD_TYPE_UBNT_E100) {
```
stintel: And this should go in the first commit, as that commit already changes things for E100. The commit that adds support for USG3P should only contains changes related to E120.

MartB: that slipped through will get fixed.

MartB:
Im not sure i understand the part about the patches correctly. 
None of these patches have any additional info i wrote them myself.
I thought just including the patches like the other guys did here https://github.com/lede-project/source/blob/master/target/linux/octeon/patches-4.9/100-ubnt_edgerouter2_support.patch is fine ?

stintel:
Is this you: Written by: Aaro Koskinen <aaro.koskinen@iki.fi>?
Also, I missed it before, but your SoB line is invalid and should contain your full real name and must match the commit author. See https://lede-project.org/submitting-patches.

MartB:
Uhm no but thats a new file with a header from the previous file i overwrote.
That needs to be deleted and replaced with the proper header i guess?

stintel:
Yes, and you should send this upstream for review, following the guidelines for kernel contributions, and then take that patch and drop it in patches-4.9/.

blogic:

PR #1371, opened 2017-09-24
kernel: Backport Quectel EC25 support from 4.9

wildy:
Backport changes from v4.9 to make Quectel EC25 LTE modem work.

072-v4.9-0001-usbnet-raw-l2-headers.patch - Backport from 4.9, needed
for this modem to work in Raw IP mode.

Note that you'll have to do 'echo Y > /sys/class/net/<iface>/qmi/raw_ip'
to make raw IP mode work.

072-v4.9-0002-qmi_wwan-backport.patch - Backports the entire qmi_wwan
driver from upstream 4.9

Signed-off-by: Yuri Gorshkov <me@koshaq.net>

### DIFF
--- a/target/linux/generic/pending-4.4/072-v4.9-0001-usbnet-raw-l2-headers.patch
+++ b/target/linux/generic/pending-4.4/072-v4.9-0001-usbnet-raw-l2-headers.patch
@@ -1,0 +1,31 @@
+commit 81e0ce79f2919dbd5f025894d29aa806af8695c7
+Author: Bjørn Mork <bjorn@mork.no>
+Date:   Thu Dec 3 19:24:20 2015 +0100
+
+    usbnet: allow mini-drivers to consume L2 headers
+    
+    Assume the minidriver has taken care of all L2 header parsing
+    if it sets skb->protocol.  This allows the minidriver to
+    support non-ethernet L2 headers, and even operate without
+    any L2 header at all.
+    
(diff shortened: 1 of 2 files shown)


oaaeruak:
This is probably not needed because Lede would jump on all devices to kernel 4.9 before releasing the next stable release in few days. For some chipsets it already at kernel 4.9.

mkresin:
> This is probably not needed because Lede would jump on all devices to kernel 4.9 before releasing the next stable release in few days.

That information is wrong. We might release a new 17.01.x. The kernel of the 17.01. branch will not be upgraded to kernel 4.9.

There is no upcoming major release (with kernel 4.9) planed at the moment.

mkresin:
@wildy On which target do you intend to use the Quectel EC25? A lot of targets are already updated to kernel 4.9 in master.

wildy:
This device is a GL-MiFi, which is ar71xx-based. When I last checked, porting ar71xx to 4.9 was still a huge work in progress, is this still the case?

blogic:
the patches need to go into the backport-4.9 folder. please change the PR in that regard.

wildy:
I thought backport-4.9 was for changes backported into the 4.9 branch? (I need to backport 4.9 changes into 4.4 for the time being until ar71xx gets migrated to 4.9)

blogic:
@wildy sorry my bad, misread the patch

nbd168:
There's no need to apply this. ar71xx will switch to 4.9 by default very soon.

PR #1372, opened 2017-09-25
kernel: kmod-fs-cifs-smb2 and kmod-fs-cifs-smb311

jcadduono:
This will build empty packages while adding the
newer protocol support to the kmod-fs-cifs kernel
module (cifs.ko) package instead.

Note that samba36-server can only support up to
SMB2. From my testing, in order for SMB2 support
to be present, smb.conf template must be modified
to include the following options:

[global]
  min protocol = NT1
  max protocol = SMB2

Signed-off-by: James Christopher Adduono <jc@adduono.com>

I wasn't sure if empty kmod packages were the way to go for this, it seems the system is smart enough to not bother building a package with no files so this worked well enough for me.
Samba 4 support is probably quite a ways away, eh? Perhaps best to only have SMB2 option and remove SMB311 from patch. Thoughts?

### DIFF
--- a/package/kernel/linux/modules/fs.mk
+++ b/package/kernel/linux/modules/fs.mk
@@ -125,6 +125,40 @@ endef
 
 $(eval $(call KernelPackage,fs-cifs))
 
+define KernelPackage/fs-cifs-smb2
+  SUBMENU:=$(FS_MENU)
+  TITLE:=SMB2 and SMB3 support
+  KCONFIG:= \
+	CONFIG_CIFS_SMB2=y \
+	CONFIG_CIFS_SMB311=n \
+	CONFIG_KEYS=y
+  DEPENDS+= \
+    +kmod-fs-cifs \
+    +kmod-fs-fscache \
+    +kmod-dnsresolver
+endef
+
+define KernelPackage/fs-cifs-smb2/description
+ Say Y here if you want to include support for
+ the SMB2 and SMB3 protocols in the CIFS module
+endef
+
+$(eval $(call KernelPackage,fs-cifs-smb2))
+
+define KernelPackage/fs-cifs-smb311
+  SUBMENU:=$(FS_MENU)
+  TITLE:=SMB3.1.1 support (experimental)
+  KCONFIG:=CONFIG_CIFS_SMB311=y
+  DEPENDS+=+kmod-fs-cifs-smb2
+endef
+
+define KernelPackage/fs-cifs-smb311/description
+ Say Y here if you want to include support for
+ the SMB3.1.1 protocol in the CIFS module
+endef
+
+$(eval $(call KernelPackage,fs-cifs-smb311))
+
 
 define KernelPackage/fs-configfs
   SUBMENU:=$(FS_MENU)


neheb:
Why not just edit the CIFS module?

jcadduono:
@neheb can you do that from config?

dissent1:
I’ve heard that nt1 protocol performs faster than smb2 and if I remember correctly the time when I’ve tested - It’s true. So is it really needed to switch to smb2 on router devices that are usually not powered enough?
Does your benchmark differs when using those protocols?

hauke:
We can not add SMB2 as an extra module, I would prefer if the kmod-fs-cifs module would always include SMB2 and SMB3 support as long as this has no negative influence to the normal kernel size.
I do not like adding kernel packages which are changing other kernel packages.

The KConfig help says that SMB2 is faster than SMA3 and SMB1 is also insecure.

jcadduono:
I don't think it would be a good idea to include SMB2 support in the base CIFS package due to the drastic size differences
SMB311 support seems like a very small difference on top of SMB2/3.
Perhaps they can be 2 mutually exclusive packages?

kmod-fs-cifs depends on ! kmod-fs-smb2 includes cifs.ko
kmod-fs-smb2 depends on ! kmod-fs-cifs includes cifs.ko

```
SMB1/CIFS only:
246732	build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/linux-ipq806x/packages/ipkg-arm_cortex-a15_neon-vfpv4/kmod-fs-cifs/lib/modules/4.9.51/cifs.ko

with SMB2/SMB3:
387136	build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/linux-ipq806x/packages/ipkg-arm_cortex-a15_neon-vfpv4/kmod-fs-cifs/lib/modules/4.9.51/cifs.ko

with SMB311:
388568	build_dir/target-arm_cortex-a15+neon-vfpv4_musl_eabi/linux-ipq806x/packages/ipkg-arm_cortex-a15_neon-vfpv4/kmod-fs-cifs/lib/modules/4.9.51/cifs.ko
```

neheb:
~150kB is not that bad for a non default module.

jcadduono:
Benchmarks appear to not be helpful due to randomness of location on disk, maybe?

```
/media/disk2: (325GB/3.7TB) WD Red 4TB via ASM1351 w/ USB3.0 + UAS kernel driver connected to R7800 IPQ8065 @ 1Gbit

iozone -c -e -s 128m -r 4m -o -f /media/disk2/test
                                                               random   random     bkwd    record    stride                                    
              kB  reclen    write  rewrite    read    reread     read    write     read   rewrite      read   fwrite frewrite    fread  freread

NT1:      131072    4096    41003    42334    23036    22183    22287    47403    32742     55245     24277    45245    56076    44063    44012
          131072    4096    40390    49648    15892    18214    20017    52506    29771     36349     23548    61676    51537    40352    42069
          131072    4096    42285    49609    36891    35518    32048    53706    33270     55796     23231    65557    63539    41807    41531
          131072    4096    41283    52896    19885    22697    39488    55874    33546     54175     33072    84817    81470    42432    39637
          131072    4096    42582    21471    22629    22282    25771    41017    20813     71496     23255    54155    72928    41385    40543

SMB2:     131072    4096    45056    44114    23000    38240    24693    39619    30890     39951     22165    74265    66492    35278    35137
          131072    4096    35442    45993    22500    35279    34760    47487    24073     43333     30391    63360    74832    35416    35838
          131072    4096    36012    40156    23308    22203    22409    38013    22583     41899     24602    55940    37104    36049    38216
          131072    4096    40124    47006    29785    37648    36245    44396    28171     49896     23910    58699    64429    36462    36796
          131072    4096    40211    38675    27729    28321    23716    42821    24397     39987     26071    66958    66077    35759    35091

iozone -c -e -s 4m -r 4k -o -f /media/disk2/test
                                                               random   random     bkwd    record    stride                                    
              kB  reclen    write  rewrite    read    reread     read    write     read   rewrite      read   fwrite frewrite    fread  freread

NT1:        4096       4      617      603    17602    18616     1107      548      793       619      3771     3078    15422     1158     1200
            4096       4      603      555    14804    15227     1182      559      827       554      3963     8575    20562     1125     1088
            4096       4      698      545    13781    13459     1101      590      942       652      3439     3599    23752     1219     1149
            4096       4      735      565    16893    19580     1165      564      828       575      3771     3923    20182     1469     1673
            4096       4      683      534    14252    19571     1160      500      796       545      3544     1934     4873     1104     1228

SMB2:       4096       4      489      586    12891    19840     1037      453      697       455      3289     9726    23377      813      891
            4096       4      676      437    14993    20127     1048      452      696       458      3370     5654    19120      865     1380
            4096       4      453      465    15459    15342     1092      494      696       479      3209     2112    19647      827      874
            4096       4      455      476    19572    19006     1108      461      703       454      3387     2924    23452      868      836
            4096       4      486      487    18366    22371     1636      460      717       462      3429     6030    16891      861      844
```

jcadduono:
as of at least Linux 4.13, SMB2 and SMB3 have been merged into core CIFS support and no longer a separate flag, so it looks like simply adding it in to kmod-fs-cifs will be the way to go

neheb:
+1

hauke:
It think we should activate the CONFIG_CIFS_SMB2=y option for the normal kmod-fs-cifs module.

neheb:
I just had a thought. Can we remove SMB1 support instead? or is it not possible?

jcadduono:
@neheb only possible if you patch it out in kernel sources (preventing mount pre2.0, remove unused code) since smb2 support depends on it
you can modify the shipped smb conf template to set min to smb2 but I think it would be better to keep smb1 support for legacy devices on the network such as smart TVs and NAS that don't support smb2 protocol

hauke:
I do not like this patch because it will not work with binary only images like release images.
When the SMB2 or SMB311 option is activated the normal SMB package will change.
I wuld prefer if we activate SMB2 support always for the kmod-smb package.

PR #1380, opened 2017-09-29
prereq-build.mk: add m4 when setup host command

hbl0307106015:
Since some packages need m4 tool during the building procedure, I add m4 when the sdk setup host command. For more detail see below snapshot.

deepinscreenshot_20170927121522


nbd168:
This does not make any sense. `m4` is already built as part of `tools/`, so there should be no need to add it to `prereq-build.mk`

PR #1381, opened 2017-09-29
net: uqmi: fix blocking in endless loops when unplugging device

commodo:
If you unplug a QMI device, the /dev/cdc-wdmX device
disappears but uqmi will continue to poll it endlessly.

Then, when you plug it back, you have 2 uqmi processes,
and that's bad, because 2 processes talking QMI to the
same device [and the same time] doesn't seem to work well.

### DIFF
--- a/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
+++ b/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
@@ -65,6 +65,7 @@ proto_qmi_setup() {
 	[ -n "$delay" ] && sleep "$delay"
 
 	while uqmi -s -d "$device" --get-pin-status | grep '"UIM uninitialized"' > /dev/null; do
+		[ -e "$device" ] || return 1
 		sleep 1;
 	done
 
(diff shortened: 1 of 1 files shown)


lynxis:
Hi @commodo 
can you revert reducing the sleep value? I don't see a problem with polling every 5 seconds for a new network state. Even 5 seconds is quite fast. The rest looks good.

package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh
```diff
@@ -102,7 +103,8 @@ proto_qmi_setup() {
 
 	echo "Waiting for network registration"
 	while uqmi -s -d "$device" --get-serving-system | grep '"searching"' > /dev/null; do
-		sleep 5;
+		[ -e "$device" ] || return 1
+		sleep 1;
```
lynxis: please keep `sleep 5`

commodo: ack ; no strong preference from me

commodo:
@lynxis 
re-spinned

lynxis:
@commodo thanks. merged into my staging treegg

PR #1383, opened 2017-09-29
sunxi: improve A20 Lime2 upload speed

lucize:
is not a perfect fix but it helps

Tested on r4950-e64463ebde


hauke:
Thank you for your patch, it was applied to the LEDE master branch.

PR #1388, opened 2017-09-30
kernel: allow unpatched builds as workaround

pprindeville:
Sometimes you want to do a quick build with no patches to see if any of the LEDE-specific patches are causing a regression.  Having an easy way to do this is handy.

Ideally you'd be able to do this with the tree as-is, but there are some incidental problems that interfere with it:

https://bugs.lede-project.org/index.php?do=details&task_id=503

This offers a trivial workaround, and this PR can be reverted when FS#503 gets fixed.

### DIFF
--- a/config/Config-devel.in
+++ b/config/Config-devel.in
@@ -118,3 +118,12 @@ menuconfig DEVEL
 		default "-fno-caller-saves"
 		help
 		  Extra target-independent optimizations to use when building for the target.
+
+	config UNPATCHED_KERNEL
+		bool
+		default n
+		help
+		  Disable kernel patches if you want to generate a baseline image
+		  or suspect a patch might be the source of a regression.  This
+		  may cause some functionality to fail.
+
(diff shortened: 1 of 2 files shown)


blogic:
@pprindeville 
1) there seems to be no prompt
2) this might work for a very select few targets but the moment you enable certain wifi chips or need swconfig it wont work.

pprindeville:
@blogic:

Not use what you mean by (1).

As for (2), that's acceptable.  It's for testing bare kernel functionality.  Sometimes it's useful to see the delta between what's upstream and what all of our local patches do.

nbd168:
I don't think this should be merged, since enabling it will break almost any configuration. If you want to test without patches, you can simply remove the patch directories in your source tree instead.

blogic:
@nbd168 agreed. removing the patches dir in a local build is a valid work around.
@pprindeville sorry cant merge this one

PR #1330, opened 2017-09-06
Revert "utils/util-linux: Update to 2.30.1"

rmounce:
This reverts commit e505f59bd946a85c7eb8d5b19d796ed4d45218df.

Fixes inconsistent sysupgrade on targets with f2fs overlayfs on a block device. Reported on x86_64 and mvebu/Turris Omnia (like ClearFog).

PR #1389, opened 2017-09-30
util-linux: update to 2.30.2

rmounce:
Compile/run tested on mvebu.

Not affected by issue reported in https://github.com/lede-project/source/pull/1330 and fixed by e64463e.


hauke:
Thank you for your patch, it was applied to the LEDE master branch.

PR #1392, opened 2017-10-01
Fix 5Ghz wifi for Sitecom WLR-8100

DavideFioravanti:
Fix the 5Ghz wireless by adding the right calibration data.

Signed-off-by: Davide Fioravanti <pantanastyle@gmail.com>


blogic:
merged, Thanks ! ... after fixing up the subject prefix

DavideFioravanti:
I'm sorry, this was my first commit. Next time I will be more careful. :-)

PR #1394, opened 2017-10-01
ar71xx: add support for Mikrotik RB750P-PBr2

robimarko:
This PR adds support for the MikroTik RouterBOARD Powerbox
https://mikrotik.com/product/RB750P-PBr2

Specifications:
- SoC: Qualcomm QCA9531 (650MHz)
- RAM: 64MB
- Storage: 16MB NOR SPI flash
- Ethernet: 5x100M (1 PoE in, 4 PoE out)
- Outdoor use ready

This ethernet router is based on the same platform as the hEX PoE lite.

Full installation instructions are included in the commit message.

Signed-off-by: Robert Marko <robimarko@gmail.com>

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
@@ -7,6 +7,7 @@
  *  - MikroTik RouterBOARD 952Ui-5ac2nD
  *  - MikroTik RouterBOARD 962UiGS-5HacT2HnT
  *  - MikroTik RouterBOARD 750UP r2
+ *  - MikroTik RouterBOARD 750P-PBr2
  *  - MikroTik RouterBOARD 750 r2
  *  - MikroTik RouterBOARD LHG 5nD
  *
@@ -763,7 +764,8 @@ static void __init rb952_setup(void)
  * Init the hEX (PoE) lite hardware (QCA953x).
  * The 750UP r2 (hEX PoE lite) is nearly identical to the hAP, only without
  * WLAN. The 750 r2 (hEX lite) is nearly identical to the 750UP r2, only
- * without USB and POE. It shares the same bootloader board identifier.
+ * without USB and POE. The 750P Pbr2 (Powerbox) is nearly identical to hEX PoE
+ * lite, only without USB. It shares the same bootloader board identifier.
  */
 static void __init rb750upr2_setup(void)
 {
@@ -775,6 +777,10 @@ static void __init rb750upr2_setup(void)
 	/* differentiate the hEX lite from the hEX PoE lite */
 	if (strstr(mips_get_machine_name(), "750UP r2"))
 		flags |= RBSPI_HAS_USB | RBSPI_HAS_POE;
+	
+	/* differentiate the Powerbox from the hEX lite */
+	else if (strstr(mips_get_machine_name(), "750P r2"))
+		flags |= RBSPI_HAS_POE;
 
 	rbspi_952_750r2_setup(flags);
 }
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -267,6 +267,7 @@ rb-750)
 	ucidef_set_led_switch "port5" "port5" "rb750:green:port5" "switch0" "0x02"
 	;;
 rb-750-r2|\
+rb-750p-pbr2|\
 rb-750up-r2)
 	ucidef_set_led_timer "user" "USER" "rb:green:user" "1000" "1000"
 	ucidef_set_led_netdev "port1" "port1" "rb:green:port1" "eth0"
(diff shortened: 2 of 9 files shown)


target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
```diff
@@ -775,6 +777,10 @@ static void __init rb750upr2_setup(void)
 	/* differentiate the hEX lite from the hEX PoE lite */
 	if (strstr(mips_get_machine_name(), "750UP r2"))
 		flags |= RBSPI_HAS_USB | RBSPI_HAS_POE;
+	
+	/* differentiate the Powerbox from the hEX lite */
+	if (strstr(mips_get_machine_name(), "750P r2"))
```
f00b4r0: Just a nitpick but this should be an `else if ()`: both tests are mutually exclusive.

robimarko: Thanks for the catch.
Corrected it

f00b4r0:
LGTM

mkresin:
If you add support for new hardware: Include in your commit message short description of the hardware and **how to install LEDE on it**. Have a look at the [recent additions](https://git.lede-project.org/?p=source.git&a=search&h=HEAD&st=commit&s=add+support+for) for some examples.

mkresin:
>  I made a wiki page for it on OpenWRT wiki with full instructions, can I just link that?

Nope, please add the instructions to the commit. External pages tend to go offline, change or move. By adding the instructions to the commit, you have them right along the required code without being depend on any external site.

> The method is the same as other RB750 series and is pretty long.

To my knowledge there is no character limit for a git commit message. Feel free to be chatty.

robimarko:
Ok,will include all relevant instructions in commit message

f00b4r0:
Looks like your methods are windows-centric. It's also not listing the relatively easy reset button method listed here: https://wiki.openwrt.org/toh/mikrotik/common

Note that this webpage incorrectly says it cannot be done when routeros is still installed: dnsmasq supports bootp.

I think you shouldn't need to explain how to install/run a dhcp/bootp/tftp server and simply mention that the installation is done via tftp-booted initramfs, and only detail the methods to trigger than network boot on the device.

robimarko:
That's the reason why I did not include instructions at first.
They are the same for almost all Mikrotik devices and there are multiple ways.

For me reset button method did not work and I did not mention it for that reason while TFTPD method did work without issues.

f00b4r0:
The reset button always works (it's the boot loader recovery) but you have to keep the button depressed for a long time and keep an eye on the LED pattern.

robimarko:
Yes, you have to wait until after flashing for second time ACT LED goes out.

robimarko:
@mkresin What do you think about instructions in commit?
Should I include only device specific stuff or full like now

mkresin:
@robimarko Yeah, my bad, I told you to be as chatty as necessary. But I agree with @f00b4r0 that the install instructions shouldn't be client OS specific.

I would propose to use the following commit message instead:

```
ar71xx: add support for Mikrotik RB750P-PBr2

Specifications:
- SoC: Qualcomm QCA9531 (650MHz)
- RAM: 64MB
- Storage: 16MB NOR SPI flash
- Ethernet: 5x100M (1 PoE in, 4 PoE out)
- Outdoor use ready

This ethernet router is based on the same platform as the hEX PoE lite.

Installation

1. login to the Mikrotik WebUI to backup your licence keys
2. setup a DHCP/BOOTP Server with:
     * DHCP-Option 66 (TFTP server name) pointing to a local TFTP
       Server within the same subnet of the DHCP range
     * DHCP-Option 67 (Bootfile-Name) matching the initramfs filename
       of the to be booted image
3. connect the port labled internet to your local network
4. keep the reset button pushed down and power on the board

The board should load and start the initramfs image from the TFTP
Server. Login as root/without password to the started LEDE via ssh
listing on IPv4 address 192.168.1.1. Use sysupgrade to install LEDE.

Revert to RouterOS

Use the "rbcfg" package on in LEDE:
  * rbcfg set boot_protocol bootp
  * rbcfg set boot_device ethnand
  * rbcfg apply

Open Netinstall and reboot routerboard. Now netinstall sees routerboard
and you can install RouterOS. If NetInstall gets stuck on Sending offer
just wait for it to timeout and then close and open Netinstall again.

Click on install again.

In order for RouterOS to function properly, you need to restore license
for the device. You can do that by including license in NetInstall

Signed-off-by: Robert Marko <robimarko@gmail.com>
```

You can either ACK this commit message (I''ll change the commit message during merge) or update the PR with your own commit message (in case my proposed commit message/install instruction is wrong/incomplete).

robimarko:
@mkresin I ACK that message.

mkresin:
Merged to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Will push it to master during the next days.

PR #1395, opened 2017-10-01
sunxi: backport support for Allwinner Security System PRNG

lucize:

tested on A20, needs to be tested on A10


target/linux/sunxi/patches-4.9/0054-crypto-sun4i-ss_support_the_Security_System_PRNG.patch
```diff
@@ -0,0 +1,198 @@
+diff --git a/drivers/crypto/Kconfig b/drivers/crypto/Kconfig
```
hauke: As this is only the one commit, could you also add the header to it:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/patch/?id=b8ae5c7387ad075ee61e8c8774ce2bca46bc9236

lucize:
added

hauke:

PR #1397, opened 2017-10-02
hostapd: enable fast math library

BigNerd95:
Enable internal math fast library for hostapd-full
like wpa_supplicant-full

Signed-off-by: Lorenzo Santina <lorenzo.santina@edu.unito.it>

### DIFF
--- a/package/network/services/hostapd/files/hostapd-full.config
+++ b/package/network/services/hostapd/files/hostapd-full.config
@@ -297,7 +297,7 @@ CONFIG_INTERNAL_LIBTOMMATH=y
 # At the cost of about 4 kB of additional binary size, the internal LibTomMath
 # can be configured to include faster routines for exptmod, sqr, and div to
 # speed up DH and RSA calculation considerably
-#CONFIG_INTERNAL_LIBTOMMATH_FAST=y
+CONFIG_INTERNAL_LIBTOMMATH_FAST=y
 
 # Interworking (IEEE 802.11u)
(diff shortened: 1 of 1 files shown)


diizzyy:
I think it would be nice if we could link it (libtommath) dynamically for both dropbear and hostapd in that case.

BigNerd95:
Good idea.
Dropbear is using internal libtommat too:
https://github.com/lede-project/source/blob/master/package/network/services/dropbear/Makefile#L82
Original reference: https://github.com/mkj/dropbear/blob/master/configure.ac#L478

diizzyy:
@BigNerd95 @ldir-EDB0 
This might be a good starting point?
https://github.com/ldir-EDB0/lede-source/commit/afbb5885dc3945c2057ecd92f874556b5dcc0801

BigNerd95:
How much this package increase the firmware size?

BigNerd95:
Did you test if hostapd/dropbear is faster with external libtom than internal libtom?

diizzyy:
@BigNerd95 
No idea, you have to evaluate it yourself.

hauke:
can you please explain which operations will be faster and provide some numbers if possible. 
It would also be nice to have some numbers on the size increase.

BigNerd95:
`At the cost of about 4 kB of additional binary size`
`faster routines for exptmod, sqr, and div to speed up DH and RSA calculation considerably`

blogic:

PR #1399, opened 2017-10-04
build: refactor TP-Link V2/3 image creation and firmware utils

f00b4r0:
This patch series aims at improving the way TP-Link V2/3 firmware images are handled in the LEDE build system:

- Board-specific magic values are moved back to the makefile's device definition 
- Hardcoded values are removed from mktplinkfw2 source code, and the matching system is also removed (it was broken anyway)
- A large chunk of duplicate code shared between mktplinkfw.c and mktplinkfw2.c is aggregated in a single file (mktplinkfw-lib.c)

Note: no attempt to clean, fix, improve or rewrite mktplinkfw* code has been made, it's simply a matter of collating duplicate code in a single place


target/linux/ramips/image/mt7620.mk
```diff
@@ -69,7 +78,9 @@ define Device/ArcherMR200
   $(Device/Archer)
   DTS := ArcherMR200
   SUPPORTED_DEVICES := mr200
-  TPLINK_BOARD_ID := ArcherMR200
+  TPLINK_FLASHLAYOUT := 8Mmtk
```
heinzek: Here is typo. MR200 use 8MLmtk flash layout.

f00b4r0: Good catch, fixed thanks!

mkresin:
merged, thanks!

PR #1401, opened 2017-10-06
ramips: revised reset button gpios for Omnima Mini Linux boards

omnima:
Revised Reset button GPIO, also fixed linux,code = <KEY_RESTART> line.
Removed wps, mode-one and mode-two gpio definitions so that these can be used as general purpose i/o lines.

### DIFF
--- a/target/linux/ramips/dts/MINIEMBPLUG.dts
+++ b/target/linux/ramips/dts/MINIEMBPLUG.dts
@@ -19,7 +19,7 @@
 
 		mobile {
 			label = "miniembplug:green:mobile";
-			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio0 10 GPIO_ACTIVE_LOW>;
 		};
 	};
 
@@ -29,28 +29,10 @@
 		#size-cells = <0>;
 		poll-interval = <20>;
 
-		wps {
-			label = "wps";
-			gpios = <&gpio0 0 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
 		reset {
 			label = "reset";
-			gpios = <&gpio0 10 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
-		mode-one {
-			label = "mode1";
-			gpios = <&gpio0 11 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
-		mode-two {
-			label = "mode2";
-			gpios = <&gpio0 12 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+            linux,code = <KEY_RESTART>;
 		};
 	};
 };
--- a/target/linux/ramips/dts/MINIEMBPLUG.dts
+++ b/target/linux/ramips/dts/MINIEMBPLUG.dts
@@ -32,7 +32,7 @@
 		reset {
 			label = "reset";
 			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
-            linux,code = <KEY_RESTART>;
+			linux,code = <KEY_RESTART>;
 		};
 	};
 };


mkresin:
If the PCB design/GPIOs has been changed for the 2017/2018 batch of the MiniEMBWiFi, you should add a rev B image instead of breaking pre 2017 MiniEMBWiFi boards.

Am I right that the wps and mode-* buttons were never physically present?

Please use your full name in the Signed-off-by and add a prefix to the commit title as described in [submitting patches](https://lede-project.org/submitting-patches).

omnima:
Thanks for reviewing this PR.

We often do custom Linux board design and there are a number of revisions, where gpios are used for different purposes but the overall design is the same. 

In order to avoid having to add all the revisions separately to LEDE/OpenWrt, over the years we've included only one board revision per CPU type. For example, MINIEMBWIFI was based on RT3050, HPM on RT3662, MINIEMBPLUG on RT5350, also planned for inclusion into LEDE is a board support for MT7620, one MT7688 and one for AR9331. 

Adding support for multiple revisions is likely to increase the time needed for LEDE devs to maintain all the variants when including new or revised features for an entire platform so we tried to keep to just one revision in LEDE/OpenWrt so far.

This allows a developer to get the base firmware built and running on our boards quickly. Then local changes can be made easily to the dts file to configure the pins as required for each different board revision according to our docs.

I've updated the PR title, hope it's ok now. We use a company wide github account, but for future PR's can create inidividual logins.

If it is preferable to add multiple board revisions into the source tree we can follow this approach.

mkresin:
> Adding support for multiple revisions is likely to increase the time needed for LEDE devs to maintain all the variants when including new or revised features for an entire platform so we tried to keep to just one revision in LEDE/OpenWrt so far.

Yeah makes perfect sense to me.

I'm still a bit concerned about changing the reset gpio for existing boards. But as far as I understand, Omnima does not build off the shelf products, no one runs a plain LEDE on your boards and it is rather that LEDE is used as a framework that needs to be customised anyway.

In the end it is up to you to decide what is the best for your boards/customers.

> We use a company wide github account, but for future PR's can create inidividual logins.

It is fine to use a single GItHub account. Just make sure to fix the author setting for your local git client, so that a realname is used in From line + SoB and both are identical. Have a look at the PR [as patch file](https://github.com/lede-project/source/pull/1401.patch) to see what I'm talking about.

I've merged your PR to [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary), squashed the commits into a single one, fixed the author/form and slightly rewrote the commit message.

Please ACK the changes or update this PR with the version you would like to see committed.

mkresin:
> Please ACK the changes or update this PR with the version you would like to see committed.

So, 12 days later I still don't have a ACK/NAK for my changes. Please do what I asked for within the next seven days, or I will drop the PR from my staging tree.

mkresin:
Unstaged the following commit from my staging tree:

```diff
From 193de218f5ec5319402b303485ce5f6679ca4da6 Mon Sep 17 00:00:00 2001
From: Michael Lewis <info@omnima.co.uk>
Date: Thu, 5 Oct 2017 21:11:21 +0100
Subject: [PATCH] ramips: fix Omnima MiniEMBPlug buttons

The new Omnima 2017/18 boards have changed the gpio connected to the
reset button. While at it, fix the linux,keycode as well.

Also remove the gpio allocations for the switches and wps button so
that they can be used as general purpose i/o lines when required.

Signed-off-by: Michael Lewis <info@omnima.co.uk>
---
 target/linux/ramips/dts/MINIEMBPLUG.dts | 24 +++---------------------
 1 file changed, 3 insertions(+), 21 deletions(-)

diff --git a/target/linux/ramips/dts/MINIEMBPLUG.dts b/target/linux/ramips/dts/MINIEMBPLUG.dts
index 2f58eaf..5520ade 100644
--- a/target/linux/ramips/dts/MINIEMBPLUG.dts
+++ b/target/linux/ramips/dts/MINIEMBPLUG.dts
@@ -19,7 +19,7 @@
 
 		mobile {
 			label = "miniembplug:green:mobile";
-			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+			gpios = <&gpio0 10 GPIO_ACTIVE_LOW>;
 		};
 	};
 
@@ -29,28 +29,10 @@
 		#size-cells = <0>;
 		poll-interval = <20>;
 
-		wps {
-			label = "wps";
-			gpios = <&gpio0 0 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
 		reset {
 			label = "reset";
-			gpios = <&gpio0 10 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
-		mode-one {
-			label = "mode1";
-			gpios = <&gpio0 11 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
-		mode-two {
-			label = "mode2";
-			gpios = <&gpio0 12 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
 		};
 	};
 };
-- 
2.7.4
```

Please reopen the PR if you have the time to do what I asked for.

PR #1402, opened 2017-10-06
ramips: add support for Kimax U-25AWF 2.5" SATA WIFI HDD Enclosure

danielkucera:
Kimax U-25AWF is is a 2,5" HDD Enclosure with Wi-Fi/Eth conection
and battery, based on MediaTek MT7620N. Patch rewritten from:
https://forum.openwrt.org/viewtopic.php?pid=305643

Specification:

- MT7620N CPU
- 64 MB of RAM
- 16 MB of FLASH
- 802.11bgn WiFi
- 1x 10/100 Mbps Ethernet
- USB 2.0 Host
- UART for serial console

more on: https://wikidevi.com/wiki/Kimax_U25AWF

Flash instruction:
1. Download lede-ramips-mt7620-u25awf-squashfs-sysupgrade.bin
2. Open webinterface a upgrade
3. After boot connect via ethernet to ip 192.168.1.1

Tested on:

- Blueendless U-35WF which is the same hardware but without battery
and with 3,5" drive enclosure

Signed-off-by: Daniel Kucera <daniel.kucera@gmail.com>

### DIFF
--- a/target/linux/ramips/dts/U25AWF.dts
+++ b/target/linux/ramips/dts/U25AWF.dts
@@ -1,0 +1,108 @@
+/dts-v1/;
+
+#include "mt7620n.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/{
+	compatible = "kimax,u25awf","ralink,mt7620n-soc";
+	model = "Kimax U-25AWF";
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+		reset {
+			label = "reset";
+			gpios = <&gpio2 3 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+		wifi {
+			label = "u25awf:red:wifi";
+			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
+		};
+
+		lan {
+			label = "u25awf:green:lan";
+			gpios = <&gpio2 4 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+
+};
+
+&gpio2 {
+	status = "okay";
+};
+
+&gpio3 {
+	status = "okay";
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x30000>;
+			read-only;
+		};
+
+		partition@30000 {
+			label = "u-boot-env";
+			reg = <0x30000 0x10000>;
+			read-only;
+		};
+
+		factory: partition@40000 {
+			label = "factory";
+			reg = <0x40000 0x10000>;
+			read-only;
+		};
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
+		};
+	};
+};
+
+&ehci {
+	status = "okay";
+};
+
+&ohci {
+	status = "okay";
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0x4>;
+	ralink,port-map = "wllll";
+};
+
+&wmac {
+	ralink,mtd-eeprom = <&factory 0>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		default {
+			ralink,group = "wled","ephy";
+			ralink,function = "gpio";
+		};
+	};
+};
+
--- a/target/linux/ramips/image/mt7620.mk
+++ b/target/linux/ramips/image/mt7620.mk
@@ -193,6 +193,13 @@ define Device/gl-mt300a
 endef
 TARGET_DEVICES += gl-mt300a
 
+define Device/u25awf
+  DTS := U25AWF
+  IMAGE_SIZE := 16064k
+  DEVICE_TITLE := Kimax U-25AWF (16MB)
+endef
+TARGET_DEVICES += u25awf
+
 define Device/gl-mt300n
   DTS := GL-MT300N
   IMAGE_SIZE := $(ralink_default_fw_size_16M)
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -368,6 +368,10 @@ tl-wr841n-v13)
 	ucidef_set_led_switch "lan4" "lan4" "$board:green:lan4" "switch0" "0x10"
 	ucidef_set_led_switch "wan" "wan" "$board:green:wan" "switch0" "0x01"
 	;;
+u25awf)
+	set_wifi_led "u25awf:red:wifi"
+	ucidef_set_led_netdev "eth" "eth" "u25awf:green:lan" "eth0"
+	;;
 vocore-8M|\
 vocore-16M)
 	ucidef_set_led_netdev "eth" "ETH" "vocore:orange:eth" "eth0"
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -172,6 +172,7 @@ ramips_setup_interfaces()
 	rt-n14u|\
 	tl-wr840n-v4|\
 	tl-wr841n-v13|\
+	u25awf|\
 	ubnt-erx|\
 	ubnt-erx-sfp|\
 	ur-326n4g|\
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -227,6 +227,9 @@ get_status_led() {
 	tew-638apb-v2|\
 	tew-691gr|\
 	tew-692gr|\
+	u25awf)
+		status_led="$board:red:wifi"
+		;;
 	ur-326n4g|\
 	ur-336un|\
 	wf-2881)
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -505,6 +505,9 @@ ramips_board_detect() {
 	*"TL-WR841N v13")
 		name="tl-wr841n-v13"
 		;;
+	*"U-25AWF")
+		name="u25awf"
+		;;
 	*"UBNT-ERX")
 		name="ubnt-erx"
 		;;
--- a/target/linux/ramips/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/base-files/lib/upgrade/platform.sh
@@ -144,6 +144,7 @@ platform_check_image() {
 	tew-714tru|\
 	timecloud|\
 	tiny-ac|\
+	u25awf |\
 	ur-326n4g|\
 	ur-336un|\
 	v22rw-2x2|\


mkresin:
I only had a brief look at the PR, but it still has some of the issues I already outlined in PR#826.

- the gpios parameter in the DTS still contains invalid values
- the GPIO_ACTIVE_* macros are not used
- the key code macros are still not used

And you managed to introduce a new issue by removing the ";;" in the vocore-* section if 01_leds.

Please either close this PR, fix the issues by your self or find someone who is willing to do so.

danielkucera:
@mkresin I'm trying... now it compiles ok. What are "key code macros" plesae?

mkresin:
> What are "key code macros" plesae?

[dt-bindings/input/input.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h)

Please reference nodes defined in the mt7620n.dtsi instead of redefine them:

```
&gpio2 {
	status = "okay";
};
```

vs.

```
palmbus@10000000 {
	gpio2: gpio@660 {
		status = "okay";
	};
};
```

Please have  a look at the other mt7620n based dts files (like ZBT-CPE102.dts) and stick to the way it is done there.

Please be prepared for more change requests. Might be that I find more if I have a closer look and do not get distracted by the obvious stuff all the time :-).

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+		reset {
+			label = "reset";
+			gpios = <&gpio2 3 GPIO_ACTIVE_HIGH>;
+			linux,code = <0x198>;
```
f00b4r0: Please use the proper KEY_ value here as listed in [the file @mkresin pointed you to](https://github.com/torvalds/linux/blob/275490680c3dede8550f5ee3e59e2a618d2adb19/include/uapi/linux/input-event-codes.h#L476)

See how it's done [here](https://github.com/lede-project/source/blob/master/target/linux/ramips/dts/ZBT-CPE102.dts)

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+	model="KimaxU-25AWF";
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+		reset {
+			label = "reset";
+			gpios = <&gpio2 3 GPIO_ACTIVE_HIGH>;
```
f00b4r0: This is the incorrect GPIO_ACTIVE level, you had '1' in the original patch which is `GPIO_ACTIVE_LOW`.

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+			linux,code = <0x198>;
+		};
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+		wifi {
+			label="u25awf:red:wifi";
+			gpios=<&gpio3 0 GPIO_ACTIVE_HIGH>;
```
f00b4r0: here too

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+		};
+
+		lan {
+			label="u25awf:green:lan";
+			gpios=<&gpio2 4 GPIO_ACTIVE_HIGH>;
```
f00b4r0: here too

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+	status = "okay";
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0x4>;
+	ralink,port-map = "wllll";
+};
+
+&wmac {
+	ralink,mtd-eeprom = <&factory 0>;
```
f00b4r0: Have you confirmed that the device gets the correct MAC address for the wireless interface?

target/linux/ramips/image/mt7620.mk
```diff
@@ -193,6 +193,15 @@ define Device/gl-mt300a
 endef
 TARGET_DEVICES += gl-mt300a
 
+define Device/u25awf
+  DTS := U25AWF
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
+  IMAGES += factory.bin
```
mkresin: please drop the IMAGE and IMAGE/factory.bin line. No need to build a factory image that is identical to the sysupgrade image

target/linux/ramips/image/mt7620.mk
```diff
@@ -193,6 +193,15 @@ define Device/gl-mt300a
 endef
 TARGET_DEVICES += gl-mt300a
 
+define Device/u25awf
+  DTS := U25AWF
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
```
mkresin: `ralink_default_fw_size_16M` is 16121856 bytes (0xf60000) which doesn't match the size if your firmware partition. Please use 16064k as IMAGE_SIZE.

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
+/dts-v1/;
+
+#include "mt7620n.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/{
+	compatible="u25awf","ralink,mt7620n-soc";
```
mkresin: please use `compatible = "kimax,u25awf", "ralink,mt7620n-soc";` as compatible string.

target/linux/ramips/dts/U25AWF.dts
```diff
@@ -0,0 +1,108 @@
...
+	model="KimaxU-25AWF";
```
mkresin: please use `model = "Kimax U-25AWF";` here

Mynzer:
@f00b4r0 Hi, I can confirm that the device boot with the changes you ask to @danielkucera and I can confirm, that the mac address in lede is the same mac address reported in my boot loader (I have breed boot loader in my device), the enclosure doesn't have any mac address written on it, I add a boot log from my device with @f00b4r0 changes. I'll add a new boot log when @danielkucera do the new changes requested by @mkresin
[dmesg.txt](https://github.com/lede-project/source/files/1364183/dmesg.txt)

danielkucera:
@Mynzer please run `git log` in source directory and paste here the hash of latest commit to make sure you are on the latest version.

Mynzer:
Here is the hash output from git log

5b975153fe6502db7a4cabe7d3e6b264d55c2c27

Just tell me if you need to pull again the new changes

danielkucera:
@Mynzer please do `git pull` and make sure you are on `59fd9d11e49545fce393ea6cb28976d4d3dec3e6`, then `make` and test again please.

Mynzer:
Ok, now `git log` show me that I'm in commit `59fd9d11e49545fce393ea6cb28976d4d3dec3e6`
here is the new boot log

[dmesg.txt](https://github.com/lede-project/source/files/1366423/dmesg.txt)

danielkucera:
@Mynzer can you please confirm that all drivers are working correctly (LEDs, reset button, USB, ethernet, wifi)?

Mynzer:
Almost all work fine, Wifi and Ethernet leds work fine, for usb and internal enclosure hard drive, I have installed samba and create a share in LEDE, one form the internal hard drive of the enclosure and one for the usb memory on the usb port, and works good getting access and writing for both shares from my laptop, the only thing that do not work well is the reset button, it simply not reset to factory defaults.

I don't know if it helps, but for breed boot loader I'm using breed-mt7620-reset13.bin

mkresin:
> the only thing that do not work well is the reset button, it simply not reset to factory defaults.

You need to keep the reset button pressed for > 5sec to start a factory reset. If you press the button for < 5 sec a reboot is done.

If the button can not distinguish between long and short press, the polarity (HIGH vs. LOW) is usually wrong.

The easiest way to test if the button works at all, is to (short) press the button during boot at the time the LED blinks really fast. The device should switch to failsafe. The diag LED (here wlan) should blink  insane fast and ssh connections to 192.168.1.1 are accepted.

danielkucera:
@Mynzer if you can enter breed command mode, please test the button using `btntst` command 
btntst

Mynzer:
Hi again, @mkresin I've tried the steps that you recommend for testing the reset button, but unfortunately I could not put in failsafe the device (pressing the reset button on boot time), once that the device fully boot in LEDE, I've tried pressing the reset button of 3 seconds to make it reboot, and I've tried pressing for 6 seconds for factory reset but for now, both steps not work in the UA25AWF device.
@mkresin Using this steps all this steps, effectively put my other device (a Xiaomi Mi Wifi Router) on fail safe (press reset on boot), reboots it (press reset for 3 seconds), an return to factory defaults (press reset for 6 seconds)

@danielkucera Here is the output from btntst from breed
breed-btntst
When i press reset button, change to 0, when i release it, change to 1



`

mkresin:
Seem to me the reset button gpio is wrong in the DTS.

@Mynzer would you please build an image form my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary) and test again!

mkresin:
I have more one more question. Does the Kimax U-25AWF have more than one network port? There is currently a lan + wan port config applied by default. If the board has only a single lan port, we can make it way simpler.

Maybe better to check this first, before another round of testing is required.

Mynzer:
@mkresin The Kimax U-25AWF has only a single lan port

danielkucera:
@Mynzer for testing mkresins code you can do following in our working folder:
```
git checkout master
git remote add mkresin https://git.lede-project.org/lede/mkresin/staging.git
git pull mkresin master
make
```
you don't need to recompile the whole project again.

mkresin:
@Mynzer please pull my staging tree now.

@danielkucera with with my changes?

danielkucera:
@mkresin yes, without objections.

Mynzer:
@danielkucera thanks for the clear instructions, that save a lot of time, and I successfully pull from mkresin staging tree

@mkresin from commit `72b98f2749560eac49d20576f95bdf6ad12d5c78` now the device could enter in fail safe pressing the reset button on boot, and once the device fully boot in LEDE, now it can reset to factory defaults when pressing the reset button for more than 5 seconds

v_20171009_102517_gif

Here is the boot log from this commit

[dmesg.txt](https://github.com/lede-project/source/files/1369203/dmesg.txt)

f00b4r0:
Took a quick look at the original firmware bootlog [here](https://forum.openwrt.org/viewtopic.php?pid=305643#p305643) and I notice this:

```
[    0.860000] 0x000000050000-0x000000ff0000 : "firmware"
[    1.470000] 0x000000ff0000-0x000001000000 : "hui-ai"
```

I don't know if that's relevant but I am concerned that by making the firmware partition in LEDE extend over that `hui-ai` one in the original firmware, you might be making it impossible to revert to stock firmware.

It also appears from that post and from your dmesg that there are several variants of that device and that the Blueendless and Kimax versions aren't exactly the same: @danielkucera mentions MT7620N, in line with the bootlog linked above which has DDR1 RAM, while @Mynzer dmesg output shows an MT7620A with DDR2 RAM. The commit message might want to reflect that.

I wonder if that could also explain the discrepancy with the reset button. It might be a good idea if @danielkucera could confirm the last change works for him and provide dmesg output.

danielkucera:
@f00b4r0 my device is U-35WF but the original patch worked for me except the reset button - it didn't work as far as I remember.

f00b4r0:
@danielkucera that btntst excerpt you posted, is it from _your_ device? Because if it is, it shows GPIO 12. @Mynzer  shows GPIO 13.
Hence me wondering if you are trying to reconcile two _different_ versions of that hardware. It's possible they're clones implemented differently, given their "origins".

danielkucera:
@f00b4r0 no, it's not from mine. I found the picture on internet.

Mynzer:
@f00b4r0 Let me check if i could found the stock firmware from this device and check if i could go back to stock. My device It's marked on front Kimax, in the back Blueenndless and bottom BS-25AWF

v_20171009_135309_segment_0_gif

v_20171009_140625_gif

Maybe the difference is the revision of the board, in my device it's printed on board
`U25AWF-7620A-1153-V2.0`

Mynzer:
@f00b4r0 I could go back to stock using firmware from here using the luci web interface using HSY-02 firmware
https://www.dropbox.com/sh/lmiogomcxqnaghv/AABORq743r3X7hEV70978Suua?dl=0
All seems to work well on stock firmware, and I could go back to LEDE from stock using the image generated by @mkresin build environment.

Mynzer:
When i see there was 2 stock firmwares version in the dropbox link, I decide to take time to test both firmwares in my device

The HSY-02 firmware boot on my device and all woks fine except the reset button that not work, the same thing that was happening before @mkresin do the change in the dts file in the reset button. When i see the U25AWF stock firmware boot log  firmware from [openwrt forum](https://forum.openwrt.org/viewtopic.php?pid=305643) I see all the things that @f00b4r0 mention before, that the original Kimax version have MT7620N and DDR1 but also notice this
`[    0.000000] MIPS: machine is HSY HSY-02`

The H1 firmware is the original firmware that came in my device, with this firmware all work fine, including the reset button, there are other U25AWF users in the openwrt forum [here](https://forum.openwrt.org/viewtopic.php?pid=353144#p353144) and [here](https://forum.openwrt.org/viewtopic.php?pid=340924#p340924) and [this](https://mysku.ru/blog/china-stores/51765.html)  Russian guy that have the U25AWF H1 version with MT7620A and DDR2 memory.

So my conclusions is that there are 2 versions of the Kimax U25AWF

- Version 1 have MT7620N, DDR1 RAM, on board it's printed HSEN-KI-300M-HDD-V3.0 and uses stock Kimax HSY-02 firmware
The original patch was written for this device, i think so, because in openwrt forum, the creator of this patch [say that all work ](https://forum.openwrt.org/viewtopic.php?pid=305720#p305720) including the reset button

- Version 2 have MT7620A, DDR2 RAM, on board it's printed U25AWF-7620A-1153-V2.0 and uses stock Kimax H1 firmware
So far, the original patch on this device works except for the reset button, but with the changes that made @mkresin this device could have support.

With this information, i don't know if could be possible add support for the 2 versions of this device, it seems that the only difference, in the software side, it's the different gpio for the reset button.

mkresin:
Oh my godness, it's getting complicated.

To summarize, there are two different revisions of the Kimax U25AWF

- MT7620N with DDR1 RAM
  - board is labelled HSEN-KI-300M-HDD-V3.0
  - stock firmware is named HSY-02

- MT7620A with DDR2 RAM
  - board is labelled U25AWF-7620A-1153-V2.0
  - stock firmware is named H1

Both of you are having the U25AWF-7620A-1153-V2.0 (MT7620A) board? If yes, we need to touch the DTS file again, since it includes the mt7620n.dtsi and has a compatible string for the mt7620n-soc.

Is it confirmed that both revisions are named U25AWF? Is there any way do distinguish the devices from the outside (suffix H1/HSY-02, revision number or similar on the case)? How does one know which stock firmware needs to be installed?

I would like to refrain from adding support for a revision which can not be tested at the moment. But I would like to name the board/image in a way that it is possible to understand it is meant for the MT7620A revision.

@f00b4r0 thanks for bringing up the issue

@Mynzer thanks for doing the research

Mynzer:
@mkresin Yes, my device it's U25AWF-7620A-1153-V2.0 MT7620A board, this [user claims](https://forum.openwrt.org/viewtopic.php?pid=353144#p353144) that have the same board, and this  [user says](https://forum.openwrt.org/viewtopic.php?pid=340924#p340924) that have an U25AWF with MT7260A chipset.

Both versions are named U25AWF, there is another version, the U25WF, but the case it's different and it use RT5350F chipset.

It's not clear how to know which U25AWF version is each one from the outside, the external case is similar for both versions, and the info printed in the case i don't think that be reliable, because it change between sub manufacturers as mention in the openwt forum.

The way that I think that could be more reliable to identify the device version could be checking the stock firmware web page configuration.

This is the stock Kimax U25AWF MT7620N configuration page
hsy-02

This is the stock Kimax U25AWF MT7620A configuration page
h1

And then unmount to check the revision board
Kimax U25AWF MT7620N board with DDR1 RAM
MT7620N

Kimax U25AWF MT7620A board with DDR2 RAM
MT7620A

i don't think that unmount the board could be a big problem, because you need to disassembly the case to insert a sata hard drive, once you follow the same steps that you use to put the sata drive, you only need to remove two additional screws to pull out the board.

Maybe the firmware image could be named u25awf-hsy02 for the MT7260N version (when it gets support in the future) and u25awf-h1 for the MT7620A version.

F5BJR:
I have 2 * U25-CWF ( only wifi ) and i have a problem ...
For moment i have connected a serial port for start with this command 👍 
uci set wireless.@wifi-device[0].disabled=0; uci commit wireless; wifi👍 
How i start firmware directitly in wifi mode ?

I think that is posible to modify in the firmware but ???

Thanks for infos

mkresin:
@Mynzer I've reworked the commit in my staging tree. The board is now named Kimax U25AWF-H1 and uses the mt7620a stuff.

Would you please do your testing once more (and hopefully for the last time).

Mynzer:
@mkresin I don't have problems testing all the necessary times that could be needed.
I pulled the new changes from your staging tree, but now it seems that the new Kimax U25AWF-H1 target don't appear in the `make menuconfig ` menu.

Maybe i'm wrong, I'm still learning, but it could be that in the mt7620.mk file in the target/linux/ramips/image folder
`TARGET_DEVICES += u25awf`
should be
`TARGET_DEVICES += u25awf-h1`

mkresin:
@Mynzer 

On spot! That was the change I missed. Seem to me now you're reviewing my changes ;-).

It is fixed in my staging tree. You might need to remove the tmp directory within the LEDE directory to see the changed board name.

Mynzer:
@mkresin Now I can see the change in the board name, here is the boot log from your staging tree, commit `8c4c2cb407a0f8a39986badb39806b325cf17c7d`

[dmesg.txt](https://github.com/lede-project/source/files/1384137/dmesg.txt)

All is working fine on the device (leds, internal hard drive, wifi, usb port, Ethernet, reset button).
Tell me if i can assist you with some other specific tests on the device and post results.

mkresin:
> Tell me if i can assist you with some other specific tests on the device and post results.

Nothing left to do. I've merged the commit. Thanks to all!

cristipatty:
Hello,
I have this device: https://www.aliexpress.com/item/Wireless-Hdd-Docking-station-for-2-5-3-5-SATA-hard-disk-USB3-0-wWifi-HDD/32623783372.html .
Device model is **BS-HD03WF**, board's chipset is **MT7620N**.
On default firmware web interface it says "**Router model HSY-02**".

As you can see it has a different form-factor but considering the HW is the same I',m tempted to try this FW. The only thing different is that it has 2 x RJ45 - WAN and LAN instead of just one. Otherwise USB3.0 is there, USB2.0 (with card-reader) is there, maybe leds might be different but they aren't that important for now.

Do you think I could flash this FW from stock web-update section?
I'm interested in LAN and Wifi to work and be able to access mounted storages (HDD sata or USB with card-reader).
Do you think the firmware will work and device won't break?

Thanks.

Mynzer:
Hi @cristipatty 
Sorry for the long delay. Kimax/Blueendless company seem to use the same "Router model HSY-02" firmware for various devices as stated in the [manual for our devices](https://www.dropbox.com/sh/lmiogomcxqnaghv/AACKb0t4V3CQo6TsmHSaBTPIa/original%201.3.5?dl=0&preview=WiFi+Storage+Manual.pdf).

Mr. @danielkucera [seem to be using](https://blog.danman.eu/lede-added-support-for-kimax-u25awf-h1/) this firmware on his U-35WF. The stock HSY-02 firmware don't make any verification in the firmware upgrade process, so you can flash it and **maybe it could** work with your BS-HD03WF device.

The bad news is that I could not find any information, any internal pics or any information how your Router/HDDcase could be recovered, **so in case of brick of your device, you are alone.**.

Now, in the best of the cases that the hardware is the same, and the device boot with OpenWRT/LEDE firmware, you will end with a device with not correct switch configuration (maybe 1 or none of the Ethernet ports will work). In case that you want full support for your device, you will need to provide information from your device and open a new PR, (maybe @danielkucera could help you opening the new PR if you kindly ask him for help and only if you are willing to take the risk and provide the info that the developers need).

If you are interested in continue anyway,, I think that the best place to continue the discussion is in the forums, I don't think that the developers like to bring back closed PR, and maybe find there more people interested in bring support for your device.

Greetings.

markosc:
> Hi @cristipatty Sorry for the long delay. Kimax/Blueendless company seem to use the same "Router model HSY-02" firmware for various devices as stated in the [manual for our devices](https://www.dropbox.com/sh/lmiogomcxqnaghv/AACKb0t4V3CQo6TsmHSaBTPIa/original%201.3.5?dl=0&preview=WiFi+Storage+Manual.pdf).
> 
> Mr. @danielkucera [seem to be using](https://blog.danman.eu/lede-added-support-for-kimax-u25awf-h1/) this firmware on his U-35WF. The stock HSY-02 firmware don't make any verification in the firmware upgrade process, so you can flash it and **maybe it could** work with your BS-HD03WF device.
> 
> The bad news is that I could not find any information, any internal pics or any information how your Router/HDDcase could be recovered, **so in case of brick of your device, you are alone.**.
> 
> Now, in the best of the cases that the hardware is the same, and the device boot with OpenWRT/LEDE firmware, you will end with a device with not correct switch configuration (maybe 1 or none of the Ethernet ports will work). In case that you want full support for your device, you will need to provide information from your device and open a new PR, (maybe @danielkucera could help you opening the new PR if you kindly ask him for help and only if you are willing to take the risk and provide the info that the developers need).
> 
> If you are interested in continue anyway,, I think that the best place to continue the discussion is in the forums, I don't think that the developers like to bring back closed PR, and maybe find there more people interested in bring support for your device.
> 
> Greetings.
I have the same model bs-HD03WF...........have you been able to try that update?

markosc:
> Hello guys @danielkucera @mkresin @Mynzer , I have this device: https://www.aliexpress.com/item/Wireless-Hdd-Docking-station-for-2-5-3-5-SATA-hard-disk-USB3-0-wWifi-HDD/32623783372.html . Device model is **BS-HD03WF**, board's chipset is **MT7620N**. On default firmware web interface it says "**Router model HSY-02**".
> 
> As you can see it has a different form-factor but considering the HW is the same I',m tempted to try this FW. The only thing different is that it has 2 x RJ45 - WAN and LAN instead of just one. Otherwise USB3.0 is there, USB2.0 (with card-reader) is there, maybe leds might be different but they aren't that important for now.
> 
> Do you think I could flash this FW from stock web-update section? I'm interested in LAN and Wifi to work and be able to access mounted storages (HDD sata or USB with card-reader). Do you think the firmware will work and device won't break?
> 
> Thanks.

I have the same model and clearly tempted to try, the problem is that it can become a brick, for the moment I only use it as sata to usb.

PR #1404, opened 2017-10-09
lantiq: rename nas0/ptm0 to dsl0

sch-m:
This change makes it possible to configure the wan/dsl ppp interface
settings independantly from the used TC-Layer (ATM/PTM).

Now you can move a device from an ADSL/ATM port to an VDSL/PTM port
without any configuration changes for example.

@mkresin

### DIFF
--- a/package/network/config/ltq-vdsl-app/Makefile
+++ b/package/network/config/ltq-vdsl-app/Makefile
@@ -58,10 +58,11 @@ CONFIGURE_ARGS += \
 #CONFIGURE_ARGS += --enable-model=debug
 
 define Package/ltq-vdsl-app/install
-	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin $(1)/etc/hotplug.d/dsl
+	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin $(1)/etc/hotplug.d/dsl $(1)/etc/hotplug.d/net
 	$(INSTALL_BIN) ./files/dsl_control $(1)/etc/init.d/
 	$(INSTALL_BIN) ./files/10_atm.sh $(1)/etc/hotplug.d/dsl
 	$(INSTALL_BIN) ./files/10_ptm.sh $(1)/etc/hotplug.d/dsl
+	$(INSTALL_BIN) ./files/10-xdsl_rename $(1)/etc/hotplug.d/net
 
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/src/dsl_cpe_control $(1)/sbin/vdsl_cpe_control
 	$(INSTALL_BIN) ./files/dsl_cpe_pipe.sh $(1)/sbin/
--- a/package/network/config/ltq-vdsl-app/files/10-xdsl_rename
+++ b/package/network/config/ltq-vdsl-app/files/10-xdsl_rename
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+if [ "$ACTION" = add ]; then
+	[ "$DEVICENAME" = "nas0" ] ||[ "$DEVICENAME" = "ptm0" ] || exit
+
+	ip link set $DEVICENAME name dsl0
+fi
--- a/target/linux/lantiq/base-files/etc/uci-defaults/02_migrate_xdsl_iface
+++ b/target/linux/lantiq/base-files/etc/uci-defaults/02_migrate_xdsl_iface
@@ -1,0 +1,57 @@
+#!/bin/sh
+
+. /lib/functions.sh
+. /lib/functions/lantiq.sh
+
+IFNAME_CHANGED=0
+
+rename_xdsl_ifname()
+{
+	local cfg="$1"
+	local section="$2"
+	local option="$3"
+	local name
+
+	config_get name ${section} $option
+	case $name in
+		nas0*)
+			name=${name/nas0/dsl0}
+			;;
+		ptm0*)
+			name=${name/ptm0/dsl0}
+			;;
+		*)
+			return
+			;;
+	esac
+
+	uci set ${cfg}.${section}.$option=$name
+	IFNAME_CHANGED=1
+}
+
+migrate_network_xdsl_ifname()
+{
+	rename_xdsl_ifname network "$1" ifname
+	rename_xdsl_ifname network "$1" name
+}
+
+migrate_led_xdsl_ifname()
+{
+	rename_xdsl_ifname system "$1" dev
+}
+
+lantiq_is_vdsl_system || exit 0
+
+config_load network
+config_foreach migrate_network_xdsl_ifname
+
+[ "$IFNAME_CHANGED" = "1" ] && uci commit network
+
+IFNAME_CHANGED=0
+
+config_load system
+config_foreach migrate_led_xdsl_ifname led
+
+[ "$IFNAME_CHANGED" = "1" ] && uci commit system
+
+exit 0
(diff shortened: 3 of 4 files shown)


package/network/config/ltq-vdsl-app/files/10_atm.sh
```diff
@@ -14,3 +14,8 @@ fi
 modprobe ltq_atm_vr9
 
 /etc/init.d/br2684ctl reload
+
+sleep 2
```
mkresin: That looks error prone to race conditions. Can't we use something hotplug based?

target/linux/lantiq/base-files/etc/uci-defaults/02_network_migration
```diff
@@ -0,0 +1,17 @@
...
+# Copyright (C) 2017 OpenWrt.org
+#
+
+. /lib/functions.sh
+. /lib/functions/lantiq.sh
+
+if lantiq_is_vdsl_system; then
+	config_load network
+	config_get iface wan ifname
+	[ "$iface" = "nas0" ] || [ "$iface" = "ptm0" ] || exit
```
mkresin: This will not migrate configs using a vlan tag. You need to check for begins with and replace only the nas0/ptm0 part afterwards.

This should do the job:

```
#!/bin/sh

. /lib/functions.sh
. /lib/functions/lantiq.sh

WAN_IFACE_CHANGED=0

if lantiq_is_vdsl_system; then
	config_load network
	config_get iface wan ifname

	case $iface in
		nas0*)
			iface=${iface/nas0/dsl0}
			WAN_IFACE_CHANGED=1
			;;
		ptm0*)
			iface=${iface/ptm0/dsl0}
			WAN_IFACE_CHANGED=1
			;;
	esac

	uci set network.wan.ifname=$iface
fi

[ "$WAN_IFACE_CHANGED" = "1" ] && uci commit network

exit 0
```
```

File: target/linux/lantiq/base-files/etc/uci-defaults/02_network_migration [hunk 3]
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+#
+# Copyright (C) 2017 OpenWrt.org
```
mkresin: Please drop the copyright or add it to your self.

target/linux/lantiq/base-files/etc/uci-defaults/02_network_migration
mkresin: 02_migrate_xdsl_iface *might be* a better name for the script.

mkresin: One more thing. You need to take care of the wan_dev device config as well (the one that sets the mac address). It might work using the old interface name as well, but I prefer to have everything migrated to the same values used by a fresh config.

sch-m: Oh yes, you are right. I've forgot the vlan variants.

sch-m: OK, will also have a look at the wan_dev section

target/linux/lantiq/base-files/etc/uci-defaults/02_migrate_xdsl_iface
```diff
@@ -0,0 +1,28 @@
...
+			iface=${iface/nas0/dsl0}
+			WAN_IFACE_CHANGED=1
+			;;
+		ptm0*)
+			iface=${iface/ptm0/dsl0}
+			WAN_IFACE_CHANGED=1
+			;;
+	esac
+
+	uci set network.wan.ifname=$iface
```
sch-m: Shouldn't it be enough to add a 'uci set network.wan_dev.name=dsl0' here ?

mkresin: Nope. It would drop the vlan tag again. For example ptm0.7 need to be dsl0.7 to make t-com connections working.

mkresin: Bummer, you are talking about wan_dev.

Yes, dsl0 should be enough to match the default config. But you never know what people have in their config.

They might have changed it to a vlan interface (not sure if this is even supported) or they are using this section for a complete different purpose. To be on the safe side, it should checked if the name matches the expections and we should expect vlan interfaces here as well.

sch-m: OK, then I will insert the complete check again here.

package/network/config/ltq-vdsl-app/Makefile
```diff
@@ -58,10 +58,11 @@ CONFIGURE_ARGS += \
 #CONFIGURE_ARGS += --enable-model=debug
 
 define Package/ltq-vdsl-app/install
-	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin $(1)/etc/hotplug.d/dsl
+	$(INSTALL_DIR) $(1)/etc/init.d $(1)/sbin $(1)/etc/hotplug.d/dsl $(1)/etc/hotplug.d/net
 	$(INSTALL_BIN) ./files/dsl_control $(1)/etc/init.d/
 	$(INSTALL_BIN) ./files/10_atm.sh $(1)/etc/hotplug.d/dsl
 	$(INSTALL_BIN) ./files/10_ptm.sh $(1)/etc/hotplug.d/dsl
+	$(INSTALL_BIN) ./files/10-xdsl_rename.sh $(1)/etc/hotplug.d/net
```
mkresin: > install: cannot stat './files/10-xdsl_rename.sh': No such file or directory

The file doesn't have a file extension.

package/network/config/ltq-vdsl-app/files/10-xdsl_rename
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ "$ACTION" = add ]; then
+	[ "$DEVICENAME" = "nas0" ] ||[ "$DEVICENAME" = "ptm0" ] || exit
+
+	ip link set $DEVICENAME name dsl0
```
dedeckeh: The ltq-vdsl-app package now needs to depend on ip

hauke: does the busybox ip command not support this feature?

mkresin: Yes, the busybox ip command provides this feature:

```
root@LEDE:/# /bin/busybox ip link set wlan0 name mk1
[  197.621145] ath10k_pci 0000:02:00.0 mk1: renamed from wlan0
```
If I add a `DEPENDS:= +ip` to the package, ip-tiny is autoselected without any need. Seem to me the ip depend doesn't work as expected with the busybox applet.

sch-m:
So, is there anything left to be done/changed?

mkresin:
> So, is there anything left to be done/changed?

Not at the moment. But I prefer to wait for comment on the ip issue by @dedeckeh. Could be that I miss something.

Nevertheless, I've merged your commit with a minor bugfix to my staging tree.

sch-m:
Oh yes, i forgot to add the 01_leds file. Sorry. Should I also update this PR?

mkresin:
> Should I also update this PR?

No need to, as it is already fixed in my staging tree.

package/network/config/ltq-vdsl-app/files/10-xdsl_rename
dedeckeh: Busybox now provides "ip" (https://git.lede-project.org/?p=source.git;a=commit;h=1cec4d4ef0145b48094ec0df40c18b8fbb315cf0); specifying ip depend should not select ip-tiny anymore now

mkresin:
merged with the added dependency on `ip`. Thanks to all!

PR #1405, opened 2017-10-09
openvpn: add support to start/stop single instances

sch-m:
This would be already used by the openvpn luci app, so this should be fixed:

https://github.com/openwrt/luci/blob/master/applications/luci-app-openvpn/luasrc/model/cbi/openvpn.lua#L111

### DIFF
--- a/package/network/services/openvpn/files/openvpn.init
+++ b/package/network/services/openvpn/files/openvpn.init
@@ -60,7 +60,7 @@ openvpn_add_instance() {
 	local dir="$2"
 	local conf="$3"
 
-	procd_open_instance
+	procd_open_instance "$name"
 	procd_set_param command "$PROG"	\
 		--syslog "openvpn($name)" \
 		--status "/var/run/openvpn.$name.status" \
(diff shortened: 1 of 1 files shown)


package/network/services/openvpn/files/openvpn.init
```diff
@@ -105,26 +105,31 @@ start_instance() {
...
 	. /usr/share/openvpn/openvpn.options
 	config_load 'openvpn'
-	config_foreach start_instance 'openvpn'
 
-	local path name
-	for path in /etc/openvpn/*.conf; do
-		if [ -f "$path" ]; then
-			name="${path##*/}"; name="${name%.conf}"
+	if [ -n "$1" ]; then
+		start_instance "$1"
```
dedeckeh: A check is missing verifying if the specified instance name is actually a valid openvpn uci section.
Now start_instance can be called with an invalid instance name which is not desired.
Have a look at dnsmasq init script how this is handled (https://git.lede-project.org/?p=source.git;a=blob;f=package/network/services/dnsmasq/files/dnsmasq.init;h=3d3d83334e0073b1092320a25a553513ca25c3db;hb=HEAD#l1081)

dedeckeh:
Thanks; merged into master (https://git.lede-project.org/?p=source.git;a=commit;h=e2f25e607d2092cffa45196e7997854feb464232)

PR #1406, opened 2017-10-10
layerscape: fix build issue for usb

yangbolu1991:
When build firmware for layerscape target with CONFIG_ALL_NONSHARED=y,
there would be a compile issue of usb ehci-fsl driver. Actually this
driver was for PPC platforms initially and was not ready for non-PPC
now, but a kernel kconfig patch removed PPC dependency for it. So that
kernel patch should be reverted.


yangbolu1991:
I put a kernel patch 819-Revert-usb-kconfig-remove-dependency-FSL_SOC-for-ehc.patch into patches-4.9 to fix the usb build issue.
Once this issue is fixed on NXP LSDK linux, I will update patches-4.9/817-usb-support-layerscape.patch, and remove this patch.
Then we still could keep one patch for one IP module.

Thanks.

hauke:

PR #1407, opened 2017-10-10
layerscape: only support 64-bit for ls1088ardb/ls2088ardb

yangbolu1991:
This is no requirement and plan to support 32-bit for ls1088ardb
and ls2088ardb. Current 32-bit firmware for them couldn't work,
so only keep 64-bit support for these two boards in menuconfig.


hauke:

PR #1408, opened 2017-10-11
ar71xx: Fix MAC partition of Meraki MR12/MR16

oleeander:
Introduced by Chris Blake in 28dd52b079d39fa80dd801e2864bea42251fd50b,
LEDE uses a custom partition to store the MAC address.
That solution required setting the device's primary MAC address manually.

With this patch, an additional partition (config) with the correct location
of the MAC address stored in the SPI is used.

Because the new MAC / config partition is within the old rootfs / firmware
area, you need to reinstall LEDE when upgrading.
To install LEDE within the changed flash layout, use to following commands
on the serial console:

	tftpboot 0x80010000 lede-mr16-squashfs-sysupgrade.bin
	erase 0xbf0a0000 +0xf40000
	cp.b 0x80010000 0xbf0a0000 0xf40000
	setenv bootcmd bootm 0xbfda0000; saveenv; boot

When upgrading from Chris's workaround with the MAC address set manually,
you are also required to copy it to the right location using these commands
_before_ copying the firmware. The config partition is marked read-only:

	erase 0xbf080000 +0x20000
	cp.b 0xbffd0000 0x80010066 0x20000
	cp.b 0x80010000 0xbf080000 0x20000

But because installation is currently only possible with access to the
serial console and that has never been different,
these steps look acceptable to me.

This was tested, and confirmed working only on the MR16.

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-mr12.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-mr12.c
@@ -42,7 +42,9 @@
 
 #define MR12_WAN_PHYMASK    BIT(4)
 
-#define MR12_CALDATA0_OFFSET            0x21000
+#define MR12_MAC			0xbf090000
+#define MR12_ART			0xbfff0000
+#define MR12_CALDATA0_OFFSET		0x1000
 
 static struct gpio_led MR12_leds_gpio[] __initdata = {
 	{
@@ -89,8 +91,10 @@ static struct gpio_keys_button MR12_gpio_keys[] __initdata = {
 
 static void __init MR12_setup(void)
 {
-	u8 *mac = (u8 *) KSEG1ADDR(0xbffd0000);
-	u8 wlan_mac[ETH_ALEN];
+	u8 *mac = (u8 *) KSEG1ADDR(MR12_MAC);
+	u8 *art = (u8 *) KSEG1ADDR(MR12_ART);
+
+	u8 wlan0_mac[ETH_ALEN];
 
 	ath79_register_mdio(0,0x0);
 
@@ -98,6 +102,8 @@ static void __init MR12_setup(void)
 	ath79_eth0_data.phy_if_mode = PHY_INTERFACE_MODE_RGMII;
 	ath79_eth0_data.phy_mask = MR12_WAN_PHYMASK;
 	ath79_register_eth(0);
+
+	ath79_init_mac(wlan0_mac, mac, 1);
 
 	ath79_register_m25p80(NULL);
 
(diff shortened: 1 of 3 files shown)


mkresin:
@riptidewave93 would you please have a look at and ACK/NAK the PR

riptidewave93:
ACK and approve. My only concern is making sure people are aware of this change before upgrading to prevent booting issues.

The reason for going the way I did with the last MAC partition change was to allow people to upgrade without any changes to the bootcmd. While I agree this patch is the correct way (as it uses the stock MAC address location) it may be worth looking at adding a sysupgrade wrapper to prevent upgrading to a firmware containing this change without a manual reflash via u-boot.

oleeander:
I wrote that patch a while ago and currently have no device at hands to compare with. But isn't the address in the bootcmd still the same? See https://github.com/riptidewave93/Openwrt-MR12-MR16 for reference.

Do you know if it is it possible to change MTD partitions after boot? Then it should be possible to write mac address and firmware without access to the console.

riptidewave93:
@oleeander Ah, I missed the fact these boards are rootfs before kernel. It's been a while since ive worked on these specific devices. Thanks for pointing that out :). Due to this, I don't have as much concern and can greenlight this PR. :+1: 

As for changing MTD partitions, are you referring to modifying the mtd layout after boot? I know there are ways for enabling RW on RO partitions [0], but I personally am not aware of a way to change the mtd layout after boot.

Also it may be worth at just leaving the new "config" partition rw so users can manually set their mac from userspace if they are sysupgrading from an older firmware. Just my 2c.

[0]: https://github.com/jclehner/mtd-rw

mkresin:
If I got you right, due to the move of the kernel/rootfs partition a simple sysupgrade would brick the board/isn't possible.

Since both boards are already supported by a release image, a bricking sysupgrade is something that should be prevented in all cases. If it isn't possible => NAK.

riptidewave93:
@mkresin I will spend some time within the next week verifying this PR does not affect upgrading the MR12/MR16 from a stable release and will report my findings here.

riptidewave93:
@mkresin This patch does require a full reflash through U-Boot due to the "running" version not being aware of the new partition layout, which affects placement of the rootfs and kernel. 

If you don't flash this change via U-Boot, your device will fail to boot. Tested on a Meraki MR12. Giving this patch a NAK. :-1:

mkresin:
Your approach is correct and it should have been done this way from the beginning. But we can't merge the fix due to the breakage of existing installations.

I don't see a way to inform users of already deployed boards that an upgrade need to be performed via u-boot. Furthermore, it might be difficult to get UART access to deployed MR12/MR16 depending on their location/position.

rotanid:
what about adding the fixed version using a slightly different name?
i don't have this device, but this idea came to my mind when reading here, maybe it could workaround the issue.

NeoRaider:
I also talked with @jow- about this issue at the WBM, and we agreed that it should be fixed in some way if possible without breaking upgrades. While this will need some more work, I propose the following solution:

* Add a new image with a changed name as @rotanid suggested. The old image will need to be provided for "a while" (probably one LEDE/OpenWrt release cycle), so people have a clean upgrade path, after that the old version can be dropped
* If possible without breaking anything, change the magic numbers of the new image so old versions of platform.sh will reject the upgrade
* Make sysupgrade distingiush the two images in platform.sh: old->old and new->new upgrades are handled the usual way, old->new will flash the image with an offset and write the MAC address to the offset it will be expected at after the upgrade. Implementing support for new->old upgrades is probably not worth the effort.

rotanid:
fixed in OpenWrt Commit af9dee336dacbc61e056bce0c0dbf508dd7453a2 for target ath79

PR #1410, opened 2017-10-12
toolchain/gcc: update 5.x to 5.5.0

rmounce:
This is the final bugfix release in the gcc-5 series.

Compile and run tested on macOS 10.13 (Xcode 9), mvebu/ar71xx.

Removed redundant patch for macOS (backported upstream by yours truly)


hauke:

PR #1411, opened 2017-10-13
build: use KERNEL_CROSS for kernel file compilations

KarlVogel:
The build system already defines KERNEL_CROSS which defaults to TARGET_CROSS.
Make use of this variable for kernel makefiles.


hauke:
I think many can even bre replaced with 
KERNEL_MAKE_FLAGS like it was done in c6c1f6a5480b12c938ae4154a55a08a559e8e46b

KarlVogel:
Updated to use KERNEL_MAKE_FLAGS. Test compiled for target x86.

hauke:

PR #1412, opened 2017-10-13
[RFC] layerscape: Add Traverse LS1043 Family

mcbridematt:
This is an experimental patch set that adds support for the Traverse LS1043 product line.
The hardware is still a work in progress, so far LS1043V(dsl) boards exist,
and the LS1043S(fp), Five64 (appliance) boards are in development.

Each board (as well as NXP's LS1043ARDB) should be compatible enough that
I would like to have a single .itb for all of them, and select the correct
device tree via FIT configurations (I will attempt this in RFCv2)
We also have infrastructure to support EFI boot which may be another route
we could take.

These patches (like the hardware) are a work in progress, but any comments
and suggestions are appreciated, especially as prototypes for the next two boards
will have a wider distribution.

Given that target/linux/layerscape was recently updated to 4.9, I think now is a good time to raise this so the smaller patches can get included, and I can gather feedback in case anyone suggests a better solution (this is a 'greenfield' platform - so we aren't constrained by what is out in the field already)

Of particular interest is layerscape/image/Makefile, as our set up is very
different to the reference boards (ubifs images etc.), and I have had a bit of difficulty
making it all work (i.e separate squashfs and ubifs kernel itbs are generated unnecessarily)

Major differences to note with existing layerscape boards (i.e LS1043ARDB):
	Traverse boards do not have NOR flash, and boot (load RCW+u-boot) from NAND by default.
	(SD/MMC booting is also possible, but we only recommend for development or unbrick)
	We use ubifs for our NAND partitioning.
	We are developing on u-boot mainline, not the NXP LSDK tree (for access to bleeding-edge features)
	We don't plan on supporting 32bit kernel/boot on our boards, YMMV

Known issues:
- mac80211 does not compile - NXP's patchset includes a function change backport from mainline, causing a compile error - will create an issue for this.

More information on the boards:
https://traverse.com.au/products/five64-arm64-platform/
https://gitlab.com/traverse-nxp-qoriq


mcbridematt:
layerscape: Remove netfilter and LZ4 config symbols has been separated into #1427 

I have updated my own branch "traverse-layerscape-wip-v1e" to reduce the number of commits involved, will resubmit when #1427 is finished.

PR #1414, opened 2017-10-13
at91 config options, laird modules, and associated ath6kl wifi packages

BWhitten:
I've had these hanging around for a bit and would be good to get these upstream so I submit them for consideration.

Firstly we refresh the kernel patches then add UBI block support and DMAs.
UBI block support in the kernel is required to be able to mount the ro rootfs portion of the squashfs images that are generated. DMAs are just handy for the speed of the system.

The UBIFS and UBINIZE parameters whilst work as globals would rapidly become unwieldy with additional devices, best just convert them to a default 'common' config per subtarget and adjust per board if necessary. This would be worth a review from @SandeepSM to ensure that the Xplained boards have the correct memory options.

The laird modules include Qualcomm Atheros 600X radios which is missing from the packages. Whilst they select other potentially unneeded packages, being FullMAC chips, I don't know how to extract them without surgery so they will just be unused.

Then finally add the two laird modules one based on the legacy chips from atmel the other their slightly newer offerings.
The DTSs have come from quite an old branch and likely could do with a review as some of the nodes may have been deprecated in newer kernels. They also probably need a more liberal usage of labels instead of re-defining when fixing up options.
The reason for splitting the 50 into an include and a top level is that the 50 include details the module itself whilst the top level is the development platform it is evaluated with. Same treatment could be applied to the 45.

Thanks!


hauke:
Thank you for the patches they were applied to master.

PR #1415, opened 2017-10-16
[refresh of PR #686] add powerpc64 support

commodo:
From the initial PR #686, I've only removed the addition of the ppc64 subtarget to the `mpc85xx` target.
I've tested this successfully on a T4240 variant ; it booted via a initiramfs image, and I could run a simple `udhcpc -i <iface>`.

Since the initial spin of PR #686, musl added BE support, but with elfv2 ABI only:
https://git.musl-libc.org/cgit/musl/commit/?id=5f3b652afe423dd2bd6f266535f79f685909cf60

Getting that in LEDE master will take a bit more effort: backporting and testing some those patches.
I've started that effort, and it will take me a few weeks to finish.
If anyone has a similar one, please shout-out.

Signed-off-by: Florian Larysch <fl@n621.de>

### DIFF
--- a/include/kernel.mk
+++ b/include/kernel.mk
@@ -85,6 +85,8 @@ else ifneq (,$(findstring $(ARCH) , armeb ))
   LINUX_KARCH := arm
 else ifneq (,$(findstring $(ARCH) , mipsel mips64 mips64el ))
   LINUX_KARCH := mips
+else ifneq (,$(findstring $(ARCH) , powerpc64 ))
+  LINUX_KARCH := powerpc
 else ifneq (,$(findstring $(ARCH) , sh2 sh3 sh4 ))
   LINUX_KARCH := sh
 else ifneq (,$(findstring $(ARCH) , i386 x86_64 ))
--- a/include/site/powerpc64
+++ b/include/site/powerpc64
@@ -1,0 +1,26 @@
+#!/bin/sh
+. $TOPDIR/include/site/linux
+ac_cv_c_littleendian=${ac_cv_c_littleendian=no}
+ac_cv_c_bigendian=${ac_cv_c_bigendian=yes}
+
+ac_cv_sizeof_char=${ac_cv_sizeof_char=1}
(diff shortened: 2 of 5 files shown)


commodo:
@nbd168

diizzyy:
Is there any hardware that makes sense to run LEDE on at all?  I doubt the OpenPOWER systems makes any sense in that regard ;-)

bobafetthotmail:
> Is there any hardware that makes sense to run LEDE on at all?

He mentioned "a T4240 variant", that SoC is a high-end network device SoC https://www.nxp.com/products/microcontrollers-and-processors/power-architecture-processors/qoriq-platforms/t-series/qoriq-t4240-t4160-t4080-multicore-communications-processors:T4240

I think these things are found in commercial router/firewall appliances, which are within most company's reach at least.

commodo:
yep; we run LEDE [ with our package-set + patch-set ] on top of a T4240 variant;
This patch-set will run on a T4240 RDB.
https://www.nxp.com/docs/en/fact-sheet/T4240RDBFS.pdf
I actually built an image for a T4240 RDB and tested it on our variant [since it does not differ all that much] ; and I was lazy to spin-up the RDB.

There's T1,T2 & T4 series, which replace the older P-series [ P1, P2, etc ].
LEDE/OpenWrt's mpc85xx target is actually supporting the P series ; there was a subtarget called P1020.
We also have P1020 variant and we also have P2020 variant.
The P series was only 32 bit [ AFAIK ]

The official reference build-system for the T2/T4 series that NXP is/was Yocto:
https://www.yoctoproject.org/downloads/bsps/dizzy171/ppce6500
[ what are the odds that `dizzy` has relation to `diizzyy` ? ]

Personally, I did not like the Yocto env much ; I find it too permissive/flexible which allows for some messy-ness.
Similar to Perl: too flexible: you can write nice code, but a lot of people choose not to.

commodo:
@nbd168 
ping

nbd168:
I've cleaned up the site file commit, please test if the updated version still works.

commodo:
will try it ;
thanks :)

PR #1417, opened 2017-10-16
mpc85xx: Add Aerohive HiveAP-330 to Default Images

riptidewave93:
This is needed to fix the snapshot builds as they are not generating the factory image for this device by default.

### DIFF
--- a/target/linux/mpc85xx/image/Makefile
+++ b/target/linux/mpc85xx/image/Makefile
@@ -85,6 +85,7 @@ define Image/Build/Profile/TLWDR4900
 endef
 
 define Image/Build/Profile/Default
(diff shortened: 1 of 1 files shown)


mkresin:
Hey Chris,

IMHO the mpc85xx image build code should be migrated to the new style as already used in/for most targets. It makes the image build way better to read an therefore less error prone.

riptidewave93:
@mkresin I have plans to re-write the Makefile once I get time. This is more of a "hold over" patch until that work is complete so snapshots get built for this new device.

riptidewave93:
RFC for the rewrite is at https://github.com/lede-project/source/pull/1429, but I would prefer to get this merged in first to get snapshot builds going.

mkresin:
Superseded by #1429

PR #1427, opened 2017-10-19
layerscape: Fix CONFIG_ALL_KMODS / mac80211 build

mcbridematt:
This pull request solves CONFIG_ALL_KMODS not working on the layerscape target, after mac80211 was updated to backports-4.14-rc2

1)  LZ4, netfilter and netlink options were compiled into the kernel breaking the standalone kmod- packages

2) The NXP LSDK kernel backported a change for ndo_get_stats64 from mainline ("net: make ndo_get_stats64 a void function"), this causes the mac80211 build to fail as backports still expects the old definition against 4.9:

```
backports-2017-10-06/net/mac80211/iface.c:1224:21: error: initialization from incompatible pointer type
error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
  .ndo_get_stats64 = bp_ieee80211_get_stats64,
```
In iface.c:1224 
```
#if LINUX_VERSION_IS_GEQ(4,11,0)
>---.ndo_get_stats64>---= ieee80211_get_stats64,
#else
>---.ndo_get_stats64 = bp_ieee80211_get_stats64,
#endif
```

To resolve this, revert the backport (this greatly reduces the content of patch 601) and create a shim for drivers that depend on this function (such as fsl-dpaa2)


target/linux/layerscape/patches-4.9/601-net-support-layerscape.patch
```diff
@@ -2223,6 +447,15 @@ Signed-off-by: Yangbo Lu <yangbo.lu@nxp.com>
  /* Make sure a field is enclosed inside headers_start/headers_end section */
  #define CHECK_SKB_FIELD(field) \
  	BUILD_BUG_ON(offsetof(struct sk_buff, field) <		\
+@@ -849,7 +875,7 @@ EXPORT_SYMBOL(napi_consume_skb);
+ 	BUILD_BUG_ON(offsetof(struct sk_buff, field) >		\
+ 		     offsetof(struct sk_buff, headers_end));	\
+ 
+-static void __copy_skb_header(struct sk_buff *new, const struct sk_buff *old)
++void __copy_skb_header(struct sk_buff *new, const struct sk_buff *old)
```
nbd168: Why did you add this and the above change?

mcbridematt: It appears I have refreshed on the wrong skbuff.c, the original patchset does drop the static but to copy_skb_header only. I will rebuild the patch and submit again.

PR #1429, opened 2017-10-19
[RFC] mpc85xx: Move to new build code style

riptidewave93:
The following moves the mpc85xx target (generic & P1020) to the new
build code style.

Note that this current form is an RFC for feedback/review & improvement.

To-Do:
 * Fix image generation for the TL-WDR4900

@mkresin


target/linux/mpc85xx/image/Makefile
```diff
@@ -1,107 +1,92 @@
...
+define Device/TLWDR4900
+  DEVICE_TITLE := TP-Link TL-WDR4900
+  DEVICE_PACKAGES := kmod-ath9k wpad-mini
+  DEVICE_PROFILE := tl-wdr4900-v1
+  DEVICE_DTS := $$(DEVICE_PROFILE)
+  BLOCKSIZE := 128k
+  TPLINK_HWID := 0x49000001
+  TPLINK_HWREV := 1
+  TPLINK_LAYOUTID := 16Mppc
+  KERNEL := dtb | copy-file $(LINUX_DIR)/arch/$(LINUX_KARCH)/boot/cuImage.$$(DEVICE_PROFILE)
```
riptidewave93: I originally was trying to get KERNEL_NAME working, but as cuImage is within the boot directory, it is never exported. The way the old style dealt with this was by using https://github.com/lede-project/source/blob/master/target/linux/mpc85xx/Makefile#L27 and https://github.com/lede-project/source/blob/master/include/kernel-defaults.mk#L126. If anyone has a good way to deal with this besides the method above, please let me know.

target/linux/mpc85xx/image/Makefile
```diff
@@ -1,107 +1,92 @@
...
+  KERNEL := dtb | kernel-bin | gzip | uImage gzip
 endef
 
-define factoryname
-$(call imgname,$(1),$(2))-factory.bin
-endef
-
-zImage:=$(BIN_DIR)/$(IMG_PREFIX)-zImage
-
-DTS_TARGETS = hiveap-330 fsl/p1010rdb-pa tl-wdr4900-v1 fsl/p1020rdb
```
riptidewave93: While this may look like I removed support for p1010rdb-pa and p1020rdb, these boards were never actually supported within the code for this target besides the dtb generation. Thus, they were not moved over to the new style.

mkresin:
Thanks! Merged to my staging tree,

PR #1431, opened 2017-10-20
toolchain: Test for supported versions of GCC

diizzyy:
Only test for supported versions of GCC
The version bump requirement for GCC is because gdb doesn't build with older
versions.

### DIFF
--- a/include/prereq-build.mk
+++ b/include/prereq-build.mk
@@ -27,32 +27,36 @@ $(eval $(call TestHostCommand,proper-umask, \
 	umask | grep -xE 00[012][012]))
 
 $(eval $(call SetupHostCommand,gcc, \
-	Please install the GNU C Compiler (gcc), \
-	$(CC) --version | grep gcc, \
-	gcc --version | grep gcc, \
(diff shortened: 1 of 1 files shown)


diizzyy:
Hold on... fixing it...

hauke:
Why do you require make 4.0 now and fail for version 3.81 for example?

hauke:
Please add to the commit comment that gdb now needs at least GCC 4.8

diizzyy:
@hauke 
My reasoning is that the "oldest" tested platform is Debian "jessy" which ships with make 4.0 (8.1+) and this is what the LEDE buildbots at least used to run (I think they run Debian 9 now). As there were specific comments regarding 3.81 being buggy and for all I know it hasn't been tested for a long time I figured it would be a good idea to bump the requirement in order to avoid unnecessary bugs reports etc.

diizzyy:
@hauke 
Removed make version check per your request
Good to go?

hauke:
Thank you for the patch it was added to master.

PR #1433, opened 2017-10-20
hostapd: ssid2 and utf8 support

BigNerd95:
SSID option now can be interpreted as:
- utf8 string
- double quoted string
- hexdump
- printf-escaped string

[Source](https://w1.fi/cgit/hostap/tree/hostapd/hostapd.conf#n87)

`ssid_type` specify the ssid format (default 0)
- 0: `ssid` is iterpreted as ssid (default)
- 1: `ssid` is iterpreted as ssid2 double quoted string
- 2: `ssid` is iterpreted as ssid2 hexdump
- 3: `ssid` is iterpreted as ssid2 printf-escaped string

If you use `option ssid_type '0'` you can specify:
`option ssid_utf8 '1'` to support UTF8 

For example now you can use an SSID with emoticons:
```
option ssid_type '3'
option ssid '\xF0\x9F\x98\x8DLEDE\xF0\x9F\x98\x8D'
```
[Preview](https://i.imgur.com/GC48SMy.png)

### DIFF
--- a/package/network/services/hostapd/files/hostapd.sh
+++ b/package/network/services/hostapd/files/hostapd.sh
@@ -141,7 +141,7 @@ EOF
 
 hostapd_common_add_bss_config() {
 	config_add_string 'bssid:macaddr' 'ssid:string'
-	config_add_boolean wds wmm uapsd hidden
+	config_add_boolean wds wmm uapsd hidden ssid_utf8
 	config_add_int ssid_type
 
 	config_add_int maxassoc max_inactivity
@@ -382,6 +382,10 @@ hostapd_set_bss_options() {
 	case "$ssid_type" in
 		0)
 			append bss_conf "ssid=$ssid" "$N"
+
+			json_get_vars ssid_utf8
+			set_default ssid_utf8 0
+			append bss_conf "utf8_ssid=$ssid_utf8" "$N"
 		;;
 		1)
 			append bss_conf "ssid2=\"$ssid\"" "$N"
--- a/package/network/services/hostapd/files/hostapd.sh
+++ b/package/network/services/hostapd/files/hostapd.sh
@@ -142,6 +142,7 @@ EOF
 hostapd_common_add_bss_config() {
 	config_add_string 'bssid:macaddr' 'ssid:string'
 	config_add_boolean wds wmm uapsd hidden
+	config_add_int ssid_type
 
 	config_add_int maxassoc max_inactivity
 	config_add_boolean disassoc_low_ack isolate short_preamble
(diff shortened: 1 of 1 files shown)


mkresin:
UTF-8 SSIDs (including emojis) is already supported as I proved in [FS#1045](https://bugs.lede-project.org/index.php?do=details&task_id=1045). You can set them via LuCI and SSH if your ssh client uses UTF-8 encoding.

The commit message lacks an explanation why double quoted string, hexdump or printf-escaped string is required or what issues are fixed by supporting them.

BigNerd95:
I'm not able to set UTF8 chars when using `vi` in SSH.
My SSH client uses UTF8 encoding, infact if I try to write a file with `echo 😘 > /tmp/test` it works.
But when I open `vi /etc/config/wireless` and I try to write an emoticon, it writes `....`

Using ssid2 I'm able to write unicode chars also if my ssh client doesn't support UTF8 encoding.

Then, why ssid supports UFT8 chars also without utf8_ssid option enabled? It's not strange?

BigNerd95:
Well, just discovered it's only a `vi` problem when showing unicode chars.
If I do `cat /etc/config/wireless` I can see emoticons.

mkresin:
>  But when I open vi /etc/config/wireless and I try to write an emoticon, it writes ....

It rather seem to me you are trying to fix the wrong package. Smells like you issue is vi.

BigNerd95:
Yes, you are right
[vi is not Unicode-capable](https://github.com/lede-project/source/blob/master/package/utils/busybox/config/editors/Config.in#L102)

So the only way to set an SSID with unicode is using uci or luci.
But I still don't understand why it supports utf8 without `utf8_ssid` flag.

mkresin:
> Yes, you are right vi is not Unicode-capable

Even if vi shows gibberish, the UTF8 SSID is used and broadcasted.

> But I still don't understand why it supports utf8 without utf8_ssid flag.

Wild guess, 'utf8_ssid' is required to interpret hex sequences like \xF0\x9F\x98\x8 as UTF8 and not to broadcast exactly the string  \xF0\x9F\x98\x8.

Anyway, would you please close the PR if vi was/is your issue. AS far as I can see, there is nothing that need to be fixed in hostapd/wireless config.

BigNerd95:
> Wild guess, 'utf8_ssid' is required to interpret hex sequences like \xF0\x9F\x98\x8 as UTF8 and not to broadcast exactly the string \xF0\x9F\x98\x8.

`\xF0\x9F\x98\x8D` is broadcasted with both `utf8_ssid '0'` and `utf8_ssid '1'`

PR #1434, opened 2017-10-20
mpc85xx: Update HiveAP-330 initrd partition

riptidewave93:
initramfs is not the proper name for this, as it stores a boot ramdisk and not a filesystem. Update the name to reflect it's usage correctly.

@mkresin


mkresin:
Thanks! Merged to my staging tree,

PR #1435, opened 2017-10-20
ar71xx: fix network setup for UniFi

p-wassi:
Compile tested: n/a
Run tested: Ubiquiti UniFi (LR)

For the UniFi access points, the board.d/02_network has no distinct case,
so execution falls through to the default case, which is eth0=lan, eth1=wan.
Since the UniFi devices just have eth0, we end up with a network configuration
file that references a non-existent interface wan@eth1.

Just add UniFi to the list of devices that only have eth0 for lan.


pepe2k:
@p-wassi,

Cheers,
Piotr

PR #1436, opened 2017-10-21
ar71xx: Add support for TP-LINK Archer C7 v4 RU

Adonai:
This adds another line to the list of supported versions.
Special ID was extracted from TP-Link firmware for RU version of this
router.

With it firmware file is successfully recognized by router after TFTP upload.
Unfortunately I haven't tested upload from stock Web UI.

@nbd168 can you please take a look at this?


pepe2k:
Hello @Adonai,

I pulled into my staging tree different patch, with more IDs for this board (the one added by you is also there): https://github.com/lede-project/source/pull/1490

Cheers,
Piotr

Adonai:
@pepe2k, good, thanks!

PR #1437, opened 2017-10-21
ramips: fix Youku-YK1 support

RoEdAl:
- Remove the ephy-pins from the ethernet device tree node. The ephy-pins are useed to controll the ePHY LEDs and this board doesn't have these. Instead one of the ePHY pins is used in GPIO mode to control the WAN LED.
- Use the switch LED trigger to control the WAN LED. Move the power LED handling to diag.sh to show the boot status via this LED.
- Add the missing kernel packages for USB and microSD card reader to the default package selection.
- Fix the maximum image saize value. The board has a 32MByte flash chip.

Fixes: FS#1055

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -461,9 +461,9 @@ zte-q7)
 	set_wifi_led "$board:blue:status"
 	;;
 youku-yk1)
-	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
 	set_wifi_led "$board:blue:air"
 	set_usb_led "$board:blue:usb"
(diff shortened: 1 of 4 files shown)


mkresin:
Hey Edmunt,

I merged your PR to my [staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary) .

I reworded your commit message in way I would like to see it with your next patch. I added an explaination for the image size change as well.

Furthermore, I cleaned up the pinmux. Would you please boot an image build from staging tree to make sure I don't introduce new issues with the change.

Please ACK the rewrite of your commit message.

RoEdAl:
I confirm that build from staging tree works correctly on Youku-YK1 router. I checked ethernet, leds, USB, microSD card reader and everything seems to work properly.

mkresin:
merged, thanks!

PR #1438, opened 2017-10-21
ramips: fix default usb support for nexx wt3020-8M

bobafetthotmail:
the nexx wt3020-8M has a usb 2.0 port, add usb 2.0 support packages to its default package list.

I discovered the hard way that this device does not have these packages installed by default https://github.com/openwrt/packages/issues/4984#issuecomment-338434466

Might be a good idea to cherry-pick this for LEDE release too.

Signed-off-by: Alberto Bursi alberto.bursi@outlook.it

### DIFF
--- a/target/linux/ramips/image/mt7620.mk
+++ b/target/linux/ramips/image/mt7620.mk
@@ -485,6 +485,7 @@ define Device/wt3020-8M
   IMAGE/factory.bin := $$(sysupgrade_bin) | check-size $$$$(IMAGE_SIZE) | \
 	poray-header -B WT3020 -F 8M
   DEVICE_TITLE := Nexx WT3020 (8MB)
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci
 endef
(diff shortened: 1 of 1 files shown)


mkresin:
Shouldn't we add the same packages to the wt3020-4M as well or gets the image to big with the usb packages added?

mkresin:
Please fix the e-mail address in your SoB. It misses `<>`.

Please word wrap your commit message:

- commit description must have less than 75 characters per line

bobafetthotmail:
updated.

> Shouldn't we add the same packages to the wt3020-4M as well or gets the image to big with the usb packages added?

That device does not have a USB port (has the same board, but USB is not populated in the 4M version), so I'd say no.

mkresin:
> That device does not have a USB port (has the same board, but USB port is not soldered in the 4M version, also the casing lacks the opening for the port), so I'd say no.

Thanks a lot for this extra information. I staged a commit which disables USB for the -4M and added you as reporter.

I've merged your PR to my staging tree as well.

PR #1439, opened 2017-10-21
mtd-utils: backport fix for glibc version 2.26+

gitbugged:
Fixes: FS#1015 - mtd-utils-1.5.2 does not compile due to missing sysmacros.h 

Signed-off-by: Bob Brooks <gitbugged@cool.fr.nf>

### DIFF
--- a/package/utils/mtd-utils/patches/020-glibc-compat.patch
+++ b/package/utils/mtd-utils/patches/020-glibc-compat.patch
@@ -1,0 +1,32 @@
+mtd-utils: backport fix for glibc version 2.26+
+
+Include sys/sysmacros.h for major/minor/makedev. These functions have
+always been defined in sys/sysmacros.h under Linux C libraries.  For
+some, including sys/types.h implicitly includes that as well, but glibc
+wants to deprecate that, and some others already have.  Include the
+header explicitly for the funcs. Credit Mike Frysinger <vapier@gentoo.org>
+for creating the original patch for this bug.
+
+Fixes: FS#1015
+
+Signed-off-by: Bob Brooks <gitbugged@cool.fr.nf>
+---
+ include/common.h | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/include/common.h b/include/common.h
+index fb0ca83..8cb3142 100644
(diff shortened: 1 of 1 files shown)


mkresin:
All your sysmacro patches have styling issues:

- all our patches have a numeric prefix
- the commit title need to have a prefix matching the package that is patched

A better naming for the patch would be something like 020-glibc-compat.patch. Some of the issues are already mentioned in our [submitting patches](https://lede-project.org/submitting-patches) guideline.

The commit message for this commit should be something like this:

```
mtd-utils: backport fix for recent glibc versions

Include sys/sysmacros.h for major/minor/makedev. These functions have
always been defined in sys/sysmacros.h under Linux C libraries.  For
some, including sys/types.h implicitly includes that as well, but glibc
wants to deprecate that, and some others already have.  Include the
header explicitly for the funcs.

Fixes: FS#1015

Signed-off-by: Bob Brooks <gitbugged@cool.fr.nf>
```
If it isn't a backport, the commit title should be "[package]: add ...".

It would be interesting to know with which glibc version the issues occurs. Is it related to the latest update of glibc in LEDE to 2.26?

Please use ```git commit --amend``` to add changes to your commit. Use ```git push --force``` to overwrite your github repository/branch. The PR will afterwards update automatically. Please don't close this PR and open a new one.

gitbugged:
Thank you, I will work to get this resolved asap.

gitbugged:
> It would be interesting to know with which glibc version the issues occurs. Is it related to the latest update of glibc in LEDE to 2.26?

I've caught this bug on any host that is using glibc 2.26>. Currently Arch Linux fails to compile without the patch.

Please review updated patch and advise if further editing is needed. I read that the commit subject "must be less than 50 characters long", so the details are included in the patch file itself.

mkresin:
I'm still now sure if this patch is really required. You're patching the packages/mtd-utils instead of tools/mtd-uitls.

Would you please try to reproduce the issue with the commits I pushed a few minutes ago. If it's fixed, please close the PR.

gitbugged:
Seems this PR is not needed. Thanks.

PR #1443, opened 2017-10-22
 uboot-ar71xx: backport fix for glibc version 2.26+

gitbugged:
Fixes  FS#1047 - u-boot-nbg460n_550n_550nh bails out on CONFIG_ENV_SIZE error 

Signed-off-by: Bob Brooks <gitbugged@cool.fr.nf>

### DIFF
--- a/package/boot/uboot-ar71xx/Makefile
+++ b/package/boot/uboot-ar71xx/Makefile
@@ -1,8 +1,7 @@
 #
 # Copyright (C) 2010 OpenWrt.org
 #
-# This is free software, licensed under the GNU General Public License 
-v2.
+# This is free software, licensed under the GNU General Public License v2.
 # See /LICENSE for more information.
 #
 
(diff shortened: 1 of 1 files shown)


gitbugged:
I'm not sure this is the "best" fix for this issue, but I followed instructions on FS#1047, and this was the working end-result.

eamaclean:
This has nothing to do with glibc, and just seems to break building uboot-ar71xx entirely.

This is a conflict with a library called ragel (which I'll bet you have), u-boot includes a file called config.h, and expects it to be in the search path, and ragel happens to provide such a file earlier in the path than u-boot's own.

I'm not entirely sure what the correct resolution for this is, other than defensive programming lessons for both parties..

mkresin:
Yes, it is completely unrelated to glibc. I'm able to trigger the compile error if I add a file `/usr/include/config.h` on my system and your change doesn't look like the appropiate solution for the issue.

mkresin:
@gitbugged any plans to come up with a proper solution?

As outlined, it is unrelated to glibc. Reverting 83783f296a18e6e481f23b2019575e45c437db4e is more like using a hammer than a proper fix. It will (re)introduce the issue of u-boot isn't build with the images by default which was fixed by the commit.

gitbugged:
No idea how to fix this unfortunately. Closing.

PR #1444, opened 2017-10-22
base-files: add 'interval' option to netdev LED trigger configuration

RoEdAl:


mkresin:
Please bump the PKG_RELEASE in package/base-files/Makefile as well.

RoEdAl:
Done.

package/base-files/files/bin/config_generate
```diff
@@ -318,6 +318,7 @@ generate_led() {
 			uci -q batch <<-EOF
 				set system.$cfg.trigger='netdev'
 				set system.$cfg.mode='$mode'
+				set system.$cfg.interval='50'
```
mkresin: Please drop this line. 50 is the default value anyway and I don't see the point in adding a fixed value here.

package/base-files/files/etc/init.d/led
```diff
@@ -72,6 +72,8 @@ load_led() {
 			[ -n "$dev" ] && {
 				echo $dev > /sys/class/leds/${sysfs}/device_name
 				echo $mode > /sys/class/leds/${sysfs}/mode
+				[ -n "$interval" ] && \
```
mkresin: Please remove the `[ -n "$interval" ] && \` condition. Due to the `config_get interval $1 interval "50"` at the beginning of the the `load_led()` function, `$interval` is set to 50 if not defined,

RoEdAl:
Simplified.

mkresin:

PR #1445, opened 2017-10-22
Make network.*.ip6assign default to 64

toreanderson:
The IPv6 Adressing Architecture (RFC 4291) mandates that subnet prefixes
must be 64 bit long. More discussion about this requirement is found in RFC
7421.

This updates the default value of network.*.ip6assign correspondingly.

Signed-off-by: Tore Anderson <tore@fud.no>

### DIFF
--- a/package/base-files/files/bin/config_generate
+++ b/package/base-files/files/bin/config_generate
@@ -101,7 +101,7 @@ generate_network() {
 				set network.$1.proto='static'
 				set network.$1.ipaddr='$ipad'
 				set network.$1.netmask='$netm'
-				set network.$1.ip6assign='60'
+				set network.$1.ip6assign='64'
 			EOF
 		;;
 
--- a/target/linux/adm5120/base-files/etc/config/network
+++ b/target/linux/adm5120/base-files/etc/config/network
@@ -19,7 +19,7 @@ config interface lan
 	option proto	static
 	option ipaddr	192.168.1.1
 	option netmask	255.255.255.0
-	option ip6assign 60
+	option ip6assign 64
 
 
 #### WAN configuration
--- a/target/linux/adm8668/base-files/etc/config/network
+++ b/target/linux/adm8668/base-files/etc/config/network
@@ -10,7 +10,7 @@ config interface lan
 	option proto	static
 	option ipaddr	192.168.1.1
 	option netmask	255.255.255.0
-	option ip6assign 60
+	option ip6assign 64
 
 config interface wan
 	option ifname	eth1
--- a/target/linux/rb532/base-files/etc/config/network
+++ b/target/linux/rb532/base-files/etc/config/network
@@ -16,7 +16,7 @@ config interface lan
 	option proto	static
 	option ipaddr	192.168.1.1
 	option netmask	255.255.255.0
-	option ip6assign 60
+	option ip6assign 64
 
 config interface wan6
 	option ifname   eth0
--- a/target/linux/zynq/base-files/etc/config/network
+++ b/target/linux/zynq/base-files/etc/config/network
@@ -11,7 +11,7 @@ config interface lan
 	option proto	static
 	option ipaddr	192.168.1.1
 	option netmask	255.255.255.0
-	option ip6assign 60
+	option ip6assign 64
 
 config globals globals
 	option ula_prefix auto


mkresin:
NAK.

Your interface should have a /64 address anyway. This option is used to specify the prefix length that can be/is delegated to to downstream routers connected to this network (search for DHCPv6-PD for more details).

Do you have a particular issue that get fixed by this change?

toreanderson:
> Your interface should have a /64 address anyway.

It doesn't - that's precicely the bug this patch fixes.

With the current default `network.lan.ip6assign=60`, `br-lan` gets configured with a /60, **not** a /64.

This is a direct violation of [RFC 7084](https://tools.ietf.org/html/rfc7084#section-4.3), by the way:

>   L-2:   The IPv6 CE router MUST assign a separate /64 from its
>          delegated prefix(es) (and ULA prefix if configured to provide
>          ULA addressing) for each of its LAN interfaces.

jow-:
Yeah but the ip6assign refers to the delegated prefix, not to the "separate /64 from its
delegated prefix(es)"

dedeckeh:
On purpose ip6assign has been changed from 64 into 60 in commit https://github.com/lede-project/source/commit/4e8f7f37edc7c0e1088d002cbb16f76509f4151c#diff-ffb41a566f05204a44f4b34e198b0e6d to support (cascaded) DHCPv6 prefix delegation.
We want to keep out of the box support for DHCPv6 prefix delegation; therefore it does not make sense to switch this back to 64

toreanderson:
I'll try to be a bit more clear: the issue I am trying to fix has really nothing to do with (cascaded) DHCPv6 Prefix Delegation. It is regarding the addresses that are assigned directly to the network interface in question, which the IPv6 protocol standards unequivocally require to be a `/64`, however I observed that this happens instead:

```console
root@x:~# uci get network.lan.ip6assign
60
root@x:~# ip -6 address show scope global dev br-lan
5: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 2a02:fe0:c420:3300::1/60 scope global noprefixroute dynamic 
       valid_lft 3556sec preferred_lft 3556sec
    inet6 fd1d:6d1b:e37::1/60 scope global noprefixroute 
       valid_lft forever preferred_lft forever
```

As you can see, with `network.lan.ip6assign=60`, what is being assigned directly to the interface is `/60`. Although it works just fine, it is a violation of the IPv6 protocol standards.

With after setting `network.lan.ip6assign=64` and rebooting, it looks better and aligned with :

```console
root@x:~# uci get network.lan.ip6assign
64
root@x:~# ip -6 address show scope global dev br-lan
5: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 2a02:fe0:c420:3300::1/64 scope global noprefixroute dynamic 
       valid_lft 1076721sec preferred_lft 471921sec
    inet6 fd1d:6d1b:e37::1/64 scope global noprefixroute 
       valid_lft forever preferred_lft forever
```

zorun:
I can confirm that the LAN is configured with a `/60`:

```
root@LEDE:~# ip a
5: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP qlen 1000
    inet6 2001:db8:18ff:ff00::1/60 scope global dynamic 
       valid_lft 325sec preferred_lft 225sec
    inet6 fdfe:9f22:b7a8::1/60 scope global 
       valid_lft forever preferred_lft forever
```

However, strangely enough, the device route has the correct `/64` (second line):

```
root@LEDE:~# ip -6 r
default from 2001:db8:18ff:ff00::/56 via fe80::20d:b9ff:fe46:66c5 dev eth1  metric 4096 
2001:db8:18ff:ff00::/64 dev br-lan  metric 1024 
unreachable 2001:db8:18ff:ff00::/56 dev lo  metric 2147483647  error -148
fdfe:9f22:b7a8::/64 dev br-lan  metric 1024 
unreachable fdfe:9f22:b7a8::/48 dev lo  metric 2147483647  error -148
fe80::/64 dev eth1  metric 256 
fe80::/64 dev br-lan  metric 256 
fe80::/64 dev wlan0  metric 256 
unreachable default dev lo  metric -1  error -128
ff00::/8 dev br-lan  metric 256 
ff00::/8 dev eth1  metric 256 
ff00::/8 dev wlan0  metric 256 
unreachable default dev lo  metric -1  error -128
```

```
root@LEDE:~# ifstatus wan6
	"ipv6-prefix": [
		{
			"address": "2001:db8:18ff:ff00::",
			"mask": 56,
			"preferred": 263,
			"valid": 363,
			"class": "wan6",
			"assigned": {
				"lan": {
					"address": "2001:db8:18ff:ff00::",
					"mask": 60
				}
			}
		}
	],
```

mkresin:
Even if it violates the mentioned RFCs, it can't be done in the suggested way if you want to have prefix delegation to downstream interfaces enabled (which was enabled on purpose as outlined by Hans). The interface need to be in the range that it will delegate.

To speak in IPv4. Using a DHCPv4 pool of 192.168.1.0/24 with a default gateway of 192.168.2.1 will not work either.

toreanderson:
> The interface need to be in the range that it will delegate.

That is simply not the case. DHCPv6 can, and usually does, use link-local IPv6 addresses. In fact, having global addresses on the link between the DHCPv6 server and the DHCPv6 client is completely optional.

My previous ISP had exactly this set-up, there was only link-local addresses on the WAN link. The home gateway learned the next hop of the default route using standard ICMPv6 Router Advertisements, while the ISP's access router would use the link local address of the DHCPv6 client as the next-hop of the delegated prefix.

My current ISP does have global addresses on the WAN link, but this is a /64 and has in any case nothing to do with prefix delegation - this uses only link-local addresses.

dedeckeh:
@toreanderson I think you misinterpreted the comments by Mathias; he tried to explain that you cannot assign 192.168.2.1 out of 192.168.1.0/24 range. Similar if an IPv6 prefix of 64 is assigned to the downstream lan interface it gets impossible to support the cascaded prefix delegation use case as there're no subprefixes left anymore;  this will trigger new bug reports by people using such scenarios as it will be seen as a regression issue

toreanderson:
@dedeckeh The IPv4 analogy is not applicable here, because IPv6 is very different from IPv4 in this respect. For starters, IPv4 DHCP simply does not support Prefix Delegation to begin with.

Also, unlike IPv4 DHCP, DHCPv6 can not assign a default gateway. In IPv6, those are assigned by way of ICMPv6 Router Advertisements - completely independent of DHCPv6, and also completely independent of global prefixes being present on the link or not.

The DHCPv6 protocol, Prefix Delegation included, only requires link-local addresses to be present on the link for correct operation. There exists simply **no requirement whatsoever** that the link must have global addresses, and most definitely not prefixes shorter than /64.

[This random PCAP of a DHCPv6-PD exchange](https://github.com/aosp-mirror/platform_external_tcpdump/blob/master/tests/dhcpv6-ia-pd.pcap) demonstrates this. The resulting outcome from this exchange is that the router will have a route added à la `2a00:1:1:100::/56 via fe80::201:2ff:fe03:405`. The `/56` is _**delegated to**_ the client, i.e., it is used on one or more links _**behind**_ the client. It is **not** supposed to be used on the link that connects the server and the client.

I'll also note that when LEDE is working in Homenet mode, this actually works as expected (at least last I tested it). All LAN links are configured with individual /64s drawn from the aggregated prefix delegated from the ISP using DHCPv6-PD, while DHCPv6 clients on the internal interfaces can at the same time obtain delegations for separete prefixes (assuming there are unused ones left).

jow-:
Your fix still breaks DHCPv6 downstream delegation as the ip6assign value not only defines the prefix size applied to the interface but also the pool size used to draw delegated downstream prefixes from.

A proper fix would require changes to netifd and possibly odhcpd to decouple the interface prefix setting logic from the prefix pool size calculation.

Merely changing ip6assign to 64 might make the interface prefix RFC compliant but it will break existing configurations and thus introduce regressions.

toreanderson:
@jow- The way you describe it, `ip6assign` does indeed seem defective by design. That is, it conflates two different and completely orthogonal settings. In that case, I agree it should probable be fixed more properly.

Come to think of it, the entire knob seems rather pointless, actually. The «what prefix length to assign to local interfaces» meaning is an IPv6 protocol constant: `/64`. No need for a configuration setting for changing this.

Regarding the «what size pool to draw prefixes from» meaning, then that is really decided by the ISP. That is, if your ISP gives you a `/48` then you there's no reason why this entire `/48` couldn't be used for assigning prefixes to local interfaces and for delegating prefixes to downstream DHCPv6-PD clients. There's no reason to artificially limit this to a `/60` by default - at least none that I can think of.

NeoRaider:
I agree with @toreanderson here. While a fix might break some (weird) configurations, I've never seen a sane reason to assign anything bigger than a /64 to an interface. Always assigning /64 seems like the right thing to do.

jow-:
Sure, but that still needs fixes to the downstream prefix allocation logic. CHanging this to 64 like this very patch proposes will not only break weird configuration but all redelegations.

jow-:
> Regarding the «what size pool to draw prefixes from» meaning, then that is really decided by the ISP. That is, if your ISP gives you a /48 then you there's no reason why this entire /48 couldn't be used for assigning prefixes to local interfaces and for delegating prefixes to downstream DHCPv6-PD clients. 

The point of this setting is to be able to partition the total available prefix into certain chunks, e.g. divide an ISP-assigned /48 into /56 chunks and distribute them among downstream ifaces.

The settings also allows you to put on upper limit on the size of PD prefixes downstream clients can obtain, e.g. if you only want to delegate at most a /56 chunk of your /48.

The wrong prefix size applied to the interface (or the fact that a prefix is applied at all) can be seen as wrong side-effect of the ip6assign setting.

toreanderson:
The  main problem I see with that approach is that it won't work very well that ISPs that are stingy with the prefixes they hand out. The largest IPv6-enabled wireline ISP in the world, Comcast, only hands out a `/60` to each residential subscriber. As I understand it, the current LEDE default would be to burn that entire `/60` on the first LAN interface and its _potential_ downstream PD clients, leaving nothing available for use with any subsequent LAN interfaces and their PD clients.

PR #1446, opened 2017-10-22
ar71xx: add support for TP-Link WA901ND-v5

p-wassi:
Compile-tested: ar71xx
Run-tested: ar71xx/tl-wa901nd-v5

The TL-WA901ND-v5 has the same hardware as -v4, although the PCB
has a different layout. The TFTP installation method from v4 still applies,
however, the v4 image is not accepted (hardware ID).

For installation: setup TFTP server, that provides the factory image as
"wa901ndv4_tp_recovery.bin" (sic!) at IP 192.168.0.66.
Hold down the reset button and power up the device. After a few seconds
it will request the image and flash LEDE.


p-wassi:
Rebased on current master (Merge conflicts)

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -574,7 +574,16 @@ define Device/tl-wa901nd-v4
   TPLINK_HWID := 0x09010004
   IMAGE/factory.bin := append-rootfs | mktplinkfw factory -C EU
 endef
-TARGET_DEVICES += tl-wa901nd-v1 tl-wa901nd-v2 tl-wa901nd-v3 tl-wa901nd-v4
+
+define Device/tl-wa901nd-v5
```
pepe2k: Could you please simplify this definition (we both don't like code duplication)?
You can do something like:

```
define Device/tl-wa901nd-v5
  $(Device/tl-wa901nd-v4)
  ...here overwrite only different values...
endef
```

Cheers,
Piotr

p-wassi:
> Could you please simplify this definition (we both don't like code duplication)?
> You can do something like

Ah, nice :-) Done.

pepe2k:
@p-wassi,

Thanks again, merged into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1448, opened 2017-10-23
uboot-kirkwood: update to 2017.09

p-wassi:
Compile-tested: kirkwood
Run-tested: kirkwood/dockstar

Upgrade U-Boot for the Kirkwood target to upstream release 2017.09.

A lot of configuration options have been moved from defines in the header files
to Kconfig. Without catching up on this, code would not compile, as dependencies
are not resolved then. Further, we'd have many warnings for re-defines.

My updated logs from the run-test on a Dockstar are given in [1].

@bobafetthotmail Are you willing to test this on your NSA3xx devices, as I touched
rather large parts of your 007 and 008 patches?

[1]:
https://pwassi.privatedns.org/lede/dockstar/#log_uboot

### DIFF
--- a/package/boot/uboot-kirkwood/Makefile
+++ b/package/boot/uboot-kirkwood/Makefile
@@ -7,10 +7,10 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_VERSION:=2017.03
+PKG_VERSION:=2017.09
 PKG_RELEASE:=1
 
-PKG_HASH:=f54baf3f9325bf444c7905f3a5b6f83680edb1e6e1a4d5f8a5ad80abe885113f
+PKG_HASH:=b2d15f2cf5f72e706025cde73d67247c6da8cd35f7e10891eefe7d9095089744
 
 include $(INCLUDE_DIR)/u-boot.mk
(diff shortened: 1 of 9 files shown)


bobafetthotmail:
Works fine on nsa310 and nsa325. And also on a goflexnet I have laying around.

That said, I'll try upstreaming these patches in the next months so any further updates can be done painlessly. Thanks for your work on aligning them to upstream. :)

mkresin:

p-wassi:
Thanks @mkresin for merging.
What I was wondering: the U-Boot images are only built for snapshots (e.g. [1]).
Is is intention, that they're not built for released LEDE versions?
(Scroll down to the "Supplementary Files" section)

[1]:
https://downloads.lede-project.org/snapshots/targets/kirkwood/generic/

mkresin:
> Thanks @mkresin for merging.

Wasn't that difficult. You have searched for testers on your own and collected tested-bys.

>  Is is intention, that they're not built for released LEDE versions?

AFAIR, it is a bug which was fixed in master and affects all package/boot/uboot-* packages. No idea if the fix can be easily backported or requires a bunch of changes will may break stable.

PR #1449, opened 2017-10-23
build: Allow to get just a subset of packages from feeds

miska:
Sometimes it is useful to get just some subset of packages from feed. This
commit changes scripts/feeds in a way that you can specify as the last optional
argument which packages should be used from the feed.

Signed-off-by: Michal Hrusecky <michal.hrusecky@nic.cz>

### DIFF
--- a/scripts/feeds
+++ b/scripts/feeds
@@ -198,7 +198,7 @@ sub update_feed_via($$$$$$) {
 	}
 
 	if($filter) {
-		system("cd '$safepath'; find ./[a-zA-Z0-9]* -name Makefile -not -regex '" . $filter . "' -exec dirname \\{\\} \\; | xargs rm -rf");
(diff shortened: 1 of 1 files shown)


jow-:
Your approach looks quite dangerous and enables shell injection through the `find | xargs rm` command chain. Also the use of `basename` is unnescessary and can be replaced with a `-printf` directive.

Apart from that; what is the use case for this change? In which situations is this preferred over feeds update with selective install?

miska:
Fixed the find/xarags.

Regarding use-case, it doesn't do anything that can't be done manually. But if you have multiple feeds providing same packages and you want use just part of the feed (mainly thinking about programming languages), you would have to do it manually with long list of packages and document it somewhere if somebody is expected to be able to build the same image again. This way it is documented and easier to do.

blogic:

PR #1453, opened 2017-10-24
ar71xx: base-files: alphabetical reordering

p-wassi:
Re-order the cases of base-files/* alphabetically.
Actually, not the cases themselves are ordered, but the multiple options
in each single case.


p-wassi:
Now also merge some cases in etc/diag.sh

mkresin:
merged, thanks!

PR #1454, opened 2017-10-24
ar71xx: fix LED names for GL-AR150

p-wassi:
Compile-tested: n/a
Run-tested: ar71xx/gl-ar150

Add the respective colour to the LED's names for the GL-AR150 to be conform
to the kernel. Also add netdev triggers for the LAN and WAN LED.

NOTE: The other GL-Inet boards still have LED names without colour, but as I just have
the AR150 lying around, I can only test and add colour on this device.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -387,7 +387,11 @@ gl-mifi)
 	ucidef_set_led_netdev "lan" "LAN" "$board:green:lan" "eth1"
 	ucidef_set_led_netdev "3gnet" "3GNET" "$board:green:net" "3g-wan"
 	;;
-gl-ar150|\
+gl-ar150)
+	ucidef_set_led_wlan "wlan" "WLAN" "$board:orange:wlan" "phy0tpt"
+	ucidef_set_led_netdev "wan" "WAN" "$board:green:wan" "eth0"
+	ucidef_set_led_netdev "lan" "LAN" "$board:green:lan" "eth1"
+	;;
 gl-ar300)
 	ucidef_set_led_wlan "wlan" "WLAN" "$board:wlan" "phy0tpt"
 	;;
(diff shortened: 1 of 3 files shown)


mkresin:
Please add a migration scripts to update existing configs. Have a look at target/linux/ar71xx/base-files/etc/uci-defaults/04_led_migration.

p-wassi:
Hi @mkresin ; thanks for commenting.
I've added a migration for the WLAN LED, as the others (LAN+WAN) weren't used anyway
by default.
So when do we need migration and when can we skip this step?
In [1] the LEDs for the GL-AR150 were also renamed.

[1]:
https://git.lede-project.org/?p=source.git;a=blobdiff;f=target/linux/ar71xx/files/arch/mips/ath79/mach-gl-ar150.c;h=e0d1ea4381396b138a435f2f8b9e480a11787607;hp=310182cc8e9c22e2bfcb168996d4f01d117dc103;hb=c142df309420b153253ff739ec5511501878b5b0;hpb=8689078a9b9754ae6163488aaa73959714529b6d

mkresin:
>  So when do we need migration and when can we skip this step?

A migration script is required each time the name of a LEDs is changed. Consider the commit you link to as "missing migration script not noticed by committer".

> I've added a migration for the WLAN LED, as the others (LAN+WAN) weren't used anyway
by default.

If I rename something, I add a migration script even for LEDs not used by default. You never know which config the user has and not breaking a users config seem to me a good idea.

p-wassi:
> A migration script is required each time the name of a LEDs is changed.

So should we add a migration from _gl_ar150_ to _gl-ar150_ retrospective?

> a migration script even for LEDs not used by default

Ok, here we go :-)

mkresin:
> So should we add a migration from gl_ar150 to gl-ar150 retrospective?

Since we are already shipping a stable version including this change, I guess it wouldn't fix that many boards. But it is up to you.

p-wassi:
Ah, ok. Then I think, we'll just do the migration from the state of the first stable release.

PR #1455, opened 2017-10-24
 libiconv-full: fix compile-time linking error GCC7

staehle:
[LEDE Flyspray Task 1091](https://bugs.lede-project.org/index.php?do=details&task_id=1091)
Fix libiconv-full 'undefined reference' compile linker error using GCC7 Musl
Tested with targets x86 (i386 and x86_64)
Addition of CFLAGS "std=gnu89" fixes the linker issues, credit to harrylwc
Issue found with 'minidlna' package, which depends on 'libiconv-full'
Error in compile log:
../lib/.libs/libiconv.so: undefined reference to `aliases_lookup'
../lib/.libs/libiconv.so: undefined reference to `aliases2_lookup'
Makefile:64: recipe for target 'iconv_no_i18n' failed

Signed-off-by: Jake Staehle <jacob@staehle.us>

### DIFF
--- a/package/libs/libiconv-full/Makefile
+++ b/package/libs/libiconv-full/Makefile
@@ -51,7 +51,7 @@ define Package/iconv
   TITLE+= utility
 endef
 
-TARGET_CFLAGS += $(FPIC) -DUSE_DOS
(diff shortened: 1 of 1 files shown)


hauke:

PR #1456, opened 2017-10-24
ipq806x: fix Zyxel NBG6817 WiFi button

tolga9009:
Zyxel NBG6817 features a WiFi button, which becomes functional by setting
correct GPIO. It is a switch-type button, so it emits KEY_RFKILL on each ON
and OFF state. This is achieved by setting input-type to EV_SW.

Signed-off-by: Tolga Cakir <tolga@cevel.net>


mkresin:
merged, thanks!

PR #1458, opened 2017-10-25
ar71xx: fix LED config for DIR-869 A1

H4ndl3:
This fixes the LED configuration for the D-Link DIR-869 A1. The pins for
orange and white were swapped and one active_low was set wrong.

### DIFF
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-dir-869-a1.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-dir-869-a1.c
@@ -42,8 +42,8 @@
 
 #define DIR869A1_GPIO_ENABLE_SWITCH	11
 
-#define DIR869A1_GPIO_LED_POWER		15
-#define DIR869A1_GPIO_LED_ORANGE	16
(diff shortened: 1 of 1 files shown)


mkresin:
Would it be possible that you're a bit more verbose in your commit message. Especially with the comment you removed in mind, it isn't that easy to get what was wrong/which former made assumption was wrong.

To my understanding the DIR-869 A1 has either a dual colour LED or two status LEDs. Pulling GPIO#16 to low, lighten the white LED/the white colour of the dual colour LED. Pulling the GPIO to high switches the white LED/white colour off.

Same applies to GPIO#15 and the orange LED.

Both GPIOs pulled to high == LED off
Both GPIOs pulled to low == ?

Having such an explanation about the status led would make reviewing way easier.

H4ndl3:
@mkresin I updated the commit message in order to make the changes more clear.

mkresin:
Thanks a lot for updating the commit message. Makes reviewing way easier. Merged to [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary).

PR #1460, opened 2017-10-26
layerscape: fix CONFIG_ALL_KMODS / mac80211 build (v2)

mcbridematt:
Rebuilt version of #1427 

This pull request solves CONFIG_ALL_KMODS not working on the layerscape 64b target, after mac80211 was updated to backports-4.14-rc2

LZ4, netfilter and netlink options were compiled into the kernel breaking the standalone kmod- packages

The NXP LSDK kernel backported a change for ndo_get_stats64 from mainline ("net: make ndo_get_stats64 a void function"), this causes the mac80211 build to fail as backports still expects the old definition against 4.9:

```
backports-2017-10-06/net/mac80211/iface.c:1224:21: error: initialization from incompatible pointer type
error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
  .ndo_get_stats64 = bp_ieee80211_get_stats64,
```
In iface.c:1224

```
#if LINUX_VERSION_IS_GEQ(4,11,0)
>---.ndo_get_stats64>---= ieee80211_get_stats64,
#else
>---.ndo_get_stats64 = bp_ieee80211_get_stats64,
#endif
```
To resolve this, revert the backport - this greatly reduces the content of patch 601, so the patches that made up 601 previously (as seen at https://github.com/qoriq-open-source/linux/tree/linux-4.9/net/ ) have been disaggregated. 

A small change to the DPAA1&2 drivers is also needed to change them back to the old struct * return type.


hauke:
This looks ok,
did you try to apply the same changes to the arm32 config as well?
The config file probably needs even more cleanup, probably 70% of the options can get removed.
Does this also boot on a board, someone on IRC reported that it was not booting for him.

mcbridematt:
I have not tried the ARM32 build yet, in theory the config changes should be the same. 

This is working on our (Traverse) LS1043 board (see #1412), but I do have a LS1043ARDB I can try with the current NOR(mtd) configuration as well.

hauke:
Thank you for the patch it was applied with some small changes and I also made the same changes for the arm_36b subtarget. This config file probably needs some more clean up later.

PR #1464, opened 2017-10-26
ar71xx: fix kernel size partitions on tp-link boards

heinzek:
This patch fix kernel partition on tplink:
 * archer-c58-v1
 * archer-c60-v1
 * tl-wr902ac-v1
 * tl-wr942n-v1

### DIFF
--- a/target/linux/ar71xx/image/tp-link.mk
+++ b/target/linux/ar71xx/image/tp-link.mk
@@ -123,7 +123,7 @@ define Device/archer-c58-v1
   TPLINK_BOARD_ID := ARCHER-C58-V1
   DEVICE_PROFILE := ARCHERC58V1
   IMAGE_SIZE := 7936k
-  MTDPARTS := spi0.0:64k(u-boot)ro,64k(mac)ro,1344k(kernel),6592k(rootfs),64k(tplink)ro,64k(art)ro,7936k@0x20000(firmware)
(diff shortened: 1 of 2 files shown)


ptpt52:
good job!

robimarko:
You are missing signed off

f00b4r0:
Are the resulting images still flashable from OFW?
Can OFW still be reverted back?

heinzek:
> Are the resulting images still flashable from OFW?
Yes

> Can OFW still be reverted back?

On TP-Link devices that use a tplink-safeloader revert back is only available by recovery mode.
Firmware like "back to stock" is not working at all because during flashing from OEM firmware to LEDE other "partition" like "soft-version" is updated by "soft_ver" tag.
Revert back firmware to OEM firmware wihout rewrite this partition causes an update is blocked because firmware can't compare soft ver.

xssa:
Just for meditating - can we separate device specific flash layots from tplink-safeloader.c code and give it as command line arguments? Thus we will have opportunity to take actual kernel size during build process and dynamicaly calculate size of os-image and file-system partitions. This will be a very perspective change in particular for a future increase, or a reduction in the size of the kernel due to the use of user configurations or increasing of the code size of the newer kernels.

pepe2k:
@heinzek,

Cheers,
Piotr

PR #1466, opened 2017-10-27
adb: fix package description

cibomahto:
Fixed a syntax error in the package Makefile.

Signed-off-by: Matt Mets <matt@blinkinlabs.com>


hauke:
Thank you for the patch it was applied.

PR #872, opened 2017-02-23
kirkwood: fix include in etc/board.d/02_network

bobafetthotmail:
the mtd_get_mac_ascii function called within this script requires the inclusion of /lib/functions/system.sh

Signed-off-by: Alberto Bursi <alberto.bursi@outlook.it>

PR #1467, opened 2017-10-27
mktplinkfw* implement kernel offset

f00b4r0:
Several (all?) images built with mktplinkfw or mktplinkfw2 currently shipped in snapshot (and some in **17.01** release) will **soft brick** TP-Link devices because in the image header the end of the kernel binary collides with rootfs data. I suspect this causes a partial or invalid load with the bootloader, but couldn't confirm this hypothesis.

This patch attempts to address the issue, but I **couldn't test it yet**. One user on IRC is attempting to unbrick his C50 and will test and report if/when he can.

I have found one FS bug that appears to be directly related [(872)](https://bugs.lede-project.org/index.php?do=details&task_id=872) but there might be others

### DIFF
--- a/tools/firmware-utils/src/mktplinkfw-lib.c
+++ b/tools/firmware-utils/src/mktplinkfw-lib.c
@@ -33,6 +33,7 @@
 extern char *ofname;
 extern char *progname;
 extern uint32_t kernel_len;
+extern uint32_t kernel_ofs;
 extern struct file_info kernel_info;
 extern struct file_info rootfs_info;
 extern struct flash_layout *layout;
@@ -207,7 +208,6 @@ inline void inspect_fw_pmd5sum(const char *label, const uint8_t *val, const char
 	printf(" %s\n", text);
 }
 
-// header_size = sizeof(struct fw_header)
 int build_fw(size_t header_size)
 {
 	int buflen;
@@ -216,7 +216,7 @@ int build_fw(size_t header_size)
 	int ret = EXIT_FAILURE;
 	int writelen = 0;
 
-	writelen = header_size + kernel_len;
+	writelen = kernel_ofs + kernel_len;
 
 	if (combined)
 		buflen = writelen;
@@ -230,7 +230,7 @@ int build_fw(size_t header_size)
 	}
 
 	memset(buf, 0xff, buflen);
-	p = buf + header_size;
+	p = buf + kernel_ofs;
 	ret = read_to_buf(&kernel_info, p);
 	if (ret)
 		goto out_free_buf;
--- a/tools/firmware-utils/src/mktplinkfw.c
+++ b/tools/firmware-utils/src/mktplinkfw.c
@@ -94,6 +94,7 @@ struct file_info kernel_info;
 static uint32_t kernel_la = 0;
 static uint32_t kernel_ep = 0;
 uint32_t kernel_len = 0;
+uint32_t kernel_ofs = 0;
 struct file_info rootfs_info;
 uint32_t rootfs_ofs = 0;
 uint32_t rootfs_align;
@@ -198,6 +199,7 @@ static void usage(int status)
 "  -C <country>    set region code to <country>\n"
 "  -F <id>         use flash layout specified with <id>\n"
 "  -k <file>       read kernel image from the file <file>\n"
+"  -K <offset>     overwrite kernel offset with <offset> (hexval prefixed with 0x)\n"
 "  -r <file>       read rootfs image from the file <file>\n"
 "  -a <align>      align the rootfs start on an <align> bytes boundary\n"
 "  -R <offset>     overwrite rootfs offset with <offset> (hexval prefixed with 0x)\n"
@@ -273,6 +275,10 @@ static int check_options(void)
 			kernel_la = layout->kernel_la;
 		if (!kernel_ep)
 			kernel_ep = layout->kernel_ep;
+		if (!kernel_ofs)
+			kernel_ofs = sizeof(struct fw_header);	/* kernel after image header */
+		else
+			ERR("warning custom kernel offset may not be supported by target");
 		if (!rootfs_ofs)
 			rootfs_ofs = layout->rootfs_ofs;
 
@@ -306,9 +312,8 @@ static int check_options(void)
 			return ret;
 
 		if (rootfs_align) {
-			kernel_len += sizeof(struct fw_header);
-			rootfs_ofs = ALIGN(kernel_len, rootfs_align);
-			kernel_len -= sizeof(struct fw_header);
+			/* align rootfs offset after kernel data */
+			rootfs_ofs = ALIGN(kernel_ofs + kernel_len, rootfs_align);
 
 			DBG("rootfs offset aligned to 0x%u", rootfs_ofs);
 
@@ -369,7 +374,7 @@ void fill_header(char *buf, int len)
 
 	hdr->kernel_la = htonl(kernel_la);
 	hdr->kernel_ep = htonl(kernel_ep);
-	hdr->kernel_ofs = htonl(sizeof(struct fw_header));
+	hdr->kernel_ofs = htonl(kernel_ofs);
 	hdr->kernel_len = htonl(kernel_len);
 
 	if (!combined) {
@@ -538,7 +543,7 @@ int main(int argc, char *argv[])
 	while ( 1 ) {
 		int c;
 
-		c = getopt(argc, argv, "a:H:E:F:L:m:V:N:W:C:ci:k:r:R:o:xX:ehsjv:");
+		c = getopt(argc, argv, "a:H:E:F:L:m:V:N:W:C:ci:k:K:r:R:o:xX:ehsjv:");
 		if (c == -1)
 			break;
 
@@ -581,6 +586,9 @@ int main(int argc, char *argv[])
 			break;
 		case 'k':
 			kernel_info.file_name = optarg;
+			break;
+		case 'K':
+			sscanf(optarg, "0x%x", &kernel_ofs);
 			break;
 		case 'r':
 			rootfs_info.file_name = optarg;
--- a/tools/firmware-utils/src/mktplinkfw2.c
+++ b/tools/firmware-utils/src/mktplinkfw2.c
@@ -99,6 +99,7 @@ static int sver_hi;
 struct file_info kernel_info;
 static uint32_t kernel_la = 0;
 static uint32_t kernel_ep = 0;
+uint32_t kernel_ofs = 0;
 uint32_t kernel_len = 0;
 struct file_info rootfs_info;
 uint32_t rootfs_ofs = 0;
@@ -175,6 +176,7 @@ static void usage(int status)
 "  -w <hwveradd>   use additional hardware version specified with <hwveradd>\n"
 "  -F <id>         use flash layout specified with <id>\n"
 "  -k <file>       read kernel image from the file <file>\n"
+"  -K <offset>     overwrite kernel offset with <offset> (hexval prefixed with 0x)\n"
 "  -r <file>       read rootfs image from the file <file>\n"
 "  -a <align>      align the rootfs start on an <align> bytes boundary\n"
 "  -R <offset>     overwrite rootfs offset with <offset> (hexval prefixed with 0x)\n"
@@ -241,6 +243,10 @@ static int check_options(void)
 		kernel_la = layout->kernel_la;
 	if (!kernel_ep)
 		kernel_ep = layout->kernel_ep;
+	if (!kernel_ofs)
+		kernel_ofs = sizeof(struct fw_header);	/* kernel after image header */
+	else
+		ERR("warning custom kernel offset may not be supported by target");
 	if (!rootfs_ofs)
 		rootfs_ofs = layout->rootfs_ofs;
 
@@ -257,7 +263,7 @@ static int check_options(void)
 
 	if (combined) {
 		if (kernel_info.file_size >
-		    layout->fw_max_len - sizeof(struct fw_header)) {
+		    layout->fw_max_len - kernel_ofs) {
 			ERR("kernel image is too big");
 			return -1;
 		}
@@ -272,20 +278,19 @@ static int check_options(void)
 			return ret;
 
 		if (rootfs_align) {
-			kernel_len += sizeof(struct fw_header);
-			rootfs_ofs = ALIGN(kernel_len, rootfs_align);
-			kernel_len -= sizeof(struct fw_header);
+			/* align rootfs offset after kernel data */
+			rootfs_ofs = ALIGN(kernel_ofs + kernel_len, rootfs_align);
 
 			DBG("rootfs offset aligned to 0x%u", rootfs_ofs);
 
-			if (kernel_len + rootfs_info.file_size >
-			    layout->fw_max_len - sizeof(struct fw_header)) {
+			if (rootfs_info.file_size >
+			    layout->fw_max_len - rootfs_ofs) {
 				ERR("images are too big");
 				return -1;
 			}
 		} else {
 			if (kernel_info.file_size >
-			    rootfs_ofs - sizeof(struct fw_header)) {
+			    rootfs_ofs - kernel_ofs) {
 				ERR("kernel image is too big");
 				return -1;
 			}
@@ -350,7 +355,7 @@ void fill_header(char *buf, int len)
 	hdr->kernel_la = htonl(kernel_la);
 	hdr->kernel_ep = htonl(kernel_ep);
 	hdr->fw_length = htonl(layout->fw_max_len);
-	hdr->kernel_ofs = htonl(sizeof(struct fw_header));
+	hdr->kernel_ofs = htonl(kernel_ofs);
 	hdr->kernel_len = htonl(kernel_len);
 	if (!combined) {
 		hdr->rootfs_ofs = htonl(rootfs_ofs);
@@ -533,7 +538,7 @@ int main(int argc, char *argv[])
 	while ( 1 ) {
 		int c;
 
-		c = getopt(argc, argv, "a:H:E:F:L:V:N:W:w:ci:k:r:R:o:xhsjv:y:T:e");
+		c = getopt(argc, argv, "a:H:E:F:L:V:N:W:w:ci:k:K:r:R:o:xhsjv:y:T:e");
 		if (c == -1)
 			break;
 
@@ -576,6 +581,9 @@ int main(int argc, char *argv[])
 			break;
 		case 'k':
 			kernel_info.file_name = optarg;
+			break;
+		case 'K':
+			sscanf(optarg, "0x%x", &kernel_ofs);
 			break;
 		case 'r':
 			rootfs_info.file_name = optarg;


f00b4r0:
[1013](https://bugs.lede-project.org/index.php?do=details&task_id=1013) is also likely related

jow-:
@mkresin @pepe2k @NeoRaider

f00b4r0:
mtdsplit-tplink.c apparently also assumes that the rootfs starts at the end of the kernel length, which is presumably an incorrect assumption that happened to work with the previous code, I'll submit a patch for this too.

f00b4r0:
OK "Fireburner" on IRC confirmed that he was able to successfully flash and boot current master with these two commits included (fix for mktplinkfw and fix for mtdsplit-tplink). Seems like a good sign :)

f00b4r0:
possibly related: 
- [880](https://bugs.lede-project.org/index.php?do=details&task_id=880) (17.01.2, kernel and rootfs overlap at 0x00100000);

f00b4r0:
@mkresin pointed me at [target/linux/ar71xx/files/drivers/mtd/tplinkpart.c](/lede-project/source/blob/master/target/linux/ar71xx/files/drivers/mtd/tplinkpart.c)

It turns out this code "falls back" to the correct way of reading the rootfs offset:

```
        squashfs_offset = offset + sizeof(struct tplink_fw_header) +
                          be32_to_cpu(header->kernel_len);

        ret = tplink_check_rootfs_magic(master, squashfs_offset);
        if (ret == 0)
                rootfs_offset = squashfs_offset;
        else
                rootfs_offset = offset + be32_to_cpu(header->rootfs_ofs);
```

The first computation uses the assumption that the rootfs is stuffed immediately after the kernel, and it tries to detect a valid squashfs signature. If it can't, it then reads the rootfs offset value from the header field.

With my change, this splitter will work, although it will report a kernel length that will be larger than the actual kernel length (as recorded in the header), because of this:

```
        parts[1].name = "kernel";
        parts[1].offset = offset;
        parts[1].size = rootfs_offset - offset;
```

It's worth mentioning that this splitter claims to support v2 firmwares, but it doesn't (it only knows about v1 structure).

pkgadd:
I've successfully tested this pull request on these devices (thanks to a trimmed down kernel config, I didn't trigger the error condition on any of my devices with kernel 4.9 before either):

- TP-Link TL-WR941nd v2.6 (AP81)
    File name              : ./bin/targets/ar71xx/generic/lede-ar71xx-generic-tl-wr941nd-v2-squashfs-sysupgrade.bin
    File size              : 0x00360004 /  3538948 bytes
    Version 1 Header size  : 0x00000200 /      512 bytes
    Header MD5Sum1         : e8 c5 60 33 5a 5e 36 72 18 f7 ec f3 fb 2c 14 2a (ok)
    Header MD5Sum2         : 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 (purpose yet unknown, unchecked here)
    
    Vendor name            : OpenWrt
    Firmware version       : r5201+3-8b9cdebc9c
    Hardware ID            : 0x09410002
    Hardware Revision      : 0x00000002
    Region code            : 0x00000000
   
    Kernel data offset     : 0x00000200 /      512 bytes
    Kernel data length     : 0x00148095 /  1343637 bytes
    Kernel load address    : 0x80060000
    Kernel entry point     : 0x80060000
    Rootfs data offset     : 0x00148298 /  1344152 bytes
    Rootfs data length     : 0x002135fc /  2176508 bytes
    Boot loader data offset: 0x00000000 /        0 bytes
    Boot loader data length: 0x00000000 /        0 bytes
    Total firmware length  : 0x003c0000 /  3932160 bytes
- TP-Link TL-WR1043nd v1 (AP83)
    File name              : ./bin/targets/ar71xx/generic/lede-ar71xx-generic-tl-wr1043nd-v1-squashfs-sysupgrade.bin
    File size              : 0x00600004 /  6291460 bytes
    Version 1 Header size  : 0x00000200 /      512 bytes
    Header MD5Sum1         : 3f ef 5b a5 6a 41 c6 3f cb 77 84 bc 80 41 70 d6 (ok)
    Header MD5Sum2         : 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 (purpose yet unknown, unchecked here)
    
    Vendor name            : OpenWrt
    Firmware version       : r5201+3-8b9cdebc9c
    Hardware ID            : 0x10430001
    Hardware Revision      : 0x00000001
    Region code            : 0x00000000
    
    Kernel data offset     : 0x00000200 /      512 bytes
    Kernel data length     : 0x00148095 /  1343637 bytes
    Kernel load address    : 0x80060000
    Kernel entry point     : 0x80060000
    Rootfs data offset     : 0x00148298 /  1344152 bytes
    Rootfs data length     : 0x004b1f10 /  4923152 bytes
    Boot loader data offset: 0x00000000 /        0 bytes
    Boot loader data length: 0x00000000 /        0 bytes
    Total firmware length  : 0x007c0000 /  8126464 bytes
- TP-Link Tl-WDR3600 (DB120)
    File name              : ./bin/targets/ar71xx/generic/lede-ar71xx-generic-tl-wdr3600-v1-squashfs-sysupgrade.bin
    File size              : 0x00610004 /  6356996 bytes
    Version 1 Header size  : 0x00000200 /      512 bytes
    Header MD5Sum1         : 5c f8 96 bf cc fb 46 7a 18 70 e9 b4 30 70 e6 59 (ok)
    Header MD5Sum2         : 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 (purpose yet unknown, unchecked here)
    
    Vendor name            : OpenWrt
    Firmware version       : r5201+3-8b9cdebc9c
    Hardware ID            : 0x36000001
    Hardware Revision      : 0x00000001
    Region code            : 0x00000000
    
    Kernel data offset     : 0x00000200 /      512 bytes
    Kernel data length     : 0x0014623d /  1335869 bytes
    Kernel load address    : 0x80060000
    Kernel entry point     : 0x80060000
    Rootfs data offset     : 0x00146440 /  1336384 bytes
    Rootfs data length     : 0x004c7f54 /  5013332 bytes
    Boot loader data offset: 0x00000000 /        0 bytes
    Boot loader data length: 0x00000000 /        0 bytes
    Total firmware length  : 0x007c0000 /  8126464 bytes
- TP-Link TL-WDR4300 (DB120)
    File name              : ./bin/targets/ar71xx/generic/lede-ar71xx-generic-tl-wdr4300-v1-squashfs-sysupgrade.bin
    File size              : 0x00600004 /  6291460 bytes
    Version 1 Header size  : 0x00000200 /      512 bytes
    Header MD5Sum1         : c2 8c 97 d6 7e 24 8f c7 a3 89 18 58 03 de 7e e6 (ok)
    Header MD5Sum2         : 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 (purpose yet unknown, unchecked here)
    
    Vendor name            : OpenWrt
    Firmware version       : r5201+3-8b9cdebc9c
    Hardware ID            : 0x43000001
    Hardware Revision      : 0x00000001
    Region code            : 0x00000000
    
    Kernel data offset     : 0x00000200 /      512 bytes
    Kernel data length     : 0x0014623d /  1335869 bytes
    Kernel load address    : 0x80060000
    Kernel entry point     : 0x80060000
    Rootfs data offset     : 0x00146440 /  1336384 bytes
    Rootfs data length     : 0x004b6e98 /  4943512 bytes
    Boot loader data offset: 0x00000000 /        0 bytes
    Boot loader data length: 0x00000000 /        0 bytes
    Total firmware length  : 0x007c0000 /  8126464 bytes

Tested-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>

diizzyy:
```
[    0.646853] 5 tp-link partitions found on MTD device spi0.0
[    0.652553] Creating 5 MTD partitions on "spi0.0":
[    0.657429] 0x000000000000-0x000000020000 : "u-boot"
[    0.663940] 0x000000020000-0x000000179890 : "kernel"
[    0.670633] 0x000000179890-0x0000007f0000 : "rootfs"
[    0.677250] mtd: device 2 (rootfs) set to be root filesystem
[    0.683079] 1 squashfs-split partitions found on MTD device rootfs
[    0.689359] 0x0000003b0000-0x0000007f0000 : "rootfs_data"
[    0.696475] 0x0000007f0000-0x000000800000 : "art"
[    0.702932] 0x000000020000-0x0000007f0000 : "firmware"
....
[   36.142149] jffs2: notice: (955) jffs2_build_xattr_subsystem: complete building xattr subsystem, 0 of xdatum (0 unchecked, 0 orphan) and 0 of xref (0 dead, 0 orphan) found.
```

Tested on: TP-Link TL-WDR3600

nbd168:
LGTM

f00b4r0:
Device bricking is apparently real, here's another report on C50(EU):
https://forum.lede-project.org/t/lede-on-tp-link-archer-c50-but-the-install-instructions-is-in-polish/3446/44

mkresin:
It is a bug for sure!

Nevertheless, I still have no explanation why only the C50 (since it is the only board where we have an confirmation that this PR fixes a boot issue) seems to be affect by the bug.

I prefer to understand the root cause to ensure we don't workaround/hide the real issue.

I'm able to reproduce the wrong header with a TD-W8980 (lantiq/xrx200) image:

```
Kernel data offset     : 0x00000200 /      512 bytes
Kernel data length     : 0x001abed4 /  1752788 bytes
Kernel load address    : 0x80002000
Kernel entry point     : 0x80002000
Rootfs data offset     : 0x00140000 /  1310720 bytes
Rootfs data length     : 0x00357506 /  3503366 bytes
Boot loader data offset: 0x00000000 /        0 bytes
Boot loader data length: 0x00000000 /        0 bytes
Total firmware length  : 0x007a0000 /  7995392 bytes
```

But the image works just fine. Dumping the boards memory shows an uncompressed and a **not**   at offset 0x00140000 corrupted kernel. The common tp-link firmware splitter works due to the same bug fine as well:

```
mtd0: 00020000 00010000 "u-boot"
mtd1: 007a0000 00010000 "firmware"
mtd2: 001ac0d4 00010000 "kernel"
mtd3: 005f3f2c 00010000 "rootfs"
mtd4: 00290000 00010000 "rootfs_data"
mtd5: 00010000 00010000 "config"
mtd6: 00030000 00010000 "boardconfig"


cat /sys/class/mtd/

mtd2/offset:  131072 (0x020000)
mtd2/size:   1753300 (0x1AC0D4)

mtd3/offset: 1884372 (0x1CC0D4)
mtd3/size:   6242092 (0x5F3F2C)
```

I'm out of ideas what kind of issue the Archer C50 has. Is it the bootloader which does different checks or just behaves different? Is it the TP-Link webinterface firmware upgrade routine/TFTP recover which overwrites the kernel based on the rootfs offset value? Or maybe something completely different?

Unfortunately my u-boot doesn't have a TFTP recover and the  TP-Link webinterface requires a RSA signed image. Hence I'm unable to check for these.

f00b4r0:
I suspect the bootloader does different checks or behave differently.

Since this is a bug, and since this proposed patch fixes this bug for the affected hardware (until we get feedback for FS872 and other linked bugreports), my hypothesis is that the bootloader in some (recent?) TP-Link hardware either checks for partition overlap or computes the kernel load length differently.

We have a bug. We have a fix for the bug that implements a correct behavior of the tool, and which does not break already working hardware (as confirmed by the above testing). I'm not sure what else you're looking for here? If the tool had be correctly coded to begin with we wouldn't be having that conversation...

mkresin:
@heinzek sponsored bootlogs which are showing in which cases the the bug gets triggered.

The upgrade from a running LEDE is fine with stable and master.

The issue is the installation from the TP-Link webinterface using the factory image:

> Now I try to go back to OEM firmware, update to the latest EU version (ArcherC50v1-squashfs-factory-eu.bin) and use the LEDE factory-EU image:

```
U-Boot 1.1.3 (Aug 31 2015 - 16:29:46)

Board: Ralink APSoC DRAM:  64 MB
relocate_code Pointer at: 83fb0000
enable ephy clock...done. rf reg 29 = 5
SSC disabled.
******************************
Software System Reset Occurred
******************************
------------------
Archer C50 v1.0.0
------------------
spi_wait_nsec: 29
spi device id: ef 40 17 0 0 (40170000)
find flash: W25Q64BV
============================================
Ralink UBoot Version: 4.1.2.0
--------------------------------------------
ASIC 7620_MP (Port5<->None)
DRAM component: 512 Mbits DDR, width 16
DRAM bus: 16 bit
Total memory: 64 MBytes
Flash component: SPI Flash
Date:Aug 31 2015  Time:16:29:46
============================================
icache: sets:512, ways:4, linesz:32 ,total:65536
dcache: sets:256, ways:4, linesz:32 ,total:32768

##### The CPU freq = 580 MHZ ####
estimate memory size =64 Mbytes

continue to starting system.                                                                                                                                                                       0
disableEthForward(1158):disable switch forward...

3: System Boot system code via Flash.(0xbc020000)
do_bootm:argc=2, addr=0xbc020000
## Booting image at bc020000 ...
   Uncompressing Kernel Image ... LZMA ERROR 1 - must RESET board to recover
```

> Checked uboot versions:
> - U-Boot 1.1.3 (Sep 15 2014 - 11:20:17)
> - U-Boot 1.1.3 (Aug 31 2015 - 16:29:46)

@f00b4r0

Would you please add a note to the commit message that the factory images + the upgrade from the TP-Link webinterface is the culprit. Might be a good idea to include the actual error message.

Please change `Probably Fixes: #872, #1013` to `Fixes: #872`. FS#1013 is most likely a different issue.

Feel free to add a: `Tested-by: Mathias Kresin <dev@kresin.me>` to the commit message (and use Stefans tested-by as well)

Please add a commit message to the 2nd commit as well.

f00b4r0:
I have updated the commit messages (patch content is otherwise untouched).

Please note:
- that the second commit references the hash of the first one, whoever merges this will need to update the commit message with the final hash reference
- that this fix will need to be specifically backported to 17.01 (it can't be cherry-picked) as 17.01 does not use the reworked mktplinkfw code base. I can provide such a backport.

f00b4r0:
I have updated the commit messages (patch content is otherwise untouched).

Please note:

- that the second commit references the hash of the first one, whoever merges this will need to update the commit message with the final hash reference
- that this fix will need to be specifically backported to 17.01 (it can't be cherry-picked) as 17.01 does not use the reworked mktplinkfw code base. I can provide such a backport.

heinzek:
Hi @mkresin @f00b4r0 
New uboot is OK
Problem is in the OEM upgrade page.
I try install LEDE factory-EU from u-boot command line and it works OK
Then I make a full flash dump.
Next I back to stock EU firmware and again install LEDE Factory-EU - thisnot working.
Then again make a full flash dump and compare it:
 image

dump1 - OK
dump2 - LZMA error
lede-fac.bin - latest snapshot ArcherC50v1-squashfs-factory-eu.bin

mkresin:
@heinzek can you send me the dumps via mail or a download link.

heinzek:
No problem :)
https://tplinkforum.pl/t/temp-archer-c50v1-dump/9844

mkresin:
@heinzek I don't understand what you mean with " New uboot is OK". If the TP-Link webinterface firmware upgrade is the issue, the u-boot version shouldn't matter.

Would it be possible that you are doing the same with the patches from the PR applied. I now we have a similar issue where a buggy tp-link webinterface firmware upgrade routine writes some values to fixed flash addresses.

heinzek:
I mean "New uboot is OK" because LEDE is working.

Now i applied patch from this PR, make image and install it.
This same error.
Check memory:
image

mkresin:
@heinzek Last question. Do you know/can you test if older TP-Link firmware versions have the same bug of writing 0x00000002/0x02000000 to offset  0x2023C of the image? This bug isn't new and we had at least two affected TP-Link boards in the last months.

I had a look at Archer_C50v1_0.9.1_3.0_up_boot(160411)_2016-04-11_13.38.13.bin to see why this behaviour doesn't seem to cause any harm to the stock firmware image. It has already 0x00000002/0x02000000 at the very same position:

screenshot

Long story short, the dumps provided by heinzek are identical beside the value at image offset 0x2023C. It is either a lucky coincident that the image booted for the C50 tester or he has done the upgrade from an older TP-Link firmware version.

There isn't any chance that this PR fixes the C50 factory upgrade issue, sorry.

f00b4r0:
@mkresin what you're looking at is a duplicate of the header:

```
hexdump -C -n 64 Archer_C50v1_0.9.1_3.0_up_boot\(160411\)_2016-04-11_13.38.13.bin 
00000000  03 00 00 00 76 65 72 2e  20 32 2e 30 00 ff ff ff  |....ver. 2.0....|
00000010  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00000030  ff ff ff ff c7 50 00 01  00 00 00 45 00 00 00 02  |.....P.....E....|
```

```
hexdump -C -n 64 -s 0x20200 Archer_C50v1_0.9.1_3.0_up_boot\(160411\)_2016-04-11_13.38.13.bin
00020200  03 00 00 00 76 65 72 2e  20 32 2e 30 00 ff ff ff  |....ver. 2.0....|
00020210  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00020230  ff ff ff ff c7 50 00 01  00 00 00 45 00 00 00 02  |.....P.....E....|
```

What this means, as I was afraid of, is that the mktplinkfw tools are still not "correct". This is a separate bug from the one the PR attempts to fix

f00b4r0:
The 0x00000002 value that gets written at 0x4023C in @heinzek dumps apparently comes from the hw_rev_add value normally found at 0x23C and 0x2023C on a working system: 0x2 is the expected value on EU devices.

It appears the OFW flashing routine improperly stamps this value at the wrong offset (0x4023C is 0x2023C + 0x20000, where 0x20000 is the expected offset between the two copies of the header normally found on the flash).

I can think of a couple ways to work around this bug (assuming TP-Link has no intention of fixing it), this will require more testing.

f00b4r0:
@heinzek could you try the PR on an **Archer C50v1 EU** again with the last three commits I added? Flashing from OFW GUI, with the OFW version that ended in a failed boot for you.

@mkresin here's what this is doing:
I'm implementing full support for kernel offset in the image. For testing purposes, for the factory-eu.bin Archer C50 image I am moving the kernel to 0x20400 (which is 0x20000 + sizeof(header)), so that when the routine writes the kernel from 0x20000 and clobbers the 4 bytes at 0x4023C, the kernel remains unaffected.

This wastes 128KB of flash when flashing from OFW, however that space can eventually be recovered when sysupgrading because the firmware partition is unaffected (in DTS).

Note:
- I have updated the tplink splitter to accept the resulting header values and correctly assign the kernel partition within the firmware one
- This is a workaround which I am not keen on generalizing for all boards (hence my current choice of doing it via a Makefile quirk), if only because of the space penalty, and because TP-Link might break other things in the non-bootloader payload upgrade path (which they apparently do not use) and we probably want to keep track of that
- Once everything is confirmed to be working, I will implement the same code in v1 (including updating tplinkpart.c) and will update the commit messages so that we can confirm this doesn't break working devices.
- Theoretically, TP-Link could eventually totally break this firmware upgrade routine (or remove it entirely), at which point the only option will be to bundle a bootloader to the factory image, as they do.

HTH

heinzek:
Nope. This not working:
image

image
header:
image
kernel start:
image

f00b4r0:
the error output is quite terse. Hard to tell what's going on:

- is it ignoring the kernel offset value?
- is the kernel image broken for some other reason?

If it's ignoring the value of the kernel offset field, the only option left (short of bundling the bootloader) is to add a lzma loader stub at 0x20200 that will load the kernel from it's actual offset, assuming we can make such a stub fit within 0x3B bytes...

heinzek:
I think this ignore kernel offset and try load kernel from address 0x20200.
Error " Uncompressing Kernel Image ... LZMA ERROR 1 - must RESET board to recover" is instant. In previous EU image error is after 0,5sec.

I check RU OEM firmware and it has the same error. ( in 0x4023C have 00 00 00 01)
US firmware for C50v1 is the same like first UN firmware. Upgrade works with factory, factory-eu and factory-us.

heinzek:
Now I test normal LEDE image with empty header:
image
its look like uboot first load kernel then read kernel offset value.
When I erase 0x20000 to 0x30000 i have LZMA error.

f00b4r0:
OK thanks for this extra info: it tends to confirm my suspicion. The bootloader ignores the value of kernel offset in the header.

So we have only two options at this point:
- cram a secondary loader between 0x200 and 0x20000 in the image (that's 0x20200 - 0x40000 on the flash)
- bundle our own bootloader

f00b4r0:
After some more investigation by @mkresin and myself, it appears that the tp-link-modified uboot has hardcoded load address for loading the image into RAM and calling the LZMA decoder, meaning that:
- it will always load (map?) the entire firmware data (up to the length of the kernel segment?) from a fixed RAM position (0xbc020000 on C50)
- it will always try to LZMA decompress that loaded data from a 0x200 (header size) offset of that fixed position into a fixed RAM destination (0x80002000 on C50), which it will then jump to, regardless of the actual offset of the LZMA kernel within that data stream

f00b4r0:
@mkresin I have updated the commit messages of the first two patches (content is untouched). Please note that the 2nd references the 1st, you'll want to update the reference if possible

The next 3 patches are a cleaner version of the code that's been tested so far, with explicit commit messages. The last commit is WIP for reasons explained in the commit message (I am in particular concerned about endianness handling). These 3 patches are independent from each other and can be merged in any order. I think they are nevertheless testing ready.

Comments welcome :)

f00b4r0:
Still WIP as far as I'm concerned:
- the initial problem isn't solved and we still ship images that will brick users' devices;
- I could polish/fix tplinkpart.c provided that I get some feedback on its operation...

f00b4r0:
I have moved the splitter cleanups to #1524 since they're orthogonal to this PR.

rotanid:
@f00b4r0 what made you quit working on this? I know some people who would really thank you for finishing this!

PR #1468, opened 2017-10-28
ramips: add support for D-Link DAP-1522 A1 

george-hopkins:
D-Link DAP-1522 is a wireless bridge/access point with 4 LAN ports and a dual-band wireless chipset.

Specifications:
- Ralink RT2880
- 32 MB of RAM
- 4 MB of Flash
- 4x 10/100/1000 Mbps Ethernet (RTL8366SR)
- 802.11abgn (RT2850)

Flash Instructions:
1. Download lede-ramips-rt288x-dap-1522-a1-squashfs-factory.bin
2. Open the web interface and upload the image


mkresin:
Given the fact that we are already [warn about using board with only 4MByte flash and 32MByte ram](https://lede-project.org/meta/infobox/432_warning), I'm not sure if we should add support for this board.

@lede-project/developer any opinions?

bobafetthotmail:
The warning is for less-technical users that usually want to install additional packages after flashing the device and don't know of the size limitations (and ram required to run additional services).

The devices with 4MB and 32MB of ram work fine as pure network devices (router/access point/repeater).

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
+/dts-v1/;
+
+#include "rt2880.dtsi"
+
+#include <dt-bindings/input/input.h>
+
+/ {
+	#address-cells = <1>;
```
mkresin: the cells properties can be dropped here.

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
...
+	#size-cells = <1>;
+	compatible = "DAP-1522-A1", "ralink,rt2880-soc";
```
mkresin: should be `compatible = "dlink,dap-1522-a1", "ralink,rt2880-soc";`

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
+/dts-v1/;
+
+#include "rt2880.dtsi"
+
+#include <dt-bindings/input/input.h>
```
mkresin: Please include <[dt-bindings/gpio/gpio.h](http://lxr.free-electrons.com/source/include/dt-bindings/gpio/gpio.h?v=4.4)> here as well and keep alphabetical order.

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
...
+			label = "ap";
+			gpios = <&gpio2 13 1>;
+			linux,code = <BTN_0>;
+			linux,input-type = <EV_SW>;
+		};
+
+		bridge {
+			label = "bridge";
+			gpios = <&gpio2 14 1>;
+			linux,code = <BTN_0>;
```
mkresin: Using the `BTN_0` keycode doesn't look right.

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
...
+		};
+	};
+};
+
+&gpio0 {
+	status = "okay";
+};
+
+
+&gpio1 {
```
mkresin: no gpio from the 2nd gpio bank is used. You can drop  it.

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -164,6 +164,7 @@ get_status_led() {
 	hc5962)
 		status_led="$board:white:status"
 		;;
+	dap-1522-a1|\
```
mkresin: please move the whole block up to keep alphabetical order.

target/linux/ramips/rt288x/config-4.9
```diff
@@ -134,6 +134,7 @@ CONFIG_MTD_SPI_NOR=y
 CONFIG_MTD_SPLIT_FIRMWARE=y
 CONFIG_MTD_SPLIT_LZMA_FW=y
 CONFIG_MTD_SPLIT_UIMAGE_FW=y
+CONFIG_MTD_SPLIT_WRGG_FW=y
```
mkresin: Not tested yet, but I'm quite sure you need to add a `# CONFIG_MTD_SPLIT_WRGG_FW is not set` to `target/linux/generic/config-*`.

Otherwise building any other target or a different subtarget than rt288x should cause a build error because of the new config symbol.

target/linux/ramips/image/rt288x.mk
```diff
@@ -30,6 +30,22 @@ define Device/ar725w
 endef
 TARGET_DEVICES += ar725w
 
+define Device/dap-1522-a1
+  DTS := DAP-1522-A1
+  BLOCKSIZE := 64k
+  IMAGE_SIZE := $(ralink_default_fw_size_4M)
```
mkresin: ralink_default_fw_size_4M (3866624) doesn'T match the size if your factory partition (0x3a0000 == 3801088 bytes)

target/linux/ramips/dts/DAP-1522-A1.dts
```diff
@@ -0,0 +1,145 @@
...
+};
+
+&gpio2 {
+	status = "okay";
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "i2c", "spi", "uartlite", "pci";
```
mkresin: drop the spi group here. none of the pins of the spi group is used as gpio.

mkresin:
Any reason why you don't add a default (vlan) switch config for the RTL8366SR?

target/linux/ramips/rt288x/config-4.9
george-hopkins: As the splitter was just extended, `CONFIG_MTD_SPLIT_WRGG_FW` was already part of `target/linux/generic/config-*`

george-hopkins:
@mkresin Thank you for reviewing! Please let me know if there are any further issues that need to be addressed.

Unfortunately, I couldn't get the VLAN to work. Frames originating from the CPU were left tagged when they left the ports. With the current configuration, the 4 LAN ports work without problems (there is no WAN port). As I don't have any experience debugging VLANs/switches, I'm kind of stuck. Any pointers as to what should be checked/tested might help to resolve the issue.

blogic:

george-hopkins:
Thank you for the update! The patch has been rebased and was submitted in openwrt/openwrt#614.

PR #1469, opened 2017-10-28
kernel: add NFS4 client support

marcin1j:
Adds NFS4 client support:
 1. Package `kmod-fs-nfs` is split into `kmod-fs-nfs` (`nfs.ko`) and `kmod-fs-nfs-v3` (`nfsv3.ko`).
 2. A new package `kmod-fs-nfs-v4` (`nfsv4.ko`) is created.
 3. Package `kmod-fs-nfs-common-v4` is renamed to `kmod-fs-nfs-rpcsec` and includes additional module `rpcsec_gss_krb5.ko`. `CONFIG_NFS_V4` goes into `kmod-fs-nfs-v4`, `CONFIG_NFSD_V4` (NFS4 server) is removed. Missing kernel module `oid_registry.ko` needed by `auth_rpcgss.ko` is added to the package.

A new package `kmod-crypto-cts` needed by `rpcsec_gss_krb5.ko` is also created.

NFS4 client was compile and runtime tested on ramips (Nexx WT3020) and ar71xx (TP-Link TL-WR710N).

### DIFF
--- a/package/kernel/linux/modules/fs.mk
+++ b/package/kernel/linux/modules/fs.mk
@@ -317,24 +317,54 @@ $(eval $(call KernelPackage,fs-msdos))
 
 define KernelPackage/fs-nfs
   SUBMENU:=$(FS_MENU)
-  TITLE:=NFS filesystem support
+  TITLE:=NFS filesystem client support
   DEPENDS:=+kmod-fs-nfs-common +kmod-dnsresolver
   KCONFIG:= \
 	CONFIG_NFS_FS \
 	CONFIG_NFS_USE_LEGACY_DNS=n \
 	CONFIG_NFS_USE_NEW_IDMAPPER=n
   FILES:= \
-	$(LINUX_DIR)/fs/nfs/nfs.ko \
-	$(LINUX_DIR)/fs/nfs/nfsv3.ko
-  AUTOLOAD:=$(call AutoLoad,40,nfs nfsv3)
+	$(LINUX_DIR)/fs/nfs/nfs.ko
+  AUTOLOAD:=$(call AutoLoad,40,nfs)
 endef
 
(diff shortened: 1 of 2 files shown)


package/kernel/linux/modules/fs.mk
```diff
@@ -356,13 +386,13 @@ $(eval $(call KernelPackage,fs-nfs-common))
...
-	CONFIG_SUNRPC_GSS\
-	CONFIG_NFS_V4=y\
-	CONFIG_NFSD_V4=y
-  DEPENDS:= @BROKEN
+	CONFIG_SUNRPC_GSS
   FILES+=$(LINUX_DIR)/net/sunrpc/auth_gss/auth_rpcgss.ko
-  AUTOLOAD=$(call AutoLoad,30,auth_rpcgss)
+  FILES+=$(LINUX_DIR)/lib/oid_registry.ko
+  AUTOLOAD=$(call AutoLoad,31,auth_rpcgss)
+  AUTOLOAD=$(call AutoLoad,31,oid_registry)
```
hauke: I think you do not need the AUTOLOAD lines at all and just let it be loaded from the NFS library.

marcin1j: You're right that these modules are unnecessary for NFS4 without RPC GSS authentication which lacks userspace support anyway.

This package should be renamed to `kmod-fs-nfs-rpcsec` or something similar and:
  1. the dependency from `kmod-fs-nfs-v4` removed,
  2. module `rpcsec_gss_krb5.ko` should be included.
That would provide separate package with kernel support for `sec=krb5`.

marcin1j:
It seems that I messed with package titles and descriptions so I'll need to update this pull request anyway.

hauke:

PR #1470, opened 2017-10-29
kernel: coretemp and cpuid packages

pprindeville:
Add support for the Intel Coretemp (on-CPU thermal sensors) and the CPUID instruction.

### DIFF
--- a/package/kernel/linux/modules/hwmon.mk
+++ b/package/kernel/linux/modules/hwmon.mk
@@ -78,6 +78,21 @@ endef
 $(eval $(call KernelPackage,hwmon-adt7475))
 
 
+define KernelPackage/hwmon-coretemp
+  TITLE:=Pentium thermal monitoring support
+  KCONFIG:=CONFIG_SENSORS_CORETEMP
+  FILES:=$(LINUX_DIR)/drivers/hwmon/coretemp.ko
+  AUTOLOAD:=$(call AutoProbe,coretemp)
+  $(call AddDepends/hwmon)
+endef
+
+define KernelPackage/hwmon-coretemp/description
+ Kernel module for Intel Pentium and above CPU temperature
+endef
+
+$(eval $(call KernelPackage,hwmon-coretemp))
+
+
 define KernelPackage/hwmon-ina209
   TITLE:=INA209 monitoring support
   KCONFIG:=CONFIG_SENSORS_INA209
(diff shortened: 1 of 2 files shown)


package/kernel/linux/modules/hwmon.mk
```diff
@@ -78,6 +78,21 @@ endef
 $(eval $(call KernelPackage,hwmon-adt7475))
 
 
+define KernelPackage/hwmon-coretemp
+  TITLE:=Pentium thermal monitoring support
+  KCONFIG:=CONFIG_SENSORS_CORETEMP
+  FILES:=$(LINUX_DIR)/drivers/hwmon/coretemp.ko
+  AUTOLOAD:=$(call AutoProbe,coretemp)
+  $(call AddDepends/hwmon)
+endef
```
hauke: This is at least activated by default in target/linux/x86/64/config-default 
Should this also be activated by default for more x86 configurations?

package/kernel/linux/modules/other.mk
```diff
@@ -135,6 +135,18 @@ endef
...
 
+define KernelPackage/cpuid
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=Intel CPUID support
+  KCONFIG:=CONFIG_X86_CPUID
+  DEPENDS:=@TARGET_x86
+  FILES:=$(LINUX_DIR)/arch/x86/kernel/cpuid.ko
+  AUTOLOAD:=$(call AutoLoad,20,cpuid)
+endef
+
```
hauke: This is activated by default for target/linux/x86/geode/config-default
Should this also be activated by default for more x86 configurations?

nbd168:
Yes

nbd168:
Yes, this should be enabled in the relevant kernel configs instead of being packaged as kmod-* packages.

package/kernel/linux/modules/hwmon.mk
pprindeville: > Should this also be activated by default for more x86 configurations?

@hauke We could, but then (a) `hwmon` would need to be baked into the kernel, and do we do it for all sensors, i.e. `CORETEMP`, `K10TEMP`, `K8TEMP`, `VIA_CPUTEMP`?

nbd168: Yes

package/kernel/linux/modules/other.mk
pprindeville: > Should this also be activated by default for more x86 configurations?

@Hauke it's not a big driver and more processors support it than not.

pprindeville:
> Yes, this should be enabled in the relevant kernel configs instead of being packaged as kmod-* packages.

@nbd168 Done, see competing PR #1471.

PR #1471, opened 2017-10-29
kernel: add CPU temperature sensors to configs

pprindeville:
Most x86 CPU's, even low-end embedded CPU's like the Via C7, support
on-chip monitoring of CPU temperature.  Building these in involves
less than 16K of extra kernel memory.

Most modern x86 CPU's also support the CPUID instruction, so bake
that in as well.


diizzyy:
LGTM

pprindeville:
> LGTM

@diizzy Thanks!

nbd168:
Please also refresh the legacy and geode subtarget kernel configs

pprindeville:
> Please also refresh the legacy and geode subtarget kernel configs

@nbd168 How?  The Geode LX has the CPUID instruction, as I remember.  As do some Pentiums (Pentium 4 and on).  Should I undefine SENSORS_CORETEMP for both of these?

nbd168:
I'm not talking about explicitly enabling/disabling any options for those subtargets. You touched the top-level config and that affects the subtargets. Because of that, you need to refresh them, e.g. using make kernel_oldconfig

pprindeville:
> Because of that, you need to refresh them, e.g. using make kernel_oldconfig

I recloned my tree, and checked out the same branch.

Then I ran `make menuconfig`, and selected `x86` as the Target System and `Geode` as the Subtarget.  Then I ran `make kernel_oldconfig` and `git diff`.  No output.

I changed the Subtarget, again via `menuconfig`, and redid the `make kernel_oldconfig`.  No change.

So I'm not sure how I'm supposed to refresh things...  What am I missing?

nbd168:
Use `make kernel_oldconfig CONFIG_TARGET=subtarget`. Otherwise it defaults to changing the target config.
Maybe it would also make more sense to only enable the sensors in the subtargets that you actually need them in, e.g. generic and 64, and leaving the others untouched.
You can do that by reverting your changes and using `make kernel_menuconfig CONFIG_TARGET=subtarget`

pprindeville:
> You can do that by reverting your changes and using `make kernel_menuconfig CONFIG_TARGET=subtarget`

Are these steps documented somewhere?  I think I'm missing a step or two.

nbd168:
https://lede-project.org/docs/guide-developer/use-buildsystem#kernel_configuration_optional

pprindeville:
@dangowrt Please review

pprindeville:
> Maybe it would also make more sense to only enable the sensors in the subtargets that you actually need them in, e.g. generic and 64, and leaving the others untouched.

@nbd168 I believe I've done this.  Can you have a look, please?

hauke:

pprindeville:
Thanks!

PR #1473, opened 2017-10-30
kernel: Silence xhci TD warnings

diizzyy:
Some devices such as card readers triggers this
"harmless" warning and spams logs.

FS#952

Source: https://kernel.opensuse.org/cgit/kernel-source/plain/patches.suse/0001-xhci-silence-TD-warning.patch?id=b52251da0b37509c00365ef21446f4d9d25b7059

### DIFF
--- a/target/linux/generic/hack-4.9/912-silence-xhci-td-warnings.patch
+++ b/target/linux/generic/hack-4.9/912-silence-xhci-td-warnings.patch
@@ -1,0 +1,32 @@
+From a6eac48b24096503358b6d4ab0d40deb004e5dfd Mon Sep 17 00:00:00 2001
+From: Oliver Neukum <oneukum@suse.com>
+Date: Mon, 30 Nov 2015 16:50:34 +0100
+Subject: [PATCH] xhci: silence TD warning
+Patch-Mainline: Never (proper fix would break kABI)
+References: bnc#939955
+
+This warning triggers for things that were cleaned up by
+the ring expansion code. As we cannot have that in SP3, lowering
+the urgency by a notch to not spam logs.
+
+Signed-off-by: Oliver Neukum <oneukum@suse.com>
+---
+ drivers/usb/host/xhci-ring.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/drivers/usb/host/xhci-ring.c b/drivers/usb/host/xhci-ring.c
+index 63735b5..2f7c86c 100644
+--- a/drivers/usb/host/xhci-ring.c
++++ b/drivers/usb/host/xhci-ring.c
+@@ -2492,7 +2492,7 @@ static int handle_tx_event(struct xhci_hcd *xhci,
+ 			 */
+ 			if (!(trb_comp_code == COMP_STOP ||
+ 						trb_comp_code == COMP_STOP_INVAL)) {
+-				xhci_warn(xhci, "WARN Event TRB for slot %d ep %d with no TDs queued?\n",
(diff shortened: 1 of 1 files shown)


diizzyy:
Seen on MT7621 and mvebu platforms but according to Google it pretty much affects any platform.

karlp:
What's the ring expansion code referred to? that patch is _old_ why isn't this fixed upstream yet?

target/linux/generic/hack-4.9/912-silence-xhci-td-warnings.patch
```diff
@@ -0,0 +1,33 @@
+From a6eac48b24096503358b6d4ab0d40deb004e5dfd Mon Sep 17 00:00:00 2001
+From: Daniel Engberg <daniel.engberg.lists@pyret.net>
```
KanjiMonster: Why did you replace the original author? This patch wasn't written by you.

Keep the original commit/patch intact as much as possible.

KanjiMonster:
Also: what about 4.4, does it make sense there as well?  What about 17.01?

diizzyy:
"Patch-Mainline: Never (proper fix would break kABI)" (source url)
Unfortunately debugging the driver is beyond my scope.

target/linux/generic/hack-4.9/912-silence-xhci-td-warnings.patch
diizzyy: Calm down, as I stated on irc this is my first kernel patch submission and I did refer to the source if you read a few lines below. I'll change that to Oliver....

diizzyy:
@KanjiMonster 
Fixed?
I have no 4.4 devices left, can't test.
I don't use 17.01 nor should be in there as it's more of a cosmetic fix than a bug/security fix (depending on how you look at it I guess).

KanjiMonster:
> I have no 4.4 devices left, can't test.

You can still read the code and check if the same code is already present in 4.4.

mkresin:
From the original patch:

>This warning triggers for things that were cleaned up by the ring expansion code.

Seem to me the cleaned up ring expansion code is the proper fix for this warning. I'm a bit concerned that this change would suppress the warning in cases where it should be displayed and makes debugging/reporting unnecessary harder.

diizzyy:
Seems related... ?
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1713691
https://lkml.org/lkml/2017/2/14/773
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/drivers/usb/host/xhci-ring.c

diizzyy:
@mkresin 
It just requires you to enabled debugging

mkresin:
>  It just requires you to enable debugging

Which isn't enabled by default and hence most likely missing in bug reports from casual users.

diizzyy:
At least SUSE deemed it's worth it instead of having logs flooded (in my case it's about 3 lines every 10th second) and having logread/dmesg useless. Do note that the devices actually work just fine.

diizzyy:
@KanjiMonster 
So this bugs also seems to be in 4.4...

That said, is this something that potentially can be merged or would I just be wasting my time?

diizzyy:
Anyone, ping?

mkresin:
> Anyone, ping?

Has anything changed since I pointed to the "cleaned up ring expansion code" as a proper fix and expressed my concerns about hiding the warning?

diizzyy:
According to the original author that breaks kABI so I'm honestly not sure if that's something we want to do (ie most likely better to wait and use a newer kernel branch). If there's such strong concern I'll just drop the PR and add a comment "won't fix" on fs.

diizzyy:
Closing this as I have no further interest.

PR #1476, opened 2017-10-30
nbg6817: support booting from and updating the alternate boot partition(s)

pkgadd:
The nbg6817 is dual-boot capable and holds two independent sets of boot partitions (mmcblk0p[3-5] vs mmcblk0p[6-8]), this pull requests adds support for booting from the alternate and sysupgrading it. It does not yet toggle the bootflag and always (sys)upgrades the currently booted partition set, instead of the non-active one.

Toggling the dual-boot flag from a running system (OEM and LEDE) is [understood,](https://forum.lede-project.org/t/web-ui-to-reboot-to-another-partition-for-linksys-dual-partition-routers-and-to-power-off-power-down/3423/94) however it is not obvious so far how to manually toggle it in the failure case (non-booting firmware, button presses, etc.). Adding support for alternating the sysupgrade targets is trivial (and probably follows in a later pull request), but it would be nice if there'd be a generic configuration/ flag to choose between alternating or writing to the active boot partition set.

Signed-off-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>

### DIFF
--- a/target/linux/ipq806x/base-files/lib/upgrade/zyxel.sh
+++ b/target/linux/ipq806x/base-files/lib/upgrade/zyxel.sh
@@ -74,11 +74,22 @@ zyxel_do_upgrade() {
 
 	[ -b "${rootfs}" ] || return 1
 	case "$board" in
-	nbg6817)
-		kernel=mmcblk0p4
+		nbg6817)
+			case "$rootfs" in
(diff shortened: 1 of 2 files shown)


blogic:
merged, Thanks !

PR #1477, opened 2017-10-30
[RFC] [WIP] Mainlining netdev led triggers

BWhitten:
This PR is a work in progress cleanup of the netdev led triggers with a view to upstreaming, as such the prefix is wrong atm (and wrong for LEDE).
I have tried to address the various issues raised on the old patch submissions to the mailing list [1] and again [2].
My aim will be to have a round of reviews and tweaks here then go ahead and engage on the netdev/kernel mailing list.

An overview:
- One sysfs entry per parameter raised by Greg KH, for link, tx, and rx.
- Cleaning up the error checking in registration and unwind.
- Some suggestions on reformatting the logic.
- Converted the mode to flags to use atomics and avoid locks.
- Moving trigger to trigger subdirectory in kernel.

What could be looked over
- Locking, needs double checking.

[1] https://lkml.org/lkml/2010/11/14/116
[2] https://lkml.org/lkml/2010/11/28/187

### DIFF
--- a/package/base-files/files/etc/init.d/led
+++ b/package/base-files/files/etc/init.d/led
@@ -70,8 +70,11 @@ load_led() {
 		case "$trigger" in
 		"netdev")
 			[ -n "$dev" ] && {
+				local m
 				echo $dev > /sys/class/leds/${sysfs}/device_name
-				echo $mode > /sys/class/leds/${sysfs}/mode
+				for m in $mode; do
+					echo 1 > /sys/class/leds/${sysfs}/$m
+				done
 			}
 			;;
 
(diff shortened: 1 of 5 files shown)


blogic:
@BWhitten LGTM, thanks for investing the time. !

BWhitten:
For reference the patch is on the linux-kernel, linux-leds, and netdev mailing lists for review: https://www.spinics.net/lists/kernel/msg2660008.html

mkresin:
@BWhitten I guess it is fine to close the PR? Your v5 looks pretty nice and earlier version collected already some ACKs.

BWhitten:
Yep, all good this can be closed.
When it goes in I'll get a backport patch up on PR.

PR #1482, opened 2017-11-01
ar71xx: mach-rbspi: fix 74x164 support

f00b4r0:
The platform data was missing the num_registers element which is now
mandatory in linux 4.9

Without this patch, the gpio probing would fail with:
`gpio gpiochip1: (74x164): tried to insert a GPIO chip with zero lines`

Fixes: FS#1106

Tested-by: Robert Marko <robimarko@gmail.com>


mkresin:

PR #1457, opened 2017-10-25
ramips: add support to Barionet 1000 HW 1.1.A9

murilobarix:
Thanks for your contribution to the LEDE project!

To help keep the codebase consistent and readable,
and to help people review your contribution,
we ask you to follow the rules you find in the wiki at this link
https://lede-project.org/submitting-patches

Please remove this message before posting the pull request.

PR #1483, opened 2017-11-01
ramips: base-files: alphabetical order + merge cases

p-wassi:
@mkresin  As PR #1457 brought my attention to the base-files of ramips, here comes the alphabetical
re-ordering of the cases in ramips' base-files. Also merge the double cases.
(Btw: learned my lesson on that PR ;-) )

However, there are further things to discuss (I'd merge the commits if applicable):
1.) merge explicitly tagged with (implicitly) tagged switch:
`"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6@eth0"` with
`"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6t@eth0"`
This should be doable without risk - IIRC, there was some discussion going on, that for the first
case (not explicitly tagged), the tag is automatically added for the CPU port.

2.) merge switches with same port configuration in different order:
`"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6t@eth0"` with
`"1:lan" "2:lan" "3:lan" "4:lan" "0:wan" "6t@eth0"`.
The port numbers are the same, but I'd expect that during first network setup,
WAN is assigned eth0.1 in the first case, and eth0.2 in the second case.
(LAN is the other VLAN vice versa)
Is this ok for a merge, as I do not alter **existing** configs?

3.) In [1], @blogic  added support for two (router) devices and introduced a hotplug script
for usb/motion (?). What's the purpose of this script? (It just prints a line to the syslog)
Maybe this script can be moved to the motion package if it's needed there,
similar to the hotplug script of mjpeg-streamer at [2]

[1]:
https://git.lede-project.org/?p=source.git;a=commit;h=43de7c1cfa8b6afd15c226676ec64a0fa765fc6b

[2]:
https://git.lede-project.org/?p=feed/packages.git;a=blob;f=multimedia/mjpg-streamer/files/mjpg-streamer.hotplug;hb=HEAD

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -139,6 +139,7 @@ ramips_setup_interfaces()
 		;;
 	ar670w|\
 	ar725w|\
+	kn_rf|\
 	rt-ac51u)
 		ucidef_add_switch "switch0" \
 			"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6t@eth0"
@@ -255,10 +256,6 @@ ramips_setup_interfaces()
 	y1s)
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "5:lan" "0:wan" "6@eth0"
-		;;
-	kn_rf)
-		ucidef_add_switch "switch0" \
-			"0:wan" "1:lan" "2:lan" "3:lan" "4:lan" "6@eth0"
 		;;
 	kng_rc)
 		ucidef_add_switch "switch1" \
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -48,14 +48,14 @@ ramips_setup_interfaces()
 	dcs-930|\
 	dcs-930l-b1|\
 	ht-tm02|\
-	linkits7688 | \
-	linkits7688d | \
+	linkits7688|\
+	linkits7688d|\
 	m2m|\
 	microwrt|\
 	mpr-a2|\
 	ncs601w|\
-	omega2 | \
-	omega2p | \
+	omega2|\
+	omega2p|\
 	timecloud|\
 	w150m|\
 	widora-neo|\
@@ -176,8 +176,8 @@ ramips_setup_interfaces()
 	ubnt-erx-sfp|\
 	ur-326n4g|\
 	wrtnode|\
-	wrtnode2p | \
-	wrtnode2r | \
+	wrtnode2p|\
+	wrtnode2r|\
 	wt3020-4M|\
 	wt3020-8M|\
 	zbt-wa05)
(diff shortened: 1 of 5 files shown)


mkresin:
> 1.) merge explicitly tagged with (implicitly) tagged switch:

ACK to this change. But shouldn't it be done for the whole file? The related commits are

- 73d923e base-files: emit tagged switch configuration by default
- 5e0441a base-files: uci-defaults: support requesting untagged switch port configuration

Please keep these changes in an extra commit.

> 2.) merge switches with same port configuration in different order:

NAK  to this change. As far as I can see, the wan on eth0.1 is aligned to the WLLLL portmap. The most left port type will be eth0.1, all other eth0.2. The portmap is used to isolate ports right at the time the switch driver is loaded. Sure, it is possible to overwrite/redefine the vlans from userspace, but it might be confusing. Furthermore, I'm afraid of introducing regressions - which I can't think of at the moment - with this change.

One issue we have for sure with this change applied are different configs. Boards using/preserving their old config will have eth0.2 for lan. New installation will have eth0.1 for lan. This mismatch will most likely create some confusion if it comes to bugreports. A proper way to prevent this, is an update of existing configs using a migration script in target/linux/ramips/base-files/etc/uci-defaults. But I doubt it is worth the work.

> In 43de7c1cfa8b6afd15c226676ec64a0fa765fc6b @blogic added support for two (router) devices and introduced a hotplug script for usb/motion (?)

This commit seem to me more or less unrelated to the cleanup. I would suggest to send this question to lede-dev@lists.infradead.org. Given the amount of notification generated by github, it is very likely that the people who know about the reason for this file, never saw your well hidden question.

pepe2k:
@p-wassi,

I did similar cleanup in ar71xx some time ago but I split reordering and style fixes into two separate commits, have a look: https://github.com/lede-project/source/commit/a75cfac1f681776eb17e9beb238e1c81fbf82c43, https://github.com/lede-project/source/commit/edcdb93467e7a4e8373ca980dc15fa36c83d178a. With this approach it was easier for me to check for possible mistakes. Are you still able to split this (maybe you have separated commits in your local tree)?

Cheers,
Piotr

mkresin:
> but I split reordering and style fixes into two separate commits

I was going to suggest the same.

I have some ramips base files cleanup commits in one of my local branches but never managed to finish them. Might be a good idea to compare your and my result to spot any mistakes. I've pushed what I have so far to the [ramips branch in my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=shortlog;h=refs/heads/ramips).

p-wassi:
@mkresin 
Regarding your commit 12017532b51cc77f39046190ce5b9b50357404c7 (merge duplicated LEDs):
What I spotted (still duplicate cases):
- vocore2lite
- mir3g
(This does not cover style and alphabetical order!)

@pepe2k 
I'll maybe make another branch locally and split (style+alphabetical) vs. duplicate code
 merging for better reviewability. But, I'd better wait for @mkresin 's commits to be merged into
master..?

mkresin:
> Regarding your commit 12017532b51cc77f39046190ce5b9b50357404c7 (merge duplicated LEDs):
>What I spotted (still duplicate cases):
>
> -   vocore2lite
> -    mir3g
>    (This does not cover style and alphabetical order!)

As said, work in progress and never finished.

> But, I'd better wait for @mkresin 's commits to be merged into master..?

No, just go ahead! Use what ever you need/consider as valuable from my commits.

I had them in my local tree since April 2017 and never managed to finish them. Don't expect them to be committed to master nearby soon (if ever). I've just pushed them to my staging tree to have them at a place where I can link to.

PR #1484, opened 2017-11-01
gcc: libgcj removed in >=gcc 7

antonlacon:
Prevent selecting gcc's java compiler when using GCC 7, as it no longer exists in that version.


nbd168:
I've removed libgcj support entirely in my staging tree. It was incomplete anyway.

PR #1485, opened 2017-11-02
xtables-addons: add package for geoip scripts

pprindeville:
The iptables-mod-geoip is usually used in conjunction with some
wrapper scripts which manipulate the GeoIP database and then kick out
one or more iptables rules.  This package contains (1) the script to
download the most recent version of the MaxMind freemium database and
(2) another script which mangles the database into sets up iptables
rules.

### DIFF
--- a/package/network/utils/xtables-addons/Makefile
+++ b/package/network/utils/xtables-addons/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=xtables-addons
 PKG_VERSION:=2.14
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_HASH:=d215a9a8b8e66aae04b982fa2e1228e8a71e7dfe42320df99e34e5000cbdf152
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
@@ -120,6 +120,25 @@ define Package/iptaccount/install
 endef
 
 
+define Package/iptgeoip
+  $(call Package/xtables-addons)
+  CATEGORY:=Network
+  TITLE:=iptables-mod-geoip support scripts for MaxMind GeoIP databases
+  # we could also use wget-nossl but that's more complicated than our
+  # syntax of dependencies permits...
+  DEPENDS:=iptables +iptables-mod-geoip \
+		+perl +perlbase-getopt +perlbase-io +perl-text-csv_xs \
+		+!BUSYBOX_CONFIG_WGET:wget +!BUSYBOX_CONFIG_GZIP:gzip +!BUSYBOX_CONFIG_UNZIP:unzip
+endef
+
+define Package/iptgeoip/install
+	$(INSTALL_DIR) $(1)/usr/lib/xtables-addons
+	$(CP) \
+		$(PKG_INSTALL_DIR)/usr/lib/xtables-addons/xt_geoip_{build,dl} \
+		$(1)/usr/lib/xtables-addons/
+endef
+
+
 #$(eval $(call BuildTemplate,SUFFIX,DESCRIPTION,EXTENSION,MODULE,PRIORITY,DEPENDS))
 
 $(eval $(call BuildTemplate,compat-xtables,API compatibilty layer,,compat_xtables,+IPV6:kmod-ip6tables))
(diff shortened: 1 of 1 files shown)


pprindeville:
Note this requires [PR #5052](https://github.com/openwrt/packages/pull/5052).

NeoRaider:
NAK. A package in the LEDE base repo can't depend on a community package.

NeoRaider:
I wonder if it would make sense to move the whole xtables-addons package to openwrt/packages?

package/network/utils/xtables-addons/Makefile
```diff
@@ -120,6 +120,23 @@ define Package/iptaccount/install
 endef
 
 
+define Package/iptgeoip
+  $(call Package/xtables-addons)
+  CATEGORY:=Network
+  TITLE:=iptables-mod-geoip support scripts for MaxMind GeoIP databases
+  DEPENDS:=iptables +iptables-mod-geoip \
+		+perl +perlbase-getopt +perlbase-io +perl-text-csv_xs \
+		+wget +gzip +unzip
```
pprindeville: Dumb question, but how do you specify a dependency on `wget` (either standalone wget, or Busybox wget, or ...)?

pprindeville: > NAK. A package in the LEDE base repo can't depend on a community package.

Well, I could rely on Busybox's wget, unzip, and gunzip and problem solved...

bobafetthotmail: LEDE has a uclient-fetch package that disguises as wget and is installed by default.

NeoRaider: * wget: just depend on uclient-fetch like opkg does
* gzip: is enabled by default in busybox. We generally expect users not to remove busybox applets without replacement, so you can just ignore this dependency
* unzip: you can use a conditional dependency like in https://github.com/freifunk-gluon/packages/blob/6a2d2880064ae580c9c39cd691d4fb76c420412d/admin/autoupdater/Makefile#L13

The remaining issue is perl. Perl is in openwrt/packages, so you can't depend on it from a base package, therefore my suggestion to move the whole thing to openwrt/packages.

pprindeville:
> LEDE has a uclient-fetch package that disguises as wget and is installed by default, you can probably specify dependency for that instead than for wget.

@bobafetthotmail Unfortunately, `uclient-fetch` is broken and doesn't work correctly with the scripts.  I found this out last night and had to spend an hour debugging it.  I've submitted a fix for it, but I don't know if it's complete or how long it will take for a new release of `uclient-fetch` to hit LEDE.

> wget: just depend on uclient-fetch like opkg does

@NeoRaider This is a larger problem, though...  That of _meta packages_ or capabilities.  If there are 3 packages which all provide a `wget` utility, it would be nice to be able to depend on the capability while being provider- (or package-) agnostic.  Similarly, if 2 or more packages provide the same capability, it would also be good to detect the potential conflict.

As it is, I'm building with `/bin/wget` from `uclient-fetch` and `/usr/bin/wget` from Busybox.

> The remaining issue is perl. Perl is in openwrt/packages, so you can't depend on it from a base package, therefore my suggestion to move the whole thing to openwrt/packages.

I'm okay with that, but it's not my call.  Indeed, I wondered myself why `xtables-addons` wasn't put into `packages`.

> unzip: you can use a conditional dependency like [...]

If I wanted to call out an explicit dependency on `wget`, where there are 3 different versions of `wget`, how would I do that?  There's `BUSYBOX_CONFIG_WGET`, then there's `wget` and `wget-nossl`.  In the script's case (currently, anyway, this may change), the URL's are `http:` so SSL isn't required, but you can of course use the SSL capable `wget` to get those files too...

NeoRaider:
@pprindeville We have initial support for "alternatives" in LEDE for a while (still used by only a few packages). Another step towards easily depending on "anything providing a specific feature" is better PROVIDES support (implemented in https://git.lede-project.org/?p=lede/neoraider/staging.git;a=commit;h=084a4105cab18dd59d0b6616d70d10a789dd4b27 , but unfortunately bitrot by now, I hope I'll find time to rebase it onto master soon). Both features will need to be implemented by the specific packages you want to depend on as well.

The uclient-fetch fix is in the LEDE master already, so just depend on that like opkg. Let's not make this more complicated than necessary.

pprindeville:
So... I'm guessing this PR is on hold until the project gets moved to `packages` instead.  Is that correct?

NeoRaider:
@pprindeville Correct.

@jow-, you're the maintainer of this package. ACK for moving the package to openwrt/packages to allow depending on perl?

pprindeville:
@bobafetthotmail The `uclient-fetch` issue has been fixed.

@jow- Can we get a call on moving `xtables-addons` to `packages`?

pprindeville:
@jow- Hi Jo.  @hauke and I were discussing whether `account` (`xt_ACCOUNT`) should select `iptaccount`... since it's useless without it.  And whether `geoip` should also select `iptgeoip`.  What do you think?

pprindeville:
Moved to https://github.com/openwrt/openwrt/pull/602.

PR #1486, opened 2017-11-02
6rd: use pure shell 6rdcalc

yangfl:
Signed-off-by: David Yang <mmyangfl@gmail.com>


dedeckeh:
What's the added value to use a 6rdcalc sh implementation compared to the C utility ?
The 6rdcalc sh implementation fails to convince me to make the switch as the implementation is more bloated and slower than the 6rdcalc C utility

mkresin:
Even if I love to do stuff in pure shell, the shell code is way harder to read than the existing c code. I don't see the benefit either.

yangfl:
Shell codes are much more portable compared to C, and that's not so slow according to my test. If you are really care about speed then you won't write a single bit shell.
As for the style, I can clean it up further and I'm looking for your advice.
@mkresin

dedeckeh:
We've invested quite a lot in the past to write C daemons/utilities to replace shell code this is a move in the opposite direction which is not the way we want to go. So for me this is a NACK

mkresin:
I don't see how all the invocation if `sed` and `awk` can be made more readable. I mean, stuff like

```
eval $1=$(echo $2 | cut -d/ -f1 | sed "s/::/$(echo $2 |
	awk -F':' '{for(i=1;i<=10-NF;i++)printf ":"}')/" |
	awk -F':' '{for(i=1;i<=NF;i++)x=x""sprintf("%04s", $i);gsub(/ /,"0",x);printf x}')
```

and the bin_to_hex/hex_to_bin functions will be a pain to debug if something doesn't work as expected.

As long as we don't have any portability issues with the C code, I don't see a reason for replacing a working and well tested solution.

PR #1488, opened 2017-11-02
hostapd: reduced size of mini versions disabling debug

BigNerd95:
1) Allows hostapd and wpa_supplicant to be launched with `-s` flag, even when compiled without `CONFIG_DEBUG_SYSLOG` flag.

2) Disable debug log in mini versions to decrease size:
wpa_supplicant-mini: reduced of `21 KB`
hostapd-mini: reduced of `25 KB`
wpa-mini: reduced of `34 KB`

### DIFF
--- a/package/network/services/hostapd/Makefile
+++ b/package/network/services/hostapd/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=hostapd
-PKG_RELEASE:=4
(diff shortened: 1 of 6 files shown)


BigNerd95:
wpad-mini now is 350 KB

hauke:
You can please compare this to setting the "CONFIG_WPA_MSG_MIN_PRIORITY" option to 4 for example. It would be nice to have some more comparisons.

PR #1489, opened 2017-11-02
netdevices.mk: add niu device

lucize:
Compile/Run tested: x86_64 Reboot (SNAPSHOT, r5226-986d9deb3b)

### DIFF
--- a/package/kernel/linux/modules/netdevices.mk
+++ b/package/kernel/linux/modules/netdevices.mk
@@ -304,6 +304,22 @@ endef
 $(eval $(call KernelPackage,r6040))
 
 
+define KernelPackage/niu
+  SUBMENU:=$(NETWORK_DEVICES_MENU)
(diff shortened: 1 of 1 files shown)


package/kernel/linux/modules/netdevices.mk
```diff
@@ -559,6 +575,25 @@ endef
 $(eval $(call KernelPackage,ixgbe))
 
 
+define KernelPackage/ixgbevf
+  SUBMENU:=$(NETWORK_DEVICES_MENU)
```
hauke: You add two packages , but the commit title only says you add one. Please extend the commit message

lucize:
I thought about that too, I'll split them

lucize:
I'll make another pr only with the sort because is giving me branch conflict if I do it in this one and I'm not so good at git

hauke:

PR #1491, opened 2017-11-03
build: allow defining license information per binary package

mhei:
At the moment, license information can only be specified on a
"per source package" level while other metadata fields (e.g. maintainer)
can be given for each binary package. Apply the same logic for license
fields as well. This can be used e.g. in cases where a library is
distributed under some license while related tools are distributed
under a different one.

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


hauke:
Thank you for the patch it was added to master.

PR #1493, opened 2017-11-05
[RFC] ar71xx: Fix 5GHz wifi for Sitecom WLR-8100 and mac addresses

DavideFioravanti:
This patch adds Sitecom WLR-8100 to ath10k caldata extraction list and fix the random MAC
address for LAN, WAN and 5GHz interface by extracting it from the u-boot-env partition.

In the stock firmware 2.4GHZ and LAN interface had the same mac address.

It's stored only in ascii format so it must be obtained using the mtd_get_mac_ascii function.

While this method is quite common for LAN/WAN and 5GHz wireless interface, I didn't found
any device using this method for 2.4GHz interface. So if someone could give me an advice how
to do this, I could add it in a future commit.

(The first part of this patch is similar to https://github.com/lede-project/source/pull/1392
but it isn't still merged so I have to commit again to fix the mac address. I don't know if
it's the right way).

Signed-off-by: Davide Fioravanti <pantanastyle@gmail.com>

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -544,6 +544,10 @@ ar71xx_setup_macs()
 	mynet-n750)
 		wan_mac=$(mtd_get_mac_ascii devdata "wanmac")
 		;;
+	wlr8100)
+		lan_mac=$(mtd_get_mac_ascii u-boot-env "ethaddr")
+		wan_mac=$(mtd_get_mac_ascii u-boot-env "wanaddr")
+		;;
 	wpj344|\
 	wpj558)
 		wan_mac=$(mtd_get_mac_binary u-boot 0x2e018)
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom
@@ -91,6 +91,10 @@ case "$FIRMWARE" in
 		ath9k_eeprom_extract "caldata" 4096 2048
 		ath9k_patch_firmware_mac $(mtd_get_mac_binary caldata 0)
 		;;
+	wlr8100)
+                ath9k_eeprom_extract "art" 8192 2048
+                ath9k_patch_firmware_mac $(mtd_get_mac_ascii u-boot-env "ethaddr")
+                ;;
 	rambutan)
 		ath9k_eeprom_extract "art" 4096 2048
 		;;
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom
@@ -92,7 +92,7 @@ case "$FIRMWARE" in
 		ath9k_patch_firmware_mac $(mtd_get_mac_binary caldata 0)
 		;;
 	wlr8100)
-                ath9k_eeprom_extract "art" 8192 2048
+                ath9k_eeprom_extract "caldata" 4096 2048
                 ath9k_patch_firmware_mac $(mtd_get_mac_ascii u-boot-env "ethaddr")
                 ;;
 	rambutan)
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -135,6 +135,10 @@ case "$FIRMWARE" in
 	rb-962uigs-5hact2hnt)
 		ath10kcal_from_file "/sys/firmware/routerboot/ext_wlan_data" 20480 2116
 		;;
+	wlr8100)
+		ath10kcal_extract "art" 24576 2116
+		ath10kcal_patch_mac $(macaddr_add $(mtd_get_mac_ascii u-boot-env ethaddr) +1)
+		;;
 	esac
 	;;
 "ath10k/pre-cal-pci-0000:00:00.0.bin")
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -136,7 +136,7 @@ case "$FIRMWARE" in
 		ath10kcal_from_file "/sys/firmware/routerboot/ext_wlan_data" 20480 2116
 		;;
 	wlr8100)
-		ath10kcal_extract "caldata" 2048 2116
+		ath10kcal_extract "caldata" 20480 2116
 		ath10kcal_patch_mac $(macaddr_add $(mtd_get_mac_ascii u-boot-env ethaddr) +1)
 		;;
 	esac
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -136,7 +136,7 @@ case "$FIRMWARE" in
 		ath10kcal_from_file "/sys/firmware/routerboot/ext_wlan_data" 20480 2116
 		;;
 	wlr8100)
-		ath10kcal_extract "art" 24576 2116
+		ath10kcal_extract "caldata" 2048 2116
 		ath10kcal_patch_mac $(macaddr_add $(mtd_get_mac_ascii u-boot-env ethaddr) +1)
 		;;
 	esac
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-wlr8100.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-wlr8100.c
@@ -49,11 +49,6 @@
 #define WLR8100_KEYS_POLL_INTERVAL	20	/* msecs */
 #define WLR8100_KEYS_DEBOUNCE_INTERVAL	(3 * WLR8100_KEYS_POLL_INTERVAL)
 
-#define WLR8100_MAC0_OFFSET		0
-#define WLR8100_MAC1_OFFSET		6
-#define WLR8100_WMAC_CALDATA_OFFSET	0x1000
-#define WLR8100_PCIE_CALDATA_OFFSET	0x5000
-
 static struct gpio_led wlr8100_leds_gpio[] __initdata = {
 	{
 		.name		= "wlr8100:amber:status",
@@ -138,7 +133,6 @@ static struct mdio_board_info wlr8100_mdio0_info[] = {
 
 static void __init wlr8100_common_setup(void)
 {
-	u8 *art = (u8 *) KSEG1ADDR(0x1fff0000);
 
 	ath79_register_m25p80(NULL);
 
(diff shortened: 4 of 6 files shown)


mkresin:
> While this method is quite common for LAN/WAN and 5GHz wireless interface, I didn't found
any device using this method for 2.4GHz interface. So if someone could give me an advice how
to do this, I could add it in a future commit.

I'm not sure if I understand your question correctly, so this could be rubbish.

Assuming your 2.4GHz wireless is an ath9k chip, drop the `ath79_register_wmac()` call from `target/linux/ar71xx/files/arch/mips/ath79/mach-wlr8100.c` and add support for your board to `target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom`.

You can use the `ath9k_patch_firmware_mac` shell function to patch the mac address of the 2.4GHz caldata.

DavideFioravanti:
> I'm not sure if I understand your question correctly, so my reply could be rubbish.

Yes, you understood perfectly my question! 👍

> You can use the ath9k_patch_firmware_mac shell function to patch the mac address of the 2.4GHz caldata.

Thank you. I managed to get it working using your method with these steps:
_target/linux/ar71xx/files/arch/mips/ath79/mach-wlr8100.c_
1. Replaced `ath79_register_wmac()` with `ath79_register_wmac_simple()`
2. Removed `ath79_init_mac()`
3. Replaced `ap91_pci_init()` with `ap91_pci_init_simple()`

_target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom_
Added these lines in the `soc_wmac.eeprom` case:
```
        wlr8100)
                ath9k_eeprom_extract "art" 8192 2048
                ath9k_patch_firmware_mac $(mtd_get_mac_ascii u-boot-env "ethaddr")
```

However, digging in the wlr8100 code, i probably found a bigger problem that stopped me from adding this new patch. I was curious about the strange addresses of art partition with this board and I found out that the flash layout of this board seems to be wrong.
This is the original OEM flash layout of the router that I extracted (I have also added it to the openwrt wiki) :
```
# cat /proc/mtd
dev:    size   erasesize  name
mtd0: 01000000 00010000 "ALL"		= ‭16.777.216‬	= 16384k
mtd1: 00030000 00010000 "Bootloader"	=    196.608	=   192k
mtd2: 00010000 00010000 "Config"	=     ‭65.536‬	=    64k
mtd3: 00f10000 00010000 "Kernel"	= ‭15.794.176	= ‬15424k
mtd4: 00c40000 00010000 "apps"		= ‭12.845.056‬	= 12544k
mtd5: 00040000 00010000 "manufacture"	= ‭   262.144‬	=   256k
mtd6: 00010000 00010000 "backup"	=     ‭65.536‬	=    64k
mtd7: 00050000 00010000 "storage"	=‭    327.680‬	=   320k
mtd8: 00010000 00010000 "caldata"	=     ‭65.536‬	=    64k
```
(`apps` partition should be inside `Kernel`)
This is the current Lede layout:
`wlr8100_mtdlayout=mtdparts=spi0.0:192k(u-boot)ro,64k(u-boot-env)ro,1408k(kernel),14080k(rootfs),192k(unknown)ro,64k(art)ro,384k(unknown2)ro,15488k@0x40000(firmware)`
With this layout the `firmware` partition overwrite partially `manufacture`, and we use the `backup` partition (which contains from 0 to 0xfff a backup of `u-boot-env` and from 0x1000 to 0xffff a backup of `caldata`) as `art` partition instead of using the right one. Thats why my addresses are shifted by 0xfff (4096 byte) compared to the other router.

So I think that the right flash layout should be:
`wlr8100_mtdlayout=mtdparts=spi0.0:192k(u-boot)ro,64k(u-boot-env)ro,1408k(kernel),14016k(rootfs),256k(manufacture)ro,64k(backup)ro,320k(storage)ro,64k(caldata)ro,15424k@0x40000(firmware)`

Now, what do you think I should do? Add the patches to get the real mac addresses and then create a new PR to correct the flash layout (editing the addresses in my patches)? Or close this PR, edit the flash layout and then create again the patches to get the mac address?

However changing the flash layout could create problems to the people that already have flashed Lede on their WLR8100?

If you want to double check my findings, I can provide you all the mtd partition that I extracted from the original firmware.
I am sorry if I'm bothering you with all these questions and this "old" router, but I am currently learning how LEDE works, but I hope to help better this project in the future. :)

mkresin:
> So I think that the right flash layout should be:
wlr8100_mtdlayout=mtdparts=spi0.0:192k(u-boot)ro,64k(u-boot-env)ro,1408k(kernel),14016k(rootfs),256k(manufacture)ro,64k(backup)ro,320k(storage)ro,64k(caldata)ro,15424k@0x40000(firmware)

Even if the size of the firmware/rootfs partition is changed, it shouldn't cause any issues. So please do the change.

> Now, what do you think I should do? Add the patches to get the real mac addresses and then create a new PR to correct the flash layout (editing the addresses in my patches)? Or close this PR, edit the flash layout and then create again the patches to get the mac address?

Please add multiple commits to this PR/split your changes into multiple commits:

1. fix the partition layout and adjust the offsets accordingly in the mach file
2. fix the eeprom and mac addresses
3. add kmod-ath10k ath10k-firmware-qca988x

I had a look at the mach file and it might not be necessary to do the fixups in userspace. You can pass a mac address with the call to `ath79_register_wmac`/`ap91_pci_init` as it can be seen in https://git.lede-project.org/?p=source.git;a=blob;f=target/linux/ar71xx/files/arch/mips/ath79/mach-mc-mac1200r.c line 110 and following. Use  `ath79_init_mac()` to create a (temporary) variable with the mac address incremented.

I'm not entirely sure, but it might be possible to pass a pointer to the ath10k caldata as well. It rather looks like the WLR8100_PCIE_CALDATA_OFFSET variable has a wrong value. According to your change the ath10k caldata is at offset 24576 (0x6000) of the art partition. But WLR8100_PCIE_CALDATA_OFFSET is set to 0x5000.

DavideFioravanti:
> I had a look at the mach file and it might not be necessary to do the fixups in userspace.

I've already tried to do that, but I had a huge problem: As far as I know, there isn't a method to extract the mac address stored as an ASCII string inside the u-boot-env partition without using userspace methods, and unfortunately the mac address stored in caldata is not the right one.

> I'm not entirely sure, but it might be possible to pass a pointer to the ath10k caldata as well.

Same problem as above.

> According to your change the ath10k caldata is at offset 24576 (0x6000) of the art partition. But WLR8100_PCIE_CALDATA_OFFSET is set to 0x5000.

WLR8100_PCIE_CALDATA_OFFSET value was correct and that's why:

Remembering the right flash layout:
```
[    3.650907] 0x000000000000-0x000000030000 : "u-boot"
[    3.657088] 0x000000030000-0x000000040000 : "u-boot-env"
[    3.663726] 0x000000040000-0x0000001a0000 : "kernel"
[    3.670051] 0x0000001a0000-0x000000f50000 : "rootfs"
[    3.676366] mtd: device 3 (rootfs) set to be root filesystem
[    3.682152] mtdsplit: no squashfs found in "rootfs"
[    3.687158] 0x000000f50000-0x000000f90000 : "manufacture"
[    3.693885] 0x000000f90000-0x000000fa0000 : "backup" -> Old (and wrong) caldata partition
[    3.700212] 0x000000fa0000-0x000000ff0000 : "storage"
[    3.706638] 0x000000ff0000-0x000001000000 : "caldata" -> Right caldata partition starting at 16.711.680 with 5GHz starting at 16.732.160
[    3.713006] 0x000000040000-0x000000f50000 : "firmware"
```
Code inside mach-wlr8100.c
```
u8 *art = (u8 *) KSEG1ADDR(0x1fff0000);  -> FF0000 = 16.711.680
ath79_register_wmac(art + WLR8100_WMAC_CALDATA_OFFSET, NULL); -> FF5000 = ‭16.732.160‬
```
So in this case we were using the right address for the calibration data in the right partition (the latest)
while in the flash layout, the caldata partition was wrong (it was the backup one, containing the u-boot-env and caldata backup)

I hope to have explained well the situation.

In the meantime I added the patches to fix the 2.4GHz mac address and then I fixed the partition layout with the right name.

pepe2k:
@DavideFioravanti,

I have merged your patches into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) with several changes. I also included your patch from #1392 as I think it got lost somewhere and hasn't been actually merged. I left `art` name for calibration partition as it's a general rule to name it this way even if vendor firmware uses different one.

Would you please test the changes from my tree? I moved this device from legacy to generic building code and dropped kernel/rootfs partitions - this will bring back snapshot images for this device.

Cheers,
Piotr

DavideFioravanti:
@pepe2k 

Thank you. I'll test your changes in the following days! :)

DavideFioravanti:
@pepe2k 

I compiled your staging tree and everything seems good except for:

1. the patch for `target/linux/ar71xx/base-files/etc/board.d/02_network` is missing (this patch gives the correct mac addresses for LAN/WAN ports)
2. in `/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom` the address should be 4096 instead of 8192

For what concerns mtd partitions it seems to detect automatically the orginal flash layout inside the firmware partition plus another small partition (rootfs_data) but I don't know if this is a normal behaviour:

```
[    3.614333] m25p80 spi0.0: found mx25l12805d, expected m25p80
[    3.620738] m25p80 spi0.0: mx25l12805d (16384 Kbytes)
[    3.625923] 7 cmdlinepart partitions found on MTD device spi0.0
[    3.631922] Creating 7 MTD partitions on "spi0.0":
[    3.636798] 0x000000000000-0x000000030000 : "u-boot"
[    3.642995] 0x000000030000-0x000000040000 : "u-boot-env"
[    3.649649] 0x000000040000-0x000000f50000 : "firmware"
[    3.682830] 2 uimage-fw partitions found on MTD device firmware
[    3.688871] 0x000000040000-0x000000310000 : "kernel"
[    3.694940] 0x000000310000-0x000000f50000 : "rootfs"
[    3.701254] mtd: device 4 (rootfs) set to be root filesystem
[    3.707069] 1 squashfs-split partitions found on MTD device rootfs
[    3.713340] 0x000000740000-0x000000f50000 : "rootfs_data"
[    3.720065] 0x000000f50000-0x000000f90000 : "manufacture"
[    3.726862] 0x000000f90000-0x000000fa0000 : "backup"
[    3.733121] 0x000000fa0000-0x000000ff0000 : "storage"
[    3.739556] 0x000000ff0000-0x000001000000 : "art"
```

I did my tests loading an initramfs image in a never flashed WLR-8100.

pepe2k:
@DavideFioravanti

> 1. the patch for target/linux/ar71xx/base-files/etc/board.d/02_network is missing (this patch gives the correct mac addresses for LAN/WAN ports)

Fixed!

> 2. in /target/linux/ar71xx/base-files/etc/hotplug.d/firmware/10-ath9k-eeprom the address should be 4096 instead of 8192

Fixed!

> For what concerns mtd partitions it seems to detect automatically the orginal flash layout inside the firmware partition plus another small partition (rootfs_data) but I don't know if this is a normal behaviour:

Yes, that's expected.

Thanks for test! I updated my staging tree.
Let me know if you have any other comments.

Cheers,
Piotr

DavideFioravanti:
@pepe2k 

> Thanks for test! I updated my staging tree.
Let me know if you have any other comments.

I tested your updated tree and now everything seems perfect! :)

In the future I'll try to fix the port disposition and the leds so the porting will be complete, but for now we can close this PR.

Thank you for your help! :)

pepe2k:
@DavideFioravanti,

> I tested your updated tree and now everything seems perfect! :)

Again, thanks for test and feedback.

> In the future I'll try to fix the port disposition and the leds so the porting will be complete, but for now we can close this PR.

OK, feel free to open another PR with more fixes.

Cheers,
Piotr

PR #1494, opened 2017-11-06
ramips: add support for TP-Link TL-WR840N v5

robimarko:
TP-Link TL-WR840N v5 is simple N300 router with
5-port FE switch and non-detachable antennas, based on MediaTek MT7628NN (aka MT7628N) WiSoC.

Specification:

- MT7628N/N (580 MHz)
- 64 MB of RAM (DDR2)
- 4 MB of FLASH
- 2T2R 2.4 GHz
- 5x 10/100 Mbps Ethernet
- 2x external, non-detachable antennas
- UART (J1) header on PCB (115200 8n1)
- 1x LED (GPIO-controlled), 1x button

* LED in TL-WR840N v5 is a dual-color, dual-leads type which isn't
  (fully) supported by the gpio-leds driver. This type of LED requires both
  GPIOs state change at the same time to select a color or turn it off.
  For now, we support/use only the green part of the LED.
 Orange LED is registered so you can later use it for your own purposes.
  
  Flash instruction:
  
  Unlike TL-WR840N v4 flashing through WEB UI works in v5.
  1. Download lede-ramips-mt76x8-tl-wr840n-v5-squashfs-sysupgrade.bin image.
  2. Go to 192.168.0.1
  3. Flash the sysupgrade image through Firmware upgrade section of WEB UI.
  4. Wait until the green LED stops flashing and use the router.

Notes:
TFTP recovery is broken since TP-Link reused bootloader code for v4 and that does not take into account only 4 MB of flash and bricks the device.
So do not use TFTP Recovery or you will have to rewrite SPI flash.
They fixed it in later GPL code, but it is unknown which version of bootloader you have.

After manually compiling and flashing bootloader from GPL sources TFTP recovery works properly.

Signed-off-by: Robert Marko robimarko@gmail.com

### DIFF
--- a/target/linux/ramips/dts/TL-WR840NV5.dts
+++ b/target/linux/ramips/dts/TL-WR840NV5.dts
@@ -1,0 +1,111 @@
+/dts-v1/;
+
+#include "mt7628an.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "tplink,tl-wr840n-v5", "mediatek,mt7628an-soc";
+	model = "TP-Link TL-WR840N v5";
+
+  	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x4000000>;
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio1 6 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+	/* LED used is dual-color,dual lead LED */
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power {
+			label = "tl-wr840n-v5:green:power";
+			gpios = <&gpio1 9 GPIO_ACTIVE_LOW>;
+		};
+
+		orange {
+			label = "tl-wr840n-v5:orange:power";
+			gpios = <&gpio1 11 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,chunked-io = <32>;
+
+		partition@0 {
+			label = "boot";
+			reg = <0x0 0x20000>;
+			read-only;
+		};
+
+		partition@20000 {
+			label = "firmware";
+			reg = <0x20000 0x3c0000>;
+		};
+
+		partition@3e0000 {
+			label = "config";
+			reg = <0x3e0000 0x10000>;
+			read-only;
+		};
+
+		factory: partition@3f0000 {
+			label = "factory";
+			reg = <0x3f0000 0x10000>;
+			read-only;
+		};
+	};
+};
+
+&ehci {
+	status = "disabled";
+};
+
+&ohci {
+	status = "disabled";
+};
+
+&wmac {
+	status = "okay";
+	mtd-mac-address = <&factory 0xf100>;
+	mediatek,mtd-eeprom = <&factory 0x10000>;
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xf100>;
+	mediatek,portmap = "wllll";
+};
+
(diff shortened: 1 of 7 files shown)


target/linux/ramips/dts/TL-WR840NV5.dts
```diff
@@ -0,0 +1,111 @@
...
+#include "mt7628an.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "tplink,tl-wr840n-v5", "mediatek,mt7628an-soc";
+	model = "TP-Link TL-WR840N v5";
+
+  chosen {
```
mkresin: missing indention.

target/linux/ramips/dts/TL-WR840NV5.dts
```diff
@@ -0,0 +1,111 @@
...
+		reset {
+			label = "reset";
+			gpios = <&gpio1 6 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+	
+	gpio-leds {
+		compatible = "gpio-leds";
+
```
mkresin: please add a comment here that it is a dual colour led.

target/linux/ramips/dts/TL-WR840NV5.dts
```diff
@@ -0,0 +1,111 @@
...
+		power {
+			label = "tl-wr840n-v5:green:power";
+			gpios = <&gpio1 9 GPIO_ACTIVE_LOW>;
+		};
+
+		orange {
+			label = "tl-wr840n-v5:orange:orange";
```
mkresin: shouldn't it be `tl-wr840n-v5:orange:power`

target/linux/ramips/dts/TL-WR840NV5.dts
```diff
@@ -0,0 +1,111 @@
...
+
+&ethernet {
+	mtd-mac-address = <&factory 0xf100>;
+	mediatek,portmap = "llllw";
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "p0led_an", "p2led_an", "perst", "refclk", "wdt", "wled_an";
```
mkresin: only the pins of the perst, p0led_an and p2led_an group are used as gpio.  drop the other.

target/linux/ramips/dts/TL-WR840NV5.dts
```diff
@@ -0,0 +1,111 @@
...
+
+&wmac {
+	status = "okay";
+	mtd-mac-address = <&factory 0xf100>;
+	mediatek,mtd-eeprom = <&factory 0x10000>;
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xf100>;
+	mediatek,portmap = "llllw";
```
mkresin: According to your config in 02_network the wan port is on switch port 0. But your portmap says the wan port is on switch port 4. Which one is right?

robimarko:
@mkresin Thanks for reviewing.
Changes have been made as requested.

target/linux/ramips/dts/TL-WR840NV5.dts
robimarko: Port 0 is the WAN port.

mkresin: Than it has to be `wllll`

mkresin:
>  Unlike TL-WR840N v4 flashing through WEB UI works in v5.
>  1. Download lede-ramips-mt76x8-tl-wr840n-v5-squashfs-sysupgrade.bin image.
>  2. Go to 192.168.0.1
>  3. Flash the sysupgrade image through Firmware upgrade section of WEB UI.
>  4. Wait until green LED stops flashing and use the router.

Are you sure it works using the sysupgrade image? Why do we create a factory image if it works this way.

target/linux/ramips/dts/TL-WR840NV5.dts
robimarko: I have changed it to that

robimarko:
Yes, I am sure.
Tested it on mine and image got accepted and flashed without any hiccups.

mkresin:
Than drop the factory image from the image build code if it isn't required.

Please do a full test with the requested changes applied. Might be I've suggested something not working.

robimarko:
Code for factory image dropped.
I have built and tested the firmware before dropping the factory image code and everything worked.

Will test it again today and report back

robimarko:
@mkresin Tested latest version and everything works fine.

mkresin:
merged, thanks!

stan-thomas:
I happened to test the trunk version today and the wifi tx power is erratic. It is very less usually and keeps varying.

Anyone else facing this issue? 

Thanks. Much appreciated.

PR #1496, opened 2017-11-06
mac80211: mwl8k: expand 5G channels above 48

waveletboy:
mac80211: mwl8k: expand 5G channels above 48.

Originally only 4 channels (36, 40, 44, 48) are supported for 5G. Add upper channels including DFS and non-DFS channels.
Non DFS channel (149-165) can be used on AP mode. DFS channel (52-144) can be used on Client mode.

Discussion is in 
https://forum.openwrt.org/viewtopic.php?pid=367894#p367894
https://forum.lede-project.org/t/linksys-ea3500-enable-all-5g-channels/8365/2

### DIFF
--- a/package/kernel/mac80211/patches/941-mwl8k_expand_5g_channels.patch
+++ b/package/kernel/mac80211/patches/941-mwl8k_expand_5g_channels.patch
@@ -1,0 +1,39 @@
+--- a/drivers/net/wireless/marvell/mwl8k.c
++++ b/drivers/net/wireless/marvell/mwl8k.c
+@@ -199,7 +199,7 @@
+ 	struct ieee80211_channel channels_24[14];
+ 	struct ieee80211_rate rates_24[13];
+ 	struct ieee80211_supported_band band_50;
+-	struct ieee80211_channel channels_50[4];
++	struct ieee80211_channel channels_50[25];
+ 	struct ieee80211_rate rates_50[8];
+ 	u32 ap_macids_supported;
+ 	u32 sta_macids_supported;
+@@ -383,6 +383,27 @@
+ 	{ .band = NL80211_BAND_5GHZ, .center_freq = 5200, .hw_value = 40, },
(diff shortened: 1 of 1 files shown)


mkresin:
From https://forum.openwrt.org/viewtopic.php?pid=367894#p367894:

> Even you can choose channel 52 - 144 in luci for AP mode, the wlan1 interface will not enable because the chip doesn't support DFS.

If these channels require DFS to be used, you violate law by forcefully enable without using DFS. Furthermore, send such chances upstream first.

If another @lede-project/developer agrees, please close the PR.

waveletboy:
The change will not enable DFS channels on AP mode because the DFS check fails. But for client mode they can be used like all others. I can remove DFS channels anyway.
I will seek the opportunity to modify it in upstream. But not sure it is worthy of the time. People like me making such change and compiling only need less than 1 hour...

waveletboy:
@mkresin 
I am not confident to get the approval on upstream either because I actually don't know upper non-DFS channels can work for other routers than my EA3500. If anyone likes the change can just pick it up and make a personal build. 
Close the PR. Thanks!

waveletboy:
@mkresin 
I removed DFS channel and submit the patch to upstream as https://patchwork.kernel.org/patch/10060609/

Dantekazuma:
without this patch the EA3500 is useless device when Lede is installed , this should be included in the kernel in the 18.06.0 since on stock firmware it was not limited to only the 4 lowest 5ghz channels. this needs to be addressed.

bobafetthotmail:
Complaining here (or in general) will not solve the issue.
Open a bug report in https://bugs.openwrt.org/ or open a new PR yourself in OpenWrt github https://github.com/openwrt/openwrt  where you fix this and go through the process of actually getting it merged, unlike what waveletboy did.

waveletboy:
Actually, upstream has accepted [change](https://patchwork.kernel.org/patch/10060831/). I don't wanna bother but, just don't know what I can do better.

PR #1497, opened 2017-11-07
functions.sh: fix default_postinst function

ratkaj:
When we run "opkg install" on a package that installs an uci-defaults
script, functions.sh will fail to evaluate that script in its
default_postinst function.

This happens because there is no "./" present and it searches for the
file in paths specified by the PATH variable. This would work on bash,
but it will not work on ash and some other shells like sh, zsh. This
applys to the ". filename" directive used in this case.

This patch will make the path relative to the /etc/uci-defaults
directory.


mkresin:
Fixes [FS#1021](https://bugs.lede-project.org/index.php?do=details&task_id=1021)

mkresin:
merged thanks!

PR #1499, opened 2017-11-08
kernel: tune e1000e module parameters

pprindeville:
I think I wrote the original version and then @jow- reworked it.  In any case, here it is now that `include/kernel.mk` has been reworked.

### DIFF
--- a/package/kernel/linux/modules/netdevices.mk
+++ b/package/kernel/linux/modules/netdevices.mk
@@ -495,6 +495,9 @@ define KernelPackage/e1000e
   KCONFIG:=CONFIG_E1000E
   FILES:=$(LINUX_DIR)/drivers/net/ethernet/intel/e1000e/e1000e.ko
   AUTOLOAD:=$(call AutoProbe,e1000e)
+  MODPARAMS.e1000e:= \
+    IntMode=1 \
+    InterruptThrottleRate=4,4,4,4,4,4,4,4
 endef
 
 define KernelPackage/e1000e/description


diizzyy:
You probably want to mention what it does?

pprindeville:
Done.

pprindeville:
Anyone?

diizzyy:
Pardon my n00b-ness but I guess instances = physical ports or is it is cpu cores?
http://downloadmirror.intel.com/20927/eng/e1000.htm

pprindeville:
> Pardon my n00b-ness but I guess instances = physical ports or is it is cpu cores?

Each value passed to `InterruptThrottleRate` is per-port.

pprindeville:
@jow- @stintel Can you please have a look at this?

package/kernel/linux/modules/netdevices.mk
```diff
@@ -495,6 +495,9 @@ define KernelPackage/e1000e
   KCONFIG:=CONFIG_E1000E
   FILES:=$(LINUX_DIR)/drivers/net/ethernet/intel/e1000e/e1000e.ko
   AUTOLOAD:=$(call AutoProbe,e1000e)
+  MODPARAMS.e1000e:= \
+    IntMode=1 \
+    InterruptThrottleRate=4,4,4,4,4,4,4,4
```
hauke: Could you please explain what are the "old" values and what is exactly changed with the new ones?

pprindeville: Looking at [Documentation/networking/e1000e.txt](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/Documentation/networking/e1000e.txt) we see:

```
InterruptThrottleRate
---------------------
Valid Range:   0,1,3,4,100-100000 (0=off, 1=dynamic, 3=dynamic conservative,
                                   4=simplified balancing)
Default Value: 3

The driver can limit the amount of interrupts per second that the adapter
will generate for incoming packets. It does this by writing a value to the
adapter that is based on the maximum amount of interrupts that the adapter
will generate per second.
...
The default behaviour of the driver previously assumed a static
InterruptThrottleRate value of 8000, providing a good fallback value for
all traffic types, but lacking in small packet performance and latency.
The hardware can handle many more small packets per second however, and
for this reason an adaptive interrupt moderation algorithm was implemented.
...
In simplified mode the interrupt rate is based on the ratio of TX and
RX traffic.  If the bytes per second rate is approximately equal, the
interrupt rate will drop as low as 2000 interrupts per second.  If the
traffic is mostly transmit or mostly receive, the interrupt rate could
be as high as 8000.
...
IntMode
-------
Valid Range: 0-2 (0=legacy, 1=MSI, 2=MSI-X)
Default Value: 2

Allows changing the interrupt mode at module load time, without requiring a
recompile. If the driver load fails to enable a specific interrupt mode, the
driver will try other interrupt modes, from least to most compatible.  The
interrupt order is MSI-X, MSI, Legacy.  If specifying MSI (IntMode=1)
interrupts, only MSI and Legacy will be attempted.
```

In particular, "providing a good fallback value for all traffic types, but lacking in small packet performance and latency."  Not good for a router.

The simplified most works well in potentially asymmetrical modes, and extensive testing with `iperf3` in half- and full-duplex modes confirms this.  For routers, the number of PCI devices is typically low and MSI addresses this scenario with minimal overhead (unlike desktop systems which might have multi-function cards, etc and hence need MSI-X).

pprindeville:
@hauke @jow- @nbd168 @blogic Can someone please merge this?

hauke:

pprindeville:
Thanks!

PR #1502, opened 2017-11-08
lantiq: add missing macaddr retrieval for Netgear DGN3500

danielg4:
The MAC addresses were not being set for LAN and WAN. This will now use the
same MAC mechanism as the rest of the target.

Signed-off-by: Daniel Gimpelevich <daniel@gimpelevich.san-francisco.ca.us>


mkresin:
merged, thanks!

PR #1505, opened 2017-11-09
libnl-tiny: export version

HenryGiraldo:
The different versions of libnl contain different functionalities and APIs. It is necessary to know the version of the library to determine what API and what functionalities are available.

Signed-off-by: Enrique Giraldo <enrique.giraldo@galgus.net>

### DIFF
--- a/package/libs/libnl-tiny/src/Makefile
+++ b/package/libs/libnl-tiny/src/Makefile
@@ -10,7 +10,7 @@ all: $(LIBNAME)
 %.o: %.c
 	$(CC) $(WFLAGS) -c -o $@ $(INCLUDES) $(CFLAGS) $<
 
-LIBNL_OBJ=nl.o handlers.o msg.o attr.o cache.o cache_mngt.o object.o socket.o error.o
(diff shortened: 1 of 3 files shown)


dedeckeh:
@HenryGiraldo Could you be more verbose in the git commit message explaining why you want version number export support ?

HenryGiraldo:
Done, mainly to reduce the [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell).

dedeckeh:
@HenryGiraldo The git commit message is still empty

Please use git commit --amend to add the comments to your commit. Use git push --force to overwrite your github repository/branch. The PR will afterwards update automatically. Please don't close this PR and open a new one

nbd168:
Do we really need that runtime version information? What piece of software uses this?

HenryGiraldo:
I use it for a personal project (as a good practice I check the versions of the libraries that I use). Issues regarding compatibility will likely become increasingly important, that why I suggest this PR as an improvement. If it does not seem like a good idea, there's no problem in not accepting it.

PR #1373, opened 2017-09-25
ar71xx: add support for TL-WR1043N v5.0

majik-sheff:
The hardware of the TP-Link WR-1043N v5 appears
to be to almost identical to the WR-1043NDv4, except
that the USB port has been removed and there is no
longer a removeable antenna option.

The software is more in line with the Archer series in
that is uses a nested bootloader scheme. The factory
image should install from the stock interface.

Signed-off-by: Tim Thorpe <tim@tfthorpe.net>

PR #1506, opened 2017-11-09
ar71xx: add support for TL-WR1043Nv5

Brother-Lal:
This is a cleaned up version of the stalled PR at https://github.com/lede-project/source/pull/1373

Everything should be identical, but i hope i fixed the hickups/code formatting things.

Please review and I'll make any needed changes.
I hope i' m not going against any policy by double-requesting a PR :)

---------------------------
Commit message:

TP-Link WR-1043N v5 appears to be identical to
the WR-1043NDv4, except that the USB port has
been removed and there is no longer a removeable
antenna option.

The software is more in line with the Archer series in
that is uses a nested bootloader scheme.

Signed-off-by: Tim Thorpe <tim@tfthorpe.net>
Signed-off-by: Ludwig Thomeczek <ledesrc@wxorx.net>


target/linux/ar71xx/base-files/etc/board.d/01_leds
```diff
@@ -802,14 +802,20 @@ tl-wr1043nd-v2)
 	ucidef_set_led_usbdev "usb" "USB" "tp-link:green:usb" "1-1"
 	ucidef_set_led_wlan "wlan" "WLAN" "tp-link:green:wlan" "phy0tpt"
 	;;
-tl-wr1043nd-v4)
-	ucidef_set_led_usbdev "usb" "USB" "tp-link:green:usb" "1-1"
+tl-wr1043nd-v4|\
+tl-wr1043n-v5)
```
pepe2k: This doesn't look for me like a correct alphabetical order.
Please, fix it here and in all other places.

Brother-Lal: Thanks for the review!
I prefer the natural sorting order, because its the one that comes after the 1043v4 - damn tpl for changing their antennas!
anyway - I hope it's fixed in all files :smile:

target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c
```diff
@@ -62,6 +63,11 @@
 #define TL_WR1043_V4_EEPROM_ADDR		0x1fff0000
 #define TL_WR1043_V4_WMAC_CALDATA_OFFSET	0x1000
 
+#define TL_WR1043N_V5_MAC0_OFFSET		0
+#define TL_WR1043N_V5_MAC1_OFFSET		6
```
heinzek: You defined two mac offsets but not used.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -669,6 +669,22 @@ define Device/tl-wr1041n-v2
 endef
 TARGET_DEVICES += tl-wr1041n-v2
 
+define Device/tl-wr1043n-v5
+  $(Device/tplink)
+  DEVICE_TITLE := TP-LINK TL-WR1043N v5
+  BOARDNAME := TL-WR1043N-v5
+  DEVICE_PROFILE := TLWR1043
+  TPLINK_HWID :=  0x10430005
+  MTDPARTS := spi0.0:128k(factory-uboot)ro,128k(u-boot)ro,1536k(kernel),13568k(rootfs),128k(product-info)ro,640k(config)ro,64k(partition-table)ro,128k(logs)ro,64k(ART)ro,15104k@0x40000(firmware)
```
heinzek: You can merge kernel and rootfs partition to firmware partition.
Other partition used in TP-Link firmware You can merge to one tp-link partition. Look to MTDPARTS on tl-wr902ac device.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -669,6 +669,22 @@ define Device/tl-wr1041n-v2
 endef
 TARGET_DEVICES += tl-wr1041n-v2
 
+define Device/tl-wr1043n-v5
+  $(Device/tplink)
```
heinzek: I think is better to use $(Device/archer-cxx)

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -669,6 +669,22 @@ define Device/tl-wr1041n-v2
...
+  DEVICE_PROFILE := TLWR1043
+  TPLINK_HWID :=  0x10430005
+  MTDPARTS := spi0.0:128k(factory-uboot)ro,128k(u-boot)ro,1536k(kernel),13568k(rootfs),128k(product-info)ro,640k(config)ro,64k(partition-table)ro,128k(logs)ro,64k(ART)ro,15104k@0x40000(firmware)
+  IMAGE_SIZE := 15104k
+  TPLINK_BOARD_ID := TLWR1043NV5
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade | \
+	append-metadata | check-size $$$$(IMAGE_SIZE)
+  IMAGE/factory.bin := append-rootfs | tplink-safeloader factory
```
heinzek: If You switch to (Device/archer-cxx) please remove these four lines.

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -669,6 +669,22 @@ define Device/tl-wr1041n-v2
...
+  BOARDNAME := TL-WR1043N-v5
+  DEVICE_PROFILE := TLWR1043
+  TPLINK_HWID :=  0x10430005
+  MTDPARTS := spi0.0:128k(factory-uboot)ro,128k(u-boot)ro,1536k(kernel),13568k(rootfs),128k(product-info)ro,640k(config)ro,64k(partition-table)ro,128k(logs)ro,64k(ART)ro,15104k@0x40000(firmware)
+  IMAGE_SIZE := 15104k
+  TPLINK_BOARD_ID := TLWR1043NV5
+  KERNEL := kernel-bin | patch-cmdline | lzma | uImageArcher lzma
+  IMAGES := sysupgrade.bin factory.bin
+  IMAGE/sysupgrade.bin := append-rootfs | tplink-safeloader sysupgrade | \
+	append-metadata | check-size $$$$(IMAGE_SIZE)
```
heinzek: You use metadata. Those need entry: SUPPORTED_DEVICES := "tl-wr1043n-v5"

target/linux/ar71xx/image/tp-link.mk
```diff
@@ -669,6 +669,22 @@ define Device/tl-wr1041n-v2
 endef
 TARGET_DEVICES += tl-wr1041n-v2
 
+define Device/tl-wr1043n-v5
+  $(Device/tplink)
+  DEVICE_TITLE := TP-LINK TL-WR1043N v5
+  BOARDNAME := TL-WR1043N-v5
+  DEVICE_PROFILE := TLWR1043
+  TPLINK_HWID :=  0x10430005
```
heinzek: this entry is used by mktplinkfw but You use only uImageArcher and tplink-safeloader

majik-sheff:
Wow! These are all phenomenal comments that would have been helpful a while ago.  Glad to see some insight from someone who more fully understands what all of this does.

I admit right up front that I treated a large part of the build system as a stack of black boxes just for the sake of my sanity.  Is there a document anywhere that actually maps this stuff out?

Brother-Lal:
Thank you for the comments+fixes @heinzek !

I incorporated most of them, except the firmware partition/additional partitions merger.
I would like to have the info in there where the factory partitions are located.
If it results in problems, just say it, and I'll change it, but if there is a chance to leave it this way, I would think it nice to preserve.

Also added a short hardware description to the Commit message, as i noticed the original 1043v4 has none...
------------
Specifications:
 - QCA9563 at 775 MHz
 - 64 MB RAM
 - 16 MB Flash
 - 3 (non-detachable) Antennas / 450Mbit
 - 1x/4x WAN/LAN GBit Ethernet on QCA8337
 - Reset + Wifi Buttons
------------

@majik-sheff 
I haven't found a clear explanation / map for a lot of stuff, sadly it's mostly try-and-error and experience after some time (was my impression :D )

pepe2k:
@Brother-Lal,

> I incorporated most of them, except the firmware partition/additional partitions merger.
> I would like to have the info in there where the factory partitions are located.

Vendor may change the layout with every single firmware update and you already have partition layout in `tplink-safeloader.c`, why should we have it in two places then?

> If it results in problems, just say it, and I'll change it, but if there is a chance to leave it this way, I would think it nice to preserve.

Please, follow @heinzek request and change it. Your approach with statically defined kernel/rootfs partitions in `MTDPARTS` just increases maintenance burden (in case of kernel or rootfs partition resize, you would have to do it in two places, have a look at [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=commitdiff;h=f05706ba565fb5e43f8b86873bf43acb85a979da)).

Also, please use lowercase for art partition name (`art` -> `ART`) in `MTDPARTS`.

Cheers,
Piotr

Brother-Lal:
Ok, i switched to the firmware inline definition - should i keep the separate product-info/art/part-table partitions or wrap them into one partition too?
art is now lowercase too.

Thanks btw for the explanation - can i assume this only works with the external definition because the safeloader tool/tplink flasher then create the partitions already and is not applicable for other devices without safeloader?

pepe2k:
@Brother-Lal,

Please, always use `scripts/checkpatch.pl` before submitting PR/patch.
Could you fix below minor issues?

```
./scripts/checkpatch.pl 1506.patch 
ERROR: code indent should use tabs where possible
#252: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c:249:
+^I                         tl_wr1043n_v5_leds_gpio);$

ERROR: code indent should use tabs where possible
#254: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c:251:
+^I                                ARRAY_SIZE(tl_wr1043nd_v4_gpio_keys),$

ERROR: code indent should use tabs where possible
#255: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c:252:
+^I                                tl_wr1043nd_v4_gpio_keys);$

ERROR: code indent should use tabs where possible
#260: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr1043nd-v4.c:257:
+^I                            ARRAY_SIZE(tl_wr1043nd_v4_mdio0_info));$

WARNING: line over 80 characters
#397: FILE: tools/firmware-utils/src/tplink-safeloader.c:1294:
+	if (strcasecmp(info->id, "ARCHER-C25-V1") == 0 || strcasecmp(info->id, "TLWR1043NV5") == 0) {

total: 4 errors, 1 warnings, 285 lines checked
```

> Ok, i switched to the firmware inline definition - should i keep the separate product-info/art/part-table partitions or wrap them into one partition too?

We usually keep at least `art` partition separated. For other partitions, you can follow one of already existing approaches - some people combine unused partitions in single, bit one called for example `tplink` (see C58/C59 v1), others prefer to have them separated (TL-WR942N v1).

> art is now lowercase too.

Perfect, thanks for fix.

> Thanks btw for the explanation - can i assume this only works with the external definition because the safeloader tool/tplink flasher then create the partitions already and is not applicable for other devices without safeloader?

If your questions is about mtd splitter then you should check the code:
https://github.com/lede-project/source/tree/master/target/linux/generic/files/drivers/mtd/mtdsplit

Cheers,
Pitor

Brother-Lal:
I let the checkpatch run beforehand - I wanted to align the function arguments with the function declaration end in the first 4 cases.
For the if in the safeloader i didn't want to break it apart for just 2 options.

Didn't know how catholic i had to be with the script output so took some freedom - it's fixed.

pepe2k:
@Brother-Lal,

Thank you, I merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

Cheers,
Piotr

PR #1508, opened 2017-11-09
openssl: update to 1.0.2m

tripolar:
don't set no-ssl3-method when CONFIG_OPENSSL_WITH_SSL3 is disabled otherwise the compile breaks with this error:

../libssl.so: undefined reference to `SSLv3_client_method'

Fixes CVE: CVE-2017-3735, CVE-2017-3736

Signed-off-by: Peter Wagner <tripolar@gmx.at>


lynxis:
merged into master

lynxis:
backported to lede-17.01

PR #1510, opened 2017-11-10
 brcm47xx: fix switch port mapping on Asus RT-N16

majik-sheff:
On the Asus RT-N16, the ports are not mapped the in the same way as
the RT-N12.  It is, however, the same as the Linksys E3000v1.
This fix has been tested on multiple instances of actual hardware.

Signed-off-by: Tim Thorpe <timfthorpe@gmail.com>


mkresin:
> This fix has been tested on multiple instances of actual hardware.

I'm still not convinced that changing the CPU port from `5@eth0` to `8@eth0` works. 

Could it be that you preserved your config during testing and the changed CPU port wasn't applied to your /etc/config/network? Would you please attach this file with your next comment.

I've little to no knowledge about brcm47xx. If this targets works somehow special, I could be wrong of course.

majik-sheff:
I'm only submitting this patch because I discovered that my router was suddenly unreachable after installing LEDE.  After much cursing and some work involving a screwdriver and a serial adapter, I discovered that the default settings for the switch were completely wrong.  For your perusal/approval, here is the contents of /etc/config/network after a fresh install and reset:

`
config interface 'loopback'
	option ifname 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option ula_prefix 'fd49:55c3:217a::/48'

config interface 'wan'
	option ifname 'eth0.1'
	option proto 'dhcp'

config interface 'wan6'
	option ifname 'eth0.1'
	option proto 'dhcpv6'

config interface 'lan'
	option type 'bridge'
	option ifname 'eth0.2'
	option proto 'static'
	option ipaddr '192.168.1.1'
	option netmask '255.255.255.0'
	option ip6assign '60'

config switch
	option name 'switch0'
	option reset '1'
	option enable_vlan '1'

config switch_vlan
	option device 'switch0'
	option vlan '1'
	option ports '0 8t'

config switch_vlan
	option device 'switch0'
	option vlan '2'
	option ports '1 2 3 4 8t'

`

With these settings, all ports work as expected.   As I said before, I have multiple instances of this particular device.  This is not a one-off.

I suspect the reason this hasn't been reported since the changes happened in July is simply because anyone who has one of these up until now has only upgraded from an older working install and not done a fresh install.

My PC cable is plugged into LAN2, my WAN is plugged into the WAN port.  Here's the relevant settings screen immediately after a full reset:
screenshot

I'm not sure what more I can do to convince you.

jow-:
It was actually me who introduced the bug when fixing FS#502 - sorry for that.

jow-:
Pulled into my staging tree: https://git.lede-project.org/96ed69101da254b0cb61a0dfc42bd48d27bfacb9

mkresin:
> I'm not sure what more I can do to convince you.

By mentioning in the commit message that the switch doesn't work at all.

What I got from the commit message is that the mapping of the ports  as displayed in LuCI isn't in line with the labels on the case. But even that was already some guessing. No mention of a wrong CPU port.

Anyway. Thanks for the fix!

majik-sheff:
I'll bear that in mind when I submit something next time.  Sorry for being short on details, I guess I glossed over the fact that this was a show-stopper.  Anyway, thank you for taking in the patch.  I'm going to do some more extensive testing including reverting back to bone factory stock firmware, just to be sure.  I'll report if I find anything relevant.

majik-sheff:
So i did a 30-30-30 reset after installing the latest official Asus firmware.  After doing the bare minimum to get SSH, here's what I get:

`admin@(none):/tmp/etc# nvram show|grep vlan[0-9]port
vlan2ports=0 8u
vlan1ports=1 2 3 4 8*
`

I think that pretty well removes any doubts.  For future reference, my hardware revision is A1.

Thank you for taking care of this so quickly!

PR #1511, opened 2017-11-10
ar71xx: Complete support for RB mAP 2nD

robimarko:
This patch adds support for the MikroTik RouterBOARD mAP 2nD
https://mikrotik.com/product/RBmAP2nD

Specifications:

SoC: Qualcomm QCA9531(650MHz)
RAM: 64MB
Storage: 16MB NOR SPI flash
Wireless: builtin QCA9531, 2x2:2
Ethernet: 2x100M (802.3af/at POE IN and Passive POE out on ETH2)
USB: microUSB type AB port
This patch adds missing code to fully support mAP.
Machfile already contained configuration for mAP 2nD but device-specific configuration like LEDs etc was missing.

Note: The POE LED works but doesn't turn on when POE passthrough is enabled, despite being configured with GPIO trigger.

Installation

login to the Mikrotik WebUI to backup your license keys
set up a DHCP/BOOTP Server with:
DHCP-Option 66 (TFTP server name) pointing to a local TFTP
Server within the same subnet of the DHCP range
DHCP-Option 67 (Bootfile-Name) matching the initramfs file name
of the to be booted image
connect the port labeled internet to your local network
keep the reset button pushed down and power on the board
The board should load and start the initramfs image from the TFTP
Server. Login as root/without a password to the started LEDE via ssh
listing on IPv4 address 192.168.1.1. Use sysupgrade to install LEDE.

Revert to RouterOS

Use the "rbcfg" package on in LEDE:

rbcfg set boot_protocol bootp
rbcfg set boot_device ethnand
rbcfg apply
Open Netinstall and reboot routerboard. Now, netinstall sees routerboard
and you can install RouterOS. If NetInstall gets stuck on Sending offer
just wait for it to timeout and then close and open Netinstall again.

Click on install again.

In order for RouterOS to function properly, you need to restore license
for the device. You can do that by including license in NetInstall

Signed-off-by: Robert Marko robimarko@gmail.com

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -320,6 +320,12 @@ rb-lhg-5nd)
 	ucidef_set_led_rssi "rssimediumhigh" "RSSIMEDIUMHIGH" "rb:green:rssi3" "wlan0" "60" "100" "-59" "13"
 	ucidef_set_led_rssi "rssihigh" "RSSIHIGH" "rb:green:rssi4" "wlan0" "80" "100" "-79" "13"
 	;;
+rb-map-2nd)
+	ucidef_set_led_switch "eth1" "WAN" "rb:green:eth1" "switch0" "0x02"
+	ucidef_set_led_switch "eth2" "LAN" "rb:green:eth2" "switch0" "0x04"
+	ucidef_set_led_gpio "poe" "POE" "rb:red:poe_out" "14" "0"
+	ucidef_set_led_wlan "wlan" "WLAN" "rb:green:wlan" "phy0tpt"
+	;;
 rb-mapl-2nd)
 	ucidef_set_led_default "power" "POWER" "rb:green:power" "1"
 	ucidef_set_led_netdev "lan" "LAN" "rb:green:eth" "eth0"
--- a/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
+++ b/target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
@@ -1,6 +1,7 @@
 /*
  *  MikroTik SPI-NOR RouterBOARDs support
  *
+ *  - MikroTik RouterBOARD mAP 2nD
  *  - MikroTik RouterBOARD mAP L-2nD
  *  - MikroTik RouterBOARD 941L-2nD
  *  - MikroTik RouterBOARD 951Ui-2nD
@@ -14,7 +15,6 @@
  *  Preliminary support for the following hardware
  *  - MikroTik RouterBOARD wAP2nD
  *  - MikroTik RouterBOARD cAP2nD
- *  - MikroTik RouterBOARD mAP2nD
  *  Furthermore, the cAP lite (cAPL2nD) appears to feature the exact same
  *  hardware as the mAP L-2nD. It is unknown if they share the same board
  *  identifier.
@@ -405,7 +405,7 @@ static struct gpio_led rbmap_leds[] __initdata = {
 		.active_low = 1,
 	}, {
 		.name = "rb:green:eth2",
-		.gpio = RBMAP_GPIO_LED_WLAN,
+		.gpio = RBMAP_GPIO_LED_LAN2,
 		.active_low = 1,
 	}, {
 		.name = "rb:red:poe_out",
@@ -915,13 +915,14 @@ static void __init rbcap_setup(void)
 }
 
 /*
- * Init the mAP hardware (EXPERIMENTAL).
- * The mAP 2nD has two ethernet ports, PoE output and an SSR for LED
- * multiplexing.
+ * Init the mAP hardware.
+ * The mAP 2nD has two ethernet ports, PoE output, SSR for LED
+ * multiplexing and USB port.
  */
 static void __init rbmap_setup(void)
 {
-	u32 flags = RBSPI_HAS_WLAN0 | RBSPI_HAS_SSR | RBSPI_HAS_POE;
+	u32 flags = RBSPI_HAS_USB | RBSPI_HAS_WLAN0 |
+			RBSPI_HAS_SSR | RBSPI_HAS_POE;
 
 	if (rbspi_platform_setup())
 		return;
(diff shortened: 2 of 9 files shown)


pepe2k:
@f00b4r0 if you have time, please have a look at this one, thanks!

Cheers,
Piotr

f00b4r0:
LGTM ( @robimarko had been in touch before submitting 😉 ), with maybe one little nitpick in the commit message: regarding the POE LED, to clarify I would say something like:

"Note: the POE LED works but doesn't turn on when POE passthrough is enabled, despite being configured with gpio trigger."

(so that the message isn't interpreted as a problem with the machine support code itself).

Cheers

robimarko:
Thanks, @f00b4r0 good catch.
Changed it.

pepe2k:
@robimarko 

```
./scripts/checkpatch.pl 1511.patch

ERROR: Unrecognized email address: 'Robert Marko robimarko@gmail.com'
#52: 
Signed-off-by: Robert Marko robimarko@gmail.com

WARNING: line over 80 characters
#221: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c:924:
+	u32 flags = RBSPI_HAS_USB | RBSPI_HAS_WLAN0 | RBSPI_HAS_SSR | RBSPI_HAS_POE;

WARNING: line over 80 characters
#236: FILE: target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c:938:
+				GPIOF_OUT_INIT_HIGH | GPIOF_ACTIVE_LOW | GPIOF_EXPORT_DIR_FIXED,

total: 1 errors, 2 warnings, 164 lines checked

1511.patch has style problems, please review.
```

`git commit -s` will include proper SoB line for you.

Also, please have a look at our [submitting patches rules](https://lede-project.org/submitting-patches). We have there "commit description must have less than 75 characters per line".

Cheers,
Piotr

robimarko:
Thanks, will fix SoB and too long lines in the morning.

target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
```diff
@@ -932,12 +933,19 @@ static void __init rbmap_setup(void)
 	/* GMAC1 is HW MAC, WLAN0 MAC is HW MAC + 2 */
 	rbspi_network_setup(flags, 0, 2, 0);
 
-	if (flags & RBSPI_HAS_POE)
-		gpio_request_one(RBMAP_GPIO_POE_POWER,
-				GPIOF_OUT_INIT_HIGH | GPIOF_EXPORT_DIR_FIXED,
-				"POE power");
+	/* USB power GPIO is inverted, set GPIOF_ACTIVE_LOW for consistency */
```
f00b4r0: you removed POE power, that board doesn't have POE passthrough power control?

robimarko: Yes, it does.
I have added GPIO switch to it.
That works without a issue.

f00b4r0: your last commit has removed the GPIO request for POE power. It can't work anymore.

robimarko: It can, there is no need for GPIO request If it is not needed for it to be on by default.
GPIO switch does its job.

robimarko:
@pepe2k Errors and warning produced by checkpatch have been resolved

target/linux/ar71xx/files/arch/mips/ath79/mach-rbspi.c
pepe2k: > It can, there is no need for GPIO request If it is not needed for it to be on by default.
> GPIO switch does its job.

But you probably don't want to allow user to change the GPIO direction?

f00b4r0: That, exactly. gpio_request_one() serves more purpose than what can be done in the userspace scripts.

f00b4r0: As for the "on by default" it was my understanding that this is the default behavior in RouterOS, and a desirable one, but maybe I was wrong... It can simply be changed with the  GPIOF_OUT_INIT_LOW flag if necessary.

robimarko: I was not thinking about users changing direction.
Will use gpio_request_one() for POE instead of GPIO switch.

No, RouterOS does not enable POE by default.
Otherwise, you will fry any device like a notebook when trying to configure the device first time since you can only do that through ETH2 port.

f00b4r0: i'm not sure how you could fry anything with POE, especially considering it's passthrough, but I stand corrected for the RouterOS default :)

robimarko: Changed to gpio_request_one()

robimarko: Well you can easily fry networking cards or whole devices.
Since device passes the same voltage as received for powering via DC jack or POE IN on first port.
So if you use 24V DC power brick or passive 24V POE that is commonly used and pass that to notebook or non POE powered routers and you can easily damage or destroy them

pepe2k: > i'm not sure how you could fry anything with POE, especially considering it's passthrough, but I stand corrected for the RouterOS default :)

Passive PoE doesn't do any negotiation, it just put voltage to the cable.

f00b4r0: Well sure but I still hope it does it right by putting the same voltage on each differential pair. In which case it's essentially harmless. That's how true POE ensures backward safety with non-POE compliant hardware...

robimarko: Passive POE does not do that.
It just puts voltage between 2 pairs.
Usually 4,5 is positive and 7,8 negative

f00b4r0: That's perfectly fine and harmless. POE works out autodetection by putting a voltage (2-10V) on the pairs. It's perfectly ok to have a voltage _between_ pairs, as long as there's no DC voltage between each side of a single pair.
If a device is damaged in this context, clearly the hardware does not follow the ethernet standard, which requires transformer coupling... Or there's something else I'm not understanding here.

f00b4r0: Replying to myself, I see where the problem is: the passive POE voltage might be too high for the other side to cope with if it isn't compliant. Point taken :)

target/linux/ar71xx/base-files/etc/board.d/01_leds
```diff
@@ -320,6 +320,12 @@ rb-lhg-5nd)
 	ucidef_set_led_rssi "rssimediumhigh" "RSSIMEDIUMHIGH" "rb:green:rssi3" "wlan0" "60" "100" "-59" "13"
 	ucidef_set_led_rssi "rssihigh" "RSSIHIGH" "rb:green:rssi4" "wlan0" "80" "100" "-79" "13"
 	;;
+rb-map-2nd)
+	ucidef_set_led_switch "eth1" "WAN" "rb:green:eth1" "switch0" "0x02"
+	ucidef_set_led_switch "eth2" "LAN" "rb:green:eth2" "switch0" "0x04"
```
pepe2k: Please, use `ucidef_set_led_netdev` here.

f00b4r0: not possible: it's a switch.

robimarko: Why?
Netdev does not make sense since ports are connected to internal QCA9531 fast ethernet switch?
With netdev you will always have link on LED

pepe2k: Are you guys sitting here all the time, just waiting to reply? :)

pepe2k:
@robimarko,

Thank you for your contribution! I have merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) (again, with some style fixes in your commit subject and description - I don't know what you are using as an editor for git, but vim can do spell checking and word wrapping).

> Note: The POE LED works but doesn't turn on when POE passthrough is
> enabled, despite being configured with GPIO trigger.

If it's a bug, please report it in our [bug tracker](https://bugs.lede-project.org/).

Cheers,
Piotr

PR #1512, opened 2017-11-10
Get IO Edgeport USB serial firmware from linux-firmware repo instead of kernel

bmork:
The in-kernel firmware is going away with Linux v4.14. There is no need to wait until the last minute preparing for that.  The edgeport firmware files in the kernel and in the linux-firmware repo are identical, so this is merely a packaging change.  There is no change to the installed firmware.


mkresin:
Merged into [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). Thanks!

PR #1513, opened 2017-11-10
ar71xx: Add support for TP-Link WR810n-v2

p-wassi:
@mkresin 
TP-Link WR810n-v2 is a compact WiFi router featuring two Ethernet ports.

Difference to TL-WR810n(-v1):
 - no USB port any more
 - just 32 MB RAM (instead of 64)
 - QCA9533 (instead of QCA9531)

---------------

Specification:
 - QCA9533 (650MHz)
 - 32 MB RAM
 - 8 MB Flash
 - 300 Mbps Wifi 2.4 GHz
 - 2x 10/100 Mbps Ethernet (eth0, eth1)
 - UART on Testpads on PCB
 - 1x LED (GPIO-controlled)
 - 1x Reset/WPS-button
 - 1x Slide-switch with 2 GPIO inputs

Flash instructions:
 - Flash as regular firmware update in Web UI
 - Default: 192.168.0.254 / admin / admin
 - Select *-factory.bin

When holding reset button during power-up, U-Boot requests a file
'wr810nv1_tp_recovery.bin' from TFTP server at 192.168.0.66

Everything's tested. The only issue I experienced is that UART RX is not working.
This is likely a HW issue. For -v1 one had to bridge a jumper _R107_ to make RX working.
Two testpads on the PCB are labelled RX and TX. TX is fine. R107 with 150 Ohm is mounted
as series resistor of the RX path. Even replacing the resistor with a solder blob does not make RX
working.

-----------

@pepe2k : What info is needed to make your U-Boot work on this device? As I've seen, the -v1 is
already in your list, so I guess, this should be easy (not even the GPIOs were changed)

----------

There's a forum thread discussing support on this device [1]

Prebuilt images for testing are here: 
lede-ar71xx-generic-tl-wr810n-v2-squashfs-factory.bin at [2]
lede-ar71xx-generic-tl-wr810n-v2-squashfs-sysupgrade.bin at [3]

[1]:
https://forum.lede-project.org/t/support-for-tl-wr810n-v2/4338

[2]:
https://pwassi.privatedns.org/lede/wr810nv2/lede-ar71xx-generic-tl-wr810n-v2-squashfs-factory.bin

[3]:
https://pwassi.privatedns.org/lede/wr810nv2/lede-ar71xx-generic-tl-wr810n-v2-squashfs-sysupgrade.bin


pepe2k:
@p-wassi,

Where is your `mach-tl-wr810n-v2.c`? :) And, if this device is almost a 1:1 copy of v1, please include support in existing mach file.

Have you actually test this support or your device?

> @pepe2k : What info is needed to make your U-Boot work on this device? As I've seen, the -v1 is
> already in your list, so I guess, this should be easy (not even the GPIOs were changed)

I will include it.

Cheers,
Piotr

p-wassi:
> Where is your mach-tl-wr810n-v2.c? :)

Hmmm... I should probably not only commit changed files, but also new ones.
(Anyway, the mach-wr810n-v2.c file is gone now)

> please include support in existing mach file.

It's a pleasure to me to remove redundancies. Here, I've just repeated the path which was
already taken. I.e., look at `mach-tl-wr802n.c` and `mach-tl-wr802n-v2.c`, they're doing exactly
the same functions calls with exactly the same data. So would a PR that removed the file
`mach-tl-wr802n-v2.c` and integrated the WR802n-v2 into wr802 be accepted?

Anyway, one difference to v1 and v2 left: LAN MAC is `-1` on v1 and `0` on v2 (relative to base MAC).
I've chosen my v2 with `0` to match it exactly like it is done in stock firmware. There, LAN and WLAN
have the same MAC, whereas WAN is `+1`. Should I keep it like this, or adapt to the v1, where setup
would further be simplified?

> Have you actually test this support or your device?

Sure. Device is here to play with. (And just tested the `v1`+`v2` merged mach-file on it)

> (U-Boot) I will include it.

Once included and given that I still have the device here, I'd do some testing.

PS: What's the current plan of migrating ar71xx to DTS?

pepe2k:
Hello @p-wassi,

> Hmmm... I should probably not only commit changed files, but also new ones.

:wink: 

> (Anyway, the mach-wr810n-v2.c file is gone now)

OK!

> It's a pleasure to me to remove redundancies. Here, I've just repeated the path which was
> already taken. I.e., look at mach-tl-wr802n.c and mach-tl-wr802n-v2.c, they're doing exactly
> the same functions calls with exactly the same data. So would a PR that removed the file
> mach-tl-wr802n-v2.c and integrated the WR802n-v2 into wr802 be accepted?

Definitely yes.

> Anyway, one difference to v1 and v2 left: LAN MAC is -1 on v1 and 0 on v2 (relative to base MAC).
> I've chosen my v2 with 0 to match it exactly like it is done in stock firmware. There, LAN and WLAN
> have the same MAC, whereas WAN is +1. Should I keep it like this, or adapt to the v1, where setup
> would further be simplified?

I wouldn't be surprised if MACs for v1 are assigned incorrectly (not in the same way as inside vendor firmware) and it should be done in the same way as for v2. For now, lets just keep v2 in the same way as it's in vendor firmware.

> PS: What's the current plan of migrating ar71xx to DTS?

I don't expect it to happen before next stable release. We had short talk about this during the OpenWrt Summit and we agreed we should work on it.

Anyway, thank you for your patch! I have merged it into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary) (with minor changes in commit subject and description).

Cheers,
Piotr

PR #1516, opened 2017-11-13
dnsmasq: fix swapped ubus args mac and ip

jedi7:
Fix swapped arguments "mac" and "ip" when calling function
"ubus_event_bcast".

Signed-off-by: Jaroslav Safka <devel@safka.org>


blogic:
merged, Thanks !
... saw your mail, did not have time to reply yet, thanks for sending a PR ...

PR #1519, opened 2017-11-14
ar71xx: merge mach-files for TP-Link WR802N v1 and v2

p-wassi:
@pepe2k 
As discussed in the other PR.

Here, the `mach-tl-wr802n-v2.c` is deleted and relevant parts merged with the `mach-tl-wr802n.c`
The `_v1_` (and `_v2_`) was stripped in all variable names and defines that both targets share.
The LED's name is the only difference remaining:
- `tp-link:blue:system`  for v1
- `tl-wr802n-v2:green:system` for v2

Also, proper indentation of tl_ap143_setup() function is done.

Is there any guide to what function calls belong to tl_ap143_setup? (I have moved registration
of the reset button inside that common setup routine, as this is equivalent on v1 and v2)

Compile tested, but not run-tested (no WR802N here).


pepe2k:
@p-wassi,

Thanks, merged your changes into [my staging tree](https://git.lede-project.org/?p=lede/pepe2k/staging.git;a=summary).

> Is there any guide to what function calls belong to tl_ap143_setup? (I have moved registration
> of the reset button inside that common setup routine, as this is equivalent on v1 and v2)

It's just a name, ap143 is a name of Qualcomm QCA953x platform. You could rename it to anything you want - as you can see (e.g. in https://github.com/lede-project/source/blob/master/target/linux/ar71xx/files/arch/mips/ath79/mach-cf-e316n-v2.c#L423), it's just a common setup function, shared between similar boards. [I hope I understood your question]

Cheers,
Piotr

PR #1520, opened 2017-11-15
uqmi: backport fix for pin issues on newer devices to 17.01

xback:
As discussed with @blogic


mkresin:
I've merged only the "uqmi: also try newer pin verification" commit and omitted the code style changes as they are not required.

Thanks for your contribution.

PR #1521, opened 2017-11-15
kernel: Hyper-V PCI pass through (64 bit)

lucize:

Tested on LEDE r5330-31ebbe34cc


bobafetthotmail:
might be good to enable this also in the 32 bit image.

lucize:
yes, 32 bit need attention, hyperv is not working, I'll make a pr, already tested

hauke:

PR #1522, opened 2017-11-15
kernel: add 32bit x86 HYPER-V support

lucize:

Tested LEDE 5329


hauke:

PR #1524, opened 2017-11-17
TP-Link splitters: code cleanup and improvement

f00b4r0:
These patches clean up the TP-Link splitters by:

- correctly placing the kernel partition at the beginning of the LZMA compressed kernel (skipping the TP-Link header data)
- adding an extra sanity check on the parsed header data
- removing the assertion that the kernel is at a fixed position

Before these patches the splitters would include the TP-Link header at the beginning of the kernel partition, when the actual kernel is 0x200 bytes further (currently all TP-Link images place the kernel immediately after the header, which is 0x200 bytes long).

These patches make it possible to dump valid LZMA data from the kernel partition without further processing.
They align the behavior of these splitters with that of e.g. the trx, seama or fit ones.
If it later becomes possible/desirable to support a different kernel offset, these patches also make the splitters future-proof.

The changes are mostly cosmetic and shouldn't break any of the existing setups but testing is advised.

### DIFF
--- a/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
+++ b/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
@@ -90,8 +90,8 @@ static int mtdsplit_parse_tplink(struct mtd_info *master,
 				 struct mtd_part_parser_data *data)
 {
 	struct tplink_fw_header hdr;
-	size_t hdr_len, retlen, kernel_size;
-	size_t rootfs_offset;
+	size_t hdr_len, retlen, kernel_size, rootfs_size;
+	size_t kernel_offset, rootfs_offset;
 	struct mtd_partition *parts;
 	int err;
 
@@ -105,25 +105,24 @@ static int mtdsplit_parse_tplink(struct mtd_info *master,
 
 	switch (le32_to_cpu(hdr.version)) {
 	case 1:
-		if (be32_to_cpu(hdr.v1.kernel_ofs) != sizeof(hdr))
-			return -EINVAL;
-
-		kernel_size = sizeof(hdr) + be32_to_cpu(hdr.v1.kernel_len);
+		kernel_offset = be32_to_cpu(hdr.v1.kernel_ofs);
+		kernel_size = be32_to_cpu(hdr.v1.kernel_len);
 		rootfs_offset = be32_to_cpu(hdr.v1.rootfs_ofs);
+		rootfs_size = be32_to_cpu(hdr.v1.rootfs_len);
 		break;
 	case 2:
 	case 3:
-		if (be32_to_cpu(hdr.v2.kernel_ofs) != sizeof(hdr))
-			return -EINVAL;
-
-		kernel_size = sizeof(hdr) + be32_to_cpu(hdr.v2.kernel_len);
+		kernel_offset = be32_to_cpu(hdr.v2.kernel_ofs);
+		kernel_size = be32_to_cpu(hdr.v2.kernel_len);
 		rootfs_offset = be32_to_cpu(hdr.v2.rootfs_ofs);
+		rootfs_size = be32_to_cpu(hdr.v2.rootfs_len);
 		break;
 	default:
 		return -EINVAL;
 	}
 
-	if (kernel_size > master->size)
+	/* minimal sanity check */
+	if ((kernel_size + rootfs_size) > master->size)
 		return -EINVAL;
 
 	/* Find the rootfs */
@@ -144,7 +143,7 @@ static int mtdsplit_parse_tplink(struct mtd_info *master,
 		return -ENOMEM;
 
 	parts[0].name = KERNEL_PART_NAME;
-	parts[0].offset = 0;
+	parts[0].offset = kernel_offset;
 	parts[0].size = kernel_size;
 
 	parts[1].name = ROOTFS_PART_NAME;
--- a/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
+++ b/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
@@ -28,7 +28,7 @@ struct fw_v1 {
 	char		fw_version[36];
 	uint32_t	hw_id;		/* hardware id */
 	uint32_t	hw_rev;		/* hardware revision */
-	uint32_t	unk1;
+	uint32_t	region_code;	/* region code */
 	uint8_t		md5sum1[MD5SUM_LEN];
 	uint32_t	unk2;
 	uint8_t		md5sum2[MD5SUM_LEN];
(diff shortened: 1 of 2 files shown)


pepe2k:
Hi @f00b4r0,

Thank you for work on this.
After a brief review of your changes I'm concerned about this part:

> * correctly placing the kernel partition at the beginning of the LZMA compressed kernel (skipping the TP-Link header data)
> 
> Before these patches the splitters would include the TP-Link header at the beginning of the kernel partition, when the actual kernel is 0x200 bytes further (currently all TP-Link images place the kernel immediately after the header, which is 0x200 bytes long).
> 
> These patches make it possible to dump valid LZMA data from the kernel partition without further processing.

Why someone would want/need to dump kernel without the header (in this particular case - for TP-Link images)? As it's more like a cosmetic change, please convince others that it's required and actually improves something.

I agree that your patch makes it easier to dump just the kernel but at the same time it makes it harder to dump the header (it's still possible but requires access to the `firmware` partition) which in case of TP-Link devices (in my opinion) is more important/useful than just the kernel data (e.g. HWID, etc. access).

> They align the behavior of these splitters with that of e.g. the trx, seama or fit ones.

Are you sure about [fit splitter](https://github.com/lede-project/source/blob/master/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L112)?

As you pointed out, some splitters skip header but others don't (e.g. `uimage`, `eva`). Personally, I prefer consistency, but I just don't know which approach is the correct one here as I'm not familiar with all the splitters and user space tools which rely on the mtd organization/layout.

Have you checked and confirmed that nothing accesses TP-Link header in `kernel` partition (instead of `firmware`)?

Cheers,
Piotr

f00b4r0:
Hi @pepe2k 

As you noticed the change is at this stage purely cosmetic; however I also plan to entirely rewrite mktplinkfw and enable true support for kernel at arbitrary offset (which is probably going to be a necessity eventually to work around recently broken TP-Link bootloader / flasher, see last comments in #1467. In fact I also plan to enable support for including a bootloader in the image, something that uboot-mod can certainly benefit from 😉). At this point this cosmetic change will be a necessity if the kernel mtd partition is to have any meaning at all.

That being said, to the best of my knowledge this change doesn't break anything but I don't know about every possible bit of software in existence that could actually touch the kernel partition (I'm not sure there's any, to be honest).

As for technical consistency arguments, my reasoning is as follows:

- the rooftfs partition _does_ contain the rootfs and nothing else
- the kernel partition _should_ contain the kernel and nothing else (no reason to treat it specially)
- the firmware partition contains the whole image with its header (dumping that partition will create a valid dump of the image that was flashed, including the current jffs2 overlay), and that dump can theoretically be reflashed with sysupgrade.

I believe my reasoning is correct (otherwise I wouldn't have implemented it 😉), and it is my opinion that all splitters should behave that way. (I was indeed wrong about the fit splitter, I stand corrected).

IMHO it makes no sense to include the image header in the kernel partition considering that all the data in that header refers to the whole image. I suspect it was done this way out of ~~laziness~~ oversight :)

For instance, if you dump the kernel partition only and try to use data found in the header, the offsets will be totally meaningless because half of the image is missing.

Hope this helps

f00b4r0:
@lynxis done.

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
```diff
@@ -105,25 +105,24 @@ static int mtdsplit_parse_tplink(struct mtd_info *master,
...
+		rootfs_size = be32_to_cpu(hdr.v1.rootfs_len);
 		break;
 	case 2:
 	case 3:
-		if (be32_to_cpu(hdr.v2.kernel_ofs) != sizeof(hdr))
-			return -EINVAL;
-
-		kernel_size = sizeof(hdr) + be32_to_cpu(hdr.v2.kernel_len);
+		kernel_offset = be32_to_cpu(hdr.v2.kernel_ofs);
+		kernel_size = kernel_offset + be32_to_cpu(hdr.v2.kernel_len);
```
lynxis: Shouldn't the kernel_size == kernel_len _if_ **kernel_offset** == kernel_ofs?

lynxis:
@f00b4r0 thanks! It looks much better. I'll test your patch tommorrow with 2 or 3 different devices.

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_tplink.c
f00b4r0: hmm no. That's a left-over bit of an overly aggressive sed. Will fix, thanks

pepe2k:
Hi @f00b4r0,

> That being said, to the best of my knowledge this change doesn't break anything but I don't know about every possible bit of software in existence that could actually touch the kernel partition (I'm not sure there's any, to be honest).

You want to change something which is kind of a standard. We are not talking about new splitter for a new firmware layout/header here. You want to change something which has been there for years and, as far as I know and remember, since the very beginning header was included in `kernel` mtd partition (in the same way as it's done in vendor firmware).

If it's not broken, everybody uses it as is and your change is only cosmetic and would make users confused, it's a no-go for me.

> I believe my reasoning is correct (otherwise I wouldn't have implemented it :wink:), and it is my opinion that all splitters should behave that way. (I was indeed wrong about the fit splitter, I stand corrected).

Can you imagine uImage kernel without the header?
How would you find out what compression were used and what architecture the kernel is for?

Also, your assumption about compression type used for kernel by TP-Link is wrong.
Please, have a look [here](https://github.com/lede-project/source/blob/master/target/linux/ar71xx/image/tp-link.mk#L58).

> For instance, if you dump the kernel partition only and try to use data found in the header, the offsets will be totally meaningless because half of the image is missing.

I think you are not aware of the fact that TP-Link started using their headers only for the kerrnel (rootfs offset and size are zeroed) - just have a look for example at [RE350](https://github.com/lede-project/source/blob/master/target/linux/ramips/image/mt7621.mk#L190).

To make it clear, NAK from me for this part: `correctly placing the kernel partition at the beginning of the LZMA compressed kernel (skipping the TP-Link header data)`.

Cheers,
Piotr

f00b4r0:
For the sake of clarity as I've been asked about this OOB:
- There is no such thing as a _standard_, as evidenced by differences of implementation between various splitters. For the record, it is possible to build a working image _without_ splitting a specific kernel mtd partition: the system has no direct use for it. The kernel partition is a purely cosmetic matter: changes to its specifications are bound to be of the same nature. The proposed change wouldn't confuse anyone and is quite **necessary** if the other features I mentioned are to be implemented down the road.
- The uImage format is not being discussed here. If it were it would be equally possible to correctly offset the kernel mtd partition if so desired (whether the considered uImage is a multi-file image or a kernel-only image).
- The compression assumption was purely illustrative and is also not relevant to the proposed changes. Incidentally the quoted "counter-example" appears to use an LZMA-compressed kernel anyway.
- The RE350 is not concerned by the proposed change: it uses tplink-safeloader, not mktplinkfw and thus doesn't use these splitters. The partition scheme for this device is hardcoded [here](https://github.com/lede-project/source/blob/7aa15953e1b60033fb1390b97fd6fe42daced738/tools/firmware-utils/src/tplink-safeloader.c#L853): this "counter-example" is irrelevant.
- The NAKed part is mostly mandatory for arbitrary placement of the kernel partition, whose use I explained [here](#issuecomment-345313529).

kthxbye

PR #1527, opened 2017-11-20
kernel: bump 4.9 to 4.9.63

xback:
@blogic 

Based on commit:

> bc50a97 dnsmasq: load instance-specific conf-file if exists

Refreshed all patches.

Removed upstreamed parts.

Compile-tested: cns3xxx, imx6, mvebu, layerscape
Run-tested: cns3xxx, imx6

Signed-off-by: Koen Vandeputte <koen.vandeputte@ncentric.com>


wongsyrone:
4.9.64 is released.

xback:
Already testing 4.9.64 on hardware locally here and got patch from .63 to .64 ready.

I'll post it when this one gets accepted so reviewers actually have some time to check the commit before it gets replaced by a newer one ;-)

thagabe:
@xback Just update the current commit.

hauke:

PR #1531, opened 2017-11-22
dropbear: disable MD5 HMAC and switch to sha1 fingerprints

sch-m:
As MD5 is known weak for many years and more and more
penetration test tools complain about enabled MD5 HMAC
I think it's time to drop it.

By disabling the MD5 HMAC support dropbear  will also
automatically use SHA1 for fingerprints.
This shouldn't be a problem too.

### DIFF
--- a/package/network/services/dropbear/patches/120-openwrt_options.patch
+++ b/package/network/services/dropbear/patches/120-openwrt_options.patch
@@ -39,7 +39,7 @@
  
  /* Enable "Counter Mode" for ciphers. This is more secure than normal
   * CBC mode against certain attacks. It is recommended for security
-@@ -131,9 +131,9 @@ If you test it please contact the Dropbe
+@@ -131,10 +131,10 @@ If you test it please contact the Dropbe
   * If you disable MD5, Dropbear will fall back to SHA1 fingerprints,
   * which are not the standard form. */
  #define DROPBEAR_SHA1_HMAC
@@ -47,10 +47,12 @@
 +/*#define DROPBEAR_SHA1_96_HMAC*/
  #define DROPBEAR_SHA2_256_HMAC
 -#define DROPBEAR_SHA2_512_HMAC
+-#define DROPBEAR_MD5_HMAC
 +/*#define DROPBEAR_SHA2_512_HMAC*/
- #define DROPBEAR_MD5_HMAC
++/*#define DROPBEAR_MD5_HMAC*/
  
  /* You can also disable integrity. Don't bother disabling this if you're
+  * still using a cipher, it's relatively cheap. If you disable this it's dead
 @@ -146,7 +146,7 @@ If you test it please contact the Dropbe
   * Removing either of these won't save very much space.
   * SSH2 RFC Draft requires dss, recommends rsa */
--- a/package/network/services/dropbear/Makefile
+++ b/package/network/services/dropbear/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dropbear
 PKG_VERSION:=2017.75
-PKG_RELEASE:=4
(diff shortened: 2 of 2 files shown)


neheb:
FWIW md5 HMAC is not the same as raw md5. Anyway, I don't use dropbear so my opinion doesn't mean much but

ACK

sch-m:
Ok, incremented PKG_RELEASE to 5.

package/network/services/dropbear/patches/120-openwrt_options.patch
```diff
@@ -39,18 +39,20 @@
  
  /* Enable "Counter Mode" for ciphers. This is more secure than normal
   * CBC mode against certain attacks. It is recommended for security
-@@ -131,9 +131,9 @@ If you test it please contact the Dropbe
+@@ -131,10 +131,10 @@ If you test it please contact the Dropbe
   * If you disable MD5, Dropbear will fall back to SHA1 fingerprints,
   * which are not the standard form. */
  #define DROPBEAR_SHA1_HMAC
 -#define DROPBEAR_SHA1_96_HMAC
 +/*#define DROPBEAR_SHA1_96_HMAC*/
```
hauke: Why are you deactivating SHA1_96_HMAC?

package/network/services/dropbear/patches/120-openwrt_options.patch
```diff
@@ -39,18 +39,20 @@
...
  #define DROPBEAR_SHA2_256_HMAC
 -#define DROPBEAR_SHA2_512_HMAC
+-#define DROPBEAR_MD5_HMAC
 +/*#define DROPBEAR_SHA2_512_HMAC*/
```
hauke: Why are you deactivating SHA2_512_HMAC?

package/network/services/dropbear/patches/120-openwrt_options.patch
```diff
@@ -39,18 +39,20 @@
...
- #define DROPBEAR_MD5_HMAC
++/*#define DROPBEAR_MD5_HMAC*/
```
hauke: This looks ok, OpenSSH also deactivated this by default in 7.2:
https://www.openssh.com/txt/release-7.2
I know of some people with some "old"? Microtik devices where you can not login via SSH from a recent Ubuntu any more, probably they depend on this, but I assume this is not a problem.

package/network/services/dropbear/patches/120-openwrt_options.patch
sch-m: That wasn't me. It's already deactivated:
https://github.com/lede-project/source/blob/master/package/network/services/dropbear/patches/120-openwrt_options.patch#L47

package/network/services/dropbear/patches/120-openwrt_options.patch
sch-m: Same here:
https://github.com/lede-project/source/blob/master/package/network/services/dropbear/patches/120-openwrt_options.patch#L50

package/network/services/dropbear/patches/120-openwrt_options.patch
sch-m: You maybe get confused by the "+" and "-" because this a "patch of a patch".

package/network/services/dropbear/patches/120-openwrt_options.patch
sch-m: Yes, this is also the reason why I think it is time that we can drop the MD5 HMAC support for dropbear as well.

dedeckeh:
Thx; merged into my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=7dfd2be3caa13f7a704456c8702629bfdc67550a)

PR #1532, opened 2017-11-22
linux: add kmod-i2c-i801

sch-m:

### DIFF
--- a/package/kernel/linux/modules/i2c.mk
+++ b/package/kernel/linux/modules/i2c.mk
@@ -142,6 +142,31 @@ endef
 $(eval $(call KernelPackage,i2c-piix4))
 
 
+I2C_I801_MODULES:= \
+  CONFIG_I2C_I801:drivers/i2c/busses/i2c-i801
+
+define KernelPackage/i2c-i801
+  $(call i2c_defaults,$(I2C_I801_MODULES),59)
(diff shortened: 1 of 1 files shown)


hauke:

pprindeville:
This doesn’t address the dependency on i2c-smbus.

sch-m:
I've overseen the smbus dependency and Hauke added the smbus module to to the kmod i2c-i801 package:

https://github.com/lede-project/source/blob/master/package/kernel/linux/modules/i2c.mk#L147

pprindeville:
Not to beat a dead-horse, but the SMbus driver isn't specific to the i2c-i801.

The Cavium ThunderX SoC also requires it (there are a couple of firewall platforms which use the ThunderX_NT SoC).

I think there was some out-of-tree environmental monitoring for certain AMD SoC's which also used SMbus.

sch-m:
Interestingly, the i2c-i801 also loads and works without the i2c-smbus driver.
This is why I have overseen this relationship.

But I also agree, that explicitly adding the i2c-smbus module would maybe be the cleaner way.

pprindeville:
> But I also agree, that explicitly adding the i2c-smbus module would maybe be the cleaner way.

@sch-m Since you seem to be getting faster turn-around than I do, can you please make this separation?

PR #1533, opened 2017-11-22
kernel: add it87-wdt watchdog timer module

sch-m:
The module parameters "nogameport=1" and "nocir=1" are needed,
because this is not supported on recent chips and doesn't
really tell if the system is stable.

As this features will already be removed in linux-4.13 or newer,
this module parameters can be removed in the future.


hauke:

PR #1538, opened 2017-11-24
ar71xx: fix lan ports on ens202ext

hanetzer:
Fixes FS#1084

May require a factory reset, it did in my case after a fresh install from factory.img
from stock firmware.

Signed-off-by: Marty E. Plummer <hanetzer@protonmail.com>

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/02_network
+++ b/target/linux/ar71xx/base-files/etc/board.d/02_network
@@ -176,6 +176,7 @@ ar71xx_setup_interfaces()
 	all0315n|\
 	dlan-hotspot|\
 	dlan-pro-500-wp|\
+	ens202ext|\
 	ja76pf2|\
 	rocket-m-ti|\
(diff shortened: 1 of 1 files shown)


mkresin:
Please add a proper commit message which describes what was wrong and what has been changed to fix it. At the moment it looks like a simple change from a lan + wan configuration to both ports as lan configuration.

Add the reference to the flyspray ticket to the commit message as well.

hanetzer:
@mkresin is this to your liking, or should the flyspray reference be handled differently?

mkresin:
If you change the commit message to:

```
ar71xx: fix lan ports on ens202ext

In the stock firmware both eth0 and eth1 are set to br-lan,
add this behavior to our images.

Fixes: FS#1084

Signed-off-by: Marty E. Plummer <hanetzer@protonmail.com>
```

i would say it is perfect.

mkresin:
merged, thanks!

PR #1543, opened 2017-11-25
build: accept gcc/g++ without minor version

jkilpatr:
Build dependency: Please install the GNU C Compiler (gcc) 4.8 or later cc
  -dumpversion | grep -E '(4\.[8-9]|5\.[0-9]|6\.[0-9]|7\.[0-9])'
Build dependency: Please install the GNU C++ Compiler (g++) 4.8 or later
  g++ -dumpversion | grep -E '(4\.[8-9]|5\.[0-9]|6\.[0-9]|7\.[0-9])'

Prerequisite check failed. Use FORCE=1 to override.

On my Fedora 26 machine gcc and g++ -dumpversion returns a whole number
'7' failing the regex introduced in commit:

b78de6207f6fc1a9db857942cb89f9fcf730a240

This change makes minor versions optional in the build dependency regex
for gcc and g++ whenever any minor version would be accepted and the
whole number version is sufficient as a dependency check. For versions
4.* a minor version is still required.

Signed-off-by: Justin Kilpatrick <jkilpatr@redhat.com>

### DIFF
--- a/include/prereq-build.mk
+++ b/include/prereq-build.mk
@@ -28,8 +28,8 @@ $(eval $(call TestHostCommand,proper-umask, \
 
 $(eval $(call SetupHostCommand,gcc, \
 	Please install the GNU C Compiler (gcc) 4.8 or later \
-	$(CC) -dumpversion | grep -E '(4\.[8-9]|5\.[0-9]|6\.[0-9]|7\.[0-9])', \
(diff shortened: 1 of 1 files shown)


hauke:
The author name contains your github name, could you change this to your real name please.

hauke:
Thank you for the patch it was committed to master.

PR #1549, opened 2017-11-27
kernel: enable eSATA port on Netgear D7800

steppres:
Enables eSATA port on Netgear D7800
see https://patchwork.kernel.org/patch/8686761/ for details:

```
On some SOCs PORTS_IMPL register value is never programmed by the BIOS
and left at zero value. Which means that no sata ports are avaiable for
software. AHCI driver used to cope up with this by fabricating the
port_map if the PORTS_IMPL register is read zero, but recent patch
broke this workaround as zero value was valid for nvme disks.

This patch adds ports-implemented dt bindings as workaround for this issue
in a way that DT can dictate the port_map incase where the SOCs does not
program it already.
```

### DIFF
--- a/target/linux/ipq806x/files-4.4/arch/arm/boot/dts/qcom-ipq8064-d7800.dts
+++ b/target/linux/ipq806x/files-4.4/arch/arm/boot/dts/qcom-ipq8064-d7800.dts
@@ -148,6 +148,7 @@
 		};
 
 		sata@29000000 {
+			ports-implemented = <0x1>;
 			status = "ok";
 		};
 
--- a/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-d7800.dts
+++ b/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-d7800.dts
@@ -148,6 +148,7 @@
 		};
 
 		sata@29000000 {
+ 			ports-implemented = <0x1>;
 			status = "ok";
 		};
 


mkresin:
The patches implementing "ports-implemented" are:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2fd0f46cb1b82587c7ae4a616d69057fb9bd0af7

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=17dcc37e3e847bc0e67a5b1ec52471fcc6c18682

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bb4add2ce991e4ec891b5a0287fd1ab77b631979

And they are only included in linux kernel 4.6+.

ports-implemented was added for the only upstream available ipq8064 board qcom-ipq8064-ap148.dts with commit https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=67f8425d0ee18c1047a5aaaea529d14d32ca9b43.

Which means:

1) we are overwriting a dts which is available upstream instead of patching our custom changes into the existing dts (if there are any)

2) changing the kernel 4.4 dts doesn't make any sense

3) I would be surprised that the D7800 and AP148 are the only boards affected by the issue

Beside of that you missed to mention the error you are seeing.

Last but not least, your PR has formal issues. Your commits don't have a signed-off-by and the commit message doesn't really describe anything. Linking to some RFC patches isn't enough. Please have a look at our [submitting patches guideline](https://lede-project.org/submitting-patches) for details.

The commit title should be "ipq806x: enable eSATA port on ipq8064 based boards". But "enable" most likely isn't the correct word here, since they are already enabled. It rather seems to me something gets fixed.

@ianchi @KanjiMonster

ianchi:
Formal issues aside, I agree with adding `ports-implemented` to DTS, I had in fact a pending PR to submit this change for EA8500. I think this should be added to all boards that enable SATA (d7800, ap148, ap149, r7800, ea8500, r7500, r7500v2).
I also agree that it needs to go in kernel 4.9.
In fact I don't know why **files-4.4** is kept at all. All devices in this platform use 4.9, and patches & config for 4.4 where already deleted a long time ago.

@dissent1 you have looked deep into this platform. Any comments?

dissent1:
We already have this property added into R7800 back in k4.4 and it did have an effect, SATA port has become available.
https://github.com/dissent1/r7800/commit/bf7166e54569266a7d7525742f3216b4e3064825#diff-8a61e3ed51bbb9e79031d387b2066417R167

I think it's better to put this property into ipq8064.dtsi because it seems tha all devices on this platform with sata port are affected.

PR #1551, opened 2017-11-28
ar71xx: C58/C59 fix LAN1 working incorrectly

blocktrron:
This fixes LAN Port 1 not transferring data in case no other LAN Port has active link-state on TP-Link Archer C58/C59.


NeoRaider:
Thanks, merged to my staging tree.

PR #1553, opened 2017-11-29
mbedtls: Don't override optimization

diizzyy:
There's no apparent reason why we should override default release optimization
set by upstream (-O2) as it doesn't save any space. It also provides a very minor
speed increase, verified using OpenVPN.

Runtime tested and verified on ar71xx
Runtime tested on ramips, MT7621

Source: 
https://github.com/ARMmbed/mbedtls/blob/mbedtls-2.6/CMakeLists.txt#L73

### DIFF
--- a/package/libs/mbedtls/Makefile
+++ b/package/libs/mbedtls/Makefile
@@ -50,6 +50,7 @@ endef
 PKG_INSTALL:=1
 
 TARGET_CFLAGS += -ffunction-sections -fdata-sections
+TARGET_CFLAGS := $(filter-out -O%,$(TARGET_CFLAGS))
 
 CMAKE_OPTIONS += \
 	-DCMAKE_BUILD_TYPE:String="Release" \


jow-:
So you override one forced value with another. If you trully want to fall-through to defaults you should maybe use `$(filter-out -O% ...)` ?

diizzyy:
That seems like a better idea, will update. Thanks!

diizzyy:
On arm (mvebu) the size difference is ~20kbyte fwiw

hauke:
So am I correct that this stays at 298Kbyte  for ar71xx and the speed is a little bit better?
What is the uncompressed size?

diizzyy:
No idea what happened with my test compile (I don't have the env left) but correct results for ar71xx below:

Stock:
IPK: 162090 bytes
Directory: 338503 bytes

Patch applied:
IPK: 162,067 bytes
Directory: 338503 bytes

Sorry for the inconvenience!

blogic:

PR #1554, opened 2017-11-29
fix spelling in CONFIG_DEVTMPFS

refacto:
Changed "ti" to "to", as that's the correct spelling.
Signed-off-by: Sascha Paunovic <azarus@posteo.net>


mkresin:
merged, thanks.

PR #1559, opened 2017-12-02
ipq806x: fix wireless regression on k4.9, sync with QSDK

dissent1:
In current state there's huge regression on ipq806x target that causes the device to transmit broken/malformed frames that are not corrected/detected by error control mechanisms and other less severe issues.
https://bugs.lede-project.org/index.php?do=details&task_id=1197

This finally had been narrowed down to patch 0071-pcie-qcom-fixes.patch

Meanwhile QSDK contains a handful of commits that add support for ipq806x to upstream qcom pcie driver
https://source.codeaurora.org/quic/qsdk/oss/kernel/linux-msm/log/drivers/pci/host/pcie-qcom.c?h=eggplant
Unfortunately qca developers do not bother to push it upstream.

Using those commits instead of lede 0071 patch fixes mentioned issue and probably many others as it seems that corrupted data has been originating within pcie misconfiguration.

Fixes: FS#1197 and probably others
Confirmed by at least 1 more user (besides me). 
https://forum.lede-project.org/t/netgear-r7800-exploration-ipq8065-qca9984/285/571
Need more testers to confirm that it's not just a coincidence.

According to QSDK and OEM tarballs (checked c2600, r7500v2, r7800) 2nd pci slot (pci1, 2,4 GHz card)) on ap148 based boards should operate in gen1 mode.

In previous commit we've added the support for this option, so enable it in DT for affected devices.

QSDK ref:
https://source.codeaurora.org/quic/qsdk/oss/kernel/linux-msm/commit/?h=release/endive_preview_cc&id=f3b07fe309027c52fc163149500cedddd707c506


dissent1:
1 more confirmation
https://forum.lede-project.org/t/netgear-r7800-exploration-ipq8065-qca9984/285/582

throbulator:
I can also confirm that the wireless performance of the R7800 is back to normal after applying this PR.

Tested with current trunk and @hnyman community build.

thess:
ACK from me
My R7800 (custom trunk builds) looks great. Wifi stable.

dissent1:
Moved phy-tx0-term-offset into dtsi file as discussed in irc.

frymezim:
Before patch #1271 I have kernel issue on my EA8500
> [    0.122137] 1b500000.pci supply vdda not found, using dummy regulator
> [    0.122184] 1b500000.pci supply vdda_phy not found, using dummy regulator
> [    0.122218] 1b500000.pci supply vdda_refclk not found, using dummy regulator
> [    0.122837] OF: PCI: host bridge /soc/pci@1b500000 ranges:
> [    0.122878] OF: PCI:    IO 0x0fe00000..0x0fefffff -> 0x0fe00000
> [    0.122903] OF: PCI:   MEM 0x08000000..0x0fdfffff -> 0x08000000
> [    0.231131] qcom-pcie 1b500000.pci: link up
> [    0.233204] qcom-pcie 1b500000.pci: PCI host bridge to bus 0000:00
> [    0.233228] pci_bus 0000:00: root bus resource [bus 00-ff]
> [    0.233247] pci_bus 0000:00: root bus resource [io  0x0000-0xfffff] (bus address [0xfe00000-0xfefffff])
> [    0.233263] pci_bus 0000:00: root bus resource [mem 0x08000000-0x0fdfffff]
> [    0.233314] pci 0000:00:00.0: [17cb:0101] type 01 class 0x060400
> [    0.233459] pci 0000:00:00.0: supports D1
> [    0.233470] pci 0000:00:00.0: PME# supported from D0 D1 D3hot
> [    0.233755] PCI: bus0: Fast back to back transfers disabled
> [    0.234068] pci 0000:01:00.0: [168c:0040] type 00 class 0x028000
> [    0.234298] pci 0000:01:00.0: reg 0x10: [mem 0x00000000-0x001fffff 64bit]
> [    0.235371] pci 0000:01:00.0: PME# supported from D0 D3hot D3cold
> [    0.235811] PCI: bus1: Fast back to back transfers disabled
> [    0.236088] pci 0000:00:00.0: BAR 8: assigned [mem 0x08000000-0x081fffff]
> [    0.236113] pci 0000:01:00.0: BAR 0: assigned [mem 0x08000000-0x081fffff 64bit]
> [    0.236241] pci 0000:00:00.0: PCI bridge to [bus 01]
> [    0.236264] pci 0000:00:00.0:   bridge window [mem 0x08000000-0x081fffff]
> [    0.236670] aer 0000:00:00.0:pcie002: service driver aer loaded
> [    0.237089] 1b700000.pci supply vdda not found, using dummy regulator
> [    0.237126] 1b700000.pci supply vdda_phy not found, using dummy regulator
> [    0.237161] 1b700000.pci supply vdda_refclk not found, using dummy regulator
> [    0.237761] OF: PCI: host bridge /soc/pci@1b700000 ranges:
> [    0.237796] OF: PCI:    IO 0x31e00000..0x31efffff -> 0x31e00000
> [    0.237816] OF: PCI:   MEM 0x2e000000..0x31dfffff -> 0x2e000000
> [    0.348061] qcom-pcie 1b700000.pci: link up
> [    0.350109] qcom-pcie 1b700000.pci: PCI host bridge to bus 0001:00
> [    0.350130] pci_bus 0001:00: root bus resource [bus 00-ff]
> [    0.350145] pci_bus 0001:00: root bus resource [mem 0x2e000000-0x31dfffff]
> [    0.350196] pci 0001:00:00.0: [17cb:0101] type 01 class 0x060400
> [    0.350317] pci 0001:00:00.0: supports D1
> [    0.350327] pci 0001:00:00.0: PME# supported from D0 D1 D3hot
> [    0.350576] PCI: bus0: Fast back to back transfers disabled
> [    0.350891] pci 0001:01:00.0: [168c:0040] type 00 class 0x028000
> [    0.351244] pci 0001:01:00.0: reg 0x10: [mem 0x00000000-0x001fffff 64bit]
> [    0.352398] pci 0001:01:00.0: PME# supported from D0 D3hot D3cold
> [    0.352851] PCI: bus1: Fast back to back transfers disabled
> [    0.353211] pci 0001:00:00.0: BAR 8: assigned [mem 0x2e000000-0x2e1fffff]
> [    0.353235] pci 0001:01:00.0: BAR 0: assigned [mem 0x2e000000-0x2e1fffff 64bit]
> [    0.353367] pci 0001:00:00.0: PCI bridge to [bus 01]
> [    0.353387] pci 0001:00:00.0:   bridge window [mem 0x2e000000-0x2e1fffff]
> [    0.353745] aer 0001:00:00.0:pcie002: service driver aer loaded
> [    0.354166] 1b900000.pci supply vdda not found, using dummy regulator
> [    0.354207] 1b900000.pci supply vdda_phy not found, using dummy regulator
> [    0.354247] 1b900000.pci supply vdda_refclk not found, using dummy regulator
> [    0.354845] OF: PCI: host bridge /soc/pci@1b900000 ranges:
> [    0.354880] OF: PCI:    IO 0x35e00000..0x35efffff -> 0x35e00000
> [    0.354901] OF: PCI:   MEM 0x32000000..0x35dfffff -> 0x32000000
> [    1.388947] qcom-pcie 1b900000.pci: phy link never came up
> [    1.391310] ------------[ cut here ]------------
> [    1.391330] WARNING: CPU: 0 PID: 1 at drivers/clk/qcom/clk-branch.c:97 0xc046d774
> [    1.391342] pcie2_phy_clk status stuck at 'on'
> [    1.391352] Modules linked in:
> [    1.391374] CPU: 0 PID: 1 Comm: swapper/0 Not tainted 4.9.58 #0
> [    1.391384] Hardware name: Generic DT based system
> [    1.391406] Function entered at [c02182b0] from [c0213ee4]
> [    1.391418] Function entered at [c0213ee4] from [c03f3c68]
> [    1.391431] Function entered at [c03f3c68] from [c0221df4]
> [    1.391443] Function entered at [c0221df4] from [c0221e5c]
> [    1.391456] Function entered at [c0221e5c] from [c046d774]
> [    1.391469] Function entered at [c046d774] from [c04620ec]
> [    1.391481] Function entered at [c04620ec] from [c04633a0]
> [    1.391495] Function entered at [c04633a0] from [c045f144]
> [    1.391507] Function entered at [c045f144] from [c045e7c8]
> [    1.391518] Function entered at [c045e7c8] from [c045fcf8]
> [    1.391529] Function entered at [c045fcf8] from [c04a8410]
> [    1.391541] Function entered at [c04a8410] from [c04a68d0]
> [    1.391553] Function entered at [c04a68d0] from [c04a6a44]
> [    1.391564] Function entered at [c04a6a44] from [c04a47b4]
> [    1.391576] Function entered at [c04a47b4] from [c04a5c64]
> [    1.391588] Function entered at [c04a5c64] from [c04a7160]
> [    1.391598] Function entered at [c04a7160] from [c020992c]
> [    1.391611] Function entered at [c020992c] from [c077edec]
> [    1.391623] Function entered at [c077edec] from [c06b0d54]
> [    1.391636] Function entered at [c06b0d54] from [c02102f8]
> [    1.391656] ---[ end trace dfee57b3f61dc821 ]---
> [    1.392020] qcom-pcie 1b900000.pci: PCI host bridge to bus 0002:00
> [    1.392042] pci_bus 0002:00: root bus resource [bus 00-ff]
> [    1.392059] pci_bus 0002:00: root bus resource [mem 0x32000000-0x35dfffff]
> [    1.392113] PCI: bus0: Fast back to back transfers enabled
> [    1.394762] L2 @ QSB rate. Forcing new rate.
> [    1.394947] L2 @ 384000 KHz
> [    1.395112] CPU0 @ 800000 KHz
> [    1.395124] CPU1 @ QSB rate. Forcing new rate.
> [    1.395239] CPU1 @ 384000 KHz
> [    1.397912] gsbi 16300000.gsbi: GSBI port protocol: 6 crci: 0

With this patch the issue is gone.
> [    0.122175] 1b500000.pci supply vdda not found, using dummy regulator
> [    0.122219] 1b500000.pci supply vdda_phy not found, using dummy regulator
> [    0.122254] 1b500000.pci supply vdda_refclk not found, using dummy regulator
> [    0.122878] OF: PCI: host bridge /soc/pci@1b500000 ranges:
> [    0.122921] OF: PCI:    IO 0x0fe00000..0x0fefffff -> 0x0fe00000
> [    0.122946] OF: PCI:   MEM 0x08000000..0x0fdfffff -> 0x08000000
> [    0.231165] qcom-pcie 1b500000.pci: link up
> [    0.231351] qcom-pcie 1b500000.pci: PCI host bridge to bus 0000:00
> [    0.231378] pci_bus 0000:00: root bus resource [bus 00-ff]
> [    0.231398] pci_bus 0000:00: root bus resource [io  0x0000-0xfffff] (bus address [0xfe00000-0xfefffff])
> [    0.231414] pci_bus 0000:00: root bus resource [mem 0x08000000-0x0fdfffff]
> [    0.231910] PCI: bus0: Fast back to back transfers disabled
> [    0.233971] PCI: bus1: Fast back to back transfers disabled
> [    0.234247] pci 0000:00:00.0: BAR 8: assigned [mem 0x08000000-0x081fffff]
> [    0.234274] pci 0000:01:00.0: BAR 0: assigned [mem 0x08000000-0x081fffff 64bit]
> [    0.234401] pci 0000:00:00.0: PCI bridge to [bus 01]
> [    0.234428] pci 0000:00:00.0:   bridge window [mem 0x08000000-0x081fffff]
> [    0.235237] 1b700000.pci supply vdda not found, using dummy regulator
> [    0.235273] 1b700000.pci supply vdda_phy not found, using dummy regulator
> [    0.235307] 1b700000.pci supply vdda_refclk not found, using dummy regulator
> [    0.235892] OF: PCI: host bridge /soc/pci@1b700000 ranges:
> [    0.235926] OF: PCI:    IO 0x31e00000..0x31efffff -> 0x31e00000
> [    0.235947] OF: PCI:   MEM 0x2e000000..0x31dfffff -> 0x2e000000
> [    0.346154] qcom-pcie 1b700000.pci: link up
> [    0.346318] qcom-pcie 1b700000.pci: PCI host bridge to bus 0001:00
> [    0.346339] pci_bus 0001:00: root bus resource [bus 00-ff]
> [    0.346355] pci_bus 0001:00: root bus resource [mem 0x2e000000-0x31dfffff]
> [    0.346784] PCI: bus0: Fast back to back transfers disabled
> [    0.348945] PCI: bus1: Fast back to back transfers disabled
> [    0.349302] pci 0001:00:00.0: BAR 8: assigned [mem 0x2e000000-0x2e1fffff]
> [    0.349327] pci 0001:01:00.0: BAR 0: assigned [mem 0x2e000000-0x2e1fffff 64bit]
> [    0.349457] pci 0001:00:00.0: PCI bridge to [bus 01]
> [    0.349478] pci 0001:00:00.0:   bridge window [mem 0x2e000000-0x2e1fffff]
> [    0.350245] 1b900000.pci supply vdda not found, using dummy regulator
> [    0.350282] 1b900000.pci supply vdda_phy not found, using dummy regulator
> [    0.350322] 1b900000.pci supply vdda_refclk not found, using dummy regulator
> [    0.350915] OF: PCI: host bridge /soc/pci@1b900000 ranges:
> [    0.350949] OF: PCI:    IO 0x35e00000..0x35efffff -> 0x35e00000
> [    0.350971] OF: PCI:   MEM 0x32000000..0x35dfffff -> 0x32000000
> [    1.385164] qcom-pcie 1b900000.pci: phy link never came up
> [    1.387147] qcom-pcie 1b900000.pci: hostinit failed
> [    1.389565] L2 @ QSB rate. Forcing new rate.
> [    1.389749] L2 @ 384000 KHz
> [    1.389915] CPU0 @ 800000 KHz
> [    1.389927] CPU1 @ QSB rate. Forcing new rate.
> [    1.390039] CPU1 @ 384000 KHz
> [    1.392816] gsbi 16300000.gsbi: GSBI port protocol: 6 crci: 0

dissent1:
EA8500 doesn't have pcie2 equipped, the log provided above confirms it.
I've updated the commit to also disable the pcie2 phy in EA8500 DTS.

ianchi:
@dissent1, this is great!
it’s true about EA8500 only one pcie, but back then when I added it to LEDE, I found a very weird issue, that for some reason if you disabled pcie2 you ended up loosing eth.
Perhaps this also solves that problem. And also another unsolved one, where the switch could only be made to see one cpu port. That’s why it’s in tagged mode
We have to test it before committing. I’ll test it in the weekend.

nbd168:

ianchi:
@nbd168 please don't merge this to master!!, it still kills the switch. 

@dissent1 the old problem I commented isn't fixed. Please leave pcie2 enabled for EA8500. At least until we finally figure out why it's interfering with the switch.



```
[    4.177764] init: - preinit -                                                                                        
[    4.540653] usb 3-1: new high-speed USB device number 2 using xhci-hcd                                               
[    4.719597] hub 3-1:1.0: USB hub found                                                                               
[    4.720000] hub 3-1:1.0: 4 ports detected                                                                            
[    5.068899] usb 3-1.3: new high-speed USB device number 3 using xhci-hcd                                             
[    5.231374] hub 3-1.3:1.0: USB hub found                                                                             
[    5.231664] hub 3-1.3:1.0: 4 ports detected                                                                          
Failed to connect to the switch. Use the "list" command to see which switches are available.                            
Failed to connect to the switch. Use the "list" command to see which switches are available.                            
Failed to connect to the switch. Use the "list" command to see which switches are available.                            
Failed to connect to the switch. Use the "list" command to see which switches are available.                            
[    5.473348] eth0: device MAC address 00:06:3b:01:41:00                                                               
[    5.574326] ipq806x-gmac-dwmac 37200000.ethernet: Failed to reset the dma                                            
[    5.574351] stmmac_hw_setup: DMA engine initialization failed                                                        
[    5.580155] stmmac_open: Hw setup failed                                                                             
ip: SIOCSIFFLAGS: Resource busy                                                                                         
sendto(): Network unreachable                                                                                           
Press the [f] key and hit [enter] to enter failsafe mode                                                                
```

dissent1:
Interesting, I've just downloaded ea8500 GPL and here what I've found
-msm_pcie_platform_data[1].force_gen1 = 1;
+/* patch from QCA 04-01-2015 c/o mtien@qca.qualcomm.com */
+msm_pcie_platform_data[0].force_gen1 = 1;

So instead of pcie1, for ea8500 pcie0 should operate in gen1 mode.

I'll take a deeper look into ea8500 maybe I'll find inconsistency that requires unequipped pcie slot to be enabled in order for switch to work.

dissent1:
@ianchi according to the GPL gpio63 is for board reset control (ps_hold?) and not the pcie2 reset.
Could you try out this patch on top of this PR?
https://github.com/dissent1/r7800/commit/33d546ae3c2bf6dd67386d20133797f7b65b4d9a

ianchi:
@dissent1 thanks for looking  into this!
 I'm recompiling at the moment with that change. Will let you know as soon as it's done.

I'm crossing my fingers that the bootloader sets it right. As you posted, in the GPL gpio63 is explicitly set.

ianchi:
@dissent1 It seems that the bootloader doesn't set it right:
The switch is gone:
```
Tue Dec 12 11:23:14 2017 daemon.notice procd: /etc/rc.d/S20network: Failed to connect to the switch. Use the "list" command to see which switches are available.
```

Apparently we need a way to set gpio63 to the proper value in DTS

dissent1:
@ianchi 
Try this
https://github.com/dissent1/r7800/commit/867a83d70b6ea5d62821a9fefb0f67ca4c882bf5

dissent1:
@ianchi try this patch
https://github.com/dissent1/r7800/commit/417cbc4a3fa923b51d62472fd7a0cb26dbe29b93

dissent1:
@ianchi a typo in previous commit :/
Try this one https://github.com/dissent1/r7800/commit/10af33fa966fdeac275696c7f93c038b1a0e92c2

dissent1:
@ianchi and this
https://github.com/dissent1/r7800/commit/cb82b2c813526af388c994401dda07a3a29e2c40

ianchi:
Hi, @dissent1 I'll try the last pach, but with the previous one (adding reset control) I get an error when compilling to dtb:

```
Warning (reg_format): "reg" property in /soc/nand@1ac00000/cs0 has invalid length (4 bytes) (#address-cells == 1, #size-cells == 1)
Error: arch/arm/boot/dts/qcom-ipq8064-ea8500.dts:43.23-24 syntax error
FATAL ERROR: Unable to parse input tree
scripts/Makefile.lib:313: recipe for target 'arch/arm/boot/dts/qcom-ipq8064-ea8500.dtb' failed

```

dissent1:
Hmm, strange. I'll adjust the commit tomorrow so it doesn't break compilation.

ianchi:
@dissent1 I changed `reset-control` to `reset_control` and it compiles OK.

Regarding RGMII2, the codeaurora patch adds gpio2 and gpio66 to the definition, but in the DTS you are only adding gpio2. Is this OK or shouldn't we also add gpio66?
Your are a genius, this must be the missing piece to solve cpu1 in the switch!!

I'm trying this out right now...

dissent1:
Interesting, maybe reset-control is reserved or smth.
Only gpio2 in case if EA8500, in GPL gpios end with gpio65. Gpio 2 there is for nss_gmac0_set0
Don't await a lot, it may not solve the problem :)

ianchi:
In the GPL patch I have:
```
+	ipq_mux(62, nss_gmac1_rgmii_set,		nss_gmac1_rgmii_set),
+	ipq_mux(63, pcie_rst_n,				pcie_rst_n),
+	ipq_mux(66, nss_gmac0_rgmii_set1,		nss_gmac0_rgmii_set1),
+};
+
+static struct msm_gpiomux_config ipq806x_db149_1xx_gpiomux[] = {
+	ipq_mux( 0, mdio_n,				mdio_n),
+	ipq_mux( 1, mdio_n,				mdio_n),
+	ipq_mux( 2, nss_gmac0_rgmii_set0,		nss_gmac0_rgmii_set0),
+```
I think we should keep gpio66 in DTS.

By the way, the pcie2 / gpio63 thing did work!

**Comment by dissent1** [comment]
It's the definition for the wrong board, you have to look for ipq806x_ap148_*

**Comment by ianchi** [comment]
Oh, you are right.

I couldn’t make the switch’s port 6 work, anyway.

I put ports 5&6 in the same VLAN. Port 6 shows UP in swconfig, eth1 shows UP in ip link.
But no transmission.

**Comment by dissent1** [comment]
@nbd168 here's the updated commit for gen1 mode. https://github.com/dissent1/r7800/commit/e4167709968dd0413bc7f23155f39b97c3452e8d
In EA8500 DTS I've moved gen1 from pcie1 to pcie0 according to GPL for this device and excluded pcie2 node removal.

I'll make a new PR with the fix to disable the unequipped pcie2 slot so it doesn't break the switch.

**Comment by dissent1** [comment]
@ianchi 

> I put ports 5&6 in the same VLAN. Port 6 shows UP in swconfig, eth1 shows UP in ip link.
> But no transmission.

Was it the same way before these changes?

**Comment by dissent1** [comment]
@ianchi hmm, could you try also adding gpio66 into rgmii2_pins for testing purposes?

**Comment by ianchi** [comment]
@dissent1 I think that eth1 is the same as before. But I don’t really remember well. I haven’t try again since almost two years.
Tonight I’ll try gpio66 and also going back to check differences.

**Comment by nbd168** [comment]
Merged the updated commit to my staging tree, please let me know if I should put in anything else.

**Comment by dissent1** [comment]
@nbd168 thanks! It's clear for takeoff, it doesn't break EA8500.

I'll form a new PR with EA8500 fixes upon further investigation.

**Comment by ianchi** [comment]
@dissent1 I find the exact same problem with the old rgmii2_pins, with +=gpio2 and with +=gpio2+gpio66
Couldn't tell any difference with ip link / swconfig. In all cases I have some TX bytes from this interface but 0pkts RX
Any other tool to check?
By the way, in none of the configs port6 (eth1) showed up on the LUCI interface. Not even when I included it in the VLAN via directly editing the config.

**Comment by ianchi** [comment]
Another doubt. All the changes where on RGMII, but eth1 (gmac2) is set to work on `sgmii` mode.
The one that was in `rgmii` mode is eth0 (gmac1) which allways worked...

**Comment by dissent1** [comment]
@ianchi you know, when struggling with r7800 I've understood that even at first glance unrelated stuff  may cause an issue, like gpio63 thing. gpio63 (while others use it as pcie reset) seem to be a hardware reset pin in EA8500 that gets pushed during restart and why should it have an effect on switch or gmac1? :)
Actually I couldn't find anything in GPL that could have a non default configuration for gmac2 or port6.
The only thing that I'd try is to play with this:
https://github.com/dissent1/r7800/blob/sta106/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064-ea8500.dts#L312

IMO bits 24, 25 and 26 should be disabled according to GPL (but I just could get it wrong). Although we have a forced link, who knows... Desc:

26
R/W
default 1
PAUSE_25M
Pause in the base-page of base-x and SGMII-PHY/MAC for auto-negotiation. This part is not included in the standard for SGMII.

25
R/W
default 1
ASYM_PAUSE_25M
ASYM_PAUSE in the base-page of base-x and SGMII- PHY/MAC for auto-negotiation. This part is not included in the standard for SGMII.

24
R/W
default 1
PAUSE_SG_TX_EN_ 25M
Enable transmitting pause in the base-page of base-x and SGMII-PHY/MAC for auto-negotiation.

Also try bits 23:22, desc:
23:22
R/W
default 0
MODE_CTRL_25M
MODE_CTRL signal for mode selection among BASE-X (2'h0), SGMII-PHY (2'h1), and SGMII-MAC (2'h2).

**Comment by dissent1** [comment]
@ianchi 
Btw, if interested you may try blogic's qca8k patch. It has a bit different logic attaching to switch ports. 
https://github.com/dissent1/r7800/commit/fd065e652c2d49db3fab42f800ff9002ce4993fa
Take r7800 mdio node as an example, you need to add CPU port phandles into EA8500 dts and enable port6. Or you can simply copy the whole node from R7800.

**Comment by ianchi** [comment]
@dissent1 back in the day I tried many different values for SGMII_CTRL registers, with no success.
But in fact I checked with running stock firmware and the values actually coincide.

Tonight I'll give a tray to @blogic 's qca8k driver. Why was it never merged into master anyway?
I couldn't find anything about it.

**Comment by ianchi** [comment]
Aside from the switch, I think that the other change you made to EA8500 DTS' (adding reset_control  and erasing pcie2) are ready for merging, regardless of what we acomplish with the cpu port.

**Comment by dissent1** [comment]
@ianchi blogic's patch is not ready yet. I've linked you a stripped version, that adds support for multiple CPU ports into kernel DSA driver and mdb offloading only. The full version supports hw offloading capabilities of qca 8337n switch. blogic awaits till netfilter supports such kind of offloading upstream (rfc patches are on ML since November) to push his patches further.
Full version: https://git.lede-project.org/?p=lede/blogic/staging.git;a=shortlog;h=refs/heads/qca8k

Regarding gpio63 - I'm coming up to a conclusion that this gpio may operate as ps_hold function. Could you try this as well? 
https://github.com/dissent1/r7800/commit/26b73f8501562d43a26a020216f2c5e0bd5626b0

**Comment by ianchi** [comment]
Ok, I'll try that also. But why ps_hold? In the GPL it is set as GPIO:

```
static struct gpiomux_setting reset_control = {
	.func = GPIOMUX_FUNC_GPIO,
	.drv = GPIOMUX_DRV_2MA,
	.pull = GPIOMUX_PULL_NONE,
	.dir = GPIOMUX_OUT_LOW,
};

...

static struct msm_gpiomux_config ipq806x_ap148_gpiomux[] = {
...
	ipq_mux(63, reset_control,			reset_control),
	ipq_mux(65, ipq806x_buttons_cfg,		ipq806x_buttons_cfg),
};
```

What I don't see where we replicate is the direct writing to this gpio, which GPL does at startup:

```
void msm_restart(char mode, const char ...
...
#ifdef CONFIG_MACH_LINKSYS_EA8500
	__raw_writel(1 << 1, GPIO_IN_OUT(63));
#endif 
```

**Comment by dissent1** [comment]
@ianchi I was wondering what happens here
https://github.com/torvalds/linux/blob/37cb8e1f8e10c6e9bd2a1b95cdda0620a21b0551/drivers/pinctrl/qcom/pinctrl-msm.c#L863

But at a closer look it seems that we should first put gpio63 into ps_hold group in pinctrl-ipq806x.c.
Anyway trying to directly write that value from GPL seem to be a more reasonable decision, who knows maybe that value defines some kind of hardware state that may enable the gmac2 link, considering that just pulling the gpio enabled the gmac1

**Comment by dissent1** [comment]
@ianchi when you add the gpio63 as function=gpio in what state is it exported? Active low or high? Could you check please. 
Bit 0 in gpio mask means 0=active high, 1= low, bit 1: 0 - push-pull, 1 - single-ended. Tried to understand what 1 << 1 means.
Maybe 'output-high' property is missing?

**Comment by ianchi** [comment]
now it is exported as:
active_low
direction = in
edge = none
value = 0

**Comment by dissent1** [comment]
@ianchi actually I'm pretty convinced now that this gpio63 is ps_hold alike but for some different purpose.
That write is executed during restart only - it switches the gpio into output (you can test what happens when you set gpio63 direction=out || direction=out && active high).
But this gpio has a different offset comparing to ps_hold
#define PS_HOLD_OFFSET 0x820
#define GPIO_IN_OUT(gpio)         (msm_gpio.msm_tlmm_base + 0x1004 + (0x10 * (gpio)))
So ps_hold function can't be used directly. Adjusting the kernel code to trigger that gpio doesn't seem to worth the effort considering that nothing bad happens in current state when the gpio63 is constantly pulled.

BTW I've found out that gpio2 is used for switch link state change in qsdk.

Regarding the gmac2 issue actually I'm out of clues what could it be caused by, maybe some registers are not set in hardware to default state as used in other boards, but qsdk(gpl) seem to set all available registers while upstream kernel sets just some of them relying that others are default.

For example from dd-wrt for trendnet and asrock ipq806x boards it's required to directly set a phy-handle = "&phy4" to a gmac1 node.

**Comment by ianchi** [comment]
@dissent1 I wanted to make a dirty test and replicate the same GPL write to gpio63, to see if something changed.
But I'm no sure where/how to do it.
I tried this, but linux won't even boot (no single output to console). So I'm doing it wrong, any ideas on how to do it?:
```
Index: linux-4.9.67/drivers/pinctrl/qcom/pinctrl-ipq8064.c
===================================================================
--- linux-4.9.67.orig/drivers/pinctrl/qcom/pinctrl-ipq8064.c
+++ linux-4.9.67/drivers/pinctrl/qcom/pinctrl-ipq8064.c
@@ -16,6 +16,7 @@
 #include <linux/of.h>
 #include <linux/platform_device.h>
 #include <linux/pinctrl/pinctrl.h>
+#include <linux/io.h>
 
 #include "pinctrl-msm.h"
 
@@ -633,9 +634,15 @@ static const struct msm_pinctrl_soc_data
 	.gpio_pull = &msm_gpio_pull,
 };
 
+#define MSM_TLMM_BASE		IOMEM(0xFA017000)	/* 16K	*/
+#define GPIO_IN_OUT(n) (MSM_TLMM_BASE + 0x1004 + (0x10 * n ))
+
 static int ipq8064_pinctrl_probe(struct platform_device *pdev)
 {
-	return msm_pinctrl_probe(pdev, &ipq8064_pinctrl);
+	int ret = msm_pinctrl_probe(pdev, &ipq8064_pinctrl);
+	__raw_writel(1 << 1, GPIO_IN_OUT(63));
+
+	return ret;
 }
 
 static const struct of_device_id ipq8064_pinctrl_of_match[] = {

```

Additionally, following GPL settings I added in DTS:
```
			reset_control: reset_control_pins {
 				mux {
 					pins = "gpio63";
 					function = "gpio";
 					drive-strength = <2>;
 					bias-disable;
+					bias-pull-down;
 					output-low;
 				};
 			};
```

**Comment by ianchi** [comment]
Another comment.
If I export gpio63, change directio to output and issue:
`echo 1 > value`
The switch is turned off (all lights go off).
If then I do `echo 0 > value` it seems to restart. Lights go on port by port in sequence and then starts blinking.
So it seems that this gpio enables/disables the switch

**Comment by dissent1** [comment]
@ianchi Try this as a base
https://github.com/dissent1/r7800/blob/sta106/target/linux/ipq806x/files-4.9/arch/arm/boot/dts/qcom-ipq8064.dtsi#L622
The one you are defining seem to be out of ipq806x ranges and correspond to other msm boards, non apq/ipq.
 Btw isn't that reset_control defined with GPIOMUX_PULL_NONE and not pull-down?

**Comment by dissent1** [comment]
btw have you checked eth1 state when doing all those tests?

**Comment by dissent1** [comment]
@ianchi a comment by brainslayer (dd-wrt):
there is a trick. you can measure the voltage on the
switch pins and enable / disable gpio63 to find out which pin is
triggered by it. then look in the datasheet and you will find out more

**Comment by ianchi** [comment]
@dissent1 I'll do some more tries in the evening.
So far no luck with eth1.

Regarding the base: in the GPL I took the address from:
```
+#if defined(CONFIG_ARCH_MSM8960) || defined(CONFIG_ARCH_APQ8064) || \
+	defined(CONFIG_ARCH_MSM7X27) || \
+	defined(CONFIG_ARCH_MSM7X25) || defined(CONFIG_ARCH_MSM7X01A) || \
+	defined(CONFIG_ARCH_MSM8625) || defined(CONFIG_ARCH_MSM7X30) || \
+	defined(CONFIG_ARCH_IPQ806X)
+
+/* Unified iomap */
....

+#define MSM_TLMM_BASE		IOMEM(0xFA017000)	/* 16K	*/
```
I'm not sure how this maps to the `reg = <0x800000 0x4000>;` in the DTS

And regarding the pull, in the GPL:

```
static struct gpiomux_setting reset_control = {
	.func = GPIOMUX_FUNC_GPIO,
	.drv = GPIOMUX_DRV_2MA,
	.pull = GPIOMUX_PULL_NONE,
	.dir = GPIOMUX_OUT_LOW,
};
```

I'm going to try the new dsa driver from blogic.

dissent1:
@ianchi Yeah, I know that that value is from gpl, but Ipq806x tlmm pinmux is at 0x800000 for sure. I too don't understand how that gpl value maps :)

Regarding GPIOMUX_PULL_NONE - it is biased-disabled, not pull-down :) besides bias-disabled and bias-pull-down do not coexist

dissent1:
@ianchi I've just checked apq8064 datasheet (twin brother) and those gpio registers are exactly at
0x00801000+0x10*n
So just replace the tlmm base with the value from dts

ianchi:
Ok, will try that later. And removing the pull-down (my bad).
Anyway, it is strange the address from GPL...

dissent1:
@ianchi and btw I guess it's better to rename MSM_TLMM_BASE and GPIO_IN_OUT into smth different so it doesn't redefine the current kernel code with updated base address and offset.

dissent1:
@ianchi don't bother about the gpio, here's gpio bit description
0x00801004+0x10*n GPIO_IN_OUTn, n=[0..89]

Type: Read/Write
Clock: CC_TLMM_HCLK Reset State: 0x0000
The GPIO_IN_OUT[n] controls the Out and In on the GPIO. The GPIO_IN bit is read-only.
GPIO_IN_OUTn
 
 Bits
Name
31:2
RESERVED
1
GPIO_OUT
0
GPIO_IN
 
That's from apq8064 data sheet. That gpl write just sets bit 1 to 1 thus enabling output.

ianchi:
Yes, I've just came to the same conclusion about the bits reading the driver (in pinctrl-ipq8064.c).

But I don't think that setting bit 1 to 1 *enables* output, that would be writing to the ctl_reg's oe_bit. It is actually writing the out value of the gpio to 1.

So, what I supposed was that it writes 1 to the out of gpio 63 to reset the switch on each restart. Perhaps to override some settings from the bootloader?

I succeeded in doing the `writel`.
The issue is that the `1` just turns off the switch (as expected by the tests when exporting the gpio), so that no switch driver is loaded and errors in resetting gmac are reported.
So somewhere/somehow the GPL version must be turning it on again. But I couldn't find how.

So, I tried to write 1 and then write 0, so as to restart it. The switch comes up OK this way.
But same problem as ever with eth1. It shows as UP everywhere, but there is no communication thru it.
swconfig as always shows some packtes as TX, but zero as RX in port 6.

@dissent1, for now, I think that you could send a PR with your changes to ea8500.dts, so at least the pcie2 errors are cleared. 
The gmac2/eth1 issue remains a mistery....

hnyman:
@nbd168 @dissent1 
Anything still preventing merging these wifi/pcie fixes? At least the users of my R7800 community build seem to be happy about the changes (commits 8d7faff57 and  f7f071c21a are included in that build).

frymezim:
@dissent1 @ianchi 
Any problems with this patch against EA8500? dissent1 already disabled the PCIE2 on this patch any other changes to be maid here?

dissent1:
I'm waiting till @nbd168 pushes this patch into master. After that I'll make a PR to disable the pcie on EA8500. Otherwise patches won't apply cleanly.

nbd168:
Pushed to master, sorry for the delay.

dissent1:
@ianchi 
Check this commit that had been picked in k4.9.77
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/?h=v4.9.77&id=6f237183c7cad875e5ddc7c32641094117b036c9
Maybe it fixes sgmii interface in EA8500

PR #1561, opened 2017-12-04
mvebu: Enable UAS (UASP)

diizzyy:
As USB storage support is already enabled, add UAS (UASP) as well.


diizzyy:
Compile tested: mvebu, Linksys WRT3200ACM, LEDE trunk
Run tested: mvebu, Linksys WRT3200ACM, LEDE trunk

ptpt52:
This is already export in kmod-usb-storage-uas
You could enable it by 

> make menuconfig

no need to enable by default in kernel_config

thagabe:
If mvebu has UASP then it should be auto-enabled by selecting Kernel >> Can-Storage >> Marvell-sata

ptpt52:
well, if that is your wish. just do it like this:

```
diff --git a/target/linux/mvebu/image/Makefile b/target/linux/mvebu/image/Makefile
index f0c4b09..05f1e17 100644
--- a/target/linux/mvebu/image/Makefile
+++ b/target/linux/mvebu/image/Makefile
@@ -119,7 +119,9 @@ define Device/linksys-wrt3200acm
   $(call Device/linksys,WRT3200ACM (Rango))
   $(Device/armada-385-linksys)
   DEVICE_DTS := armada-385-linksys-rango
-  DEVICE_PACKAGES += kmod-btmrvl kmod-mwifiex-sdio
+  DEVICE_PACKAGES += kmod-btmrvl kmod-mwifiex-sdio \
+                                        kmod-ata-core kmod-ata-marvell-sata \
+                                        kmod-usb-storage-uas
 endef
 TARGET_DEVICES += linksys-wrt3200acm
 
```

thagabe:
^ this looks like the right solution

neheb:
NACK

The USB interface does not support UASP at all. I've tested with 6 different enclosures and none use UASP whereas they do on my laptop.

diizzyy:
@ptpt52 @thagabe 
Both yes and no, the downside is that it doesn't work if you're booting off a USB device as the kernel needs to have the driver and not read it off a filesystem.

@neheb 
What device and chipsets did you use for testing?

neheb:
Turris Omnia. it's based on a WRT1900ACS platform minus the WiFi. The upstream kernel driver does not use streams and as such cannot support UAS. I'd love to be proven wrong though.

There are a bunch of USB related debug messages in the kernel but I have no idea how to enable them. They should show why UAS doesn't work.

thagabe:
Was this ever transferred to the new repo?

bobafetthotmail:
no, but you can install the kmod-usb-storage-uas module. It seems none above confirmed that it is actually supported by their devices.

thagabe:
Got it, thank

PR #1564, opened 2017-12-07
ar71xx: generate BR region-code factory image for TP-Link TL-WR940N

heinzek:
First commit add BR region code to mktplinkfw tool.
Second commit add factory image with BR region code for TP-Link TL-WR940N V4

### DIFF
--- a/target/linux/ar71xx/image/tp-link.mk
+++ b/target/linux/ar71xx/image/tp-link.mk
@@ -1077,9 +1077,10 @@ define Device/tl-wr940n-v4
   BOARDNAME := TL-WR940N-v4
   DEVICE_PROFILE := TLWR941
   TPLINK_HWID := 0x09400004
-  IMAGES += factory-us.bin factory-eu.bin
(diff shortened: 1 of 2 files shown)


mportela:
i’ve tested here and the factory-br image works in the tplink Web update interface. Using tplink 940n V5 BR

mkresin:
We have a similar PR#1574 sets Code 2 for region BR. Since this one is confirmed working, I suspect that PR#1574 is wrong.

mportela:
@mkresin no this PR is realy correct, the code for BR is 0, as i get explanation from the author "Matthias Schiffer" by email:
```
Adding another line {"BR", 0} should be sufficient (the second field 0/1 is
for another type of region specification that added the region as a third
field after hwid and hwrev in the firmware header, but as far as I know
this has only be used in the US, and the field is 0 for both old
"universal" firmwares as well as all non-US regions I've seen so far).
```

blogic:
merged, Thanks !

mportela:
@blogic sorry for my ignorance but did not see these commits in the lede: master .. or did they go to another branch first?

PR #1566, opened 2017-12-08
ramips: fix a typo in 02_network

dissent1:
There's is a typo in network defaults script in ramips target that prevents defaults to initialize.


mkresin:
merged, thanks!

PR #1567, opened 2017-12-09
busybox: enable find -newer

lucize:
needed for shorewall firewall, no size increase on binary


pprindeville:
You might want to bump the `Makefile` `PKG_VERSION`, though.

hauke:
It is only added very few source code lines, so it looks ok to me.

blogic:
merged, Thanks !

PR #1568, opened 2017-12-09
toolchain/musl: Use upstream configuration

diizzyy:
* Strip optimization CFLAGS provided by buildroot
* Use upstream optimization CFLAGS (configure switch)

### DIFF
--- a/toolchain/musl/common.mk
+++ b/toolchain/musl/common.mk
@@ -9,7 +9,7 @@ include $(INCLUDE_DIR)/target.mk
 
 PKG_NAME:=musl
 PKG_VERSION:=1.1.18
-PKG_RELEASE=1
+PKG_RELEASE=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_VERSION)
@@ -26,6 +26,8 @@ HOST_BUILD_DIR:=$(BUILD_DIR_TOOLCHAIN)/$(PKG_NAME)-$(PKG_VERSION)
 include $(INCLUDE_DIR)/toolchain-build.mk
 include $(INCLUDE_DIR)/hardening.mk
 
+TARGET_CFLAGS := $(filter-out -O%,$(TARGET_CFLAGS))
+
 MUSL_CONFIGURE:= \
 	$(TARGET_CONFIGURE_OPTS) \
 	CFLAGS="$(TARGET_CFLAGS)" \
@@ -35,7 +37,8 @@ MUSL_CONFIGURE:= \
 		--host=$(GNU_HOST_NAME) \
 		--target=$(REAL_GNU_TARGET_NAME) \
 		--disable-gcc-wrapper \
-		--enable-debug
+		--enable-debug \
+		--enable-optimize
 
 define Host/Prepare
 	$(call Host/Prepare/Default)


diizzyy:
https://git.musl-libc.org/cgit/musl/tree/configure#n391
https://git.musl-libc.org/cgit/musl/tree/configure#n412
Explains what both switches does.

NeoRaider:
In my opinion, our own -O flags should always have precedence over whatever projects default to. We have those flags in our TARGET_CFLAGS for a reason.

diizzyy:
While agree with you in general I fail to see why we should override upstream since as far as I can tell we are blindly overriding -O flags that aren't evaluated.

NeoRaider:
We usually want to pass -Os (many of our targets don't have much space), while most projects default to -O2 or -O3 (which makes sense for most usecases, but it's not a good default for OpenWrt/LEDE).

Removing flags from CFLAGS might be acceptable if there is a concrete problem with a specific package and optimization level, but not without a good reason.

diizzyy:
Did you have a look at musl's configure script?
It sets -Os by default, however when not using debug it uses -O3 for "internal,malloc,string" which makes sense as those are performance sensitive. I'll have a look at binary size tomorrow but I doubt it'll be much of a difference.

NeoRaider:
Thanks for the explanation, I had only a short glance at the script, and I can't say I grasp what all those variables do exactly without having a closer look. Please add a comment to the line that filters the TARGET_CFLAGS, explaining musl's default optimization flags.

I'm not entirely happy with simply dropping -O* flags, as such flags might also have been specified by the user in CONFIG_TARGET_OPTIMIZATION, but if musl really has good defaults and uses -Os for most files, doing so is probably a good idea.

I don't think CONFIG_DEBUG should disable optimization; normally, it just adds -g3 without removing any optimization flags.

diizzyy:
Ok, I was under the impression that anything above O(1) isn't really desirable if you want to do debugging but I can add both I guess.

neheb:
There is also -Og for debugging. Probably better than -g3 ¯\_(ツ)_/¯

diizzyy:
@NeoRaider 
2kbyte increase on mips (ar71xx) :-)

jow-:
2KB is significant. Any noticeable speed increase? If not I tend to NACK this.

diizzyy:
Hmmm...

https://git.musl-libc.org/cgit/musl/commit/configure?id=a80847d86a8865a78fdbebe7f9e2533f7a74e010
Followed by https://git.musl-libc.org/cgit/musl/commit/configure?id=43d2531316270633589c67cf0c4d6d01e520a397

So it appears that upstream has done performance evaluation.

While being library wide optimizations the tests as far as I can tell mostly touches the optimized parts that benefits. (GCC 7.1.0)

https://rv8.io/bench

http://l4.rv8.io/plots/optimisation-x86-64.svg
http://l4.rv8.io/plots/optimisation-x86-32.svg
http://l4.rv8.io/plots/optimisation-riscv64.svg
http://l4.rv8.io/plots/optimisation-riscv32.svg

pprindeville:
> Ok, I was under the impression that anything above O(1) isn't really desirable if you want to do debugging but I can add both I guess.

Strongly agree.

pprindeville:
My takeaway is that for integer intensive stuff, `-O2` seems to be slightly better.

pprindeville:
> 2KB is significant. Any noticeable speed increase? If not I tend to NACK this.

We're talking about a .47% delta for the overall library size.  It's not clear that's significant.

2KB for a library that was 50KB long would be more worrisome.

pprindeville:
> 2KB is significant. Any noticeable speed increase? If not I tend to NACK this.

We're talking about a .34% delta for the overall library size.  It's not clear that's significant.

2KB for a library that was 50KB long would be more worrisome.

nbd168:
Regarding --enable-debug: There may be some cases where reducing optimization may be necessary for debugging, and CONFIG_DEBUG is useful for that.
In many other cases it works with optimized builds just fine, and enabling debugging symbols does not add any overhead to the packaged binaries, so it should be enabled unconditionally.

diizzyy:
Upstream also does this https://git.musl-libc.org/cgit/musl/tree/configure#n455 which may/may not be desirable? A bit unclear to me is why we enable debugging symbols for musl and pretty much nothing else as far as I can tell by default?

nbd168:
Actually, many packages enable debug symbols implicitly by default. The kernel does as well. 
libc is special here, because it also gets distributed with things like the SDK.
I also frequently rely on being able to debug user space binaries in the default build. Whenever I get a segfault in a user space program, I simply recompile that individual package (and associated libs if necessary) with CONFIG_DEBUG override on the command line, and then debug it with gdbserver on the device.
This works even without updating binaries on the target, because typically CONFIG_DEBUG does not affect the generated code.
If I have to recompile libc and swap it out on the device (because the generated code is affected), it's really inconvenient. Even more so for SDK users.

diizzyy:
@nbd168 
Better now?

nbd168:
Yes, thanks.

PR #1569, opened 2017-12-10
ipq806x: update clk and cpufreq drivers

dissent1:
A newer clk and cpufreq drivers for ipq806x platform had been sent upstream.

A change that i have noticed is that now it's possible to set min, cur and max frequencies from sysfs (previously it was bugged and caused nothing).

Following patches are removed:
 - 0036-clk-Avoid-sending-high-rates-to-downstream-clocks-du.patch - seems it was dropped from the patchset by current committer.
 - 0044-clk-qcom-krait-Remove-CLK_IS_ROOT.patch - already applied to the driver itself in the corresponding patch.
 - 0057-clk-qcom-Add-regmap-mux-div-clocks-support.patch - seem to be irrelevant to ipq806x.

@blogic


nbd168:

PR #1573, opened 2017-12-11
treewide: add only one device when appending to TARGET_DEVICES

luaraneda:
I've encountered conflicts several times when rebasing a developing branch (ipq806x) against master.

This will avoid some conflicts when doing a git rebase or merge,
specially when adding support to a a new device.

### DIFF
--- a/target/linux/brcm47xx/image/Makefile
+++ b/target/linux/brcm47xx/image/Makefile
@@ -519,7 +519,16 @@ define Device/usrobotics-usr5461
 endef
 TARGET_DEVICES += usrobotics-usr5461
 
-TARGET_DEVICES += standard standard-noloader-gz
+define Device/legacy-standard
+  $(Device/standard)
+endef
+TARGET_DEVICES += legacy-standard
+
+define Device/legacy-standard-noloader-gz
+  $(Device/standard-noloader-gz)
+endef
+TARGET_DEVICES += legacy-standard-noloader-gz
+
 endif
 
 #################################################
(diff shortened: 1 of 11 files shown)


target/linux/brcm47xx/image/Makefile
```diff
@@ -519,7 +519,16 @@ define Device/usrobotics-usr5461
...
+define Device/legacy-standard
+  $(Device/standard)
+endef
+TARGET_DEVICES += legacy-standard
+
+define Device/legacy-standard-noloader-gz
+  $(Device/standard-noloader-gz)
+endef
+TARGET_DEVICES += legacy-standard-noloader-gz
+
```
NeoRaider: What's this, these devices are renamed? This looks like an unrelated change that should be a separate commit.

mkresin: It isn't really unrelated, it is just really hard to get from the diff why it is done.

At the moment we have the same image referenced twice. One time in `ifeq ($(SUBTARGET),legacy)` and one time in `ifeq ($(SUBTARGET),mips74k)`.

With this change it is more obvious that the same images should be build for both subtargets.

But I agree, that this change should go into an extra commit due to the rename of the images. In fact, I dropped this chunk from my local branch which I compile test at the moment.

luaraneda: Exactly, the name of the generated images remains the same, it's just a cosmetic change to identify better the subtarget, and to have a  device defined for every append to TARGET_DEVICES.
I understand that the change should be on a separate commit. Do I add another commit or should I create another pull request?

Without defining the devices, the change would be:
```diff
-TARGET_DEVICES += standard standard-noloader-gz
+TARGET_DEVICES += standard
+TARGET_DEVICES += standard-noloader-gz
```

mkresin:
I merged your PR excluding the brcm47xx changes. Please create a new PR for the brcm47xx changes if you are still interested.

PR #1575, opened 2017-12-12
libjson-c: pull newer released version 0.13

commodo:
Has more features (json pointer - RFC6901, deep copying
of objects, json-equal function, and other stuff).

But it's also bigger in size.
On an x86_64 platform the old one:
```
root@LEDE:# ls -la /usr/lib | grep json
lrwxrwxrwx    1 root     root            18 Dec  7 17:43 libjson-c.so.2 -> libjson-c.so.2.0.2
-rwxr-xr-x    1 root     root         36896 Dec  7 17:43 libjson-c.so.2.0.2
```

new one:
```
root@LEDE:# ls -la /usr/lib | grep json
lrwxrwxrwx    1 root     root            18 Dec  7 17:43 libjson-c.so.3 -> libjson-c.so.3.0.1
-rwxr-xr-x    1 root     root         49222 Dec  7 17:43 libjson-c.so.3.0.1
```

A 13k increase in size.


commodo:
[ open discussion ]
since the newer `libjson-c` has more stuff, it would be interesting to have it, especially if the new stuff will be used ;
but, regarding the size bump, i am also a bit reserved ;
i've been trying to find if OpenWrt modules/libs [libubox, procd, netifd, etc] could benefit from this by using some new functions ; at first glance, nothing stands out

i've been wanting to push to it JSON patch support, which can be pretty cool for using JSON as a communication format by sending only what's changed in a big JSON object ;
JSON patch - https://tools.ietf.org/html/rfc6902
( it seems that if you use JSON for communicating stuff, it can become very verbose, very quickly )
atm, only JSON pointer is implemented - https://tools.ietf.org/html/rfc6901

not sure yet if netjson would benefit from this in the future

package/libs/libjson-c/Makefile
```diff
@@ -8,14 +8,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=json-c
-PKG_VERSION:=0.12.1
+PKG_VERSION:=0.13-20171207
 PKG_RELEASE:=1
 
-PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION)-nodoc.tar.gz
```
wongsyrone: No need to change how to fetch tarball. Please refer to https://github.com/json-c/json-c/wiki#overview

wongsyrone:
I haven't checked the details; we can keep the current configuration if no new features being enabled by default. If someone has further need, he can enable what he wants.

package/libs/libjson-c/Makefile
commodo: ack ; thanks for pointing it out

commodo:
@wongsyrone 
re-spinned 

i did this change last week, but forgot to push it to the PR ; 

so, we could try to see about adding `disable` flags for json-c to disable what isn't needed ;
i'll see about that upstream ; 

in the meantime, i think it's best to close this and come back when i have a good solution for disabling features ;

PR #1579, opened 2017-12-12
ar71xx: TL-WR1043N(D) v2+ add support for VLAN IDs up to 4096

fzipi:
This patch solves this [old problem](https://dev.openwrt.org/ticket/22316) in TPLINK 1043ND v2+.

It permits to assign VLAN ids > 128.

Signed-off-by: Felipe Zipitría <fzipi@fing.edu.uy>

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/ar8216.h
+++ b/target/linux/generic/files/drivers/net/phy/ar8216.h
@@ -29,7 +29,7 @@
 #define AR8316_NUM_VLANS	4096
 
 /* size of the vlan table */
-#define AR8X16_MAX_VLANS	128
+#define AR8X16_MAX_VLANS	4096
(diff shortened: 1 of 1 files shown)


jow-:
The change looks odd to me, iirc this switch has 128 vlan table entries but supports vids up to 4096. By default the vid value defaults to the value of "option vlan". To configure vids great than 128 you have a configuration like this:

    config switch_vlan
      option switch switch0
      option ports "1 5t"
      option vlan 1
      option vid 1
  
    config switch_vlan
      option switch switch0
      option ports "2 5t"
      option vlan 2
      option vid 4095

fzipi:
Ups, let me check that again....

fzipi:
Seems you're completely right! Forget about this....

PR #1582, opened 2017-12-15
ramips: add full support for newifi d2

huming2207:
# Overview

Previously Newifi D2 could only use PandoraBox M1's firmware. It works fine, but LED GPIO is different. As a result, a separated DTS file for this device should be implemented.

Signed-off-by: Jackson Ming Hu <huming2207@gmail.com>

# Installation method

the same as PBR-M1 and/or Newifi D1.

Simply upload the firmware to bootloader's web console (press and hold reset button when powering it up) at 192.168.1.1.

# Hardware spec

* CPU: MTK MT7621A
* RAM: 512MB
* ROM: 32MB SPI Flash
* WiFi: MT7612EN+MT7603EN
* Other features: USB 3.0, UART, Gigabit Ethernet

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -304,6 +304,12 @@ pbr-m1)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
 	ucidef_set_led_default "sys" "sys" "$board:blue:sys" "1"
 	;;
+newifi-d2)
+	set_usb_led "$board:blue:usb"
+	ucidef_set_led_switch "internet" "internet" "$board:amber:internet" "switch0" "0x10"
+	ucidef_set_led_netdev "wlan2g" "WiFi 2.4GHz" "$board:blue:wlan2g" "wlan0"
+	ucidef_set_led_netdev "wlan5g" "WiFi 5GHz" "$board:blue:wlan5g" "wlan1"
+	;;
 psg1208)
 	set_wifi_led "$board:white:wlan2g"
 	;;
--- a/target/linux/ramips/base-files/etc/diag.sh
+++ b/target/linux/ramips/base-files/etc/diag.sh
@@ -124,6 +124,7 @@ get_status_led() {
 	d240|\
 	dap-1350|\
 	na930|\
+	newifi-d2|\
 	pbr-m1|\
 	re350-v1|\
 	rt-ac51u|\
--- a/target/linux/ramips/dts/Newifi-D2.dts
+++ b/target/linux/ramips/dts/Newifi-D2.dts
@@ -1,0 +1,165 @@
+/dts-v1/;
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "pandorabox,newifi-d2", "mediatek,mt7621-soc";
+	model = "Newifi-D2";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x10000000>;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power-amber {
+			label = "newifi-d2:amber:power";
+			gpios = <&gpio0 6 GPIO_ACTIVE_LOW>;
+		};
+
+		power-blue {
+			label = "newifi-d2:blue:power";
+			gpios = <&gpio0 15 GPIO_ACTIVE_LOW>;
+		};
+		
+		internet-amber {
+			label = "newifi-d2:amber:internet";
+			gpios = <&gpio0 4 GPIO_ACTIVE_LOW>;
+		};
+
+		internet-blue {
+			label = "newifi-d2:blue:internet";
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+		};
+
+		wlan2g {
+			label = "newifi-d2:blue:wlan2g";
+			gpios = <&gpio0 14 GPIO_ACTIVE_LOW>;
+		};
+
+		wlan5g {
+			label = "newifi-d2:blue:wlan5g";
+			gpios = <&gpio0 16 GPIO_ACTIVE_LOW>;
+		};
+
+		usb {
+			label = "newifi-d2:blue:usb";
+			gpios = <&gpio0 10 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio0 3 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+		
+		wps {
+			label = "wps";
+			gpios = <&gpio0 7 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+
+	gpio_export {
+		compatible = "gpio-export";
+		#size-cells = <0>;
+
+		power_usb3 {
+			gpio-export,name = "power_usb3";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 11 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&sdhci {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&sdhci_pins>;
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,chunked-io = <32>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x30000>;
+			read-only;
+		};
+
+		partition@30000 {
+			label = "u-boot-env";
+			reg = <0x30000 0x10000>;
+			read-only;
+		};
+
+		factory: partition@40000 {
+			label = "factory";
+			reg = <0x40000 0x10000>;
+			read-only;
+		};
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+
+	pcie0 {
+		mt76@0,0 {
+			reg = <0x0000 0 0 0 0>;
+			device_type = "pci";
+			mediatek,mtd-eeprom = <&factory 0x8000>;
+			ieee80211-freq-limit = <5000000 6000000>;
+		};
+	};
+
+	pcie1 {
+		mt76@1,0 {
+			reg = <0x0000 0 0 0 0>;
+			device_type = "pci";
+			mediatek,mtd-eeprom = <&factory 0x0000>;
+		};
+	};
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xe000>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "wdt", "rgmii2", "jtag", "uart3";
+			ralink,function = "gpio";
+		};
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -142,6 +142,16 @@ define Device/newifi-d1
 endef
 TARGET_DEVICES += newifi-d1
 
+define Device/newifi-d2
+  DTS := Newifi-D2
+  IMAGE_SIZE := $(ralink_default_fw_size_32M)
+  DEVICE_TITLE := Newifi D2
+  DEVICE_PACKAGES := \
+	kmod-mt7603 kmod-mt76x2 kmod-sdhci-mt7620 \
+	kmod-usb3 kmod-usb-ledtrig-usbport
+endef
+TARGET_DEVICES += newifi-d2
+
 define Device/pbr-m1
   DTS := PBR-M1
   IMAGE_SIZE := $(ralink_default_fw_size_16M)
--- a/target/linux/ramips/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/base-files/etc/board.d/02_network
@@ -93,6 +93,7 @@ ramips_setup_interfaces()
 	mt7628|\
 	mzk-750dhp|\
 	mzk-w300nh2|\
+	newifi-d2|\
 	nixcore-x1-8M|\
 	nixcore-x1-16M|\
 	oy-0001|\
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -376,6 +376,9 @@ ramips_board_detect() {
 	*"Newifi-D1")
 		name="newifi-d1"
 		;;
+	*"Newifi-D2")
+		name="newifi-d2"
+		;;
 	*"NCS601W")
 		name="ncs601w"
 		;;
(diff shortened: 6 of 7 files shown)


target/linux/ramips/image/mt7621.mk
```diff
@@ -142,6 +142,16 @@ define Device/newifi-d1
 endef
 TARGET_DEVICES += newifi-d1
 
+define Device/newifi-d2
+  DTS := Newifi-D2
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
+  DEVICE_TITLE := Newifi D2
+  DEVICE_PACKAGES := \
+	kmod-mt7603 kmod-mt76x2 kmod-sdhci-mt7620 \
+	kmod-usb3 kmod-usb-ledtrig-usbport wpad-mini
```
mkresin: drop wpad-mini. it is included by default.

drop kmod-usb-ledtrig-usbport. I neither see an usb led in the DTS nor a set_usb_led() call in 01_leds

target/linux/ramips/image/mt7621.mk
```diff
@@ -142,6 +142,16 @@ define Device/newifi-d1
 endef
 TARGET_DEVICES += newifi-d1
 
+define Device/newifi-d2
+  DTS := Newifi-D2
+  IMAGE_SIZE := $(ralink_default_fw_size_16M)
```
mkresin: ralink_default_fw_size_16M is wrong here. ralink_default_fw_size_16M is 16121856 bytes (0xf60000). But your firmware partition is 0xfb0000 bytes.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -304,6 +304,13 @@ pbr-m1)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
 	ucidef_set_led_default "sys" "sys" "$board:blue:sys" "1"
 	;;
+newifi-d2)
+	ucidef_set_led_netdev "internet" "internet" "$board:blue:internet" "eth0.2"
+	ucidef_set_led_default "power" "power" "$board:amber:power" "1"
```
mkresin: Drop the power led here and fix the diag.sh support for the LED.

target/linux/ramips/base-files/etc/diag.sh
```diff
@@ -124,6 +124,7 @@ get_status_led() {
 	d240|\
 	dap-1350|\
 	na930|\
+	newifi-d2|\
```
mkresin: It is for sure at the wrong location. the board doesn't have a newifi-d2:blue:power led.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		power {
+			label = "newifi-d2:amber:power";
+			gpios = <&gpio0 6 GPIO_ACTIVE_LOW>;
+		};
+
+		sys {
```
mkresin: Could it be that it is the usb led? If so, rename the LED and use set_usb_led() in 01_leds.

target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -304,6 +304,13 @@ pbr-m1)
 	ucidef_set_led_default "power" "power" "$board:blue:power" "1"
 	ucidef_set_led_default "sys" "sys" "$board:blue:sys" "1"
 	;;
+newifi-d2)
+	ucidef_set_led_netdev "internet" "internet" "$board:blue:internet" "eth0.2"
```
mkresin: please use the following instead:

```
ucidef_set_led_switch "wan" "wan" "$board:blue:internet" "switch0" "0x10"
```

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+};
+
+&ethernet {
+	mtd-mac-address = <&factory 0xe000>;
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "wdt", "rgmii2", "jtag", "mdio";
```
mkresin: no pin of the mdio group is used as gpio. drop it.

please add the uart3 group here. &gpio0 6 (newifi-d2:amber:power) is pin of the uart3 group

mkresin:
Please add the hardware spec as well as the Installation method information to the commit message. The PR description is like an envelope and is not merged into the LEDE repository.

Please be a bit more verbose in your commit message. number of button + which, number of LEDs +which and number of lan ports should be added.

target/linux/ramips/dts/Newifi-D2.dts
981213: No. The power led and the following sys led are two LEDs with different color but at the same place. There is another blue USB led using GPIO10.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio0 18 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
```
981213: The reset button is GPIO3 and there is another button using GPIO7.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+	};
+
+	gpio_export {
+		compatible = "gpio-export";
+		#size-cells = <0>;
+
+		power_usb2 {
+			gpio-export,name = "power_usb2";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 22 GPIO_ACTIVE_HIGH>;
```
981213: There is no USB2.0 on this router.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+		};
+
+		power_usb3 {
+			gpio-export,name = "power_usb3";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 25 GPIO_ACTIVE_HIGH>;
```
981213: And this one is using GPIO11

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,155 @@
...
+		sys {
+			label = "newifi-d2:blue:sys";
+			gpios = <&gpio0 15 GPIO_ACTIVE_LOW>;
+		};
+
+		internet {
+			label = "newifi-d2:blue:internet";
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+		};
+
```
981213: Missing the amber internet led on GPIO4

target/linux/ramips/dts/Newifi-D2.dts
mkresin: Please add the missing usb led to the dts and add the usb led trigger using set_usb_led().

mkresin: Why do you enable the sys LED by default? (`ucidef_set_led_default "sys" "sys" "$board:blue:sys" "1"`)

huming2207:
Hi guys, so sorry for the mistakes in the PR. I had my lunch that day after I added initial support in the script and I forgot continuing to correct those stuff after I came back. 

Now the compilation has passed but not yet tested on the real device. I'll do it tomorrow morning and I'll notice you soon after I confirm it really works.

Also special thanks to @981213 for his hardware details provided to me today. 😂

huming2207:
Hi @mkresin 

I've tested on a real Newifi D2 device and I can confirm it works fine now. You can check again and accept it into the official repo.

Regards,
Jackson

mkresin:
@981213 fine with the commit now?

The IMAGE_SIZE is still wrong. Your firmware partition has a size of 16064 Kbyte, which is far away from ralink_default_fw_size_32M=33226752.

If it is the only issue left, I will fix it during merging your commit.

huming2207:
Hi @mkresin, sorry for that I forgot to correct that value. I'll fix that if you want.

981213:
Actually the wrong thing is the mtd partition in the dts file. Newifi D2 uses W25Q256 and the actual size of the firmware partition should be 0x1fb0000 = 33226752 bytes.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,165 @@
...
+		factory: partition@40000 {
+			label = "factory";
+			reg = <0x40000 0x10000>;
+			read-only;
+		};
+
+		partition@50000 {
+			label = "firmware";
+			reg = <0x50000 0xfb0000>;
```
981213: Should be <0x50000 0x1fb0000>

mkresin: Excellent catch! And 0x1fb0000 == 33226752 == ralink_default_fw_size_32M

mkresin: I fixed it in my staging tree. No need to update the PR.

mkresin:
I merged your PR to [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary). I rebased your PR on top of my pending commits.

Would you please do a full testcycle with an image compiled from [my staging tree](https://git.lede-project.org/?p=lede/mkresin/staging.git;a=summary) and report back whether I broke something during rebasing. I'll not push the changed commit to master without further testing.

huming2207:
Hi @mkresin, I've left my office and I can't test with my device now. I'll do it tomorrow. 

I've read your changes and looks all good. But I think you should name it as `newifi-d2`, not `pandorabox-newifi-d2`, as it stays constant as Newifi D1.

mkresin:
> But I think you should name it as newifi-d2, not pandorabox-newifi-d2, as it stays constant as Newifi D1.

It is the new way of doing things. The Newifi D1 will be renamed at a later point as well.

981213:
@mkresin I'm not sure if the naming scheme is <manufacturer>-<device name>. If so, "PandoraBox" is not the name of the manufacturer, it's a firmware made by the manufacturer actually. Their official English name is "D-Team".
PS: PBR-M1 is the only one they use "PandoraBox" as a brand name. Their other products are using "Newifi" as their trademark (including Y1 and Y1S which used to be OEM products for Lenovo).

981213:
@mkresin Sorry I missed something else in the dts: There is no SD card slot on this device so the sdhci part in the dts and kmod-sdhci-mt7620 in image/mt7621.mk should be removed.

target/linux/ramips/dts/Newifi-D2.dts
```diff
@@ -0,0 +1,165 @@
...
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "pandorabox,newifi-d2", "mediatek,mt7621-soc";
+	model = "Newifi-D2";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x10000000>;
```
MiltonLai: should be <0x0 0x1c000000>, <0x20000000 0x4000000>
otherwise it won't recognize all 512M ram

huming2207:
Hi @mkresin, I think you've forgot to change the values for LEDs and ethernet switch settings and the both of these two functions are broken now.

By the way the RAM and Flash settings seem correct:

```
root@OpenWrt:/# df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/root                11.5M     11.5M         0 100% /rom
tmpfs                   250.9M    228.0K    250.7M   0% /tmp
tmpfs                   250.9M    100.0K    250.8M   0% /tmp/root
overlayfs:/tmp/root     250.9M    100.0K    250.8M   0% /
tmpfs                   512.0K         0    512.0K   0% /dev
root@OpenWrt:/# cat /proc/mtd
dev:    size   erasesize  name
mtd0: 00030000 00010000 "u-boot"
mtd1: 00010000 00010000 "u-boot-env"
mtd2: 00010000 00010000 "factory"
mtd3: 01fb0000 00010000 "firmware"
mtd4: 00194e5f 00010000 "kernel"
mtd5: 01e1b1a1 00010000 "rootfs"
mtd6: 01290000 00010000 "rootfs_data"
```

```
root@OpenWrt:/# free 
             total       used       free     shared    buffers     cached
Mem:        513848      46000     467848        228       4932      12940
-/+ buffers/cache:      28128     485720
Swap:            0          0          0
```

huming2207:
Hi @mkresin , just discussed with @981213 and here are some more details:

```
root@OpenWrt:/# ls /tmp/sysinfo/
board_name  model
root@OpenWrt:/# cat /tmp/sysinfo/board_name 
newifi-d2
root@OpenWrt:/# cat /tmp/sysinfo/model
Newifi-D2
root@OpenWrt:/# cat /etc/config/system 

config system
        option hostname 'OpenWrt'
        option timezone 'UTC'
        option ttylogin '0'
        option log_size '64'
        option urandom_seed '0'

config timeserver 'ntp'
        option enabled '1'
        option enable_server '0'
        list server '0.openwrt.pool.ntp.org'
        list server '1.openwrt.pool.ntp.org'
        list server '2.openwrt.pool.ntp.org'
        list server '3.openwrt.pool.ntp.org'

```

Also D-Team is the OEM factory of this router, not its real trademark. It should be "Newifi" I think.

MiltonLai:
@mkresin Sure I will test it later today. 
@huming2207 <0x0 0x10000000> indicates a size of 16^7 = 268,435,456 = 256MB, while according to the datasheet of mt7621, the memory map is
> 
> Start End Size Description 
> 0 1BFFFFFF 448M DRAM Direct Map
> 1C000000 1DFFFFFF 32M <<Reserved>> 
> ...
> 1FC00000 1FFFFFFF 4M ROM/SPI FLASH Direct Access  
> 20000000 23FFFFFF 64M DRAM Re-Map 
> 24000000 5FFFFFFF 960M <<Reserved>>
>

981213:
@huming2207 No. You just forgot to apply the generic board detection part in @mkresin 's staging tree.

981213:
And the naming scheme should be [manufacture]_[board] and "newifi" is just their trademark.
PS: The OEM factory of this router is Shen Zhen MTN Electronics Co.,Ltd.

mkresin:
>Hi @mkresin , just discussed with @981213 and here are some more details:
>
>```
>root@OpenWrt:/# cat /tmp/sysinfo/board_name 
>newifi-d2
>```

This look indeed wrong. Please make sure to clone my complete tree and do not cherry pick single commits.

> Also D-Team is the OEM factory of this router, not its real trademark. It should be "Newifi" I think.

As pointed out by @981213 the compatible string is `"<manufacture>,<board>"`. Within my staging branch the boardname is derived from the device tree compatible string.

As long as it isn't some OEM/ODM board (as produced by alpha-networks, senao, arcadyan ...) the manufacture should be the company which sells the board under his name.

MiltonLai:
@mkresin I have built and run the new image, everything looks good

huming2207:
Sorry @mkresin  I was using my own repo and cherry-picked your commit in it before and it was broken. Everything are fine after cloning and compiling your staging repo directly. So sorry about that.

```
root@OpenWrt:/# df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/root                 5.5M      5.5M         0 100% /rom
tmpfs                   250.9M    240.0K    250.7M   0% /tmp
tmpfs                   250.9M     80.0K    250.8M   0% /tmp/root
tmpfs                   512.0K         0    512.0K   0% /dev
/dev/mtdblock6           24.8M    728.0K     24.0M   3% /overlay
overlayfs:/overlay       24.8M    728.0K     24.0M   3% /
root@OpenWrt:/# free -m
             total       used       free     shared    buffers     cached
Mem:        513848      45320     468528        320       4644      13980
-/+ buffers/cache:      26696     487152
Swap:            0          0          0
root@OpenWrt:/# 

```

huming2207:
And here is the image I've compiled with @mkresin 's staging repo. If someone has this device, you can have a try if you wish.

[openwrt-ramips-mt7621-d-team_newifi-d2-squashfs-sysupgrade.zip](https://github.com/lede-project/source/files/1574600/openwrt-ramips-mt7621-d-team_newifi-d2-squashfs-sysupgrade.zip)

PR #1584, opened 2017-12-16
ar71xx: fix TP-Link Archer C7 v4 switch LEDs

blocktrron:
This fixes wrong switch LEDs on TP-Link Archer C7 v4.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/board.d/01_leds
+++ b/target/linux/ar71xx/base-files/etc/board.d/01_leds
@@ -58,8 +58,7 @@ sc1750|\
 sc450)
 	ucidef_set_led_wlan "wlan" "WLAN" "$board:green:wlan" "phy0tpt"
 	;;
-archer-c25-v1|\
-archer-c7-v4)
+archer-c25-v1)
 	ucidef_set_led_netdev "wan" "WAN" "$board:green:wan" "eth0"
 	ucidef_set_led_wlan "wlan" "WLAN" "$board:green:wlan2g" "phy1tpt"
 	ucidef_set_led_wlan "wlan5g" "WLAN5G" "$board:green:wlan5g" "phy0tpt"
@@ -67,12 +66,6 @@ archer-c7-v4)
 	ucidef_set_led_switch "lan2" "LAN2" "$board:green:lan2" "switch0" "0x08"
 	ucidef_set_led_switch "lan3" "LAN3" "$board:green:lan3" "switch0" "0x04"
 	ucidef_set_led_switch "lan4" "LAN4" "$board:green:lan4" "switch0" "0x02"
-	case "$board" in
-		archer-c7-v4)
-			ucidef_set_led_usbdev "usb1" "USB1" "$board:green:usb1" "1-1"
-			ucidef_set_led_usbdev "usb2" "USB2" "$board:green:usb2" "2-1"
-		;;
-	esac
 	;;
 archer-c58-v1|\
 archer-c59-v1|\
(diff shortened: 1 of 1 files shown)


tisj:
Patch is valid, does the same as [#1589](https://github.com/lede-project/source/pull/1589). However you seem to invert the led order aswell. Althoug this seems more correct, as "led1" now refers to port number one at the back side, this may be less intuitive (switching cable from right to left on the back side, changes the led from left to right at the front side) and inconsistent with other models, wdr3600, wdr4300, archer c7 v2, ...

mkresin:
I had a look at the archer-c7-v4 GPL code and the with the patch applied the LED GPIOs are matching the ones from the GPL code.

To my knowledge all TP-Link routers in LEDE have the most left LAN LED assigned to lan port 1. Which LED is used by the TP-Link firmware for LAN1?

blocktrron:
@mkresin I'm still having troubles restoring my C7 v4 to stock firmware. I can't speak for stock firmware of the v4, but the v2, LAN Port 1 matches the LAN LED on the far right. I don't think they have changed the LED order on the v4, given that the switch numbering stays the same.

The LAN Port order on the C7v2 and v4 is different from (to my knowledge) every other TP-Link router, so does the LED Order. I guess the inverting the LED order is the right approach to match the behavior of the original firmware.

mkresin:
> The LAN Port order on the C7v2 and v4 is different from (to my knowledge) every other TP-Link router, so does the LED Order. 

Add this explanation to the commit message. The change looks indeed strange. Everything mentioned in the commit message will not cause questions ;-).

>  I guess the inverting the LED order is the right approach to match the behavior of the original firmware.

To be honest, I don't really care as I most likely will never use that board. But the approach to match the tplink firmware behaviour sounds reasonable to me.

tisj:
I restored the original firmware on my unit, and behaves as originally implemented: the back-side left port (numbered 4) lids up the left led on the front panel (corresponds to LED1). Repatching the cable one port right (numbered 3), lids up the second led (one right) on the front panel (corresponds to LED2). And so fort... I can make a video if you wish ;-)

mkresin:
> I restored the original firmware on my unit, and behaves as originally implemented

I've to admit I got lost. Does it mean the PR is fine to merge? Or does "originally" means the current LEDE code is fine?

Would you please check if `echo 1 > /sys/class/leds/archer-c25-v1:green:lan1` lids up the LAN1 led (most left) in LEDE with the patch applied.

Would you mind to have a look at the wireless mac as well to provide an answer to https://github.com/lede-project/source/pull/1593#issuecomment-353218426.

blocktrron:
@tisj Could you explain how / wit thich firmware you did restore it? My device always fails the tftp process 😢

tisj:
> @tisj Could you explain how / with which firmware you did restore it? My device always fails the tftp process 

$ ls -l /tftpboot/ArcherC7v4_tp_recovery.bin 
lrwxrwxrwx 1 root root 27 Dec 21 15:46 /tftpboot/ArcherC7v4_tp_recovery.bin -> 'Archer C7(EU)_V4_170901.bin'

Downloaded from the tplink UK website as a zip 
$ ls -l Archer\ C7\(EU\)_V4_170901.zip 
-rw-r--r-- 1 15226615 Dec 21 15:45 'Archer C7(EU)_V4_170901.zip'

@mkresin: i will build a firmware again

tisj:
I plug in the LAN cable the left-most port (backside numbered 4), as expected the first lan led lights up, however to the system this seems to be LED4 now, as shown below. 

root@OpenWrt:~# cat /etc/banner
  _______                     ________        __
 |       |.-----.-----.-----.|  |  |  |.----.|  |_
 |   -   ||  _  |  -__|     ||  |  |  ||   _||   _|
 |_______||   __|_____|__|__||________||__|  |____|
          |__| W I R E L E S S   F R E E D O M
 -----------------------------------------------------
 OpenWrt SNAPSHOT, r5537-746d0aa
 -----------------------------------------------------
root@OpenWrt:~# grep . /sys/class/leds/*lan*/brightness
/sys/class/leds/archer-c7-v4:green:lan1/brightness:0
/sys/class/leds/archer-c7-v4:green:lan2/brightness:0
/sys/class/leds/archer-c7-v4:green:lan3/brightness:0
/sys/class/leds/archer-c7-v4:green:lan4/brightness:255

When echo 255 > /sys/class/leds/archer-c7-v4:green:lan1, the right-most LED lids up.

blocktrron:
So if i plug a Cable in the most left port, the most left LED lights up correctly. The port is labeled as LAN 4. 

So from my understanding, we should name this LED LAN4 (like it is currently in my patch)? I think numbering the LEDs after the corresponding LAN port instead of "left to right" is the most natural approach.

It feels kinda complicated to talk and discuss this topic 😸

mkresin:
> I think numbering the LEDs after the corresponding LAN port instead of "left to right" is the most natural approach.

Please consider LED names and LAN port triggers as two independent topic. A user can use the LED for what ever (s)he like. Hence, the LEDs names should either match the labels on the the case and in case of absence of labels they should follow some logical naming order (left to right).

The actual mapping of LAN ports to LEDs is done via `ucidef_set_led_switch` and should match the tplink firmware behaviour.

mkresin:
@tisj is this one fine to merge? I kind of lost the overview what is the correct LED order/which LED should be triggered by a particular switch port.

blocktrron:
@mkresin Switch port status is displayed accurately and the numbering of the LEDs is now from left to right like every other TP-Link device.

mkresin:
>  Switch port status is displayed accurately and the numbering of the LEDs is now from left to right like every other TP-Link device.

Ah , I see. You dropped the LED rename completely. Just to be sure, the tp-link firmware has the same port <> led mapping?

It would be really helpful if the commit message gives a hint what the stock behaviour is and what the LEDE behaviour is. "This fixes wrong switch LEDs on TP-Link Archer C7 v4" can be all or nothing.

Something like:

> Match the network port mapping to the TP-Link firmware behaviour. LAN port 1 triggers the most left LED, LAN port 2 the second from the left and so on.

would describe what the desired behaviour is.

blocktrron:
@mkresin The current PR matches the TP-Link firmwares behavior. I will adjust the commit message towards your proposal.

mkresin:
@blocktrron did you forget to push you adjusted commit message?

blocktrron:
@mkresin indeed i forgot it, sorry.  Adjusted the commit message, should be fine now (?)

hmh:
@blocktrron:
Shouldn't this PR be "resubmitted" against openwrt/openwrt (https://github.com/openwrt/openwrt), due to the LEDE/OpenWRT merge/unfork ?

mkresin:
Thanks! Merged into my [staging tree](https://git.openwrt.org/?p=openwrt/staging/mkresin.git;a=summary).

PR #1586, opened 2017-12-16
version.mk: don't use comma in version sed scripts

pprindeville:
Especially not if comma is a potential delimiter for the
value being replaced, like "Redfish Solutions, LLC" for the
manufacturer (or URLs, where commas are common).

Fixes FS#1151.

### DIFF
--- a/include/version.mk
+++ b/include/version.mk
@@ -94,10 +94,10 @@ VERSION_SED:=$(SED) 's,%U,$(VERSION_REPO),g' \
 	-e 's,%S,$(BOARD)/$(if $(SUBTARGET),$(SUBTARGET),generic),g' \
 	-e 's,%A,$(ARCH_PACKAGES),g' \
 	-e 's,%t,$(VERSION_TAINTS),g' \
-	-e 's,%M,$(VERSION_MANUFACTURER),g' \
-	-e 's,%m,$(VERSION_MANUFACTURER_URL),g' \
-	-e 's,%b,$(VERSION_BUG_URL),g' \
-	-e 's,%s,$(VERSION_SUPPORT_URL),g' \
+	-e 's\#%M\#$(VERSION_MANUFACTURER)\#g' \
+	-e 's\#%m\#$(VERSION_MANUFACTURER_URL)\#g' \
+	-e 's\#%b\#$(VERSION_BUG_URL)\#g' \
+	-e 's\#%s\#$(VERSION_SUPPORT_URL)\#g' \
 	-e 's,%P,$(VERSION_PRODUCT),g' \
 	-e 's,%h,$(VERSION_HWREV),g'
 


pprindeville:
Can I please get a review from @wigyori or @jow-?  Thanks.

NeoRaider:
As VERSION_MANUFACTURER etc. are user-provided strings, we should probably do some escaping on them anyways. At least `\`, `&` and the sed delimiter need to be escaped for sed, and we might also want to replace occurrences of `'` in the input string with `'\''` for the shell. I can especially see `&` being used in company names, but if we start fixing this, we might as well do it properly.

pprindeville:
@NeoRaider Okay, what macros exist for handling escaping in the Makefiles?

NeoRaider:
Hmm, not much I guess. I can see one use of the *subst* Make function to replace `'` with `'\''` in *rules.mk* (in the definition of *locked*).

It would be useful to introduce some generic definitions for sed escaping. I'd add them to *version.mk* for now - if we find other places where they are useful, they can be moved to *rules.mk* or something like that later.

`#` is probably not the best choice for a delimiter, as it needs escaping for Make. If we're going to escape the input string anyways, using a character that has no special meaning for Make, sh or sed would be a good choice to keep the subst expressions as simple as possible, so I'd just keep the current `,`.

pprindeville:
> `#` is probably not the best choice for a delimiter, as it needs escaping for Make. If we're going to escape the input string anyways, using a character that has no special meaning for Make, sh or sed would be a good choice to keep the subst expressions as simple as possible, so I'd just keep the current `,`.

Actually, it's because `#` is an awkward character to use for user-defined values that I picked it... because as-is it already calls out special handling.

What about the `qstrip_escape` macro in `include/version.mk`?

KanjiMonster:
Instead of trying to find a delimiter that is uncommon, how about extending the qstrip_escape function to escape the delimiter in the strings so it cant appear anymore? e.g. something like
```
comma:=,
qstrip_escape=$$(subst $$(comma),\$$(comma),$$(subst ','\'',$(call qstrip,$(1))))
```

seemed to work fine in my limited tests. For consistency sake we should probably also escape ```\``` as well.

pprindeville:
If you have:

`CONFIG_VERSION_MANUFACTURER="Sanford & Sons, Co."`

then what does `VERSION_MANUFACTURER` end up being?

NeoRaider:
Ah, I overlooked *qstrip_escape*. We should indeed extend that to handle the other characters I listed: `&`, `,` and `\` (I think that's all that matter in the substitution value?) (if we keep `,` as delimiter)

Currently, `&` would be replaced with the original pattern `%M`.

pprindeville:
@NeoRaider Do you want to take this one over?

pprindeville:
Moving to https://github.com/openwrt/openwrt/pull/690.

PR #1587, opened 2017-12-16
kernel: bump 4.4 to 4.4.106 for 17.01

bladeoner:
kernel: bump 4.4 to 4.4.106 for 17.01

Refreshed all patches.

Made the following patch for Mediatek and Oxnas compatible with kernel 4.4.106:
0072-mtd-backport-v4.7-0day-patches-from-Boris.patch

Compile-tested: ar71xx
Run-tested: ar71xx

Signed-off-by: Etienne Haarsma bladeoner112@gmail.com


bladeoner:
The fix for the following patch is not ok:
0072-mtd-backport-v4.7-0day-patches-from-Boris.patch

PR #1588, opened 2017-12-18
mwlwifi: update to version 10.3.4.0 / 2017-12-14

anomeome:
Lastest stamped version: compile / test target mvebu / rango
 - WDS AP on 88W8964
 - correct region code


lucize:
for me is stable since the release date, on rango

hauke:
Thank you for the patch, it was applied to master

hnyman:
If there is going to be 17.01.5 service release (like indicated today in https://forum.lede-project.org/t/security-meltdown-and-spectre-vulnerabilities-in-arm/10283/10 ), it would be great to see a mwlwifi driver backport in that release, as the 17.01 still has the really ancient 2017-06-06 driver.

PR #1590, opened 2017-12-18
kernel: uas

Hallo32:
Add the uas module to the modules loaded on boot.
This is necessary to be able to use uas storage for extroot.


blogic:
please add a prefix to the patch subject (in this case "kernel: ") and make the description more verbose explaining why this is needed.

blogic:
merged after fixing up the subject and description a bit. Thanks !

PR #1591, opened 2017-12-19
imagebuilder: show package list before image building

feckert:
In some situation it is useful to get a complete list of package which get installed into the image **before** image building. It is often unclear if you add package to the image how many package will get also installed into to image due package dependency.

### DIFF
--- a/target/imagebuilder/files/Makefile
+++ b/target/imagebuilder/files/Makefile
@@ -33,6 +33,7 @@ Available Commands:
 	info:	Show a list of available target profiles
 	clean:	Remove images and temporary build files
 	image:	Build an image (see below for more information).
+	list:	List image packages (see below more information).
 
 Building images:
 	By default 'make image' will create an image with the default
@@ -44,6 +45,13 @@ Building images:
 	make image FILES="<path>" # include extra files from <path>
 	make image BIN_DIR="<path>" # alternative output directory for the images
 	make image EXTRA_IMAGE_NAME="<string>" # Add this to the output image filename (sanitized)
+
+List packages:
+	List "all" packages which get installed into the image.
+	You can use the following parameters:
+
+	make list PROFILE="<profilename>" # override the default target profile
+	make list PACKAGES="<pkg1> [<pkg2> [<pkg3> ...]]" # include extra packages
 endef
 $(eval $(call shexport,Helptext))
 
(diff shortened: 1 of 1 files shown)


aparcar:
I couldn't test this yet. Does it show the to be installed package version as well?

aparcar:
@dangowrt could you please check this patch? This would be an interesting feature for the upgrade server to estimate the manifest hash before creation. 
Maybe an option to add the manifest hash as extra_image_name could be created as well.

feckert:
@aparcar thanks for reviewing my PR
yes it will show the versions as well

aparcar:
@feckert I had the time to check the patch and really like it. My use case is that I'm producing many images with variations of installed packaged. Ideally the created firmware name contains a hash of the created manifest. 
I'd think of the following: 

1. list (get packages + versions)
2. sha256sums
3.  EXTRA_IMAGE_NAME=<manifest_hash>
4. image (actually create image)

As I'm bad with Makefiles, do you think this is a possible and easy to implement without opkg reloading packages multiple times?

target/imagebuilder/files/Makefile
```diff
@@ -206,5 +221,18 @@ endif
 		$(if $(PACKAGES),USER_PACKAGES="$(PACKAGES)") \
 		$(if $(BIN_DIR),BIN_DIR="$(BIN_DIR)"))
 
-.SILENT: help info image
+list:
+ifneq ($(PROFILE),)
```
aparcar: Instead of creating this twice there could be a function called `_check_profile` used both by `image` and `list`

feckert:
@aparcar i will open this PR on the new remerge location of openwrt.
Question for clarify. The manifest file is the file that will get generated during the whole build process in the first stage?
And you want to use this file and preserve only the information for the packages which are build into the image during image generation with the imagebuilder?

aparcar:
@feckert hey I started developing on another project again and require this patch, could you please reopen it on the new project page? If not can do it as well

target/imagebuilder/files/Makefile
```diff
@@ -206,5 +221,18 @@ endif
 		$(if $(PACKAGES),USER_PACKAGES="$(PACKAGES)") \
 		$(if $(BIN_DIR),BIN_DIR="$(BIN_DIR)"))
 
-.SILENT: help info image
+list:
```
aparcar: I think this line should be `list: FORCE`

aparcar:
I created an PR https://github.com/openwrt/openwrt/pull/1198 - hope that's okay

feckert:
@aparcar no problem
Thanks to take over this feature

PR #1592, opened 2017-12-20
ar71xx: fix MAC of 5GHz WiFi on Archer C58/C59/C60

blocktrron:
This assigns the 5GHz interface a MAC-address based on the devices primary MAC. 

Previously the MAC-address was not connected with the primary MAC.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -145,6 +145,7 @@ case "$FIRMWARE" in
 		ath10kcal_extract "art" 20480 12064
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		ath10kcal_patch_mac $(macaddr_add $(cat /sys/class/net/eth1/address) -1)
 		;;
 	esac
 	;;


mkresin:
Have you confirmed that the c58 + c59 + c60 TP-Link firmware uses (eth1 mac address - 2) for the 5GHz wireless interface?

heinzek:
Hotplug script: source/target/linux/ar71xx/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
set proper MAC address to 5GHz wlan interface on these  devices.
Did You check this script?

blocktrron:
@mkresin Thanks for making me double check this, indeed it is not (eth1 mac address - 2) but (eth1 mac address - 1). Will double-check on the c7 one too!

@heinzek I weren't aware about this files existence - thanks. It also seems to do the trick on my C59 but not on my C58. This seems very odd to me, considering both are based on the same board?

mkresin:
@heinzek I neither wasn't aware that there is `etc/hotplug.d/ieee80211/10_fix_wifi_mac` beside `hotplug.d/firmware/11-ath10k-caldata`.

As far as I can see, the approach was chosen to set the mac address of the 2.4GHz and 5GHz wireless with the same script.  I would prefer to set the 2.4GHz wireless mac in the mach file (as it is done for most ar71xx boards) and the 5GHz wireless in  `hotplug.d/firmware/11-ath10k-caldata`.

Having a third place where the wireless macs are set is kind of confusing.

btw, I like the approach of using `mtd_get_mac_binary` way more than relying on the eth0/eth1 mac address. This way a possible wrong eth0/eth1 mac can be fixed without causing a wrong wireless mac.

@blocktrron what kind of wrong mac address do you have? Which mac addresses are used by the tplink firmware and which are used by LEDE? This information should go into the commit message as well.

blocktrron:
I flashed the current snapshot again on a C58 and C59 and the MAC-Address was set correctly both times. So it seems @heinzek scripts set it correctly.

PR #1593, opened 2017-12-20
ar71xx: fix Archer C7 5GHz MAC-address

blocktrron:
Currently, LEDE assumes the primary MAC for the Archer C7v4 on eth1 while eth1 is not present on the Archer C7v4. The primary MAC rests on eth0.

### DIFF
--- a/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ar71xx/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -54,11 +54,14 @@ case "$FIRMWARE" in
 		ath10kcal_extract "ART" 20480 2116
 		ath10kcal_patch_mac $(macaddr_add $(cat /sys/class/net/eth0/address) +16)
 		;;
-	archer-c7-v4|\
 	archer-c25-v1|\
 	tl-wdr6500-v2)
 		ath10kcal_extract "art" 20480 2116
 		ath10kcal_patch_mac $(macaddr_add $(cat /sys/class/net/eth1/address) -2)
+		;;
+	archer-c7-v4)
+		ath10kcal_extract "art" 20480 2116
+		ath10kcal_patch_mac $(macaddr_add $(cat /sys/class/net/eth0/address) -1)
 		;;
 	cf-e355ac|\
 	cf-e380ac-v1|\


mkresin:
Please add a commit message. Right now the commit message is empty.

Have you confirmed that the TP-Link firmware uses (eth0 mac address - 2) for the 5GHz wireless interface?

mkresin:
Same as with your other mac addresses commit. Please add some informations about the mac address used by the tplink firmware, the wrong mac address set by LEDE as well as the mac address used with your fix applied.

tisj:
Currently the mac address seems to be unchanged and set to 5C:78:2D:32:5C:78 incorrectly

root@OpenWrt:~# ifconfig wlan0
wlan0     Link encap:Ethernet  HWaddr 5C:78:2D:32:5C:78  

After the patch it is configured to the eth0 mac address minus 2

However the stock tp-link firmware seems to configure the mac adress minus 1

tisj:
Strange to see that the MAC address for eth0 is shared with wlan 2G, even so on the stock tp-link firmware. I never saw this on any other device before...

heinzek:
Most of TP-Link devices have WLAN 2,4 MAC = LAN MAC = MAC on sticker. WAN mac is +-1, WLAN 5GHz is +-2.

blocktrron:
@tisj The behaviour you described is also the one of the Archer C58/C59. I guess it's TP-Links new standard (?)

I will update the PR to set eth0-1 to match the stock firmwares behaviour.

howl:
I reported a bug about Archer C7 v2 https://bugs.lede-project.org/index.php?do=details&task_id=408&string=archer+c7 but it seems to be different in that version.

blocktrron:
Updated my PR to match the behavior described by @tisj

howl:
I have get all the mac addresses in the Archer C7 v2 and the are as follow:
eth0 (wan): ##:##:##:##:##:FE
eth1 (lan): ##:##:##:##:##:FD
wlan  2ghz: ##:##:##:##:##:FD
wlan  5ghz: ##:##:##:##:##:FC

In the v2 the eth1 (lan) address is the same that the wlan 2,5 GHz but the wan mac is +1 and the wlan 5GHz is -1. Of course in v2 exists both eth0 and eth1.

tisj:
Running my v2 on OpenWrt, I have distinct MAC addresses on all interfaces:
wlan5 (-2), wlan2  (-1), lan, wan (+1).

Op 27-dec.-2017 01:27 schreef "howl" <notifications@github.com>:

I have get all the mac addresses in the Archer C7 v2 and the are as follow:
eth0 (wan): ##:##:##:##:##:FE
eth1 (lan): ##:##:##:##:##:FD
wlan 2ghz: ##:##:##:##:##:FD
wlan 5ghz: ##:##:##:##:##:FC

In the v2 the eth1 (lan) address is the same that the wlan 2,5 GHz but the
wan mac is +1 and the wlan 5GHz is -1. Of course in v2 exists both eth0 and
eth1.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/lede-project/source/pull/1593#issuecomment-354029008>,
or mute the thread

howl:
Yes it's wrong in OpenWRT/LEDE, the mac I said are the ones that appears with factory firmware. C7 seems to need corrections to take the same macs as the factory firmware but some versions need some changes and other versions need another changes.

blocktrron:
I've updates the PR to fit the TP-Link behavior. Could someone verify it is working correctly? I've tested for myself and it seems to work fine.

adridolf:
If it's only affecting C7 v4, it would be good to include the "v4" in the commit title!

howl:
The problem is in all versions but the stock macs are calculated different, perhaps another pull for v2. Will try to see if I can do it.

blocktrron:
@adridolf Thank for the hint, i updated the commit message to reflect the hardware revision

mkresin:
Thanks a lot for the PR.

I merged your PR into the master branch of [my staging tree](https://git.openwrt.org/?p=openwrt/staging/mkresin.git;a=summary).

PR #1594, opened 2017-12-20
nftables: update to 0.8

pprindeville:
0.8 has been out a couple of months no, builds against HEAD (95fe3c5) with no changes.

Given that it's been a year since 0.7 came out, there are probably a fair number of enhancements and fixes.

Note: this PR should be merged *after* https://github.com/lede-project/source/pull/1596.


pprindeville:
@sbyx Requesting review.

hauke:
Could you please switch to the normal http download from:
https://netfilter.org/projects/nftables/files/nftables-0.8.tar.bz2

There is also a more recent libnftnl, should we also update this version?
https://netfilter.org/projects/libnftnl/downloads.html

Have you runtime tested this?

pprindeville:
> Could you please switch to the normal http download from [...]

Done.  But... this causes a separate problem...  there's a file called `tests/py/ip/hash.t.rej` in the tarball!

`-rw-r--r-- pablo/pablo     504 2016-10-27 10:56 nftables-0.8/tests/py/ip/hash.t.rej`

> Have you runtime tested this?

I have not.  I don't have rules I use with `nft`.  Is someone willing to pull down the changes and do a test build/run?

pprindeville:
Got a workaround for the build issues with `nftables-0.8` (namely, the `.rej` file that somehow found its way into the tarball) but now there's a build issue with some non-varargs functions (specifically `mpz_printf()` which only takes on argument, but is being called from `...` macros).

hauke:
Thanks for taking care of this.

pprindeville:
> Thanks for taking care of this.

Not there yet, alas.  Pablo is fixing the `.rej` and `mpz_printf()` issues in a soon-to-be-out 0.8.1 release.  I'll update as soon as that's happened and I've had a chance to test build.

pprindeville:
@hauke BTW, how do you express that the current package requires version 1.0.8 of `libnftnl` or greater, for instance?

hauke:
You can not express dependencies on version as far as I know.
Just tell in your pull request that this depends on some other pull request.

pprindeville:
Moved to https://github.com/openwrt/openwrt/pull/601.

PR #1599, opened 2017-12-22
x86: add NVMe support for motherboards w/ M.2 slot

pprindeville:
A lot of 1U x86_64 servers have NVMe support, which is lower-power
and higher speed than SSD or CFast drives, etc.  The drivers
required to make this work are trivial.


hauke:
Does this make sense for all x86 subtargets?
I assume that most devices with a NVMe disk will be used with the x86_64 subtarget.

pprindeville:
There are Atom-based SoC's (32-bit) where having NVMe seems imminent because of the significantly lower power consumption of NVMe versus SSD or CFast while still having extremely high throughput.

There's nothing inherently x86_64'ish about NVMe.

hauke:

pprindeville:
Thank you!

PR #1600, opened 2017-12-22
lantiq: remove DSL_ChipSetHWVersion from status information

sch-m:
The value DSL_ChipSetHWVersion is fetched from the dsl frontend
via the dsl_control service, but not really provided by the dsl
frontend firmware and for now always "UNKNOWN".

The lantiq support told us that this information wouldn't be
provided in the foreseeable future, so let's remove this
useless "UNKNOWN" information.


mkresin:
merged thanks!

PR #1601, opened 2017-12-22
dnsmasq: reload configuration by signal

feckert:
- If the system name was changed then the "dnsmasq" must be restarted to rewrite his config with the new hostname. Adding "/etc/config/system" to the reload trigger will trigger "dnsmasq" on "/etc/config/system" change.

- To apply the new configuration "dnsmasq" must get restarted with start/stop to rewrite the configuration and then also restart a new instance to serve the new configuration. If reload_service was triggered then only sending a signal to dnsmasq to reload his config for the running service is enough. No new instance is required to serve the new configuration.

I have added this functionality one year ago, but this introduce errors on service reload and was reverted.
ADD:
https://github.com/TDT-AG/source/commit/854459a2f923376e0e509ebc0fb8ff90e9f13c02#diff-f65b4391dc02a776f873f8e933742458
REVERT:
https://github.com/TDT-AG/source/commit/93227e4d3fb81332f1dcb6eaa158ddf914ca67a6#diff-f65b4391dc02a776f873f8e933742458

### DIFF
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=dnsmasq
 PKG_VERSION:=2.78
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=http://thekelleys.org.uk/dnsmasq/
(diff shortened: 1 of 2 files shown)


dedeckeh:
Thanks; added to my staging tree (https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=a14ec406f8b312ab7b28b3ff1e88b74825df7d3a and https://git.lede-project.org/?p=lede/dedeckeh/staging.git;a=commit;h=e56f456037f8ca2f9e66b972adb7c621d44e8743)

PR #1604, opened 2017-12-25
Layerscape: fix build issue

yangbolu1991:
These two patches are to fix build issue for layersacpe.

a86e38c layerscape: add byte_swap.py script for ls-rcw package
f7e178d layerscape: fix PKG_BUILD_DIR for packages


hauke:

PR #1605, opened 2017-12-26
odhcp6c: authdata option config + patch

fandrieu:
Allows to set DHCPv6 option 11 "Authenfication" with hex data.

Adds "authdata" uci config for dhcpv6, passed as option -A to odhcp6c.
Includes a patch implementing this option in odhcp6c.

The french ISP orange.fr requires this options to be present, I guess this new uci config may be helpful to others.
A typical configuration would look like :
```
config interface 'wan6'
	option ifname '@wan'
	option proto 'dhcpv6'
	option reqaddress 'none'
	option vendorclass '0000040e0005736167656d'
	option userclass 'FSVDSL_livebox.Internet.softathome.Livebox4'
	option authdata '00000000000000000000006674692Fxxxxxxxxxxxxxx'
```

### DIFF
--- a/package/network/ipv6/odhcp6c/files/dhcpv6.sh
+++ b/package/network/ipv6/odhcp6c/files/dhcpv6.sh
@@ -26,6 +26,7 @@ proto_dhcpv6_init_config() {
 	proto_config_add_string 'ifaceid:ip6addr'
 	proto_config_add_string "userclass"
 	proto_config_add_string "vendorclass"
+	proto_config_add_string sendopts
 	proto_config_add_boolean delegate
 	proto_config_add_int "soltimeout"
 	proto_config_add_boolean fakeroutes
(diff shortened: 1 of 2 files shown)


dedeckeh:
Hi,
 I'm not in favour of adding extra command line options per DHCPv6 option which needs to be supported; an extra command line which allow to configure DHCPv6 options in a transparent way like the udhcpd busybox - x option is the preferred approach.

Also proper support for the DHCPv6 auth option requires more logic as it's used to detect replay detection and message integrity validation; see RFC3315 §21.2.

I'm interested to see what the DHCPv6 server is actually sending in the DHCPv6 advertise and reply messages.

fandrieu:
Hi, thanks for taking the time to answer.

I agree it would be better to have a way to set any DHCPv6 option. Especially since, as you point out, this is to support the non standard behavior of a custom server. (I did it like this to come up with a minimal patch and start the discussion, I wasn't sure a -x style option would be welcome)

Anyway, with that in mind I replaced the commit:
	- odhcp6c: add a "-x" option, with the same syntax as udchpc
	- dhcp6.sh: re-use the "sendopts" dhcp uci config

The options specified with "-x" are added as-is, it lacks a system to merge them with options already handled elsewhere (unlike udhcpc). This may result in duplicated options, it's up to the user to get it right.

The above  example config would now include instead of "authdata":
```
	option sendopts '0x0B:00000000000000000000006674692Fxxxxxxxxxxxxxx'
```

(the "opt" part of -x can be specified either in base 10 or 16 = "11" or "0x0B")

fandrieu:
As for this ISP's cutom use of the auth option: As far as I know there's no auth exchange as per the RFC, it's a completely custom use.
The server just requires that the auth option is set with our username as "authentication information" and the other fields set to 0. The server replies have the same kind of auth option with "authentication information" set to a specific string.
I sent you an email with a capture of this exchange.

dedeckeh:
@fandrieu I added support in odhcp6c to configure send options via the -x command line option and updated the odhcp6c package in master (https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=26045049baf646aa2ce3dce78106da5acf4936ea)
As proposed by your patch this can be configured via the uci option sendopts
Can you give it a try on your setup ?

fandrieu:
@dedeckeh I just installed the updated package, everything works as expected.
Thank you very much for working on this and props for adding option parsing !

PR #1609, opened 2017-12-28
target: add ARM server (EFI) target

mcbridematt:
This patch series adds an 'armserver' target, for ARM machines that use EFI to boot.

These machines are generally server machines (Centriq, ThunderX, X-Gene etc.) that comply with the ARM Server Base System requirements (using ACPI).

There is also a newer ARM Embedded Boot Spec which mandates a small set of EFI services with ACPI or DT, i.e u-boot's bootefi

GRUB2 is used as the bootloader, similar to x86, but with GPT as the disk label format.

These binaries will also work under QEMU and KVM with an EDK2 BIOS file, see https://gist.github.com/mcbridematt/2e45073a0a2a489d379c7b33c6019d91 for a boot log.

Tested:
- qemu-system-aarch64 on x86, -cpu cortex-a57 -M virt
- qemu-system-aarch64 on aarch64, -cpu host -M virt (NXP LS1043)
- Traverse LS1043V board with u-boot 2018.01-rc2 (GRUB test only, this machine requires a slightly different configuration - out of scope for this patch)

I based this target on armvirt but I think having a distinct armserver target is better as the ARM server/EFI target needs to support all possible hardware in the kernel for aarch64.

Two points of note:
- The boot partition for EFI must be FAT32, which has a minimum size of 32MB/~~256MB~~ (for 512/~~4K~~ disks), I have set the kernel partition size to ~~256MB~~ 48MB on TARGET_armserver to ensure this requirement is met (mkfs.fat may not generate FAT32 if the specified size is on or around 32MB)
- This patch also introduces QCOW2 image generation, for convenience. x86 generates vdi and vmdk images but qcow2 is the prevalent aarch64 image type among other distributions (Fedora, Ubuntu)
- This target uses 4.14


blogic:
looks good at first glance. the order of patches should be reversed. the first patch should be last to ensure that the tree stays bisectable.

blogic:

mcbridematt:
As suggested, I have reordered the patches back to front.

I also changed the minimum boot/EFI partition down to 48MB, as the image is created as a 512-byte sector image anything over the minimum size for that (32MB) seems OK.

mcbridematt:
Closing this one, update with tweaks and rebase onto openwrt/openwrt incoming

PR #1610, opened 2017-12-29
ramips:Correct switch and LED configuration for Lenovo newifi-y1

981213:
There are 3 ethernet ports on Y1(LAN1 on port1, LAN2 on port0 and WAN on port4).
Use a standalone switch configuration to match this and use switch trigger so that LAN LED could indicate the connetction status for both lan ports correctly.

### DIFF
--- a/target/linux/ramips/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/base-files/etc/board.d/01_leds
@@ -453,8 +453,7 @@ y1)
 	set_usb_led "$boardname:blue:usb"
 	ucidef_set_led_netdev "wifi" "WIFI" "$boardname:blue:wifi" "wlan1"
 	ucidef_set_led_netdev "wifi5g" "WIFI5G" "$boardname:blue:wifi5g" "wlan0"
-	ucidef_set_led_netdev "lan" "LAN" "$boardname:blue:lan" "eth0.1" "tx rx"
-	ucidef_set_led_netdev "wan" "WAN" "$boardname:blue:internet" "eth0.2" "tx rx"
+	ucidef_set_led_switch "lan" "LAN" "$boardname:blue:lan" "switch0" "0x03"
 	;;
 y1s)
 	set_usb_led "$boardname:blue:usb"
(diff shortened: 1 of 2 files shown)


target/linux/ramips/base-files/etc/board.d/01_leds
```diff
@@ -453,7 +453,7 @@ y1)
 	set_usb_led "$boardname:blue:usb"
 	ucidef_set_led_netdev "wifi" "WIFI" "$boardname:blue:wifi" "wlan1"
 	ucidef_set_led_netdev "wifi5g" "WIFI5G" "$boardname:blue:wifi5g" "wlan0"
-	ucidef_set_led_netdev "lan" "LAN" "$boardname:blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_switch "lan" "LAN" "$boardname:blue:lan" "switch0" "0x03"
 	ucidef_set_led_netdev "wan" "WAN" "$boardname:blue:internet" "eth0.2" "tx rx"
```
mkresin: Would you please switch the internet led to the switchdev trigger as well. The netdev trigger will only work as long as VID 2 is used for wan which might not be the case in all configurations.

target/linux/ramips/base-files/etc/board.d/02_network
```diff
@@ -348,6 +347,10 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "0:wan" "6@eth0"
 		;;
+	y1)
+		ucidef_add_switch "switch0" \
+		"0:lan:2" "1:lan:1" "4:wan" "6@eth0"
```
mkresin: please add a tab here

target/linux/ramips/base-files/etc/board.d/01_leds
981213: There is a separated led for wan port which directly connected to ephy led pin and that LED could indicate the connection status for WAN port.I'm not sure whether I should keep the old configuration or just drop the configuration for internet LED at all.

981213: BTW in my memory the original firmware used an userspace program to turn on the internet led when there is a stable internet connection.

981213:
I've just dropped the internet led configuration because there is another WAN led for port4 and I think there isn't a suitable trigger for it currently.

blogic:

mkresin:
Thanks a lot for the PR.

I merged your PR into the master branch of [my staging tree](https://git.openwrt.org/?p=openwrt/staging/mkresin.git;a=summary).

PR #1612, opened 2017-12-29
lantiq: fix Netgear DM200 dts compatible string

tomjnixon:
This was broken in 7bab49fd ("lantiq: add compatible strings to dts files"), causing garbled serial output during boot, and likely other issues.

It looks like the same issue is present in P2601HNFX.dts, though i don't have that device to test it.

Thanks,

Tom


mkresin:
merged, thanks!

PR #1620, opened 2018-01-09
ramips: add support for TP-Link Archer C20 v4

Linaro1985:
TP-Link Archer C20 v4 is a router with 5-port FE switch and
non-detachable antennas. It's based on MediaTek MT7628N+MT7610EN.

Specification:
- MediaTek MT7628N/N (580 Mhz)
- 64 MB of RAM
- 8 MB of FLASH
- 2T2R 2.4 GHz and 1T1R 5 GHz
- 5x 10/100 Mbps Ethernet
- 3x external, non-detachable antennas
- UART (J1) header on PCB (115200 8n1)
- 7x LED (GPIO-controlled*), 2x button, power input switch

* WAN LED in this devices is a dual-color, dual-leads type which isn't
  (fully) supported by gpio-leds driver. This type of LED requires both
  GPIOs state change at the same time to select color or turn it off.
  For now, we support/use only the green part of the LED.
* MT7610EN ac chip isn't not supported by LEDE. Therefore 5Ghz won't
  work.

Flash instruction:

The only way to flash LEDE image in ArcherC20v4 is to use
tftp recovery mode in U-Boot:

1. Configure PC with static IP 192.168.0.66/24 and tftp server.
2. Rename "openwrt-ramips-mt7628-ArcherC20v4-squashfs-tftp-recovery.bin"
   to "tp_recovery.bin" and place it in tftp server directory.
3. Connect PC with one of LAN ports, press the reset button, power up
   the router and keep button pressed for around 6-7 seconds, until
   device starts downloading the file.
4. Router will download file from server, write it to flash and reboot.

Signed-off-by: Maxim Anisimov <maxim.anisimov.ua@gmail.com>
[LEDE_Bootlog.txt](https://github.com/lede-project/source/files/1614151/LEDE_Bootlog.txt)

### DIFF
--- a/target/linux/ramips/base-files/lib/ramips.sh
+++ b/target/linux/ramips/base-files/lib/ramips.sh
@@ -85,9 +85,6 @@ ramips_board_detect() {
 	*"C108")
 		name="c108"
 		;;
-	*"C20")
-		name="c20"
-		;;
 	*"C20i")
 		name="c20i"
 		;;
--- a/target/linux/ramips/dts/ArcherC20v4.dts
+++ b/target/linux/ramips/dts/ArcherC20v4.dts
@@ -1,0 +1,170 @@
+/dts-v1/;
+
+#include "mt7628an.dtsi"
+
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+
+/ {
+	compatible = "tplink,c20-v4", "mediatek,mt7628an-soc";
+	model = "TP-Link Archer C20 v4";
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		lan {
+			label = "c20-v4:green:lan";
+			gpios = <&gpio1 10 GPIO_ACTIVE_LOW>;
+		};
+
+		power {
+			label = "c20-v4:green:power";
+			gpios = <&gpio0 2 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			label = "c20-v4:green:wan";
+			gpios = <&gpio1 11 GPIO_ACTIVE_LOW>;
+		};
+
+		wan_orange {
+			label = "c20-v4:orange:wan";
+			gpios = <&gpio0 11 GPIO_ACTIVE_LOW>;
+		};
+		
+		wlan5g {
+			label = "c20-v4:green:wlan5g";
+			gpios = <&gpio1 8 GPIO_ACTIVE_LOW>;
+		};
+
+		wlan2g {
+			label = "c20-v4:green:wlan2g";
+			gpios = <&gpio1 7 GPIO_ACTIVE_LOW>;
+		};
+
+		wps {
+			label = "c20-v4:green:wps";
+			gpios = <&gpio1 9 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys-polled {
+		compatible = "gpio-keys-polled";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio1 6 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		rfkill {
+			label = "rfkill";
+			gpios = <&gpio1 5 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+		};
+	};
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x0 0x4000000>;
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	m25p80@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,chunked-io = <32>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x20000>;
+			read-only;
+		};
+
+		partition@20000 {
+			label = "firmware";
+			reg = <0x20000 0x7a0000>;
+		};
+
+		partition@7c0000 {
+			label = "config";
+			reg = <0x7c0000 0x10000>;
+			read-only;
+		};
+
+		rom: partition@7d0000 {
+			label = "rom";
+			reg = <0x7d0000 0x10000>;
+			read-only;
+		};
+
+		partition@7e0000 {
+			label = "romfile";
+			reg = <0x7e0000 0x10000>;
+			read-only;
+		};
+
+		radio: partition@7f0000 {
+			label = "radio";
+			reg = <0x7f0000 0x10000>;
+			read-only;
+		};
+	};
+};
+
+&ehci {
+	status = "disabled";
+};
+
+&ohci {
+	status = "disabled";
+};
+
+&wmac {
+	status = "okay";
+	mediatek,mtd-eeprom = <&radio 0x0>;
+	mtd-mac-address = <&rom 0xf100>;
+	mtd-mac-address-increment = <(-2)>;
+};
+
+&ethernet {
+	mtd-mac-address = <&rom 0xf100>;
+	mediatek,portmap = "wllll";
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "i2s", "refclk", "p0led_an", "p1led_an", "p2led_an", "p3led_an", "p4led_an", "wdt";
+			ralink,function = "gpio";
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+
+	pcie-bridge {
+		mt76@0,0 {
+			reg = <0x0000 0 0 0 0>;
+			device_type = "pci";
+			mediatek,mtd-eeprom = <&radio 0x8000>;
+			ieee80211-freq-limit = <5000000 6000000>;
+			mtd-mac-address = <&rom 0xf100>;
+			mtd-mac-address-increment = <(-1)>;
+		};
+	};
+};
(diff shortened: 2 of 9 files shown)


mkresin:
Please create an extra commit for renaming the c20 to c20v1. It shouldn't be in the commit adding support for the c20v4.

The rename will cause issues for people using a c20v1. But I'm fine to do so, since we don't have support for the c20v1 in stable and keeping compatibility within in the development version (master) isn't than crucial.

target/linux/ramips/base-files/lib/ramips.sh
```diff
@@ -85,8 +85,11 @@ ramips_board_detect() {
 	*"C108")
 		name="c108"
 		;;
-	*"C20")
-		name="c20"
+	*"C20 v1")
+		name="c20v1"
+		;;
+	*"C20 v4")
```
mkresin: Drop the C20 v4 here and use the compatible string as boardname in all files (tplink,c20v4).

target/linux/ramips/image/mt76x8.mk
```diff
@@ -18,6 +18,18 @@ define Device/tplink
 endef
 DEVICE_VARS += TPLINK_FLASHLAYOUT TPLINK_HWID TPLINK_HWREV TPLINK_HWREVADD TPLINK_HVERSION
 
+define Device/ArcherC20v4
```
mkresin: Please use `define Device/tplink_c20v4` here.

target/linux/ramips/image/mt76x8.mk
```diff
@@ -18,6 +18,18 @@ define Device/tplink
...
+  DTS := ArcherC20v4
+  IMAGE_SIZE := 7808k
+  DEVICE_TITLE := TP-Link ArcherC20 v4
+  TPLINK_FLASHLAYOUT := 8Mmtk
+  TPLINK_HWID := 0xc200004
+  TPLINK_HWREV := 0x283
+  TPLINK_HWREVADD := 0x4
+  TPLINK_HVERSION := 3
+endef
+TARGET_DEVICES += ArcherC20v4
```
mkresin: `TARGET_DEVICES += tplink_c20v4`

target/linux/ramips/dts/ArcherC20v4.dts
```diff
@@ -0,0 +1,170 @@
...
+
+&ethernet {
+	mtd-mac-address = <&rom 0xf100>;
+	mediatek,portmap = "wllll";
+};
+
+&pinctrl {
+	state_default: pinctrl0 {
+		gpio {
+			ralink,group = "p0led_an", "p2led_an", "perst", "refclk", "wdt", "wled_an";
```
mkresin: must be:

`ralink,group = "i2s", "refclk", "p0led_an", "p1led_an", "p2led_an",  "p3led_an", "p4led_an", "wdt";`

- power led uses a pin of the i2s group
- rfkill button uses the refclk pin
- reset button used the wdt pin
- wlan2g led uses the p4led_an pin
- wlan5g led uses the p3led_an pin
- wps led uses the p2led_an pin
- lan led used the p1led_an pin
- green wan led uses the p0led_an pin
- orange wan led uses gpio pin which can be configured only as gpio

Linaro1985:
All done! Please, check it. Thanks!

mkresin:
Looks good so far. I have one more change request. Would you please seperate boardname and version number with a minus. The same in the dts as well of course:

- tplink,c20-v1
- tplink,c20-v4

Please add a commit message to the "ramips: rename TP-Link Archer C20 to TP-Link Archer C20 v1" commit. Explain why you have done the changes and add a note that you migrated the board to the generic board detection. Have a look at #594 for some related commit messages.

Linaro1985:
I have just finished changes. I hope that everything is correct.

mkresin:
Not yet. You need to change Device/tplink_c20v4 to Device/tplink_c20-v4 as well. Same for the tplink_c20v1 and the TARGET_DEVICES variable.

Please do a full runtime test on your c20-v4 afterwards including a sysupgrade and so on.

Linaro1985:
> You need to change Device/tplink_c20v4 to Device/tplink_c20-v4 as well. Same for the tplink_c20v1 and the TARGET_DEVICES variable.
Done
> Please do a full runtime test on your c20-v4 afterwards including a sysupgrade and so on.
Unfortunately I have no device by the hand. Users of a [forum](https://forum.lede-project.org/t/tp-link-archer-c20-v4-build/8945/6) help me with testing. Therefore it will take some time.

mkresin:
Just leave a note if the test was done. A new bootlog would be handy as well.

target/linux/ramips/dts/ArcherC20v4.dts
```diff
@@ -0,0 +1,170 @@
+/dts-v1/;
+
+#include "mt7628an.dtsi"
```
heinzek: You can include TL-WR84XN.dtsi and remove:
chosen, memory, spi, ehci, ohci
In "wmac" leave only "mtd-mac-address-increment",
In "Ethernet" remove "mtd-mac-address"

ps. Maybe its good time to change filename from TL-WR84XN.dtsi to TPLINK_8M.dtsi ?

mkresin: Please don't include the TL-WR84XN.dtsi or at least not as long as the file is named TL-WR84XN.dtsi. Such DTS files are a pain to read.

TPLINK_8M.dtsi is way better. But it shouldn't happen in context of this PR. I'm, fine to merge an extra PR which does these kind of optimizations.

target/linux/ramips/image/mt76x8.mk
```diff
@@ -18,6 +18,18 @@ define Device/tplink
 endef
 DEVICE_VARS += TPLINK_FLASHLAYOUT TPLINK_HWID TPLINK_HWREV TPLINK_HWREVADD TPLINK_HVERSION
 
+define Device/tplink_c20-v4
```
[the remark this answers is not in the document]
mkresin: In general @heinzek is correct here. But I have already a commit in my staging tree which will break to alphabetical order anyway. Hence I've not requested to keep alphabetical order and it doesn't really matter till my stuff is committed as well.

target/linux/ramips/dts/ArcherC20v4.dts
Linaro1985: > Please don't include the TL-WR84XN.dtsi
Ok

target/linux/ramips/image/mt76x8.mk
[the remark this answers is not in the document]
Linaro1985: I will correct it.

target/linux/ramips/image/mt76x8.mk
```diff
@@ -18,6 +18,18 @@ define Device/tplink
...
+  $(Device/tplink)
+  DTS := ArcherC20v4
+  IMAGE_SIZE := 7808k
+  DEVICE_TITLE := TP-Link ArcherC20 v4
+  TPLINK_FLASHLAYOUT := 8Mmtk
+  TPLINK_HWID := 0xc200004
+  TPLINK_HWREV := 0x283
```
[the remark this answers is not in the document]
Linaro1985: In my region firmware HWREV is 0x283. I didn't check a HWREV code for other regions. It turns out that a code everywhere different. Is there options?

Linaro1985:
New logfile: [LEDE_Bootlog.txt](https://github.com/lede-project/source/files/1617589/LEDE_Bootlog.txt)
The user from a forum has tested new firmware release and reported that everything works normally except sysupgrade procedure via webinterface.

> test4 sysupgrade.bin
> using webinterface (Window 10, Firefox 57.0.5 (64bit)
> Error: The uploaded image file does not contain a supported format. Make sure that you choose the > generic image format for your platform.
> Command Line upgrade successful.

Any ideas?

mkresin:
>>test4 sysupgrade.bin
>>using webinterface (Window 10, Firefox 57.0.5 (64bit)
>>Error: The uploaded image file does not contain a supported format. Make sure that you choose the generic image format for your platform.
>>
>>Command Line upgrade successful.
>>
>Any ideas?

To be honest, it doesn't make any sense. If the upgrade via command line works without any errors, it should work via LuCI as well. LuCI is just a web frontend for for the sysupgrade command and doesn't do any extra checks. Nevertheless, I really would like to know what went wrong for your tester. Would you please ask for a (putty/serial) logfile of a sysupgrade done via command line. It could be of course that the tester used a wrong image with LuCI or has some kind of caching issues with the browser.

@heinzek Any advice what to do with the HWREV? You are way more familiar with the tp-link images than I.

target/linux/ramips/image/mt76x8.mk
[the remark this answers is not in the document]
heinzek: Look here: https://github.com/lede-project/source/commit/24043a0d2e01b9843c0dc529205b3b0bc7ecbbf9

>Factory image notes:
>
> These devices use version 3 of TP-Link header, fortunately without RSA
signature (at least in case of devices sold in Europe). The difference
lays in the requirement for a non-zero value in "Additional Hardware
Version" field. Ideally, it should match the value stored in vendor
firmware header on device ("0x4"/"0x13" for these devices) but it seems
that anything other than "0" is correct.

You can set a random "non-zero" value in this field and everything is OK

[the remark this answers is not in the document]
Linaro1985: But in my case difference only in HWREV code! HWREVADD is the same for EU and RU regions.
2018-01-11_214328_cr

Linaro1985:
@mkresin The user from a forum has started sysupgrade process via webinterface once again and everything has passed perfectly. So everything is all right.
[LEDE_UpgradeLog.txt](https://github.com/lede-project/source/files/1623770/LEDE_UpgradeLog.txt)

Linaro1985:
> The difference lays in the requirement for a non-zero value in "Additional Hardware Version" field. 

In case of Archer C20 v4 it's not true, because of difference only in HWREV code. EU firmware header data was seted to resolve this question. I hope that this option will ok. Please accept patches if there are no other problems. Thanks!

mkresin:
Thanks! Merged into my [staging tree](https://git.openwrt.org/?p=openwrt/staging/mkresin.git;a=summary).

PR #1638, opened 2018-02-23
Add INITRAMFS for mt7620

lars18th:
Enable the compilation of "initramfs" images for testing without flashing in devices with mt7620.


lars18th:
I close it to change to https://github.com/openwrt/openwrt/pulls

PR #1644, opened 2018-04-05
Teach iwinfo about the Marvell 88W8964 and 88W8887

ErikBAndersen:
Teach iwinfo about the Marvell 88W8964 and 88W8887 wireless chipsets,
included in the wrt3200acm.  Since the current iwinfo tree seems to only be
in the openwrt git tree, and not on github, this is submitted here as a
build-time patch.  But this should really be included in the upstream iwinfo
git tree instead.


lynxis:
Hi @ErikBAndersen,
can you please re-open this PR against https://github.com/openwrt/openwrt ?
The lede project merged with the openwrt project.

bobafetthotmail:
> . But this should really be included in the upstream iwinfo git tree instead.

You can send patches for that git tree using the mailing list.

PR #1645, opened 2018-04-06
ramips: fix USB init on MT7620 w/kernel 4.14

K900:
The old phy driver just used direct register access;
upstream uses syscon, so add it to the device tree

Signed-off-by: Ilya Katsnelson <me@0upti.me>

### DIFF
--- a/target/linux/ramips/dts/mt7620a.dtsi
+++ b/target/linux/ramips/dts/mt7620a.dtsi
@@ -35,7 +35,7 @@
 		#size-cells = <1>;
 
 		sysc: sysc@0 {
-			compatible = "ralink,mt7620a-sysc", "ralink,rt3050-sysc";
+			compatible = "ralink,mt7620a-sysc", "ralink,rt3050-sysc", "syscon";
 			reg = <0x0 0x100>;
 		};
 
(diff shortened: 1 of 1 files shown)


K900:
I am not sure this fix is fully correct, because I'm not familiar enough with syscon to tell if adding it to the DT can actually _break_ stuff, but it gets the phy initialized properly and devices are detected - tested by me on miwifi-mini, some more testing going on on IRC right now.

K900:
Also, mt7620n/mt7628an probably need the same changes, but I don't have any hardware to test on, and I'm not familiar enough with the SoC to be confident it'll work without testing, so I don't want to be the one making the change.

K900:
Actually seems like upstream already has a similar setup in the device tree for [mt7628a](https://github.com/torvalds/linux/blob/master/arch/mips/boot/dts/ralink/mt7628a.dtsi).

mrkiko:
Hello! This fixes my Archer MR200.
Without this patch, the internal LTE modem won't appear.
Now, it does, and I can talk AT with it :) so it works great!
A lot of thanks, cool work!

P.S.: MT7620N hardware like the DWR-921 (C1 or C3 variants) may suffer the same issue. Testing would be apreciated.

K900:
Re-pushed with proper indentation (2 tabs instead of 8 spaces)

CHKDSK88:
I have the same issue on DWR-921-C3. I'll add Your patch to my dtb and I'll let You know if it work on 7620N.

P.S.
Please reopen this in openwrt tree. Lede not accept new PR's.

K900:
Done: https://github.com/openwrt/openwrt/pull/851

PR #1652, opened 2018-07-12
Revert sit: reload iphdr in ipip6_rcv

hanipouspilot:
This fixes FS#1541

Initially the commit was made upstream with CC to <stable>,
then reverted upstream, but without the CC.

This has been fixed in other distos like Ubuntu.

Signed-off-by: Dmitry Tunin <hanipouspilot@gmail.com>


dedeckeh:
No PRs are accepted anymore on the lede-project. I noticed you opened a PR https://github.com/openwrt/openwrt/pull/1172 on the OpenWRT project; if you want the patch to be backported to lede-17.01 please mention this in the PR

hanipouspilot:
I added that to the other PR.

PR #1656, opened 2018-10-08
base-files: config_get: prevent filename globbing

dvlemplgk:
When config_get is called as "config_get <section> <option>" the option
is unexpectedly globbed by the shell which differs from the way options
are read to a variable with "config_get <variable> <section> <option>".
Add another layer of double quotes to fix it.

Signed-off-by: Günther Kelleter <guenther.kelleter@devolo.de>


dedeckeh:
PRs are not accepted anymore on the Lede project; please open a PR in https://github.com/openwrt/openwrt if you want the patch to be integrated in OpenWrt

dvlemplgk:
Ok, thanks.

PR #1662, opened 2019-06-17
WIP: mbim.sh: use umbim config instead of dhcp client

jcadduono:
I noticed after purchasing my Sierra EM7565 and EM7455 that there were a few issues:

1. Sierra has a weird PIN status return for a non-existent pin2. I figured this might be related to the EM7xxx dual sim capability, but upon dual sim mode I still got the error. Thus, I have added the ignore_pinfail config option.
2. The mbim connection initialization actually retrieves the IP addresses for us, so DHCP is not required.
3. Because of 2, I was failing to get an IP address for IPv4. (IP was already obtained, but not being used) It was still working for IPv6 because of IPv6 however.

Okay so there's some considerations to be had here...
I've added pdptype to mbim.sh because umbim (secretly) supports it, however, it does not appear to support it very well as upon testing, the umbim client will just get stuck when using ipv4 or ipv6 only pdptype. I know that my mobile provider supports all of these pdptypes, so maybe as a temporary solution we could simply use pdptype variable to decide if we want to ignore an IP version reported by umbim config and always assume that umbim will connect with ipv4v6.

Regarding the pin fail ignoring, perhaps that is another issue left to umbim? I don't know the protocol at all but maybe a flag is getting misinterpreted somewhere? Unlikely, I suppose.

I wrote a luci-proto-mbim module and would be happy to submit that after this PR is looked over.

Major issue: **IPv6 gateway does not propagate to clients** (dhcp server is sending the router's br-lan link local ipv6 address?!)
I'm stuck on this one and don't know how to fix it yet.
I'm hoping you guys would have some answers.

### DIFF
--- a/package/network/utils/umbim/files/lib/netifd/proto/mbim.sh
+++ b/package/network/utils/umbim/files/lib/netifd/proto/mbim.sh
@@ -11,8 +11,10 @@ proto_mbim_init_config() {
 	available=1
 	no_device=1
 	proto_config_add_string "device:device"
+	proto_config_add_string pdptype
 	proto_config_add_string apn
 	proto_config_add_string pincode
+	proto_config_add_boolean ignore_pinfail
 	proto_config_add_string delay
 	proto_config_add_string auth
 	proto_config_add_string username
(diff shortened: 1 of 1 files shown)


jcadduono:
@bmork @blogic

jcadduono:
dang i must have cloned this from lede project initially, gotta move this PR lol

PR #1666, opened 2019-11-01
Mbacklund2

mbacklund:
Added correct file for wnr2000v5.c


pocha12:
where can I get openwrt 
arch 64 A53 amlogic

PR #1673, opened 2021-02-28
build: accept GCC 11

stefanct:
Hi,

I was using the dev version of GCC 11 as host compiler and noticed that the build system checks for very specific GCC versions. I am not exactly sure why this construct is there and it clearly violates the principle to test for actual effects instead of version numbers but I have created this minimal patch that includes GCC 11 as well.


stefanct:
Apparently this was fixed in cae69d558135456976b8fc6cb08530d1358cf6d5

